PR #62, opened 2020-11-02
Cassandra query caching in Redis

benbromhead:
This PR implements a redis cache for Apache Cassandra in Redis. It should also work for most SQL style queries that loosely follow similar structures and query types that Apache Cassandra can support (once we have a source that can support SQL queries).

Caching in redis is based on sorted sets, the set name maps to the cassandra partition key and the clustering key is prepended to the stored value in the set. 

This allows for range queries as well as queries to retreive slices of the parition key from Redis.

The implementation is a write through cache, so cold restarts of the cache itself will result in additional load. 

This cache implementation is specific to SQL(CQL) -> RESP2 however it can support arbitrary transform chains on the caching side of things.

The transform supports pipelined requests, though on the update path, it currently doesn't strip out reads from the cache side of things.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1,3 +1,3 @@
 ---
-main_log_level: "info,shotover_proxy=trace"
+main_log_level: "info,shotover_proxy=info"
 observability_interface: "127.0.0.1:9001"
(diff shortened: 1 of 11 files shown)


rukai:
Looking at the existing RedisCache implementation in main I can see that ben already directly pushed this to main in https://github.com/shotover/shotover-proxy/commit/1c7af53fdb4c72f84d9d030b3f6f7947963575cc

I'll go ahead and close this PR.

The next step for RedisCache will be writing up an integration test so we can observe what does and doesnt work in the current implementation.

Issue #120, opened 2021-07-26
Thoughts on user provided transforms


rukai:
There has been a lot of work/discussion towards scriptable or plugin-driven transforms.
But this sounds like it could be the wrong approach.
The transforms have the same requirements as shotover itself which needs to be:
*   extremely fast
*   correct
*   cannot crash or otherwise drop data

The perfect candidate for this is rust, we obviously believe that because shotover itself is written in rust.

Even so we could still use rust compiled to wasm or rust dylib plugins.
That way developers can use rust or if they really really want they can use another language.

But doing so has several disadvantages:
*   poor versioning - by using a dylib or wasm we step outside of cargo's robust semantic versioning and dependency management.
*   adds complexity to the implementation resulting in a large maintenance burden and surface area for bugs
*   when writing plugins in another language:
    + mismatch between the provided API and the way code is written in the plugin language - can be reduced by bindings that provide abstractions suitable for the plugin language - but again, requires effort to implement + maintain
*   when writing plugins in rust:
    + for dylib we lose our type safety because we have to expose an FFI API
    + for wasm we have to serialize and deserialize our types into bincode as we can only communicate over an array of bytes between the host and wasm

**So how can we give shotover the flexibility it needs without plugins/wasm?**

We can expose shotover as a library crate exposing a builder pattern API.
Then the user of shotover implements their own binary crate using this shotover library.

e.g. They could run vanilla shotover with just the following
```rust

fn main() {
    Shotover::new().run();
}
```
This would be a very batteries included approach. `run` would be responsible not just for running shotover but it would also start the async runtime, read the config from disk, start the logger etc.

We could still provide a binary release of shotover by compiling this simple example. This binary release could do everything except implement custom transforms.

Then to extend shotover with their own transforms, developers implement a `Transform` trait and register it like this:
```rust
struct MyTransform { }

impl Transform for MyTransform {
    ...
}

fn main() {
    Shotover::new()
        .register_transform(MyTransform::new())
        .run();
}
```

As a bonus we could allow bundling config into the binary like this:
```rust
fn main() {
    Shotover::new()
        // config guaranteed to exist at compile time
        .set_config(include_str!("config.yaml"));
        .run();
}
```

or we could even go so far as to have the config constructed in our rust builder pattern instead of yaml:
```rust
fn main() {
    Shotover::new()
        // config structure gauranteed to be correct at compile time
[...]

Issue #108, opened 2021-07-15
Create a WASM transform


benbromhead:
Using either something like:

- A lib like wasmer https://wasmer.io/posts/wasmer-2.0
- Or the way Timber IO implemented WASM support https://github.com/timberio/vector/blob/41d90b7df1995429d179c89a4f1c176d72a263f1/rfcs/2020-04-15-wasm-plugins.md

Create a Transform that will pass a Wrapper struct into a user defined WASM function, and also allow the user to call `call_next_transform` on it to run subsequent transforms. 

Initially a basic implementation of the Transform config will look something like:

```yaml
- WASMTransform:
   - path: myfunction.wasm
   - name: Func
```

rukai:
This was attempted in https://github.com/shotover/shotover-proxy/pull/136
Since then we have instead implemented custom transforms via rust static linking.
We dont have any current plans for a WASM transform, although its possible that could change in the future.

PR #136, opened 2021-08-11
Wasm transform

rukai:
This branch has working wasm plugins but requires a lot of cleanup:
* shotover-proxy should depend on shotover-types crate instead of duplicating Message etc.
* I designed the current implementation thinking we would want to be able to call into the same wasm instance concurrently, but I just noticed `Transform::transform` takes mut self so it doesn't seem like we can have multiple `Transform::transform` calls going concurrently on the same transform instance anyway.
    + I need to better understand how shotover handles concurrent connections and apply that to the design.
* loading system - need to load a wasm file specified in config.
* versioning system - call `api_version` and fail if doesnt return 0.
* address the various TODO's

Communications between the wasm transform plugin and shotover occurs as following:
1. shotover serializes Messages into bincode
2. shotover requests a memory allocation from the plugin with the size of the bincode.
3. shotover writes the bincode into the allocated memory.
4. shotover calls run on the plugin
    1. run deserializes the bincode into a Messages.
    2. run passes Messages to the custom transform which does arbitrary transformations on it and then returns it
    3. run serializes Messages back to bincode and returns the address to it. We can only return one integer so the size is stored in the first 8 bytes of the response.
5. shotover deserializes the bincode to Messages and stores that in the Wrapper
6. Continues the transform chain via call_next_transform

There are 3 different rust wasm runtimes: wasmer, wasmtime and lucet
* comparison between wasmer and wasmtime: https://wiki.alopex.li/ActuallyUsingWasm
* comparison between wasmer and lucet https://wasmer.io/wasmer-vs-lucet

Wasmer seems to be the best one but it is possible we find that another runtime is better suited to our needs.
Advantages for wasmer:
* multiple backends: cranelift and llvm (the others only have cranelift)
* cross platform (lucet only runs on linux)

Related to https://github.com/shotover/shotover-proxy/issues/120 and https://github.com/shotover/shotover-proxy/issues/108

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -35,6 +35,9 @@ jobs:
     - name: Benchmark
       run: |
         set -o pipefail # Needed to make the cargo bench fail early on e.g. compiler error.
+        cd wasm-transform
+        cargo build --release
+        cd ..
         cargo bench --bench redis_benches -- --output-format bencher --noplot | tee bench_redis.txt
         cargo bench --bench chain_benches -- --output-format bencher --noplot | tee bench_chains.txt
     - name: Store benchmark result
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,7 +64,11 @@ jobs:
     - name: Build
       run: cargo build --all-targets
     - name: Run tests
-      run: cargo test -- --include-ignored
+      run: |
+        cd wasm-transform
+        cargo build --release
+        cd ..
+        cargo test -- --include-ignored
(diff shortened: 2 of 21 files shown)


rukai:
The work in this PR could be a useful starting point for research into wasm/plugins for shotover.
Personally I would like to see a radically different approach to the one in this PR.
But even if we kept the approach mostly the same we will still need to start from scratch again given all the changes that's happened to shotover in the last 2 years, so I'll go ahead and close this and a new PR can be opened one day.

Issue #107, opened 2021-07-15
Cassandra source closes socket


benbromhead:
In writing a set of integration tests. Cassandra-rs will fail to connect to shotover configured with a Cassandra source, while cqlsh will connect succesfully.

After a quick look at wireshark, it appears that the cassandra-rs driver (a wrapper around the c/c++ datastax one) tries a few unsupported protocol version, which fail or get downgraded. Cassandra itself wont shutdown the tcp connection, whereas shotover will causeing the driver to error out. 

Given Cassandra supports this behaviour (as broken as it seems), shotover needs to support this. Protocol version errors shouldn't cause us to drop the connection.

Claude-at-Instaclustr:
Testing works locally, now trying to get automated build to work.

PR #225, opened 2021-09-28
Cassandra Peers rewrite

conorbros:
The functionality of port rewriting and emulating a single node (returning an empty system peers table) is working. Before this can be merged I need to:
- [ ] Fix some issues with the Cassandra driver
- [ ] Reconsider the ergonomics of the config file and setting the variables/using this transform. The current implement is pretty ad-hoc so I can test things.
- [ ] In the case of Shotover emulating a single node, we also need to have Shotover handle request routing to the right Cassandra nodes
- [ ] Add integration tests. This is blocked by #107 

#### Cassandra driver issues: 
* The cause of the hacky 4 magic bytes that needed to be added at the start of the body is because the driver wasn't adding the [byte representing the ResultKind](https://github.com/apache/cassandra/blob/5ee1ba2dfaac0f4424322268912f6714e89904ae/doc/native_protocol_v4.spec#L554) 
* The driver wasn't properly serialising null values. Cassandra parsers [expects a -1](https://github.com/apache/cassandra/blob/5ee1ba2dfaac0f4424322268912f6714e89904ae/doc/native_protocol_v4.spec#L230) when a value is null. The driver was just skipping them altogether.

### DIFF
--- a/shotover-proxy/examples/cassandra-rewrite-peers/docker-compose.yml
+++ b/shotover-proxy/examples/cassandra-rewrite-peers/docker-compose.yml
@@ -1,0 +1,30 @@
+version: "3.3"
+services:
+  cassandra-one:
+    image: library/cassandra:4.0.0
+    ports:
+      - "9043:9042"
+    healthcheck:
+      test: [ "CMD", "cqlsh", "-e", "describe keyspaces" ]
+      interval: 5s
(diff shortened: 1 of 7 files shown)


conorbros:
Closing because my fork has detatched when we made the repo public, I have the branch locally to resubmit.

PR #300, opened 2021-10-25
Bump version to v0.1.0

XA21X:
A new release is needed with functionality from #275 for INS-15988. I am bumping the minor version, roughly following SemVer, although we're being flexible with pre-1.0 in regards to the definition of a 'breaking change'.


rukai:
ill tag the commit

PR #328, opened 2021-10-29
Register metrics on startup

conorbros:
Register all metrics on startup so they can be retrieved from the metrics endpoint without any needed to be actually be recorded yet.

### DIFF
--- a/shotover-proxy/src/server.rs
+++ b/shotover-proxy/src/server.rs
@@ -90,6 +90,7 @@ impl<C: Codec + 'static> TcpCodecListener<C> {
         tls: Option<TlsAcceptor>,
     ) -> Self {
         register_gauge!("shotover_available_connections", Unit::Count, "source" => source_name.clone());
+        gauge!("shotover_available_connections", limit_connections.available_permits() as f64, "source" => source_name.clone());
 
         TcpCodecListener {
             chain,
--- a/shotover-proxy/src/server.rs
+++ b/shotover-proxy/src/server.rs
@@ -4,7 +4,7 @@ use crate::transforms::chain::TransformChain;
 use crate::transforms::Wrapper;
 use anyhow::{anyhow, Result};
 use futures::StreamExt;
-use metrics::gauge;
+use metrics::{gauge, register_gauge, Unit};
 use std::sync::Arc;
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::net::{TcpListener, TcpStream};
@@ -89,6 +89,8 @@ impl<C: Codec + 'static> TcpCodecListener<C> {
         shutdown_complete_tx: mpsc::Sender<()>,
         tls: Option<TlsAcceptor>,
     ) -> Self {
+        register_gauge!("shotover_available_connections", Unit::Count, "source" => source_name.clone());
+
         TcpCodecListener {
             chain,
             source_name,
--- a/shotover-proxy/src/transforms/mod.rs
+++ b/shotover-proxy/src/transforms/mod.rs
@@ -250,7 +250,11 @@ pub enum TransformsConfig {
 impl TransformsConfig {
     #[async_recursion]
     /// Return a new instance of the transform that the config is specifying.
-    pub async fn get_transforms(&self, topics: &TopicHolder) -> Result<Transforms> {
+    pub async fn get_transforms(
+        &self,
+        topics: &TopicHolder,
+        chain_name: String,
+    ) -> Result<Transforms> {
         match self {
             TransformsConfig::CassandraSinkSingle(c) => c.get_source().await,
             TransformsConfig::KafkaSink(k) => k.get_source().await,
@@ -265,7 +269,7 @@ impl TransformsConfig {
             TransformsConfig::DebugPrinter => Ok(Transforms::DebugPrinter(DebugPrinter::new())),
             TransformsConfig::Null => Ok(Transforms::Null(Null::default())),
             TransformsConfig::Loopback => Ok(Transforms::Loopback(Loopback::default())),
-            TransformsConfig::RedisSinkCluster(r) => r.get_source().await,
+            TransformsConfig::RedisSinkCluster(r) => r.get_source(chain_name).await,
             TransformsConfig::ParallelMap(s) => s.get_source(topics).await,
             //TransformsConfig::PoolConnections(s) => s.get_source(topics).await,
             TransformsConfig::Coalesce(s) => s.get_source().await,
@@ -282,7 +286,7 @@ pub async fn build_chain_from_config(
 ) -> Result<TransformChain> {
     let mut transforms: Vec<Transforms> = Vec::new();
     for tc in transform_configs {
-        transforms.push(tc.get_transforms(topics).await?)
+        transforms.push(tc.get_transforms(topics, name.clone()).await?)
     }
     Ok(TransformChain::new(transforms, name))
 }
--- a/shotover-proxy/src/transforms/query_counter.rs
+++ b/shotover-proxy/src/transforms/query_counter.rs
@@ -4,7 +4,7 @@ use crate::message::{ASTHolder, MessageDetails, QueryMessage};
 use crate::transforms::{Transform, Transforms, Wrapper};
 use anyhow::Result;
 use async_trait::async_trait;
-use metrics::counter;
+use metrics::{counter, register_counter, Unit};
 use serde::Deserialize;
 use sqlparser::ast::Statement;
 
@@ -16,6 +16,14 @@ pub struct QueryCounter {
 #[derive(Deserialize, Debug, Clone)]
 pub struct QueryCounterConfig {
     pub name: String,
+}
+
+impl QueryCounter {
+    pub fn new(counter_name: String) -> Self {
+        register_counter!("query_count", Unit::Count, "name" => counter_name.clone());
+
+        QueryCounter { counter_name }
+    }
 }
 
 #[async_trait]
@@ -72,8 +80,8 @@ impl Transform for QueryCounter {
 
 impl QueryCounterConfig {
     pub async fn get_source(&self) -> Result<Transforms> {
-        Ok(Transforms::QueryCounter(QueryCounter {
-            counter_name: self.name.clone(),
-        }))
+        Ok(Transforms::QueryCounter(QueryCounter::new(
+            self.name.clone(),
+        )))
     }
 }
--- a/shotover-proxy/src/transforms/redis/sink_cluster.rs
+++ b/shotover-proxy/src/transforms/redis/sink_cluster.rs
@@ -5,7 +5,7 @@ use async_trait::async_trait;
 use derivative::Derivative;
 use futures::stream::FuturesUnordered;
 use futures::{Future, StreamExt, TryFutureExt};
-use metrics::counter;
+use metrics::{counter, register_counter, Unit};
 use rand::prelude::SmallRng;
 use rand::SeedableRng;
 use redis_protocol::resp2::types::Frame;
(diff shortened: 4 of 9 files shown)


shotover-proxy/tests/runner/observability_int_tests.rs
```diff
@@ -1,17 +1,75 @@
...
+
+    println!("{}", body);
+
+    for s in expected {
+        assert!(body.contains(s));
+    }
+
+    // Check we still get the metrics after sending a couple requests
+
     redis::cmd("SET")
```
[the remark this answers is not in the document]
conorbros: We can't register the `"shotover_chain_latency"` histogram metric because there will be a different label for every client that connects. I can:

- register it without the client_details label
- register it with an empty client_details, but that means that there will be a useless group of metrics with `client_details="" ` remaining in the metrics.
- not register it at all 

I think option 1 is the best.

shotover-proxy/src/server.rs
```diff
@@ -89,6 +89,9 @@ impl<C: Codec + 'static> TcpCodecListener<C> {
         shutdown_complete_tx: mpsc::Sender<()>,
         tls: Option<TlsAcceptor>,
     ) -> Self {
+        register_gauge!("shotover_available_connections", Unit::Count, "source" => source_name.clone());
+        gauge!("shotover_available_connections", limit_connections.available_permits() as f64,"source" => source_name.clone());
```
rukai, line 93: missing space after f64

shotover-proxy/tests/runner/observability_int_tests.rs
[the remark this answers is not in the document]
rukai: yes, lets remove client_details, as long thats there preregistering shotover_chain_latency just adds noise.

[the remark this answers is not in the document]
conorbros: To be clear, you mean just remove it from the registration? We can't remove it entirely.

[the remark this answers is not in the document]
rukai: Yeah, I assume we need to remove it for the register to be useful.

> We can't remove it entirely.

why? is it important?

If it is needed then lets just not register it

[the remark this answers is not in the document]
rukai: ok lets remove the register then

shotover-proxy/src/transforms/query_counter.rs
```diff
@@ -18,6 +18,14 @@ pub struct QueryCounterConfig {
     pub name: String,
 }
 
+impl QueryCounter {
+    pub fn new(counter_name: String) -> Self {
+        register_counter!("query_count", Unit::Count, "name" => counter_name.clone());
```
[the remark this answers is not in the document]
conorbros: Raised an issue regarding QueryCounter #335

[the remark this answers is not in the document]
rukai: We can either:
* remove the QueryCounter changes from this PR
* add a note to the issue to add a test case for QueryCounter in test_metrics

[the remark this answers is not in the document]
conorbros: Added a note to create an int test for QueryCounter

shotover-proxy/tests/runner/observability_int_tests.rs
```diff
@@ -1,17 +1,77 @@
...
+        r#"shotover_chain_latency_sum{chain="redis_chain"}"#,
+        r#"shotover_chain_latency_count{chain="redis_chain"}"#,
+    ];
+
+    // check we get the metrics on startup
+    let mut body = http_request_metrics().await;
+
+    let lines = body.lines().collect::<Vec<&str>>().len();
+
+    for s in &expected {
```
rukai, line 68: should we also do that length check here?

Maybe pull the length check and element check into a helper function?

conorbros: I moved the element check. Couldn't find an obvious way to split out the length check, might become more clear once we add more metrics tests.

shotover-proxy/tests/runner/observability_int_tests.rs
```diff
@@ -26,16 +90,31 @@ async fn test_metrics() {
...
+    redis::cmd("GET")
+        .arg("the_key")
+        .query_async::<_, ()>(&mut connection)
+        .await
+        .unwrap();
+
+    std::thread::sleep(time::Duration::from_secs(1));
+    body = http_request_metrics().await;
+
+    let new_lines = vec![
```
rukai, line 102: Please add a TODO comment explaining that we dont actually assert on the contents of these yet, due to the randomized order of the labels but a proper solution needs to be investigated.
As is, this is very misleading!

shotover-proxy/tests/runner/observability_int_tests.rs
```diff
@@ -26,16 +90,34 @@ async fn test_metrics() {
...
-    assert!(body.contains("# TYPE shotover_available_connections gauge"));
-    assert!(body.contains("# TYPE shotover_transform_latency summary"));
-    assert!(body.contains("# TYPE shotover_chain_latency summary"));
+    redis::cmd("GET")
+        .arg("the_key")
+        .query_async::<_, ()>(&mut connection)
+        .await
+        .unwrap();
+
+    std::thread::sleep(time::Duration::from_secs(1));
```
rukai, line 99: Oh did we end up actually needing this sleep?

conorbros: I guess we could loop now that we are counting the lines?

rukai: nah, sleep is fine for now if we need it.
I just thought the issue turned out to not be a race condition.

rukai (changes requested):
LGTM

rukai (approved):

shotover-proxy/src/transforms/mod.rs
```diff
@@ -265,7 +269,7 @@ impl TransformsConfig {
             TransformsConfig::DebugPrinter => Ok(Transforms::DebugPrinter(DebugPrinter::new())),
             TransformsConfig::Null => Ok(Transforms::Null(Null::default())),
             TransformsConfig::Loopback => Ok(Transforms::Loopback(Loopback::default())),
-            TransformsConfig::RedisSinkCluster(r) => r.get_source().await,
+            TransformsConfig::RedisSinkCluster(r) => r.get_source(chain_name).await,
```
benbromhead, line 272: Why do we pass in the transform name on creation this one when we do it differently on runtime (which all other transforms then use)?

Transforms dynamically get their chain name here from `src/transforms/mod.rs`:


```rust
pub async fn call_next_transform(mut self) -> ChainResponse {
        if self.transforms.is_empty() {
            panic!("The transform chain does not end with a terminating transform. If you want to throw the messages away use a Null transform, otherwise use a terminating sink transform to send the messages somewhere.");
        }
        let transform = self.transforms.remove(0);

        let transform_name = transform.get_name();
        let chain_name = self.chain_name.clone();

        let start = Instant::now();
        let result = CONTEXT_CHAIN_NAME
            .scope(chain_name, transform.transform(self))
            .await;
        counter!("shotover_transform_total", 1, "transform" => transform_name);
        if result.is_err() {
            counter!("shotover_transform_failures", 1, "transform" => transform_name)
        }
        histogram!("shotover_transform_latency", start.elapsed(),  "transform" => transform_name);
        result
    }
```

It might actually be better to do do all the name / metrics setup by wrapping the chain setup function here:    

```rust
async fn _prep_transform_chain(&mut self, t: &mut TransformChain) -> Result<()> {}
```

Then actually calling that whenever we setup a transform chain (which will happen on startup).

shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -97,6 +83,36 @@ pub struct RedisSinkCluster {
 }
 
 impl RedisSinkCluster {
+    pub fn new(
+        first_contact_points: Vec<String>,
+        connection_count: usize,
+        tls: Option<TlsConfig>,
+        chain_name: String,
+    ) -> Result<Self> {
+        register_counter!("redis_cluster_failed_request", Unit::Count, "chain" => chain_name);
```
XA21X, line 92: Just checking that we want two ways of passing the chain name into this transform? (Task local variable used to get the name for incrementing this counter.) Was there a blocker using the existing method? Or do we want to get rid of the task local variable in the future?

XA21X (approved):

shotover-proxy/src/transforms/redis/sink_cluster.rs
benbromhead, line 92: We should seperate this out to use prep_transform_chain

shotover-proxy/src/transforms/mod.rs
benbromhead: `async fn _prep_transform_chain(&mut self, t: &mut TransformChain) -> Result<()> {}`
 
Could then also use the same chain name injection as above?

It is fairly "magic" and non-explicit... but we should do it all the same way, rather than 2 different ways.

shotover-proxy/src/transforms/redis/sink_cluster.rs
benbromhead: new only gets called on shotover startup. Whereas prep_transform_chain (at least used to) get called for each new one iirc...

benbromhead (changes requested):
We need to address the chain name mechanism before merging this. See above comments

shotover-proxy/src/transforms/redis/sink_cluster.rs
XA21X: Prep_transform_chain isn't used anymore. Do we want to revive it?

shotover-proxy/src/transforms/mod.rs
rukai: Good thinking, I agree.
I'm thinking we should refactor all transform name accesses to come from CONTEXT_CHAIN_NAME in https://github.com/shotover/shotover-proxy/pull/336
Then come back to this PR once thats merged.

rukai: oh wait, CONTEXT_CHAIN_NAME probably has the name of the chain not the transform...

rukai: ah, so you are suggesting to create a CONTEXT_TRANSFORM_NAME?

rukai: Oh I see whats happened. You are talking about chain names but accidentally said "transform name".

rukai: I actually dont think we want to use CONTEXT_CHAIN_NAME here.
CONTEXT_CHAIN_NAME was a hack because we didnt want to pass the chain name really deep into the callstack.

And I think trying trying to use it here would cause problems!
Initializing chains is recursive and CONTEXT_CHAIN_NAME doesnt function like a stack so a recursive chain setup would leave the wrong value in CONTEXT_CHAIN_NAME when it finishes.

shotover-proxy/src/transforms/redis/sink_cluster.rs
rukai: I would like to avoid it if possible.
I would want a pretty good reason for having any initialization outside of new.

shotover-proxy/src/transforms/mod.rs
benbromhead: in that case let's create a ticket to figure out "getting names into things properly and with (hopefully) one way of doing it" and I'm happy for this to get merged

shotover-proxy/src/transforms/redis/sink_cluster.rs
benbromhead: the problem is chains that get lazily initialized from a clone of an existing base chain (e.g. load_balance.rs). They will all have the same name (this happens now anyway).

Again happy to get this merged as we are not going to head down the CONTEXT_CHAIN_NAME path... and address this in a seperate ticket as this doesn't break anything, but rather warrents a discussion about how we do things.

benbromhead (approved):

shotover-proxy/src/transforms/redis/sink_cluster.rs
rukai: Ill make an issue

shotover-proxy/src/transforms/mod.rs
conorbros: I originally attempted to use CONTEXT_CHAIN_NAME but it's only available when you're running a chain, not on initialisation.

PR #331, opened 2021-10-29
Cassandra protocol version negotiation

Claudenw:
Changes to support protocol version negotiation


.github/workflows/test.yml
```diff
@@ -52,6 +52,13 @@ jobs:
       run: |
         sudo apt-get update
         sudo apt-get install -y libpcap-dev
+    - name: Install Cassandra CPP driver
+      run: |
+        wget https://downloads.datastax.com/cpp-driver//ubuntu/18.04/cassandra/v2.16.0/cassandra-cpp-driver_2.16.0-1_amd64.deb
```
conorbros, line 57: We can use the following to background them all and wait for completion instead of doing it sequentially:

```bash
wget https://downloads.datastax.com/cpp-driver//ubuntu/18.04/cassandra/v2.16.0/cassandra-cpp-driver_2.16.0-1_amd64.deb & 
wget https://downloads.datastax.com/cpp-driver/ubuntu/18.04/cassandra/v2.16.0/cassandra-cpp-driver-dev_2.16.0-1_amd64.deb & 
wget https://downloads.datastax.com/cpp-driver/ubuntu/18.04/dependencies/libuv/v1.35.0/libuv1_1.35.0-1_amd64.deb & 
wget https://downloads.datastax.com/cpp-driver/ubuntu/18.04/dependencies/libuv/v1.35.0/libuv1-dev_1.35.0-1_amd64.deb &
wait
```

rukai (changes requested):
Some minor issues, ill investigate the actual solution used tomorrow

shotover-proxy/tests/test_examples.rs
```diff
@@ -27,6 +27,8 @@ async fn test_cassandra_standalone() {
 #[tokio::test(flavor = "multi_thread")]
 #[serial]
 async fn test_cassandra_cluster() {
+    let start = std::time::Instant::now();
```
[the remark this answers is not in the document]
Claude-at-Instaclustr: removed

test-helpers/src/docker_compose.rs
```diff
@@ -98,22 +103,66 @@ impl DockerCompose {
...
+    ///
+    /// # Panics
+    /// * If `count` occurances of `log_text` is not found in the log within `time` seconds.
+    ///
+    /// # Example
+    /// ```
+    /// let _compose = DockerCompose::new("examples/redis-passthrough/docker-compose.yml")
+    ///         .wait_for("Ready to accept connections", 3, 65);
+    /// ```
+    pub fn wait_for_n_t(self, log_text: &str, count: usize, time: u64) -> Self {
```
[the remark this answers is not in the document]
Claude-at-Instaclustr: We do need this.  I have a test that ran 120 seconds, thought if it doesn't complete in 90 it probably won't complete in 120.  Since we have the code and the "magic" number of 90 is specified in a call to this method let it stand.

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+impl CassandraTestContext {
+    pub fn new() -> Self {
+        CassandraTestContext::new_with_points("127.0.0.1")
+    }
+
+    pub fn new_with_points(contact_points: &str) -> Self {
+        CassandraTestContext::new_with_points_and_port(contact_points, 9042)
+    }
+
+    pub fn new_with_points_and_port(contact_points: &str, port: u16) -> Self {
```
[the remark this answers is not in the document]
Claude-at-Instaclustr: Why does the ShotoverManager (a generic thing) know about the specifics of Redis or Cassandra (specific thing)?  Everytime a new protocol is added ShotoverManager gets extended to handle them?  This seems like a design error to me.

benbromhead:
Sorted CLA issue

shotover-proxy/tests/test_examples.rs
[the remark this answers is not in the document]
rukai: ... its still here?

test-helpers/src/docker_compose.rs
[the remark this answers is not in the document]
rukai: "Since we have the code" is not good reasoning.
Feel free to keep it around in a local branch for future reference if you think we might need it later.
But I do not want needless extra complexity in main

shotover-proxy/src/server.rs
```diff
@@ -26,9 +28,115 @@ impl<T: Decoder<Item = Messages, Error = anyhow::Error> + Clone + Send> CodecRea
...
+                            protocol_error: 1,
+                        }
+                    }
+                }
+            } else {
+                m
+            }
+        })
+        .filter(|m| m.protocol_error == 0)
+        .for_each(|m| result.push(m));
```
rukai, line 127: We should use .collect() on the iterator instead of manually pushing to result.

shotover-proxy/src/protocols/cassandra_protocol2.rs
```diff
@@ -649,8 +679,17 @@ fn get_cassandra_frame(rf: RawFrame) -> Result<Frame> {
     }
 }
 
+impl CodecErrorFixup for CassandraCodec2 {
+    fn fixup_err(
+        &self,
+        message: Message,
+    ) -> (Option<Message>, Option<Message>, Option<anyhow::Error>) {
+        (Some(message), None, None)
```
rukai, line 687: Is the idea that you will put the error handling in here but you just havent done so yet?

Claude-at-Instaclustr: No, that is the error handling.  When an error message is detected by the Cassandra processing it passes it back to the originator.  

The result of the `fixup_error()` method is tuple consisting of:

1. A message to go back to the originator of the message.
2. A message to continue down the chain.
3. An string written to the log.

In the Cassandra case the protocol error is returned to the client so that we can negotiate the protocol version correctly.

In the Redis case it simply logs an error that says there should be no Redis protocol errors.

shotover-proxy/examples/cassandra-cluster/topology1.yaml
```diff
@@ -0,0 +1,24 @@
...
+        test.simple:
+          - pk
+        test.clustering:
+          - pk
+          - clustering
+chain_config:
+  main_chain:
+    #    - Printer
+    - CassandraSinkSingle:
+        bypass_result_processing: true
```
rukai, line 19: we flipped this btw.
So now you want: result_processing: false

.github/workflows/test.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         sudo apt-get install -y libpcap-dev
         sudo apt-get upgrade -y
     - name: Install Cassandra CPP driver
-      if: ${{ matrix.name == 'Ubuntu 18.04 - Debug' }}
+      if: ${{ startsWith( matrix.name, 'Ubuntu 18.04') }}
```
rukai, line 57: you could check matrix.runner directly.

shotover-proxy/src/protocols/cassandra_protocol2.rs
```diff
@@ -554,23 +559,20 @@ impl CassandraCodec2 {
...
+            self.current_head = h;
+            r
+        })
     }
 
     fn encode_raw(&mut self, item: Frame, dst: &mut BytesMut) {
         let buffer = item.into_cbytes();
         if buffer.is_empty() {
-            info!("trying to send 0 length frame");
+            info!("{:?} trying to send 0 length frame", thread::current().id());
```
rukai, line 575: Im not really sure what all the `thread::current().id()` changes are for.
Maybe move it into a separate PR and make a case for why we need to know the opaque thread id of the thread that each log came from.

shotover-proxy/src/protocols/cassandra_protocol2.rs
```diff
@@ -584,12 +586,53 @@ impl Decoder for CassandraCodec2 {
...
+            Ok(Some(frame)) => {
+                debug!("{:?} Decoded {:?}", thread::current().id(), &frame);
+                Ok(Some(self.process_cassandra_frame(frame)))
+            }
             Ok(None) => Ok(None),
             Err(e) => {
-                warn!("decode raw error {:?}", e);
-                Err(e)
+                self.last_error = Some(e.message.as_plain());
+                debug!("{:?} CDRSError {:?}", thread::current().id(), &e);
```
rukai, line 608: cdrs? We arent even using cdrs.

.github/workflows/test.yml
```diff
@@ -63,8 +76,10 @@ jobs:
...
     - name: Ensure that tests pass
       run: cargo test ${{ matrix.cargo_profile }} -- --include-ignored --show-output
+      # can not run tests if cpp driver not installed
+      if: ${{ startsWith( matrix.name, 'Ubuntu 18.04') }}
     - name: License Check
       run: |
         cargo install --locked cargo-deny
         cargo deny check licenses
-      if: ${{ matrix.name == 'Ubuntu 20.04 - Debug' }}
+
```
rukai, line 85: We only want the license check to run on one job.
It takes a long time to build `cargo-deny` and also uses up a lot of cache.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -0,0 +1,123 @@
...
+            .unwrap()
+            .get_column(0)
+            .unwrap()
+            .get_str()
+            .ok(),
+        Some("cycling")
+    );
+}
+
+#[test]
```
rukai, line 50: Because we are using global resources like network ports we should add `#[serial]`

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -0,0 +1,123 @@
...
+fn test_basic_connection() -> Result<()> {
```
rukai, line 51: No need to return `Result<()>` here. Failures are indicated in tests by panicking.

rukai:
Im getting a hang on my local machine too.

shotover-proxy/src/server.rs
```diff
@@ -26,9 +28,114 @@ impl<T: Decoder<Item = Messages, Error = anyhow::Error> + Clone + Send> CodecRea
...
+            // One thorough the mapping if the message has a protocol_error is dropped.
+            if m.protocol_error != 0 {
+                info!(
+                    "{:?} processing protocol error: {:?}",
+                    thread::current().id(),
+                    &m
+                );
+                let (up_msg, down_msg, an_err) = codec.fixup_err(m);
+                // if there is a message for upstream send it
+                if up_msg.is_some() {
```
rukai, line 83: `if let Some(up_msg) = up_msg {`

shotover-proxy/src/server.rs
```diff
@@ -26,9 +28,114 @@ impl<T: Decoder<Item = Messages, Error = anyhow::Error> + Clone + Send> CodecRea
...
+) -> Messages {
+    // this code creates a new Vec and uses an iterator with mapping and filtering to
+    // populate it from the original Messages.message Vec.  It may be more efficient to scan the
+    // original Vec and replace or delete individual Message in place.
+    let result = messages
+        .into_iter()
+        .map(|m| {
+            // if there is a protocol error handle it otherwise return the original message.
+            // One thorough the mapping if the message has a protocol_error is dropped.
+            if m.protocol_error != 0 {
```
rukai, line 75: we seem to be just using protocol_error as a boolean, maybe it should be a boolean instead?

shotover-proxy/src/server.rs
```diff
@@ -320,21 +425,38 @@ fn spawn_read_write_tasks<
...
-                    if let Err(error) = in_tx.send(message) {
-                        warn!("failed to send message: {}", error);
-                        return;
+                    let filtered_messages = handle_protocol_error(&codec, message, &out_tx);
+                    debug!(
+                        "{:?} filtered_messages: {:?}",
+                        thread::current().id(),
+                        filtered_messages
+                    );
+                    if filtered_messages.is_empty() {
```
rukai, line 443: Only send messages when there are no messages to send?
That sounds wrong.

.github/workflows/test.yml
Claude-at-Instaclustr: done

shotover-proxy/src/server.rs
Claude-at-Instaclustr: fixed

shotover-proxy/src/server.rs
Claude-at-Instaclustr: protocol_error is intended to be non-zero on error, and the value is intended to be used by the codec to determine what to do.  For Cassandra for example there are several error values (including 0) so the error values is `0x10000 | e.error_code`, later the codec strips off the `0x10000` to determine what the actual error code is.

shotover-proxy/src/server.rs
Claude-at-Instaclustr: What is being noted here?

.github/workflows/test.yml
Claude-at-Instaclustr: Since you dependencies change from version to version, don't you need to run for at least each Ubuntu version?

shotover-proxy/src/protocols/cassandra_protocol2.rs
Claude-at-Instaclustr: I required them to debug the multi-threaded beast.  Without them you can't tell which thread is doing what as all the log messages are jumbled together.  There may be a better way to do this but this worked for me and allows me to tell, with reasonable assurance, what failed and why.

shotover-proxy/src/server.rs
conorbros: We should use `if let Some(up_msg) = up_msg` instead of `if up_msg.is_some()`

shotover-proxy/src/server.rs
rukai: Can you point out where the 0x10000 gets stripped and the error code used? I cant find it all.

benbromhead:
Folks - FYI I won't be able to spend some serious time on this PR till next week, but don't let that stop other reviewers

shotover-proxy/src/protocols/cassandra_protocol2.rs
rukai: There is a proper fix coming in https://github.com/shotover/shotover-proxy/pull/363 so lets remove those changes.

rukai:
test_basic_connection has the same failure locally as on CI.
@Claude-at-Instaclustr does it pass on your local machine?

shotover-proxy/src/server.rs
Claudenw: arrrg... I was recalling an earlier implementation the cassandra code just sets the error to the 0x10000 | error_code.  Since the value is not zero it is detected as an error.  I suspect that there are some edge cases where we will want to do some special processing particularly where the error is a "system" error (error code 0).

test-helpers/src/docker_compose.rs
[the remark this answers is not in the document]
Claude-at-Instaclustr: Recent tests have shown that we need this functionality to perform some problem determination tests.

shotover-proxy/src/server.rs
rukai: The assumption that every protocol has an error that neatly fits into a i32 is not great.
I would rather start with the simplest possible type for this (a bool) then we can easily refactor to use an i32 or enum or whatever if we deem it necessary in the future.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -0,0 +1,106 @@
...
+    test_create_keyspace(CassandraTestContext::new());
+}
+
+// this test is used for dev testing to ensure that the driver works we directly connect vi the
+// test context
+// with the Cassandra we are running against.
+//#[test] // can not use ignore on test as build builds ignored tests
+//#[serial]
+#[allow(dead_code)] // to make clippy happy
+fn test_create_keyspace_direct() {
```
rukai, line 73: we can very easily modify test_basic_connection if we need to quickly check if the driver can connect directly to cassandra, so no need to leave this dead code here.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    test_create_keyspace(CassandraTestContext::new_with_points("10.5.0.2"));
+}
+
+// this test is used for dev testing to ensure that  the cpp driver we are using actually works
+// with the Cassandra we are running against.
+//#[test] // can not use ignore on test as build builds ignored tests
+//#[serial]
+#[allow(dead_code)] // to make clippy happy
+fn test_cpp_driver() {
```
rukai, line 94: we can very easily modify test_basic_connection if we need to quickly check if the driver can connect directly to cassandra, so no need to leave this dead code here.

shotover-proxy/tests/scripts/cassandra-cluster.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
rukai, line 1: I dont think we need this.
Running `docker-compose -f docker-compose.yml up` is already pretty simple.
I'm not sure what the defer is for, docker-compose already runs down when the up is killed.

shotover-proxy/tests/test_examples.rs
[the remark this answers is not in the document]
rukai: ... its still here?

shotover-proxy/config/config.yaml
```diff
@@ -1,3 +1,3 @@
 ---
-main_log_level: "info,shotover_proxy=info"
+main_log_level: "info"
```
rukai, line 2: I think the reason this is like this is like this is to demonstrate how to set separate log levels for shotover and everything else.
I think what we should do is add a comment above explaining how to configure the log level.
But that should go in a separate PR, so for now just revert this change.

.github/workflows/test.yml
```diff
@@ -52,6 +52,19 @@ jobs:
       run: |
         sudo apt-get update
         sudo apt-get install -y libpcap-dev
+        sudo apt-get upgrade -y
+    - name: Install Cassandra CPP driver
+      if: ${{ matrix.runner == 'ubuntu-18.04' }}
+      run: |
+        echo CPP version number
+        cc --version
```
conorbros, line 60: Do we want to leave this logging stuff in the PR or is it just leftovers?

shotover-proxy/tests/scripts/cassandra-cluster-shot.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
```
rukai, line 1: This sounds like a pretty reasonable addition but I think it should live in `examples/cassandra-cluster` as it will be easier to discover there.

shotover-proxy/examples/cassandra-cluster/topology1.yaml
```diff
@@ -0,0 +1,24 @@
...
+        system.local:
+          - key
+        test.simple:
+          - pk
+        test.clustering:
+          - pk
+          - clustering
+chain_config:
+  main_chain:
+    #    - Printer
```
conorbros, line 17: Commented transform

shotover-proxy/src/protocols/cassandra_protocol2.rs
```diff
@@ -867,4 +933,47 @@ mod cassandra_protocol_tests {
         }];
         test_frame_codec_roundtrip(&mut codec, &bytes, messages);
     }
+
+    #[test]
+    fn test_process_cassandra_frame() {
+        // Frame { version: Request, flags: [], opcode: Options, stream: 0, body: [], tracing_id: None, warnings: [] }
```
conorbros, line 939: Comments

shotover-proxy/tests/redis_int_tests/basic_driver_tests.rs
```diff
@@ -764,7 +764,6 @@ async fn test_auth(connection: &mut Connection) {
             .code(),
         Some("NOAUTH")
     );
-
```
rukai, line 767: ??

shotover-proxy/src/server.rs
```diff
@@ -27,9 +29,114 @@ impl<T: Decoder<Item = Messages, Error = anyhow::Error> + Clone + Send> CodecRea
 pub trait CodecWriteHalf: Encoder<Messages, Error = anyhow::Error> + Clone + Send {}
 impl<T: Encoder<Messages, Error = anyhow::Error> + Clone + Send> CodecWriteHalf for T {}
 
+pub trait CodecErrorFixup {
+    /// An method to fix up an error.
```
conorbros, line 33: What does "fix up an error" mean?

shotover-proxy/src/server.rs
```diff
@@ -27,9 +29,114 @@ impl<T: Decoder<Item = Messages, Error = anyhow::Error> + Clone + Send> CodecRea
...
+    ///
+    /// Codecs that do not have additional error handling should return `(None, None, Optional<Err>)`
```
conorbros, line 35: I think an `enum` would be preferable to a combination of `Option`.

shotover-proxy/src/server.rs
```diff
@@ -27,9 +29,114 @@ impl<T: Decoder<Item = Messages, Error = anyhow::Error> + Clone + Send> CodecRea
...
+                    thread::current().id(),
+                    &m
+                );
+                let (up_msg, down_msg, an_err) = codec.fixup_err(m);
+                // if there is a message for upstream send it
+                if let Some(up_msg) = up_msg {
+                    // send up stream messages now
+                    debug!("{:?} Return message: {:?}", thread::current().id(), &up_msg);
+                    tx_out.send(vec![up_msg]).ok();
+                }
```
conorbros, line 88: Making the return from `fixup_err` an `enum` would probably make some of this code a bit clearer.

.github/workflows/test.yml
```diff
@@ -52,6 +52,19 @@ jobs:
...
+        cat /etc/lsb-release
+        wget https://downloads.datastax.com/cpp-driver//ubuntu/18.04/cassandra/v2.16.0/cassandra-cpp-driver_2.16.0-1_amd64.deb &
+        wget https://downloads.datastax.com/cpp-driver/ubuntu/18.04/cassandra/v2.16.0/cassandra-cpp-driver-dev_2.16.0-1_amd64.deb &
+        wget https://downloads.datastax.com/cpp-driver/ubuntu/18.04/dependencies/libuv/v1.35.0/libuv1_1.35.0-1_amd64.deb &
+        wget https://downloads.datastax.com/cpp-driver/ubuntu/18.04/dependencies/libuv/v1.35.0/libuv1-dev_1.35.0-1_amd64.deb &
+        wait
+        sudo apt -y install ./cassandra-cpp-driver_2.16.0-1_amd64.deb ./cassandra-cpp-driver-dev_2.16.0-1_amd64.deb ./libuv1_1.35.0-1_amd64.deb ./libuv1-dev_1.35.0-1_amd64.deb
```
conorbros, line 67: Can we add to the Contributing docs that you need to install `cassandra-cpp-driver`?

shotover-proxy/src/server.rs
```diff
@@ -190,6 +303,9 @@ impl<C: Codec + 'static> TcpCodecListener<C> {
                 conn_details: conn_string,
                 source_details: self.source_name.clone(),
 
+                // Initialize the connection state. This allocates read/write
```
rukai, line 306: what does this mean?

conorbros (changes requested):
Nicely done! Everything works on my machine. Some feedback:

test-helpers/src/docker_compose.rs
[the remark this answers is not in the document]
rukai: I dont see why we couldnt just change the constant here?

.github/workflows/test.yml
conorbros: And to `install_ubuntu_packages.sh` when #368 gets merged

test-helpers/src/docker_compose.rs
```diff
@@ -141,6 +170,10 @@ impl Drop for DockerCompose {
                     "ERROR: docker compose failed to bring down while already panicking: {:?}",
                     err
                 );
+                error!(
```
rukai, line 173: as explained in the comment this will do nothing due to being in a panic

.github/workflows/test.yml
```diff
@@ -52,6 +52,19 @@ jobs:
       run: |
         sudo apt-get update
         sudo apt-get install -y libpcap-dev
+        sudo apt-get upgrade -y
```
conorbros, line 55: sfsdf

shotover-proxy/Cargo.toml
```diff
@@ -52,7 +52,7 @@ halfbrown = "0.1.11"
 
 # Transform dependencies
 redis-protocol = "3.0.1"
-cassandra-proto = { git = "https://github.com/shotover/cassandra-proto", branch = "move-to-bytes", features = ["v4"] }
+cassandra-proto = { git = "https://github.com/shotover/cassandra-proto", branch = "OPTION_fix", features = ["v4"]}
```
rukai, line 55: I was able to get the test passing with https://github.com/shotover/cassandra-proto/pull/5 which only has the bare minimum of changes required.

image


cassandra-proto needs a lot of work but that should be done on the upstream repo, (or possibly a properly maintained fork if thats what it comes too)

But we cant just keep adding more changes onto bens poorly maintained fork.
So lets make the absolute minimum change to fix the version panic and then make the other cassandra-proto fixes once we land this shotover PR.

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+        cluster.set_port(port).ok();
+        cluster.set_load_balance_round_robin();
+        let mut session;
+        //let query = stmt!("SELECT keyspace_name FROM system_schema.keyspaces;");
+        let query = stmt!("SELECT release_version FROM system.local");
+
+        let attempts = 30;
+        let mut current_attempt = 0;
+
+        loop {
```
rukai, line 34: Now that we've sorted out all the cluster init problems I ran the test without the retry logic here and the test still passes.
So seems we can remove it.

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -0,0 +1,72 @@
+use cassandra_cpp::*;
+use std::thread;
+use std::thread::sleep;
+use std::time::Duration;
+use tracing::info;
+
+mod basic_driver_tests;
+
+pub struct CassandraTestContext {
```
rukai, line 9: The `CassandraTestContext` wrapper doesnt add anything over just a raw `Session`.
Lets remove it and return the `Session` directly.

shotover-proxy/tests/cassandra_int_tests/mod.rs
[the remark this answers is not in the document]
rukai: Yes, in my comment I did say "it might be worth moving it to be freestanding functions outside of ShotoverManager"

I think it makes sense to keep the test connection creators together, but dont really know what the best way to organize them would be, so for now and possibly permanently we can keep this here.

shotover-proxy/Cargo.toml
Claude-at-Instaclustr: I think that change significantly modifies the way the original cassandra_protocol2 works.  We should sit down and do a full review of cassandra_protocol2 so that we understand exactly how we want it to work.  I asked Alex but got no response.

shotover-proxy/tests/cassandra_int_tests/mod.rs
Claude-at-Instaclustr: Lets do that in a subsequent request.  This pull is already too big.

shotover-proxy/tests/scripts/cassandra-cluster.sh
Claude-at-Instaclustr: When you kill `docker-compose up` it does not always clean up everything, particularly the networks.  Running `docker-compose down` ensures that the networks are removes so that a subsequent `docker-compose up` will work.

.github/workflows/test.yml
conorbros: That's reasonable. Though I doubt the `cc` version will change on the Ubuntu images we are building with.

.github/workflows/test.yml
Claude-at-Instaclustr: Contributing docs updated.

shotover-proxy/tests/cassandra_int_tests/mod.rs
Claude-at-Instaclustr: Lets do this on another ticket plz

test-helpers/src/docker_compose.rs
Claude-at-Instaclustr: It won't write the message to the log?

shotover-proxy/src/server.rs
Claude-at-Instaclustr: It returns 3 values.  One to go up stream, one to go down stream, and one to be logged.  I don't see how you would do that with an enum.  see `codec.fixup_err` for documentation.

shotover-proxy/src/server.rs
Claude-at-Instaclustr: How?  You have to handle 3 simultaneous possibilities.  You may have to send a message back to the originator as well as a message to the destination to maintain the connection.  And you may want to log something in the server.  That is the goal here, to support all 3 simultaneously.

shotover-proxy/tests/cassandra_int_tests/mod.rs
rukai: Doing so would reduce the size of this PR.

shotover-proxy/src/server.rs
Claude-at-Instaclustr: The parser discovers an error and reports it.  However, some parser errors are part of the larger protocol results.  In Cassandra if the wrong protocol version is requested a protocol error has to be sent to the originator and not on to the destination.  The `fixup_err` method provides a mechanism for the codec to catch those messages and route them properly.

shotover-proxy/config/config.yaml
Claude-at-Instaclustr: done

shotover-proxy/tests/test_examples.rs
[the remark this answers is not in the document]
Claude-at-Instaclustr: done

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
Claude-at-Instaclustr: removed

test-helpers/src/docker_compose.rs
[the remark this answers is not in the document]
Claude-at-Instaclustr: So make our life easier and leave it in.  We have already shown that it is useful.  Why remove it?

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -0,0 +1,72 @@
+use cassandra_cpp::*;
```
conorbros, line 1: What are we importing from `cassandra_cpp`? I think being explicit would be better.

shotover-proxy/src/server.rs
conorbros: Why does ' down_msg` exist if we aren't using it?

shotover-proxy/src/server.rs
conorbros: Ah I see. Is there a scenario where we need to send a message upstream and downstream as part of error handling?

test-helpers/src/docker_compose.rs
rukai: Correct

shotover-proxy/src/protocols/cassandra_protocol2.rs
rukai: Looks fine to me
image

The reason we need to remove it is because:
* Its adding a lot of noise to the PR and making it harder to review the real changes
* Its adding misleading information to the logs, thread id is an incorrect abstraction in an async world.

test-helpers/src/docker_compose.rs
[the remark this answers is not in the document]
rukai: Ok, there is value in being able to set an appropriate timeout for each type of docker setup we have to deal with.

In order to actually get value out of it please:
*  change the default value from 90 to 30.
        - The only reason it was originally bumped from 30 to 90 was to support cassandra.
        - 30 is plenty for all our redis setups.
* The cassandra test is currently on 300 which is excessive, 90 should be enough.

shotover-proxy/Cargo.toml
rukai: I'm not sure what "that change" refers to:

* the screenshot
* cassandra-proto/move-to-bytes-version-fix
* cassandra-proto/OPTION_fix

?

shotover-proxy/tests/cassandra_int_tests/mod.rs
rukai: oh right, redis connections do have a system to handle shotover port not being ready yet.
This equivalent of a retry used by redis has a few advantages:
* its simpler
* its reusable across different protocols
* it only fails if the port is closed. This is important because if shotover can be in a state where the port is open but the connection fails then the test should fail!

So we should be using this functionality instead of a retry.

Using it just requires the following change + fixing other uses of wait_for_socket_to_open
image

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -0,0 +1,65 @@
...
+    assert_eq!(result.row_count(), 1);
+    assert_eq!(
+        result
+            .first_row()
+            .unwrap()
+            .get_column(0)
+            .unwrap()
+            .get_str()
+            .ok(),
+        Some("3.11.10")
```
rukai, line 31: Currently the error is thrown away and not displayed when a failure occurs.
Instead we should write like this so we get a panic that displays the error.
```rust
    assert_eq!(
        result
            .first_row()
            .unwrap()
            .get_column(0)
            .unwrap()
            .get_str()
            .unwrap(),
        "cycling"
    );
 ```

File: shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs [hunk 38]
```diff
@@ -0,0 +1,65 @@
...
+    );
+
+    query = stmt!("SELECT keyspace_name FROM system_schema.keyspaces;");
+    result = ctx.session.execute(&query).wait().unwrap();
+    debug!("{:?} query result {:?}", thread::current().id(), result);
+    assert_eq!(result.row_count(), 6);
+    assert_eq!(
```
rukai, line 38: Currently the error is thrown away and not displayed when a failure occurs.
Instead we should write like this so we get a panic that displays the error.
```rust
    assert_eq!(
        result
            .first_row()
            .unwrap()
            .get_column(0)
            .unwrap()
            .get_str()
            .unwrap(),
        "cycling"
    );
 ```

File: shotover-proxy/tests/scripts/cassandra-cluster-shot.sh (hunk 22 above)
Comment (Claude-at-Instaclustr) [reply shotover-proxy/tests/scripts/cassandra-cluster-shot.sh:1]: As this is not a simple case of moving the script (it has to be edited and tested), lets do this under a separate pull request.

File: shotover-proxy/src/server.rs (hunk 15 above)
Comment (Claude-at-Instaclustr) [reply shotover-proxy/src/server.rs:76]: The design is not that every protocol error fits neatly into an i32.  The design is that every protocol error can be encoded by the codec into an i32 that the codec can later decode and handle properly.

File: shotover-proxy/tests/redis_int_tests/basic_driver_tests.rs (hunk 25 above)
Comment (Claude-at-Instaclustr) [reply shotover-proxy/tests/redis_int_tests/basic_driver_tests.rs]: fixed

File: shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs (hunk 17 above)
Comment (Claude-at-Instaclustr) [reply shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs]: fixed

File: test-helpers/src/docker_compose.rs (hunk 3 above)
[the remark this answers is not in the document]
Comment (Claude-at-Instaclustr) [reply test-helpers/src/docker_compose.rs:119]: fixed

File: test-helpers/src/docker_compose.rs [hunk 39]
```diff
@@ -97,22 +97,51 @@ impl DockerCompose {
     /// * `count` - The number of times the regular expression should be found.
     ///
     /// # Panics
-    /// * If `count` occurrences of `log_text` is not found in the log within 60 seconds.
+    /// * If `count` occurances of `log_text` is not found in the log within 90 seconds.
```
conorbros, line 100: Spelling of "occurrences" here

shotover-proxy/tests/cassandra_int_tests/mod.rs
conorbros: Has this been addressed in a commit you haven't pushed yet?

shotover-proxy/Cargo.toml
Claude-at-Instaclustr: I think that if you want to change the implementation of cassandra-proto that the code uses you need to get your branch into the shotover repository.  I think we should only use "official" repositories ("AlexPikalov" in this case) or the shotover repository.

So I think the proper steps for this change are:

1. copy your branch to shotover/cassandra-proto
2. create a change ticket to change the version in use in this code.

But I think this change can stand as it is.

shotover-proxy/tests/scripts/cassandra-cluster.sh
rukai: Oh I see!
Fair enough

shotover-proxy/tests/scripts/cassandra-cluster-shot.sh
rukai: Given that we are now keeping cassandra-cluster.sh it makes sense for this to be separate to cassandra-cluster.sh and not run `docker-compose -f docker-compose.yml up`.

As for moving it, its not super important and it can stay where it is.

shotover-proxy/tests/cassandra_int_tests/mod.rs
Claude-at-Instaclustr: removed

test-helpers/src/docker_compose.rs
Claude-at-Instaclustr: fixed

shotover-proxy/src/server.rs
Claude-at-Instaclustr: yes

shotover-proxy/src/server.rs
Claude-at-Instaclustr: Do you not see any cases where the protocol error might be addressed by the codec or where the protocol error should be sent down the stream?

It appears that you and  @rukai  are not convinced that we will need to process protocol errors and send messages in multiple directions.  My experience with other systems shows otherwise.  However, if you both agree then I will do the following:

1. convert the protocol_err to a boolean flag.
2. change fixup so that it returns either
    * a message to be returned to the client
    * nothing (drops the message processing all together)

rukai: 1. yes please, it will be easy to change the type in the future if we need more information from the error.
2. While this is how I would approach the problem, I certainly lack the years of experience that you have!
I would like to hear @benbromhead thoughts on if this abstraction is appropriate.

test-helpers/src/docker_compose.rs
```diff
@@ -97,22 +97,51 @@ impl DockerCompose {
     /// * `count` - The number of times the regular expression should be found.
     ///
     /// # Panics
-    /// * If `count` occurrences of `log_text` is not found in the log within 60 seconds.
+    /// * If `count` occurrences of `log_text` is not found in the log within 90 seconds.
```
rukai, line 100: 90 -> 30

.github/workflows/test.yml
rukai: The cassandra driver installation needs to go into shotover-proxy/build/install_ubuntu_packages.sh otherwise the release.yml workflow will fail.

docs/src/contributing.md
```diff
@@ -2,33 +2,63 @@
...
 * gcc
 * g++
 * libssl-dev
 * pkg-config (Linux)
+* Cassandra CPP Driver
 
-On ubuntu you can install them via `sudo apt-get install cmake gcc g++ libssl-dev pkg-config`
+On Ubuntu you can install all except the Cassandra driver via `sudo apt-get install cmake gcc g++ libssl-dev pkg-config`
 
+### Installing Cassandra CPP Driver
```
rukai, line 21: Shouldnt this also go under `Installing Optional Tools and Libraries` as its only needed when running `cargo test`

shotover-proxy/src/server.rs
```diff
@@ -387,16 +451,16 @@ impl<C: Codec + 'static> Handler<C> {
...
-            spawn_read_write_tasks(self.codec.clone(), rx, tx, in_tx, out_rx);
+            spawn_read_write_tasks(self.codec.clone(), rx, tx, in_tx, out_rx, out_tx.clone());
         };
 
         while !self.shutdown.is_shutdown() {
             // While reading a request frame, also listen for the shutdown signal
-            trace!("Waiting for message");
-            let frame = tokio::select! {
+            debug!("{:?} Waiting for message", thread::current().id());
+            let messages: Messages = tokio::select! {
```
conorbros, line 463: Don't need to specify the type here.

shotover-proxy/src/protocols/cassandra_protocol2.rs
```diff
@@ -40,6 +42,7 @@ pub struct CassandraCodec2 {
     current_frames: Vec<Frame>,
     pk_col_map: HashMap<String, Vec<String>>,
     bypass: bool,
+    last_error: Option<String>,
```
conorbros, line 45: What is `last_error` used for? 

I can see it being used in `decode` to catch the case where an error occurs twice. Why isn't it reset to `None` on an Ok response?

shotover-proxy/src/message/mod.rs
```diff
@@ -18,7 +18,10 @@ pub type Messages = Vec<Message>;
 pub struct Message {
     pub details: MessageDetails,
     pub modified: bool,
+    /// The frame in the format defined by the protocol.
     pub original: RawFrame,
+    /// If return_to_sender true then the message is sent back to the originating system.
+    pub return_to_sender: bool,
```
conorbros, line 24: Is `return_to_sender` only used when there is a protocol error?

I think if a Message is an Protocol error there should be a variant in `MessageDetails` representing this. We don't want to get in the habit of adding boolean flags for every different type of Message.

@rukai & @benbromhead any thoughts on the right abstraction for this?

shotover-proxy/src/protocols/cassandra_protocol2.rs
Claude-at-Instaclustr: For Cassandra, we have to send a message back to the client and then close the connection.  Once we get a protocol error we can not guarantee the state of the input buffer.  For example, a higher protocol level may mean that there are more bytes in the buffer than we process (new fields added) a protocol 1 or 2 means that there are fewer than we expect for protocol 3 or 4.  So we close the connection.  But we can not do this until after we send the message back.  last_error flags the condition where we want to close the connection the next time we look for input.  It is a string so that a message can be logged indicating why we are closing the connection.

rukai: ooooh, this sort of explanation of how the protocol functions would be very valuable as a comment in the source.

shotover-proxy/src/message/mod.rs
Claude-at-Instaclustr: Changed return to sender to be a MessageDetail.

shotover-proxy/src/protocols/cassandra_protocol2.rs
Claude-at-Instaclustr: Updated documentation

shotover-proxy/src/server.rs
Claude-at-Instaclustr: removed

docs/src/contributing.md
Claude-at-Instaclustr: updated

test-helpers/src/docker_compose.rs
Claude-at-Instaclustr: Fixed

.github/workflows/test.yml
Claude-at-Instaclustr: updated

shotover-proxy/tests/cassandra_int_tests/mod.rs
benbromhead: this is a test retry mechanism cleanup - not something I want to block on given this PR has taken a bit to get through let's do this in another PR, but worth the cleanup

@rukai given you've already written the code can you just PR once this is merged.

rukai: sure, lets do that

shotover-proxy/src/message/mod.rs
benbromhead: MessageDetails shouldn't need to know about protocol errors (It should be protocol unaware... even if it knows the query language etc)

benbromhead:
FYI I've started reviewing.. but let me have a proper think / go through the error handling stuff

shotover-proxy/src/protocols/cassandra_protocol2.rs
benbromhead: yup lets pull it out of this change - knowing the thread for a log statement is good, but should be in a seperate PR

shotover-proxy/Cargo.toml
rukai: https://github.com/shotover/cassandra-proto/pull/5 has landed so you can remove this line and apply the other changes in the screenshot.

Claude-at-Instaclustr: move-to-bytes is not sufficient.  It assumes that there will be 9 bytes in the buffer.  Protocol 1 and 2 buffers do not have 9 bytes, they have 8.  The only way to be certain that we can process the buffer is to read the first byte and verify that it is in the range we want.

shotover-proxy/src/protocols/cassandra_protocol2.rs
Claude-at-Instaclustr: @benbromhead en, there is a lot of code that uses the thread for identifying the logging information.  It was critical to my debugging the original issue and figuring out exactly where things were going wrong.  I have no objection to removing it if the request info that Lucas put in provides enough detail.  

Which brings me to the next comment

@rukai In your example, why does the WARN statement (2nd line) not contain the Request info?

Also, since a single request will spin multiple threads how does the request info help us distinguish between them?  Are we assuming that two threads will not execute for the same Request at the same time?

shotover-proxy/src/message/mod.rs
Claude-at-Instaclustr: @benbromhead The change to MessageDetails is a state that indicates the message is being returned to the sender.  I the reason why it is being returned is presumed to be encoded in the message itself.  As it is strictly a routing indicator I think it may belong here.

.github/workflows/test.yml
rukai: and then removed from here.
Otherwise we would run it twice!

shotover-proxy/Cargo.toml
rukai: hmmm maybe I can just close my eyes and ignore the large change to https://github.com/shotover/cassandra-proto

We do plan to move to cdrs-tokio's protocol implementation (cassandra-protocol) so its probably not worth thinking too hard about until we move to cassandra-protocol.

shotover-proxy/src/server.rs
```diff
@@ -327,35 +365,53 @@ fn spawn_read_write_tasks<
...
+                    warn!(
+                        "{:?} failed to decode message: {}",
+                        thread::current().id(),
+                        error
+                    );
+                    return;
                 }
             }
         }
-        .in_current_span(),
```
rukai, line 351: this passes in the request span

shotover-proxy/src/protocols/cassandra_protocol2.rs
rukai: > In your example, why does the WARN statement (2nd line) not contain the Request info?

Because your PR removes the line that passes the span into that task. (I added a comment to direct you to the exact line)

> Are we assuming that two threads will not execute for the same Request at the same time?

In general yes, the message has to finish processing in one task before it can be sent onto the next one.

This assumption is not true for something like Tee or ParallelMap where the message is cloned and sent down multiple chains. For those more work needs to be done to spans so we can distinguish between the branching chains.

However your PR does not touch those transforms, so for the purposes of your PR you can assume "yes".

shotover-proxy/tests/cassandra_int_tests/mod.rs
rukai: this is addressed now

docs/src/contributing.md
```diff
@@ -2,33 +2,65 @@
...
 
+#### libpcap-dev
 Some tests will require `libpcap-dev` to be installed as well (reading pcap files for protocol tests).
 
+## Building Shotover 
+
 Now you can build Shotover by running `cargo build`. The executable will then be found in `target/debug/shotover-proxy`.
 
-## Building Shotover (release)
+### Building Shotover (release)
```
rukai, line 32: This should go above `### Installing Optional Tools and Libraries`

shotover-proxy/src/protocols/cassandra_protocol2.rs
```diff
@@ -577,12 +584,53 @@ impl Decoder for CassandraCodec2 {
         &mut self,
         src: &mut BytesMut,
     ) -> std::result::Result<Option<Self::Item>, Self::Error> {
+        // if we need to close the connection return an error.
+        if self.force_close.is_some() {
```
rukai, line 588: We can simplify this:
```rust
        if let Some(result) = self.force_close.take() {
            debug!("Closing errored connection: {:?}", &result);
            return Err(anyhow::Error::msg(result));
        }
```

shotover-proxy/src/message/mod.rs
rukai: I think MessageDetails::ReturnToSender does make sense!
If we are returning it directly to the sender then there is no need to have the generic contents of the message provided by the other MessageDetails variants.

shotover-proxy/src/protocols/cassandra_protocol2.rs
rukai: Is waiting for the next message to come in before closing the connection a hack? or just how the protocol works?

Claude-at-Instaclustr: It is a timing issue.  We send the return message in `server.rs` line 377
`let filtered_messages = perform_custom_handling(message, &out_tx);`
but if we close the connection immediately the `out_tx` queued message does not get sent.  So we need to delay long enough that the client has a high probability of receiving the message.  In the protocol error we currently are working on the error is detected after reading one byte from the buffer, so there is more data in the buffer and the reader will be called again.  On the subsequent calls it is probable that the byte will not match a valid protocol version and we will loop with another error.

- If we shut the connection before the client receives the message the client will receive a `peer closed the connection` message and may try again with the same invalid protocol and we can try to process it again.
- If the client receives the message they will start a new connection with a different protocol options if there are more options they support.

There may  be a cleaner way to do this.  The critical bits are:

1. We can not reuse the connection because we can not ensure that the entire pipe is empty and because the client may be sending keep alive messages while waiting for reply.
2. We have to hold the connection open long enough for the client to receive the message.

docs/src/contributing.md
Claude-at-Instaclustr: I disagree.  The level 2 headers are 

- Configuring the Environment
- Building Shotover
- Running the Tests
- Submitting a PR

It is a logical progression to configure the environment before running the build.  Also it is logical to place all the environmental configuration in a single section so that if the build fails there is only one place to check for missing components.

rukai: ahhh sorry, my bad.
I must of misread a `#` depth level.
What you have is what we want.

shotover-proxy/src/protocols/cassandra_protocol2.rs
rukai: > If we shut the connection before the client receives the message the client will receive a peer closed the connection message and may try again with the same invalid protocol and we can try to process it again.

Ahh, makes sense, this explains away my concern.
Might be worth adding as a comment.

Claude-at-Instaclustr: code simplified

shotover-proxy/src/protocols/cassandra_protocol2.rs
Claude-at-Instaclustr: thread stuff removed from logging

shotover-proxy/src/server.rs
Claude-at-Instaclustr: ahhh... I did not intend to remove that during a merge.  Fixed now

rukai:
Just a reminder to merge with master, might be able to merge this tomorrow when I review again.

shotover-proxy/examples/cassandra-cluster/test_schema.cql
```diff
@@ -0,0 +1,51 @@
+CREATE KEYSPACE "test" WITH replication = {'class': 'SimpleStrategy', 'replication_factor' : 1};
```
rukai, line 1: is this unused?

shotover-proxy/src/protocols/cassandra_protocol2.rs
```diff
@@ -577,12 +583,48 @@ impl Decoder for CassandraCodec2 {
         &mut self,
         src: &mut BytesMut,
     ) -> std::result::Result<Option<Self::Item>, Self::Error> {
+        // if we need to close the connection return an error.
+        if let Some(result) = self.force_close.take() {
+            self.force_close = None;
```
rukai, line 588: This is not needed, the self.force_close.take() sets force_close to None.

shotover-proxy/src/server.rs
```diff
@@ -444,15 +479,17 @@ impl<C: Codec + 'static> Handler<C> {
                 .await
             {
                 Ok(modified_message) => {
-                    out_tx.send(modified_message)?;
+                    debug!("sending message: {:?}", modified_message);
+                    // send the result of the process up stream
+                    let result = out_tx.send(modified_message)?;
```
rukai, line 484: send returns `Error<()>` and the `?` removes the `Result` leaving just `()` the [unit value](https://doc.rust-lang.org/std/primitive.unit.html).
So the result value here just contains ()
So we should remove the `let result = ` and the `debug!("Send message result {:?}", result);`

shotover-proxy/src/server.rs
```diff
@@ -27,6 +27,29 @@ impl<T: Decoder<Item = Messages, Error = anyhow::Error> + Clone + Send> CodecRea
...
+    // populate it from the original Messages.message Vec.  Any messages that require special
+    // handling are processed here.  Specifically messages with ReturnToSender details.
+    let result = messages
+        .into_iter()
+        .map(|m| {
+            // if there is a protocol error handle it.  always return the original message
+            // it will be filtered out in the next step.
+            if m.details == MessageDetails::ReturnToSender {
+                debug!("processing ReturnToSender: {:?}", &m);
+                tx_out.send(vec![m]).ok();
```
rukai, line 41: `.ok()` turns the Error into an Option, this tends to be used to stop warnings about unused errors.
But we dont want to ignore this error!
At the very least we should do something like:
```rust
if let Err(err) = tx_out.send(vec![m]) {
    error!("Failed to send return to sender message: {:?}", err);
}
```

shotover-proxy/src/server.rs
```diff
@@ -182,6 +206,7 @@ impl<C: Codec + 'static> TcpCodecListener<C> {
                 .unwrap_or_else(|_| "Unknown peer".to_string());
 
             // Create the necessary per-connection handler state.
+            debug!("New connection from {}", &conn_string);
```
rukai, line 209: I think this new connection log came from a merge conflict? I moved the new connection log down to line 240 in the span PR.

shotover-proxy/src/server.rs
```diff
@@ -224,7 +249,7 @@ impl<C: Codec + 'static> TcpCodecListener<C> {
                         error!(cause = ?err, "connection error");
                     }
                 }
-                .instrument(tracing::info_span!(
+                .instrument(tracing::debug_span!(
```
rukai, line 252: Ah, what I had was wrong, however debug_span is not what we want either.
This would cause the span to not be used when we are on the default info level.

Filtering for debug spans works the same as logs.
So info_span will only be included when we are on debug, trace or info levels.

However unlike logs, spans dont report their level.
So this means the correct span level we should use to ensure it is always used is: error_span

rukai (approved):
All of the remaining fixes are straightforward to apply and everything else looks good to me, so you can merge once you have applied the fixes.

benbromhead (approved):
I'm happy with the approach and PR as it stands. I still don't feel like `ReturnToSender` part of the MessageDetails enum is in the right spot... as it ends up being something a transform author then gets exposed. So it feels like protocol details are bleeding through the abstraction somewhat.

Having said that... I can't off the top of my head think of a better spot without more needless reworking of this patch. So let's leave it as is for now.

Thanks @Claude-at-Instaclustr for your work on this as well as all the reviewers... I know this one was a pita.

shotover-proxy/examples/cassandra-cluster/test_schema.cql
Claude-at-Instaclustr: I don't know.  It was in the branch from @benbromhead.  I have removed it

shotover-proxy/src/server.rs
Claude-at-Instaclustr: fixed

shotover-proxy/src/server.rs
Claude-at-Instaclustr: fixed

shotover-proxy/src/server.rs
Claude-at-Instaclustr: fixed

shotover-proxy/src/server.rs
Claude-at-Instaclustr: fixed

shotover-proxy/src/protocols/cassandra_protocol2.rs
```diff
@@ -40,6 +41,13 @@ pub struct CassandraCodec2 {
     current_frames: Vec<Frame>,
     pk_col_map: HashMap<String, Vec<String>>,
     bypass: bool,
+    /// if force_close is not None then the connection will be closed the next time the
```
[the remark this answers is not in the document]
Claude-at-Instaclustr: fixed

shotover-proxy/src/protocols/cassandra_protocol2.rs
Claude-at-Instaclustr: removed

conorbros (approved):

Issue #323, opened 2021-10-28
fix CI caching


rukai:
why do we have two caching systems? (AWS and github actions)

Build times seem to have gone up significantly since we disabled the AWS key
consider https://github.com/shotover/shotover-proxy/runs/4017067700?check_suite_focus=true vs https://github.com/shotover/shotover-proxy/runs/4029994005?check_suite_focus=true

Possibly we want to fix one and throw the other way.
Needs a proper investigation.

PR #333, opened 2021-10-31
rewrite CI, fixing caching and failing on warnings

rukai:
closes https://github.com/shotover/shotover-proxy/issues/323
My conclusion here is that the AWS + sccache solution would be a bit better, but using the GA cache is easier to setup and manage.
So if we find the GA cache limiting we can investigate going back to AWS + sccache.

The other changes are:
* failing on clippy warnings
  + we can do this because we are now specifying a rust version in the rust-toolchain.toml and we already have our Cargo.lock checked in so there is no way for our build to break due to warnings being added.
     - The expectation here is that we will update the rust-toolchain.toml every time a new rust version comes out to take advantage of the speed ups and UX improvements not to mention new features.
* changes the toolchain installer to be a little more efficient and added a comment explaining why we are still wasting some time installing a toolchain we dont use.
* installs [cargo-hack](https://github.com/taiki-e/cargo-hack) so that we can use it for the clippy check allowing us to test every possible feature combination.
* removes the manual docker cleanup - not sure why this existed ... surely the whole VM gets destroyed when we finish anyway.

All the actual code changes are to fix clippy warnings that we weren't hitting locally due to not including `--all-targets`.

### DIFF
--- a/.github/workflows/bench_run.yml
+++ b/.github/workflows/bench_run.yml
@@ -18,23 +18,22 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v2
-    - uses: actions/cache@v2
+    - uses: actions-rs/toolchain@v1
       with:
-        path: |
-          ./cache
-          ~/.cargo/registry
-          ~/.cargo/git
-          target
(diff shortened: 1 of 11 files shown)


conorbros (approved):
The CI speedup is very appreciated

benbromhead (approved):

PR #336, opened 2021-11-01
Fix transform names

conorbros:
Use [`IntoStaticStr`](https://docs.rs/strum/0.22.0/strum/derive.IntoStaticStr.html) to get the name of transforms instead of a method on the `Transform` trait. 

In some places we need access to the transform name from within the transform itself so I have added `get_name` methods directly to the struct.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2524,6 +2524,7 @@ dependencies = [
  "serial_test",
  "sodiumoxide",
  "sqlparser",
+ "strum_macros",
  "test-helpers",
  "thiserror",
  "threadpool",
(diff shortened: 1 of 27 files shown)


rukai (approved):
Very cool!

I think the small hit to compile times is worth the reduced hits to sanity.

Having to define a get_name on some transforms is unfortunate, maybe we could pass the name in as an arg instead?
Happy to land the PR as is though.

XA21X (approved):
I've been wanting this since day one. 😂

conorbros:
Using https://doc.rust-lang.org/std/any/fn.type_name.html might be an alternative to having `get_name` directly in the struct, I'll have a quick look and see if it's worth doing.

PR #338, opened 2021-11-02
Remove MessageDetails::Bypass

rukai:
From what I can see it is only created in `redis/sink_cluster.rs` and I cant see any reason we would want to bypass something here.
Would appreciate a thorough review to make sure i'm not missing anything here.


conorbros:
@benbromhead What was the intention of ByPass?

rukai:
a few weeks ago on slack he said:
image

I dont see any remains of that previous functionality.

benbromhead (approved):
Not being used really at the moment... easy enough to add back in if gets a real use case

XA21X (approved):

PR #351, opened 2021-11-03
Fix testing on ubuntu 18.04

rukai:
The section named "Build & Test" in the GA tagged-release workflow didnt actually do any testing!
This change fixes that, but also moves our test workflows to primarily use 18.04 as that is where our binary is built but also includes an extra 20.04 workflow so that we dont find ourselves stuck when 18.04 is EoL.

In order to fix building on ubuntu 18.04 we had to change rdkafka to build with mklove instead of cmake.
Not sure why we were using cmake in the first place as mklove seems to be shipped with rdkafka.

### DIFF
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -53,7 +53,7 @@ halfbrown = "0.1.11"
 # Transform dependencies
 redis-protocol = "3.0.1"
 cassandra-proto = { git = "https://github.com/shotover/cassandra-proto", branch = "move-to-bytes", features = ["v4"] }
-rdkafka = { version = "0.27", features = ["cmake-build"] }
+rdkafka = { version = "0.27" }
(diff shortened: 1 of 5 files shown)


conorbros (approved):

shotover-proxy/Cargo.toml
```diff
@@ -53,7 +53,7 @@ halfbrown = "0.1.11"
 # Transform dependencies
 redis-protocol = "3.0.1"
 cassandra-proto = { git = "https://github.com/shotover/cassandra-proto", branch = "move-to-bytes", features = ["v4"] }
-rdkafka = { version = "0.27", features = ["cmake-build"] }
+rdkafka = { version = "0.27" }
```
XA21X, line 56: Is there anything else still depending on cmake? Or can we get rid of it completely? (I'll look into it for the next managed service release.) 🤔

XA21X (approved):

PR #352, opened 2021-11-04
remove use of feature specific to cassandra-proto fork

rukai:
* Simplifies the implementation
* Removes dependency on feature added in our cassandra-proto fork (in upstream Row.metadata and Row.row_content are private)
* Less allocation


conorbros:
Are we able to test that function?

rukai:
We already have test_result_codec, which sends a return value decodes and encodes a return value.
That should cover this function.

benbromhead:
Can we remove dependence on our fork altogether or is this just one part.

rukai:
This is just step one on our side. I've also made some PRs for things we need on their side

benbromhead (approved):

conorbros (approved):

PR #353, opened 2021-11-04
Tee config fixes and tests

conorbros:


shotover-proxy/src/transforms/tee.rs
```diff
@@ -26,29 +26,65 @@ pub struct Tee {
 pub enum ConsistencyBehavior {
     Ignore,
     FailOnMismatch,
-    SubchainOnMismatch(Vec<TransformsConfig>),
```
rukai, line 29: Why didnt this work?
This approach seems simpler and better UX.

conorbros: Because we can't use `behavior: SubchainOnMismatch: failchain:` in YAML

rukai: That sounds wrong to me, what happens if you serialize ConsistencyBehaviour::SubchainOnMismatch to yaml

PR #354, opened 2021-11-04
cassandra unittests assert decoded value

rukai:
The removed test was removed because its now duplicated with the functionality of the improved tests.

### DIFF
--- a/shotover-proxy/src/protocols/cassandra_protocol2.rs
+++ b/shotover-proxy/src/protocols/cassandra_protocol2.rs
@@ -655,28 +655,6 @@ mod cassandra_protocol_tests {
     use std::collections::HashMap;
     use tokio_util::codec::{Decoder, Encoder};
 
-    const STARTUP_BYTES: [u8; 31] =
-        hex!("0400000001000000160001000b43514c5f56455253494f4e0005332e302e30");
-
-    const READY_BYTES: [u8; 9] = hex!("840000000200000000");
-
-    const REGISTER_BYTES: [u8; 58] = hex!(
-        "040000010b000000310003000f544f504f4c4f47595f4348414e4745
-    000d5354415455535f4348414e4745000d534348454d415f4348414e4745"
-    );
-
-    const QUERY_BYTES: [u8; 60] = hex!(
-        "0400000307000000330000002c53454c454354202a2046524f4d20737973
-    74656d2e6c6f63616c205748455245206b65793d276c6f63616c27000100"
-    );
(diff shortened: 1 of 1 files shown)


shotover-proxy/src/protocols/cassandra_protocol2.rs
```diff
@@ -675,103 +684,208 @@ mod cassandra_protocol_tests {
...
+        let mut codec = new_codec();
+        let messages = vec![Message {
+            details: MessageDetails::Unknown,
+            modified: false,
+            original: RawFrame::Cassandra(Frame {
+                version: Version::Request,
+                flags: vec![],
+                opcode: Opcode::Startup,
+                stream: 0,
+                body: vec![
```
benbromhead, line 726: we build binary blobs now two ways in our tests. One using the vec macro and one using the hex macro. We should pick one and keep it consistent.

If we want to go with vec! for all things then we can get rid of one more dependency

rukai: That's a good point, I had just copied the vec from the output of the assert_eq failure.
But using hex! for this seems like the right approach.

shotover-proxy/src/protocols/cassandra_protocol2.rs
```diff
@@ -646,132 +646,232 @@ impl Encoder<Messages> for CassandraCodec2 {
...
-        pk_map.insert("test.simple".to_string(), vec!["pk".to_string()]);
-        pk_map.insert(
-            "test.clustering".to_string(),
-            vec!["pk".to_string(), "clustering".to_string()],
-        );
-        let mut codec = CassandraCodec2::new(pk_map, false);
-        test_frame(&mut codec, &READY_BYTES);
+    fn test_codec_startup() {
+        let mut codec = new_codec();
+        let bytes = hex!("0400000001000000160001000b43514c5f56455253494f4e0005332e302e30");
```
conorbros, line 696: Wouldn't converting a string to hex be clearer? Instead of having to convert this hex to find out what it's sending.

rukai: image
They are mostly binary data.
We could possibly split them up into documented subsections so we understand what the data is, but that would be out of scope of this PR.

conorbros (approved):

rukai:
Just rebased it incase there were any problems due to the other cassandra changes.

XA21X (approved):

PR #355, opened 2021-11-05
Tee subchain fixes

conorbros:
Validate subchains in Tee and update doc example to be correct

### DIFF
--- a/shotover-proxy/src/transforms/tee.rs
+++ b/shotover-proxy/src/transforms/tee.rs
@@ -95,6 +95,25 @@ impl Tee {
 
 #[async_trait]
 impl Transform for Tee {
+    fn validate(&self) -> Vec<String> {
+        if let Some(fail_chain) = &self.fail_chain {
+            let mut errors = fail_chain
+                .original_chain
+                .validate()
+                .iter()
+                .map(|x| format!("  {}", x))
+                .collect::<Vec<String>>();
+
+            if !errors.is_empty() {
+                errors.insert(0, format!("{}:", self.get_name()));
+            }
+
+            errors
+        } else {
+            vec![]
+        }
+    }
+
     async fn transform<'a>(&'a mut self, message_wrapper: Wrapper<'a>) -> ChainResponse {
         match self.behavior {
             ConsistencyBehavior::Ignore => {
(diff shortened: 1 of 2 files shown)


shotover-proxy/src/transforms/tee.rs
```diff
@@ -167,3 +186,83 @@ impl Transform for Tee {
...
+                TransformsConfig::Null,
+            ])),
+            timeout_micros: None,
+            chain: vec![TransformsConfig::Null],
+            buffer_size: None,
+        };
+
+        let transform = config.get_source(&holder).await.unwrap();
+        let result = transform.validate();
+        let expected = vec!["Tee:", "  fail_chain:", "    Terminating transform \"Null\" is not last in chain. Terminating transform must be last in chain."];
```
rukai, line 244: its not called fail_chain anymore, did we miss a rename somewhere?

conorbros: https://github.com/shotover/shotover-proxy/blob/54b0e40956ceb46a62513de3b759751a6d0db17c/shotover-proxy/src/transforms/tee.rs#L45

It was always called fail_chain

shotover-proxy/src/transforms/tee.rs
```diff
@@ -167,3 +186,83 @@ impl Transform for Tee {
...
+    use std::collections::HashMap;
+
+    #[tokio::test]
+    async fn test_validate_no_subchain() {
+        let holder = TopicHolder {
+            topics_rx: HashMap::new(),
+            topics_tx: HashMap::new(),
+        };
+
+        let mut config = TeeConfig {
```
rukai, line 204: Can we wrap each of these in their own block instead of making the variables mutable? that way its easier to reason about.
e.g.
```rust
{
       let config = TeeConfig {
            behavior: Some(ConsistencyBehavior::Ignore),
            timeout_micros: None,
            chain: vec![TransformsConfig::Null],
            buffer_size: None,
        };
        let transform = config.get_source(&holder).await.unwrap();
        let result = transform.validate();
        assert_eq!(result, Vec::<String>::new());
}

```

shotover-proxy/src/transforms/tee.rs
rukai: ah, what I meant was, now that the config is called SubchainOnMismatch we should rename "fail_chain" to something like "mismatch_chain"

conorbros: Oh right I thought you meant the other PR. I'll fix that up

rukai (approved):
nice!

XA21X (approved):

PR #359, opened 2021-11-08
Tee int tests

conorbros:


shotover-proxy/src/transforms/tee.rs
```diff
@@ -128,11 +128,11 @@ impl Transform for Tee {
                 let chain_response = chain_result?;
 
                 if !chain_response.eq(&tee_response) {
-                    Ok(vec!(Message::new(
-                        MessageDetails::Response(QueryResponse::empty_with_error(Some(Value::Strings(
-                            "The responses from the Tee subchain and down-chain did not match and behavior is set to fail on mismatch"
+                    Ok(vec!(Message::new_response(
+                        QueryResponse::empty_with_error(Some(Value::Strings(
+                            "ERR The responses from the Tee subchain and down-chain did not match and behavior is set to fail on mismatch"
```
conorbros, line 133: Unless this `"ERR"` is here the string comes out as `"The: responses from the Tee subchain and down-chain did not match and behavior is set to fail on mismatch"`.

Setting `"ERR: "` will cause the string to come out as `"ERR:: responses from the Tee subchain and down-chain did not match and behavior is set to fail on mismatch"`.

shotover-proxy/src/transforms/internal_debug_transforms.rs
```diff
@@ -1,23 +1,66 @@
 use crate::error::ChainResponse;
-use crate::message::Messages;
-use crate::transforms::{Transform, Wrapper};
-use anyhow::anyhow;
+use crate::message::{Message, MessageDetails, Messages};
+// use crate::transforms::{
```
rukai, line 3: commented out code.

And should the DebugReturner changes go in there own PR?

conorbros: 👍 I'll move them to another PR

shotover-proxy/src/transforms/tee.rs
rukai: Is this a redis protocol thing?
Maybe you should write a comment in the code explaining whats going on.

rukai: Hmm... this isnt a redis specific transform, I'm quite confused what you mean then.

shotover-proxy/src/transforms/mod.rs
```diff
@@ -81,7 +80,6 @@ pub enum Transforms {
     RedisTimestampTagger(RedisTimestampTagger),
     RedisSinkCluster(RedisSinkCluster),
     RedisClusterPortsRewrite(RedisClusterPortsRewrite),
-    #[cfg(test)]
     DebugReturner(DebugReturner),
```
rukai, line 83: If we are adding this to our public API we will need to document it.

rukai: Looking at the returner API:
```yaml
    - DebugReturner:
        response:
          Redis: "42"
```

The `response:` line is just noise.
Can we remove it?
I think we could do this by one of:
*  writing the type as `struct DebugReturner (Response)`
*  adding a serde attribute to merge with the parent.

conorbros: Raised #370

shotover-proxy/tests/transforms/tee.rs
```diff
@@ -0,0 +1,147 @@
+use crate::helpers::ShotoverManager;
+use serial_test::serial;
+use test_helpers::docker_compose::DockerCompose;
+
+#[tokio::test(flavor = "multi_thread")]
+#[serial]
+async fn test_ignore() {
+    ignore_matches().await;
```
rukai, line 8: Why not just make these all invidual tests rather then batching them into groups of two?

conorbros: True, I originally had `DockerCompose` so it made sense to group them. I'll fix that up

shotover-proxy/src/transforms/tee.rs
conorbros: The first byte of an error response from Redis is a minus sign followed by the type of error: e.g. `"-ERR"` for a generic error. Our Redis codec is taking this, thinking the error type is `"The"` and serialising it as `"-The responses from the Tee subchain and down-chain did not match and behavior is set to fail on mismatch"`. The Redis client receives the error response, thinks it's type is `"The"` as well and formats it like `"{Error type}: message"`

This can be fixed by changing `fn build_redis_response_frame` to add `"ERR "` to the front of our generic errors.

rukai: nice find!
Should the change maybe go in redis-protocol?

conorbros: I don't think so, the issue is how we are converting our generic `QueryResponse` errors to Redis ones.

rukai: ah, doesnt look like it.
My only concern then is if anything is using the current functionality of  build_redis_response_frame to set a custom error type.

rukai: https://github.com/shotover/shotover-proxy/blob/b029e97a1b03ed0115d85c51c650f119013d834d/shotover-proxy/src/protocols/redis_codec.rs#L405 Hmm looks like we are passing redis errors straight through like this.

rukai: So I think the code as you've currently implemented it is the correct approach.
Thanks for investigating!

rukai:
I think this needs a rebase as the #[test] removals are still showing up?

shotover-proxy/src/transforms/tee.rs
conorbros: For now, but when we need to handle Cassandra errors too it won't work.

shotover-proxy/tests/transforms/tee.rs
```diff
@@ -0,0 +1,138 @@
...
+        .unwrap();
+
+    assert_eq!("42", result);
+}
+
+#[tokio::test(flavor = "multi_thread")]
+#[serial]
+async fn test_ignore_with_mismatch() {
+    let shotover_manager =
+        ShotoverManager::from_topology_file("tests/test-topologies/tee/ignore.yaml");
```
rukai, line 33: are we meant to be testing a different topology file?

shotover-proxy/tests/test-topologies/tee/subchain_with_mismatch.yaml
```diff
@@ -0,0 +1,35 @@
...
+    Redis:
+      listen_addr: "127.0.0.1:6379"
+      connection_limit: 3000000
+  redis_2:
+    Redis:
+      listen_addr: "127.0.0.1:6378"
+      connection_limit: 3000000
+chain_config:
+  redis_chain_2:
+    - RedisSinkCluster:
```
rukai, line 13: what is the second chain for?

shotover-proxy/tests/test-topologies/tee/subchain_with_mismatch.yaml
```diff
@@ -0,0 +1,35 @@
+---
```
rukai, line 1: Why is the mismatch case the one tested against a real database instance?
Shouldnt we be testing the match case against a real database to make sure it matches unless something bad happens?
And then for the mismatch case we want to simulate something bad happening so thats where use the returner.

shotover-proxy/src/transforms/tee.rs
rukai: yeah, thats something to look into when we get to the cassandra stuff.

shotover-proxy/tests/test-topologies/tee/subchain_with_mismatch.yaml
conorbros: Not sure how else to test that the subchain runs other than put a redis cluster at the end and check a value was sent to it. 

We can test the match case with a real database too though.

shotover-proxy/tests/transforms/tee.rs
conorbros: True, I need to add another topology YAML for the case where we use Ignore and matches.

shotover-proxy/tests/test-topologies/tee/subchain_with_mismatch.yaml
rukai: Well at the very least we could use a single redis instance rather than a cluster which will be faster to startup.

Possible improvements could be inspecting the log and using the log transform
Or we could create a transform that writes the message to the file system. Then we could be bank scale :laughing: 

But lets just use a single redis instance for now.
Maybe leave a comment explaining why we need a redis instance here.

shotover-proxy/tests/transforms/tee.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    // Set the value on the top level chain cluster
+    redis::cmd("SET")
+        .arg("key")
+        .arg("myvalue")
+        .query_async::<_, String>(&mut top_level_chain_cluster)
+        .await
+        .unwrap();
+
+    // // When the mismatch occurs, the value should be sent to the mismatch chain's cluster
```
rukai, line 130: * double comment
* this isnt a cluster anymore
* can we just query the redis instance directly rather than going through shotover here?

shotover-proxy/tests/transforms/tee.rs
```diff
@@ -0,0 +1,138 @@
...
+        .await
+        .unwrap();
+
+    assert_eq!(result, "42");
+}
+
+#[tokio::test(flavor = "multi_thread")]
+#[serial]
+async fn test_subchain_with_mismatch() {
+    let _compose = DockerCompose::new("examples/redis-multi/docker-compose.yml")
```
rukai, line 112: why are we using multi?
Surely pass-through would be fine?

shotover-proxy/tests/test-topologies/tee/subchain_with_mismatch.yaml
```diff
@@ -0,0 +1,35 @@
...
+        remote_address: "127.0.0.1:3333"
+  redis_chain_1:
+    - Tee:
+        buffer_size: 10000
+        behavior:
+          SubchainOnMismatch:
+            - QueryTypeFilter:
+                filter: Read
+            - RedisSinkSingle:
+                remote_address: "127.0.0.1:3333"
```
rukai, line 23: We are testing mismatch so we can just use a DebugReturner here.

shotover-proxy/tests/test-topologies/tee/subchain_with_mismatch.yaml
```diff
@@ -0,0 +1,35 @@
...
+        chain:
+          - QueryTypeFilter:
+              filter: Read
+          - DebugReturner:
+              Redis: "42"
+    - RedisSinkSingle:
+        remote_address: "127.0.0.1:3331"
```
rukai, line 30: we are testing mismatch so we can just a DebugReturner here.

shotover-proxy/tests/transforms/tee.rs
conorbros: > can we just query the redis instance directly rather than going through shotover here? 
We can, but I'll use the `redis_connection_async` method from `ShotoverManager` still.

rukai: thats fine!

shotover-proxy/tests/test-topologies/tee/subchain_with_mismatch.yaml
```diff
@@ -0,0 +1,31 @@
+---
+sources:
+  redis_1:
+    Redis:
+      listen_addr: "127.0.0.1:6379"
+      connection_limit: 3000000
+chain_config:
+  redis_chain_2:
```
rukai, line 8: this got left in

shotover-proxy/tests/test-topologies/tee/fail.yaml
```diff
@@ -0,0 +1,22 @@
+---
+sources:
+  redis_prod:
+    Redis:
+      listen_addr: "127.0.0.1:6379"
+      connection_limit: 3000000
+chain_config:
+  redis_chain:
+    - Tee:
+        behavior: Ignore
```
rukai, line 10: FailOnMismatch

shotover-proxy/tests/test-topologies/tee/ignore_with_mismatch.yaml
```diff
@@ -0,0 +1,21 @@
...
+      connection_limit: 3000000
+chain_config:
+  redis_chain:
+    - Tee:
+        behavior: Ignore
+        buffer_size: 10000
+        chain:
+          - QueryTypeFilter:
+              filter: Read
+          - Null
```
rukai, line 15: Should we be consistent with the others and use DebugReturner here?

shotover-proxy/tests/test-topologies/tee/subchain.yaml
```diff
@@ -0,0 +1,19 @@
+---
+sources:
+  redis_prod:
+    Redis:
+      listen_addr: "127.0.0.1:6379"
+      connection_limit: 3000000
+chain_config:
+  redis_chain:
+    - Tee:
+        behavior: Ignore
```
rukai, line 10: SubchainOnMismatch

shotover-proxy/tests/test-topologies/tee/subchain.yaml
```diff
@@ -0,0 +1,19 @@
...
+        buffer_size: 10000
+        chain:
+          - Null
```
rukai, line 13: DebugReturner

shotover-proxy/tests/test-topologies/tee/subchain.yaml
```diff
@@ -0,0 +1,23 @@
...
+    Redis:
+      listen_addr: "127.0.0.1:6379"
+      connection_limit: 3000000
+chain_config:
+  redis_chain:
+    - Tee:
+        behavior:
+          SubchainOnMismatch:
+            - DebugReturner:
+                Redis: "42"
```
rukai, line 13: Dont we need to hit a real redis instance so we can assert that the subchain isnt run?

shotover-proxy/tests/transforms/tee.rs
```diff
@@ -0,0 +1,146 @@
...
+
+#[tokio::test(flavor = "multi_thread")]
+#[serial]
+async fn test_ignore_matches() {
+    let shotover_manager =
+        ShotoverManager::from_topology_file("tests/test-topologies/tee/ignore.yaml");
+
+    let mut connection = shotover_manager.redis_connection_async(6379).await;
+
+    redis::cmd("SET")
```
rukai, line 13: I think we should run the same command for every test.
This SET seems fine, lets run just this one command for every test, having the tests consistent makes it easier to ensure they are all correct and makes it easier to investigate failures.

shotover-proxy/tests/transforms/tee.rs
```diff
@@ -0,0 +1,142 @@
...
+#[tokio::test(flavor = "multi_thread")]
+#[serial]
+async fn test_subchain_matches() {
+    let _compose = DockerCompose::new("examples/redis-passthrough/docker-compose.yml")
+        .wait_for("Ready to accept connections");
+
+    let shotover_manager =
+        ShotoverManager::from_topology_file("tests/test-topologies/tee/subchain.yaml");
+
+    let mut connection = shotover_manager.redis_connection_async(6379).await;
```
rukai, line 88: can we name this consistently with the other test?

shotover-proxy/tests/transforms/tee.rs
```diff
@@ -0,0 +1,142 @@
...
+        .query_async::<_, String>(&mut mismatch_chain_redis)
+        .await
+        .unwrap();
+
+    redis::cmd("SET")
+        .arg("key")
+        .arg("notmyvalue")
+        .query_async::<_, ()>(&mut connection)
+        .await
+        .unwrap();
```
rukai, line 102: We should assert that this returns 42 (and do the same for test_subchain_with_mismatch)

shotover-proxy/tests/transforms/tee.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    // Set the value on the top level chain redis
+    let mut result = redis::cmd("SET")
+        .arg("key")
+        .arg("myvalue")
+        .query_async::<_, String>(&mut shotover_redis)
+        .await
+        .unwrap();
+
+    assert_eq!(result, "41");
```
rukai, line 136: this is inconsistent with the the ignore test which has 42 <-> 41 swapped.

rukai (approved):

XA21X (approved):

PR #362, opened 2021-11-08
Debug returner transform fixes

conorbros:
Make `DebugReturnerTransform` usable from a topology file as well. Can be used like so:

```YAML
- DebugReturnerTransform:
    RedisResponse: "42"
```


shotover-proxy/src/transforms/internal_debug_transforms.rs
```diff
@@ -1,23 +1,63 @@
...
-    pub message: Messages,
+    pub message: Option<Messages>,
+    pub response: Option<String>,
     pub ok: bool,
 }
 
 #[async_trait]
 impl Transform for DebugReturnerTransform {
     async fn transform<'a>(&'a mut self, _message_wrapper: Wrapper<'a>) -> ChainResponse {
+        if let Some(response) = self.response.clone() {
```
conorbros, line 46: DebugReturnerTransform will first check if there is a response, then check if there is a message.

We should error when creating if there is both a response and message.

rukai: We should use enums so that is impossible to be in invalid state

rukai:
This is currently hardcoded to be a redis bulk string but the name of the transform doesnt reflect that.
I think your going to have to think about the design for this a bit.

I'm not sure if you thought this was ready or not but if you know a PR isnt ready but still want to publish it please make it a draft PR. Reasons you might want to do this are if you want to ask specific questions about the design or you just want to see how CI handles what you have so far.

conorbros:
I'll make the enum either a `Message` or `RedisResponse` for now, when we want to use it for other types of responses we can add those in.

shotover-proxy/src/transforms/internal_debug_transforms.rs
conorbros: Fixed

shotover-proxy/src/transforms/internal_debug_transforms.rs
```diff
@@ -1,23 +1,74 @@
 use crate::error::ChainResponse;
-use crate::message::Messages;
 use crate::transforms::{Transform, Wrapper};
-use anyhow::anyhow;
 use async_trait::async_trait;
 use rand_distr::Distribution;
 use rand_distr::Normal;
 use tokio::time::Duration;
 
+#[cfg(test)]
```
rukai, line 8: Can we move it into its own file and then do:
```rust
#[cfg(test)]
mod debug_returner
```

shotover-proxy/src/transforms/internal_debug_transforms.rs
```diff
@@ -1,23 +1,74 @@
...
+    crate::transforms::Transforms,
+    anyhow::{anyhow, Result},
+    redis_protocol::resp2::prelude::Frame,
+    serde::Deserialize,
+};
+
+#[cfg(test)]
+#[derive(Deserialize, Debug, Clone)]
+pub struct DebugReturnerTransformConfig {
+    ok: bool,
```
rukai, line 21: this should be a variant of response

shotover-proxy/src/transforms/debug/returner.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+}
+
+#[async_trait]
+impl Transform for DebugReturner {
+    async fn transform<'a>(&'a mut self, _message_wrapper: Wrapper<'a>) -> ChainResponse {
+        match &self.response {
+            Response::Message(message) => Ok(message.clone()),
+            Response::Redis(string) => {
+                let res = vec![Message {
```
rukai, line 49: use implicit return:
```rust
Ok(vec![Message { .. }])
```

rukai (approved):
lgtm

XA21X (approved):

PR #363, opened 2021-11-09
Instrument async code with tracing spans

rukai:
Concurrent logs now look like this:

image

This allows us to see through the noise and form a sequence of events from concurrent logs.

Possibly we should be using a random id instead of an incrementing value. - Discussed with kuangda: incrementing will be fine for now

Relevant docs: https://docs.rs/tracing/0.1.29/tracing/trait.Instrument.html

### DIFF
--- a/shotover-proxy/src/transforms/chain.rs
+++ b/shotover-proxy/src/transforms/chain.rs
@@ -10,7 +10,7 @@ use tokio::sync::mpsc::Sender;
 use tokio::sync::oneshot::Receiver as OneReceiver;
 use tokio::time::Duration;
 use tokio::time::Instant;
-use tracing::{debug, trace, warn};
+use tracing::{debug, trace, warn, Instrument};
 
 type InnerChain = Vec<Transforms>;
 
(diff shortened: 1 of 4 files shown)


shotover-proxy/src/transforms/chain.rs
```diff
@@ -115,43 +115,48 @@ impl TransformChain {
...
 
         let mut chain = self.clone();
-        let _jh = tokio::spawn(async move {
-            while let Some(ChannelMessage {
-                return_chan,
-                messages,
-            }) = rx.recv().await
-            {
-                #[cfg(test)]
+        let _jh = tokio::spawn(
```
rukai, line 118: Very little was actually change here, the only thing I did here was call .in_current_span() but it made rust fmt freak out.

conorbros (changes requested):
Everything looks good to me but we should add some tests before merging. 

Testing tracing: https://docs.rs/tracing-test/0.1.0/tracing_test/

rukai:
That library is way too inflexible to do anything with:
* No way to assert on spans (What this PR is actually changing)
* No way to assert on part of a log. e.g. we cant write an assertion for `New connection from 127.0.0.1:38602` because the port is random.
* No way to view the actual log after a failure.

The idea is cool but I think we'll have to revisit that later when its more than just a prototype.

conorbros (approved):
Fair enough. Would be nice to test but should be fine for the time being.

XA21X (approved):

PR #364, opened 2021-11-09
Dockerfile for Shotover

conorbros:
Example of how to use, assuming your config and topology file are in the home dir:
`docker run -t -v /home/conor:/config shotover-proxy --help`
`docker run -t -v /home/conor:/config shotover-proxy -t config/top1.yaml -c config/config.yaml`

I've registered a "shotover" user on Dockerhub and will also integrate publishing our releases in this PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM rust:latest as builder
+
+WORKDIR ./shotover-proxy
+
+COPY ./ ./
+
+# Clean out anything already built in the target dir
+RUN rm -rf target
+
+RUN cargo build
+
+FROM debian:bullseye-slim
+
+RUN apt-get update -y
+    \ && apt-get install glibc-doc -y
+    \ && rm -rf /var/lib/apt/lists/*
+
+COPY --from=builder /shotover-proxy/target/debug/shotover-proxy /shotover-proxy
+
+ENTRYPOINT ["./shotover-proxy"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,13 @@
 FROM rust:latest as builder
 
-WORKDIR ./shotover-proxy
+WORKDIR /shotover-proxy
 
 COPY ./ ./
 
-# Clean out anything already built in the target dir
-RUN rm -rf target
-
-RUN cargo build
+RUN cargo build --release
 
 FROM debian:bullseye-slim
 
-RUN apt-get update -y
-    \ && apt-get install glibc-doc -y
-    \ && rm -rf /var/lib/apt/lists/*
-
-COPY --from=builder /shotover-proxy/target/debug/shotover-proxy /shotover-proxy
+COPY --from=builder /shotover-proxy/target/release/shotover-proxy /shotover-proxy
 
 ENTRYPOINT ["./shotover-proxy"]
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM rust:latest as builder
+
+WORKDIR ./shotover-proxy
+
+COPY ./ ./
+
+# Clean out anything already built in the target dir
+RUN rm -rf target
```
XA21X, line 8: You can add a `.dockerignore` file to exclude paths from the build context (just like `.gitignore`).

conorbros:
I've tested the workflow on this repo: https://github.com/conorbros/hot-restart-proxy
Published results can be seen here: https://hub.docker.com/repository/docker/conorbros/hot-restart-proxy

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM rust:latest as builder
+
+WORKDIR ./shotover-proxy
+
+COPY ./ ./
+
+RUN cargo build
```
rukai, line 7: I was wondering why the binary was 239MB,
you forgot to build in `--release`!

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+FROM debian:bullseye-slim
+
+RUN apt-get update -y \
+    && apt-get install glibc-doc -y \
```
rukai, line 12: why do we install glibc-doc?

rukai (changes requested):

Dockerfile
conorbros: Not on `debian:bulleye-slim` by default.

Dockerfile
conorbros: Oops, yep I didn't want to wait as long in between tests.

Dockerfile
rukai: what is it though?

conorbros: Was a mistake leaving it in. Removed it as `debian:bullseye` includes by default.

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+WORKDIR /shotover-proxy
+
+COPY ./ ./
+
+RUN cargo build --release
+
+FROM debian:bullseye-slim
+
+RUN apt-get update -y \
```
rukai, line 11: isnt this basically a no-op?
updating the lists and then deleting them?
Or is there a reason for it?

conorbros: Oops, there should be an upgrade in there.

rukai: Is that a normal thing to do?
Wouldnt the debian image already have up to date packages?
Looking at redis dockerfile I can see that it doesnt run upgrade https://github.com/docker-library/redis/blob/master/6.2/Dockerfile

rukai: https://github.com/docker-library/cassandra/blob/master/4.0/Dockerfile cassandra too

conorbros: Ok sure I'll leave out the upgrading.

rukai (approved):
I dont have much experience with docker but this all seems reasonable to me:
* `docker image save shotover -o foo.tar.gz` shows nothing strange is being accidentally included.
* `docker run --network host -p 6379:6379 -v $PWD/config:/config shotover` runs fine off the default config files in the shotover repo and I can succesfully send a redis message.

XA21X (approved):

PR #371, opened 2021-11-11
Debug transform docs

conorbros:


docs/src/transforms.md
```diff
@@ -110,10 +112,37 @@ Upon receiving the configured number of responses, the transform will attempt to
...
 
-```yaml
+```YAML
 - DebugPrinter
 ```
 
+
+### DebugReturner
+
+This transform will return the supplied response.
```
rukai, line 122: I think we should also explicitly mention that the outgoing message is dropped.

rukai (changes requested):
just one small change

rukai (approved):

XA21X (approved):

Issue #281, opened 2021-10-15
Tee fixes


rukai:
Currently we only have an integration tests that use ConsistencyBehaviour::IGNORE

conorbros:
Tee should only be handling write requests. Currently `redis-cluster-dr` has a `QueryTypeFilter` that removes the read queries, we should instead integrate this directly into Tee so users don't have to remember to use the filter every time.

- [ ] Integrate `QueryTypeFilter: Read` directly with `Tee`
- [ ] Add integration tests that use ConsistencyBehaviour::FAIL and ConsistencyBehaviour::LOG

PR #372, opened 2021-11-11
Filter Read messages in Tee

conorbros:
Filter out all "Read" messages in Tee. Removes the need to put a `QueryTypeFilter::Read` in front of Tee for it to function correctly.

Last of the fixes for #281

### DIFF
--- a/docs/src/user-guide/configuration.md
+++ b/docs/src/user-guide/configuration.md
@@ -127,8 +127,6 @@ chain_config:
         buffer_size: 10000
         #The child chain, that Tee will asynchronously pass requests to
         chain:
-          - QueryTypeFilter:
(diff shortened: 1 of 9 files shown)


rukai:
Might be worth checking in with ben on monday about this before getting to far into it.

benbromhead:
I would let a filter transform do this. General purpose Transforms should (where practical) do one thing only.

rukai:
Just for the record: I think forgetting to filter reads in the tee branch is a huge footgun, so in the future we might want investigate automatically filtering reads or have validation that the tee is followed by a filter read transform.

But there are much higher priorities atm, so I think closing the PR is the right call.

PR #373, opened 2021-11-12
Fix broken doc links and enforce via mdbook-linkcheck plugin

rukai:
The setup script is adjusted to run for both netlify and local usage (previously just local usage)

I opted to delete most of the dead-links but im happy to comment them out if we want to keep them around until we bring back the pages they link to.

### DIFF
--- a/docs/mdbook.sh
+++ b/docs/mdbook.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+set -e; set -u
+
+# Setup the the mdbook binaries if they havent already been setup
+if [ ! -d "mdbook_bin" ]; then
+    mkdir -p mdbook_bin
+    pushd mdbook_bin
+    curl -L https://github.com/rust-lang/mdBook/releases/download/v0.4.13/mdbook-v0.4.13-x86_64-unknown-linux-gnu.tar.gz | tar xvz
+    wget https://github.com/Michael-F-Bryan/mdbook-linkcheck/releases/download/v0.7.6/mdbook-linkcheck.x86_64-unknown-linux-gnu.zip -O linkcheck.zip
+    unzip linkcheck.zip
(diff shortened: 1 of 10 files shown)


docs/mdbook.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+set -e; set -u
+
+# Setup the the mdbook binaries if they havent already been setup
```
conorbros, line 5: Can we install with cargo?

rukai: It needs to run on netlify so unfortunately not.

conorbros: Can we do the checks on our CI instead?

rukai: Doing that would mean we have two locations that we need to manage running mdbook.
Surely doing it all in one place would be simpler?

If we build both mdbook and mdbook-linkcheck on CI that will make CI run longer and use more of our cache.

rukai: There are some basic syntax checks that netlify is already handling for us.

rukai:
Currently blocked on https://github.com/Michael-F-Bryan/mdbook-linkcheck/pull/65
Will split out the link fixes if the issue takes too long to resolve.

rukai:
Oh wow our netlify image is on xenial (16.04) 
image

@benbromhead I dont seem to be able to set this from the config file.
Can you follow the instructions to upgrade the image here: https://answers.netlify.com/t/new-ubuntu-20-04-focal-fossa-build-image/37095

Should be completely safe as everything is statically linked (except for libc)

benbromhead:
Done!

rukai:
It works!

conorbros (approved):
LGTM!

benbromhead (approved):

PR #374, opened 2021-11-15
Always pass chain name into Wrapper

conorbros:
We wouldn't want to create a `Wrapper` outside of a chain unless testing so I have restricted the use of `new` to tests only. The other 2 `new_xxx` methods ensure that `Wrapper` should always be given a chain name.


rukai (approved):
Nice!
Changing the API to make it harder for this bug to reoccur is 💯

XA21X (approved):

PR #375, opened 2021-11-16
Run cargo udeps

conorbros:
The aftermath of running `cargo udeps` on our codebase.

If properly implementing `timeout` and `count` into `ConsistentScatter` is something we want to do I can open a ticket for it.

### DIFF
--- a/shotover-proxy/src/protocols/cassandra_protocol2.rs
+++ b/shotover-proxy/src/protocols/cassandra_protocol2.rs
@@ -37,7 +37,6 @@ use anyhow::{anyhow, Result};
 pub struct CassandraCodec2 {
     compressor: NoCompression,
     current_head: Option<FrameHeader>,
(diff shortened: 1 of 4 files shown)


rukai (approved):
Im not sure how an unused deps tool found dead code, but the results look good to me!

I suspect we might find more dead code if we set more things to private.
But possibly being pub is useful if we want to expose things as a library in the future.

conorbros:
Shouldn't only the `runner` module be public?

Assuming we would publish as a library like detailed here: https://github.com/shotover/shotover-proxy/issues/120

rukai:
Something like that yeah

rukai:
lets avoid merging this until after https://github.com/shotover/shotover-proxy/pull/331 (which should happen tonight)

benbromhead (approved):

PR #379, opened 2021-11-18
Reuse wait for socket implementation across cassandra and redis

rukai:
Changes cassandra tests to use the wait_for_socket_to_open implementation that redis uses instead of the current cassandra level protocol level wait/retry system.

wait_for_socket_to_open has a few advantages:
* its simpler
* its reusable across different protocols
* it only fails if the port is closed. This is important because if shotover can be in a state where the port is open but the connection fails then the test should fail!


conorbros (approved):
LGTM. 

I think that `cassandra_connection` should live in `helpers/mod.rs` but I won't block this PR for that.

rukai:
I left that out because it was a bit controversial, but I would like to address that in another PR

benbromhead (approved):

PR #382, opened 2021-11-23
Cassandra integration test improvements

rukai:
* unify test_pass_through/test_passthrough naming to always be test_passthrough
* improve the cassandra test infrastructure
    + The cassandra-cpp API doesnt provide a good way to access structured result data so I had to roll my own
        - https://docs.rs/cassandra-cpp/0.17.0/cassandra_cpp/struct.Value.html
        - https://docs.rs/cassandra-cpp/0.17.0/cassandra_cpp/struct.CassResult.html
    + With the addition of `assert_query_result` we can now easily make assertions about the entire result.
* test_cassandra_standalone is replaced with test_passthrough which actually runs tests on shotover instead of just starting it.


shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -2,63 +2,93 @@ use crate::helpers::ShotoverManager;
...
+            ValueType::INT => ResultValue::Int(value.get_i64().unwrap()),
+            ty => todo!(
+                "The test infrastructure hasnt implemented the type {} yet, you should add it.",
+                ty
+            ),
+        }
+    }
+}
+
+fn assert_query_result(session: &Session, query: Statement, expected_rows: &[&[ResultValue]]) {
```
conorbros, line 29: I'd put this in `cassandra_int_tests/mod.rs`

conorbros (approved):
Only one non-blocking comment. LGTM!

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
rukai: yeah, that makes sense to me

benbromhead (approved):

PR #384, opened 2021-11-23
Easy win in test runtime

rukai:
from my vaguely scientific testing I came up with the following table:

```
optimizations      get/set loop     test_pass_through time to complete
--release          1                3.5s
--release          10000            3.5s
all opt=3          1                4.0s
deps opt=3         1                4.5s
none               1                6.0s
deps opt=3         10000            15s
none               10000            30s
```

* `--release` - I ran with `test --release`, includes all the crazy release optimizations like LTO and stuff
* `deps opt=3` - set debug mode to have just the dependencies at opt-level=3, all other optimizations unchanged.
* `all opt=3` - set debug mode to have everything at opt-level=3, all other optimizations unchanged.

## get/set loop iterations
reducing the number of iterations in this loop is a very easy win.
However I've decided to keep the number of iterations at 10000 in release as it can handle that many just fine and it acts as a nice stress test.

Also opted to make a bunch of other loops make use of this constant too. This gives a nice win on the clustering integration tests which werent represented in the table.

## optimizations

using --release for local iterative testing is unusable because the LTO takes forever to build. (Although on CI it is incredibly valuable for testing how real shotover runs.)

To improve the total runtime of `cargo test` we could set debug builds to opt 3 for only the deps this is a potentially useful sweet spot that is worth considering:
*   takes 1s off every redis int test.
*   shouldnt affect iterative compile times
*   The one downside is that stacktraces in dependencies wont be as useful.

Because of this one downside I've left this change out of this PR.
However as the number of tests grow I might want to revisit this and possibly enable opt-level=3 for our dependencies in debug mode.

## Wins
Compare this recent CI run to the CI run for this PR:
image

And running `cargo test` locally on my machine:
* main branch - 10m 2s
* this branch - 4m 30s

### DIFF
--- a/shotover-proxy/tests/redis_int_tests/basic_driver_tests.rs
+++ b/shotover-proxy/tests/redis_int_tests/basic_driver_tests.rs
@@ -14,6 +14,14 @@ use crate::helpers::ShotoverManager;
 use shotover_proxy::tls::TlsConfig;
 use test_helpers::docker_compose::DockerCompose;
 
+// Debug rust is pretty slow, so keep the stress test small.
+// CI runs in both debug and release so the larger iteration count does get run there.
+#[cfg(debug_assertions)]
+const STRESS_TEST_MULTIPLIER: usize = 1;
+
+#[cfg(not(debug_assertions))]
+const STRESS_TEST_MULTIPLIER: usize = 100;
+
 async fn test_args(connection: &mut Connection) {
     redis::cmd("SET")
         .arg("key1")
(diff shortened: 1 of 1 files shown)


conorbros (approved):

benbromhead (changes requested):
The problem I have with this is we've had bugs in the past that only got uncovered when we ran tests with the `--release` flag (nasty timing based bugs).

So as it stands I'm -1 on this change.

If we could decouple `--release` tests to not block mergeing and we could run it as a daily test. I would be happy to turn this into a +1. This approach would also be good to move benchmarks into a daily pattern as well.

rukai:
I'm confused. Im not proposing to remove --release from CI.
I think its really valuable.
The only change I'm proposing at the moment are the changes in the PR.

benbromhead (approved):

PR #388, opened 2021-11-29
Join on source task

rukai:
This PR has two changes:
* Join on the sources task when shotover shutsdown - first commit
    + this is needed to prevent a race condition where shotover ends before the source's task shutsdown. This can be observed by the `redis source shutting down` sometimes not being logged.
* Remove shutdown_complete broadcast channel - second commit
    + now that we are waiting on sources join handle it seems that this is redundant.
    + Having said that I'm a little concerned as to if there is some significance behind mini-redis taking this approach. Maybe shotover has diverged enough from mini-redis that a different approach is warranted?

### DIFF
--- a/shotover-proxy/src/config/topology.rs
+++ b/shotover-proxy/src/config/topology.rs
@@ -108,13 +108,11 @@ impl Topology {
     pub async fn run_chains(
         &self,
         trigger_shutdown_rx: watch::Receiver<bool>,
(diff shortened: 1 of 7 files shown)


benbromhead:
We dont really use shutdown_complete at all - happy to remove it, but we might be reaching for it again in the future

benbromhead (approved):
lgtm

conorbros (approved):

Issue #385, opened 2021-11-25
Tee transform breaks auth isolation because BufferedChains are shared


XA21X:
## Describe the bug
This is another auth isolation bug that is very similar to #204. Authentication support in RedisSinkCluster assumes that each instance of itself is dedicated per client connection to Shotover. Transforms like Tee (and the since-removed PoolConnections) stores state in shared references (Arc) in a way that violates this assumption.

## To Reproduce
1. Add auth support to the `redis-cluster-dr` example.
2. Connect to primary cluster using redis-cli without authentication.
3. Connect to primary cluster using redis-cli with authentication.
4. Connect to secondary cluster using redis-cli with authentication.
5. Attempt to write something using (2), which should fail with a NOAUTH.
6. Check that the write did not go through using (4)... and this is is the problem. The write went through.

The successful authentication from (3) also authenticates the shared DR chain, which is also used by (2).

## Expected behaviour
- Unauthenticated connections must be prevented from writing to the secondary (DR) cluster.
- Authenticated contexts must be kept isolated between connections.
- BufferedChains possibly shouldn't be shared by default?

## Systems and Version:
 - Shotover Version 0.1.0

benbromhead:
Some thoughts:
So this is likely to be a "big" fix issue as it will require some restructuring of how we share data amongst different connections.

Previously we've discussed things like colored connections or auth contexts etc. 

I think right now the simplest could be attaching context to each Wrapper (or individual message) and one of these could be an auth context. Then on both Tee and Connection pooling we maintain a map of contexts to senders (this should be ok with just a `&` ref ?)

Contexts would need to be modifiable by transforms and then an authentication aware transform (or maybe at the protocol/server) we would inject the context into the Wrapper / message. The downside with a transform based approach to auth context tagging is it requires a user to add it to the chain, rather than just having it on by default. 

It probably needs to live on the source side of things (though a lot of the logic is done at the tcp listener). One idea I had was the concept of default transforms, or source generated transforms. Where the cassandra source will actually prepend / or modify the transform chains with it's own set of transforms it needs to do Cassandra things.

That way it keeps the Transform model a little more pure, takes a cognitive load off the user in having to build the right chain just to get a certain protocol working and keeps protocol specific code out of tcp_listener?

benbromhead:
[...]

PR #389, opened 2021-11-29
Add Wrapper::flush flag

rukai:
This is a step towards https://github.com/shotover/shotover-proxy/issues/385

This PR adds:
* A flush flag in Wrapper which indicates to transforms that they should immediately flush all buffered messages.
    + Support for this flag is added to the Coalesce transform
* coalesce shutdown logic that flushes all buffered messages.
*  Shutdown logic for sources and buffered chains that flushes all messages before dropping all their objects.

I originally planned to have a separate `Transform::shutdown` method that handles this but I had to ditch this approach because it required each transform to reimplement a lot of logic from their `Transform::transform` implementation. If for example the QueryTypeFilter transform was just simply passing on messages without filtering them that would introduce all sorts of bugs.

### DIFF
--- a/shotover-proxy/src/server.rs
+++ b/shotover-proxy/src/server.rs
@@ -244,6 +244,24 @@ impl<C: Codec + 'static> TcpCodecListener<C> {
         }
     }
 
+    pub async fn shutdown(&mut self) {
+        match self
+            .chain
+            .shutdown(Wrapper::new_with_chain_name(
+                vec![],
+                self.chain.name.clone(),
+            ))
+            .await
+        {
+            Ok(()) => {
+                info!("source {} was shutdown", self.source_name);
+            }
+            Err(e) => {
+                error!("shutdown chain processing error - {}", e);
+            }
+        }
+    }
+
     async fn create_listener(&self) -> Result<TcpListener> {
         TcpListener::bind(&self.listen_addr)
             .await
@@ -382,14 +400,8 @@ impl<C: Codec + 'static> Handler<C> {
     /// Request frames are read from the socket and processed. Responses are
     /// written back to the socket.
     ///
-    /// Currently, pipelining is not implemented. Pipelining is the ability to
-    /// process more than one request concurrently per connection without
-    /// interleaving frames. See for more details:
-    /// <https://redis.io/topics/pipelining>
-    ///
     /// When the shutdown signal is received, the connection is processed until
     /// it reaches a safe state, at which point it is terminated.
-    // #[instrument(skip(self))]
     pub async fn run(&mut self, stream: TcpStream) -> Result<()> {
         debug!("Handler run() started");
         // As long as the shutdown signal has not been received, try to read a
--- a/shotover-proxy/src/server.rs
+++ b/shotover-proxy/src/server.rs
@@ -247,18 +247,17 @@ impl<C: Codec + 'static> TcpCodecListener<C> {
     pub async fn shutdown(&mut self) {
         match self
             .chain
-            .shutdown(Wrapper::new_with_chain_name(
-                vec![],
-                self.chain.name.clone(),
-            ))
+            .process_request(
+                Wrapper::flush_with_chain_name(self.chain.name.clone()),
+                "".into(),
+            )
             .await
         {
-            Ok(()) => {
-                info!("source {} was shutdown", self.source_name);
-            }
-            Err(e) => {
-                error!("shutdown chain processing error - {}", e);
-            }
+            Ok(_) => info!("source {} was shutdown", self.source_name),
+            Err(e) => error!(
+                "source {} encountered an error when flushing the chain for shutdown: {}",
+                self.source_name, e
+            ),
         }
     }
 
--- a/shotover-proxy/src/transforms/chain.rs
+++ b/shotover-proxy/src/transforms/chain.rs
@@ -156,15 +156,16 @@ impl TransformChain {
                 );
 
                 match chain
-                    .shutdown(Wrapper::new_with_chain_name(vec![], chain.name.clone()))
+                    .process_request(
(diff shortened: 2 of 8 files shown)


shotover-proxy/src/server.rs
```diff
@@ -253,6 +253,24 @@ impl<C: Codec + 'static> TcpCodecListener<C> {
         }
     }
 
+    pub async fn shutdown(&mut self) {
+        match self
```
benbromhead, line 257: we probably want this to timeout if the whole chain is taking too long to shutdown (configurable)

shotover-proxy/src/transforms/mod.rs
```diff
@@ -475,6 +509,17 @@ pub trait Transform: Send {
     ///
     async fn transform<'a>(&'a mut self, message_wrapper: Wrapper<'a>) -> ChainResponse;
 
+    /// Called when shotover or buffered chain hosting the transform is shutting down.
+    /// The default implementation will automatically `call_next_shutdown` when non terminating.
+    /// So if you need to override this method to provide your own shutdown logic make sure to do the same.
+    async fn shutdown<'a>(&'a mut self, message_wrapper: Wrapper<'a>) -> ChainResponse {
```
benbromhead, line 515: probably worth documenting what to expect about messages that are included in the method call here (like are these messages that came in before / after the shutdown hook?), do you absolutely need to handle them? If so it could cause a double up in logic being implemented in shutdown or encouraging people to call transform from shutdown to handle the messages provided in the shutdown hook.

I would actually be ok with us dropping any messages that were in the buffer as part of shutdown. Pretty much all clients will end up getting a stream closed error or a timeout and have their own trety logic. So we could get rid of the message_wrapper param here

shotover-proxy/src/server.rs
```diff
@@ -253,6 +253,24 @@ impl<C: Codec + 'static> TcpCodecListener<C> {
...
+            .chain
+            .shutdown(Wrapper::new_with_chain_name(
+                vec![],
```
benbromhead, line 260: oh right we just pass an empty vec into the message_wrapper anyway. Let's remove that from the shutdown call

conorbros (approved):

shotover-proxy/src/transforms/mod.rs
benbromhead: We also don't need to do it the same way transform does. (where wrapper has the list of transforms to call next). At this point I don't envisage one transform needing to know about the results of the downchain transforms shutdown result or need to know about it in an ergonomic way (the reason we do the whole Wrapper thing for the transform method).

We could just call shutdown on the transform chain and just have it go through the chains transforms one by one (or map or whatever) and call shutdown.

shotover-proxy/src/server.rs
conorbros: Won't it just be SIGKILL'd if it takes too long

benbromhead: only if someone notices it's stuck... most workflows people will just see a stuck autoscaling group in AWS or a stuck stateful set in kubernetes or something

conorbros: Kubernetes will kill any processes that don't shut down in 30 seconds after a SIGTERM. Not sure about AWS but I would imagine it would do something similar if a process was refusing to shutdown.

benbromhead: ok happy to leave this up to an external scheduler to do

shotover-proxy/src/transforms/mod.rs
rukai: I hit a few issues and completely rewrote this part, I'll update the PR description with the full details.

shotover-proxy/src/transforms/mod.rs
```diff
@@ -280,6 +280,8 @@ pub struct Wrapper<'a> {
     transforms: Vec<&'a mut Transforms>,
     pub client_details: String,
     chain_name: String,
+    /// When true transforms must flush any buffered messages into the messages field
```
benbromhead, line 283: Let's expand the documentation here to explain why (e.g. a flush can be set to true on shutdown, or auth context changes)

rukai: "auth context change" doesnt sound quite right to me.
Does what I wrote instead make sense?

shotover-proxy/src/transforms/chain.rs
```diff
@@ -154,6 +154,20 @@ impl TransformChain {
                 debug!(
                     "buffered chain processing thread exiting, stopping chain loop and dropping"
                 );
+
```
benbromhead, line 157: We need to document now that transforms can get called with an empty list of messages on shutdown and flush set. If they don't expect it or handle it could lead to delayed shutdowns or bugs.

benbromhead:
LGTM - let's address the changes it brings to transform behaviour in docs and get this merged!

shotover-proxy/src/transforms/mod.rs
benbromhead: If I change the user I'm authenticated as in the middle of the session. Then we should probably flush all previous messages sent under the old auth context

shotover-proxy/src/transforms/chain.rs
rukai: Added a line to the Transform::transform docs

shotover-proxy/src/transforms/mod.rs
rukai: oh I see.
If its desirable to set flush at any time for protocol reasons then maybe we shouldnt try to exhaustively document the reasons.
What about:
`This can occur at any time but will always occur before the transform is destroyed due to either shotover or the transforms chain shutting down.`

benbromhead: sounds good

benbromhead (approved):

PR #390, opened 2021-11-29
Fix warnings that trigger in fridays upcoming rust 1.57 release

rukai:
You can see the warnings yourself by setting `channel = "beta"` in rust-toolchain.toml

I'm not sure why we have such an extensive packet parser just for one single test.
The fields on DnsPacket are giving unused warnings, the test doesnt make use of the dns parser so DnsPacket should just be entirely removed.
Although maybe we do want such powerful parsing in the future and the whole parser should be moved into a separate crate?

I removed the comment in `shotover-proxy/tests/helpers/mod.rs` I was expecting shotover-proxy/tests/helpers/mod.rs to make this code simpler but it didnt end up helping due to a mismatch in the values used by the PROFILE env var and the values accepted by `--profile`.

### DIFF
--- a/shotover-proxy/src/message/mod.rs
+++ b/shotover-proxy/src/message/mod.rs
@@ -87,14 +87,15 @@ pub struct RawMessage {
 
 #[derive(PartialEq, Debug, Clone)]
 pub enum ASTHolder {
-    SQL(Statement),
+    // Statement is boxed because Statement takes up a lot more stack space than Value.
+    SQL(Box<Statement>),
     Commands(Value), // A flexible representation of a structured query that will naturally convert into the required type via into/from traits
 }
 
 impl ASTHolder {
     pub fn get_command(&self) -> String {
         match self {
-            ASTHolder::SQL(statement) => match statement {
+            ASTHolder::SQL(statement) => match **statement {
                 Statement::Query(_) => "SELECT",
                 Statement::Insert { .. } => "INSERT",
                 Statement::Update { .. } => "UPDATE",
--- a/shotover-proxy/src/transforms/distributed/consistent_scatter.rs
+++ b/shotover-proxy/src/transforms/distributed/consistent_scatter.rs
@@ -66,7 +66,7 @@ fn get_timestamp(frag: &QueryResponse) -> i64 {
 }
 
 fn get_size(frag: &QueryResponse) -> usize {
-    frag.result.as_ref().map_or(0, |v| std::mem::size_of_val(v))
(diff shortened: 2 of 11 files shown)


shotover-proxy/src/message/mod.rs
```diff
@@ -87,14 +87,14 @@ pub struct RawMessage {
 
 #[derive(PartialEq, Debug, Clone)]
 pub enum ASTHolder {
-    SQL(Statement),
+    SQL(Box<Statement>),
```
conorbros, line 90: Can we add a comment explaining why this is a Box?

shotover-proxy/src/message/mod.rs
```diff
@@ -87,14 +87,14 @@ pub struct RawMessage {
...
     Commands(Value), // A flexible representation of a structured query that will naturally convert into the required type via into/from traits
 }
 
 impl ASTHolder {
     pub fn get_command(&self) -> String {
         match self {
-            ASTHolder::SQL(statement) => match statement {
+            ASTHolder::SQL(statement) => match &**statement {
```
conorbros, line 97: Is the borrow needed here?

conorbros (approved):
LGTM

shotover-proxy/src/transforms/distributed/consistent_scatter.rs
```diff
@@ -66,7 +66,7 @@ fn get_timestamp(frag: &QueryResponse) -> i64 {
 }
 
 fn get_size(frag: &QueryResponse) -> usize {
-    frag.result.as_ref().map_or(0, |v| std::mem::size_of_val(v))
+    frag.result.as_ref().map_or(0, std::mem::size_of_val)
```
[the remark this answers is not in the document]
rukai: nope, this line is still completely busted, just making clippy happy here.

shotover-proxy/src/message/mod.rs
rukai: yeah, the borrow is needed.
self is borrowed so we need to also borrow statement

rukai: wait, I tested it and it wasnt needed!

rukai: It seems the reference only needs to readded for `&mut`, not sure why :shrug:

benbromhead:
the packet parse code was largely grabbed from an example on using packet parse. the DNS stuff 100% doesn't need to be in it.

benbromhead (approved):

rukai:
Sweet, thats one less dependency then

PR #394, opened 2021-11-30
Cassandra keyspace int tests

conorbros:


shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1,37 +1,94 @@
...
-    assert_query_result(
-        &session,
+mod keyspace {
+    use cassandra_cpp::{stmt, Session};
+
+    use crate::cassandra_int_tests::{assert_query_result, ResultValue};
+
+    fn test_create_keyspace(session: &Session) {
+        assert_query_result(
+        session,
```
rukai, line 14: Theres some wonky formatting here, I think rustfmt is confused by the stmt! macro but if we manually fix the formatting here it wont override it.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1,37 +1,94 @@
...
+            ),
+            &[],
+        );
+    }
+
+    fn test_alter_keyspace(session: &Session) {
+        assert_query_result(session, stmt!("CREATE KEYSPACE IF NOT EXISTS alter_me WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 } AND DURABLE_WRITES = false;"), &[]);
+        assert_query_result(
+            session,
+            stmt!("ALTER  KEYSPACE alter_me WITH DURABLE_WRITES = true;"),
```
rukai, line 74: extra space

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1,37 +1,94 @@
...
+mod keyspace {
+    use cassandra_cpp::{stmt, Session};
+
+    use crate::cassandra_int_tests::{assert_query_result, ResultValue};
+
+    fn test_create_keyspace(session: &Session) {
+        assert_query_result(
+        session,
         stmt!(
             "CREATE KEYSPACE IF NOT EXISTS cycling WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 };"
```
rukai, line 16: is it possible to first delete the keyspace if it already exists?
Otherwise we might get false positives or just strange failures if a previous test case creates a keyspace with the same name.
Its still early days for these tests so happy to leave it for later if its not a simple fix.

rukai (commented):
Generally looks good, but a few minor concerns

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
conorbros: I'm thinking to change `assert_query_result` to accept a `str` for the statement and just use the macro there. Will make all the tests less verbose and fix these weird formatting issues.

rukai: I uh... dont think that will work because the macro needs a string literal not a string.
I could be wrong though, so double check.

rukai: oh, nice!

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
conorbros: I'll change the keyspace name `keyspace_tests` and similar for the other tests to prevent this. A bit of discipline with the test names we use should avoid any issues like that.

conorbros:
We cannot rely on the order of the rows so I've added `assert_query_result_contains`

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
rukai: Can we remove the `IF NOT EXISTS` so it blows up if it does happen to exist?

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1,37 +1,74 @@
...
-            &[ResultValue::Varchar("cycling".into())],
-            &[ResultValue::Varchar("system_auth".into())],
-            &[ResultValue::Varchar("system_schema".into())],
-            &[ResultValue::Varchar("system_distributed".into())],
-            &[ResultValue::Varchar("system".into())],
-            &[ResultValue::Varchar("system_traces".into())],
-        ],
-    );
+    fn test_create_keyspace(session: &Session) {
+        run_query(session, "CREATE KEYSPACE IF NOT EXISTS keyspace_tests_create WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 };");
```
rukai, line 15: I think we should use assert_query_result here as a sanity check

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -34,11 +36,21 @@ impl ResultValue {
...
 
-fn assert_query_result(session: &Session, query: Statement, expected_rows: &[&[ResultValue]]) {
-    let result = session.execute(&query).wait().unwrap();
-    let result_rows: Vec<Vec<ResultValue>> = result
+fn assert_query_result(session: &Session, query: &str, expected_rows: &[&[ResultValue]]) {
+    let result_rows = run_query(session, query);
+    assert_eq!(result_rows, expected_rows);
+}
+
+fn assert_query_result_contains(session: &Session, query: &str, row: &[ResultValue]) {
```
rukai, line 44: hmmm I initially thought this did an unordered comparison instead of checking for the existence of a single value.
Maybe we should name it assert_query_result_contains_row to make it obvious we are checking if a single row exists?

Or maybe we should replace it with a function that actually does do an unordered comparison.
That would force developers to make stronger assertions on the returned values.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
conorbros: Sanity check against what? Unwrapping on execute should catch any errors.

shotover-proxy/tests/cassandra_int_tests/mod.rs
conorbros: I think checking for the existence of a row is useful, lets us keep the test queries simple. `assert_query_result` should be checking unordered though so I will add that.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
rukai: against shotover returning values where it shouldnt.

conorbros: Ah good point! I'll fix that up

rukai: Not only does protect against bugs in shotover but helps aid people reading the tests to understand the functionality of the cassandra query.

Maybe we should have something like:
```rust
pub fn run_command(session: &Session, query: &str) {
    assert_query_result(session, query, &[]);
}
pub fn assert_query_result
fn run_query
```

shotover-proxy/tests/cassandra_int_tests/mod.rs
rukai: Hmmm, I think the broader we make our assertions the more cases we can catch where shotover is messing up.
But I realize that does get in the way a bit of the "test one thing" principle so I'll let you decide what you want to do here.

rukai: Oh right, looking at the `SELECT keyspace_name FROM system_schema.keyspaces;` style queries I can see the only sane option is to check for the existence of a specific result among many.

However for the delete case we do need to check that a specific result does NOT occur.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
conorbros: Yep I ended up with something similar

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -36,21 +36,39 @@ impl ResultValue {
...
+            if !result_rows.contains(item) {
+                panic!("{:?} missing from results", item);
+            }
+        });
+}
+
+/// Execute a `query` against the `session` and assert the result rows contain `row`
+fn assert_query_result_contains_row(session: &Session, query: &str, row: &[ResultValue]) {
+    let result_rows = execute_query(session, query);
+    assert!(result_rows.contains(&row.to_vec()));
```
rukai, line 68: We should craft an assert message that contains the expected row and the actual rows.

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -36,21 +36,39 @@ impl ResultValue {
...
     result
         .into_iter()
         .map(|x| x.into_iter().map(ResultValue::new).collect())
         .collect()
 }
+
+/// Execute a `query` against the `session` and assert that the result rows match `expected_rows`
+fn assert_query_result(session: &Session, query: &str, expected_rows: &[&[ResultValue]]) {
+    let result_rows = execute_query(session, query);
+    assert!(expected_rows.len() == result_rows.len());
```
rukai, line 52: assert_eq

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -34,11 +36,39 @@ impl ResultValue {
...
+        .iter()
+        .map(|x| x.to_vec())
+        .collect::<Vec<Vec<ResultValue>>>()
+        .iter()
+        .for_each(|item| {
+            if !result_rows.contains(item) {
+                panic!("{:?} missing from results", item);
+            }
+        });
+}
```
rukai, line 63: This is a simpler way to express the same thing:
```rust
/// Execute a `query` against the `session` and assert that the result rows match `expected_rows`
fn assert_query_result(session: &Session, query: &str, expected_rows: &[&[ResultValue]]) {
    let result_rows = execute_query(session, query);
    assert!(expected_rows.len() == result_rows.len());
    for item in expected_rows {
        if !result_rows.contains(&item.to_vec()) {
            panic!("{:?} missing from results", item);
        }
    }
}
```

But I think we can do better:
```rust
// first turn ResultValue into derive soup
#[derive(Debug, Clone, PartialOrd, Ord, PartialEq, Eq)]
enum ResultValue {
    Text(String),
    Varchar(String),
    Int(i64),
    Boolean(bool),
}

/// Execute a `query` against the `session` and assert that the result rows match `expected_rows`
fn assert_query_result(session: &Session, query: &str, expected_rows: &[&[ResultValue]]) {
    let mut result_rows = execute_query(session, query);
    result_rows.sort();

    let mut expected_rows: Vec<_> = expected_rows.iter().map(|x| x.to_vec()).collect();
    expected_rows.sort();

    assert_eq!(result_rows, expected_rows);
}
```

This is a little simpler but most importantly provides a better error message.

rukai (changes requested):

rukai (approved):
lgtm

benbromhead (approved):
lgtm

PR #397, opened 2021-12-02
Allow Cassandra passthrough to be configured in docker-compose.yml 

conorbros:
This change allows us to use docker compose volume mounting for files the example directory. 

E.g. We can just mount the PWD in `docker-compose.yml`

```YAML
volumes:
    - ./docker-entrypoint.sh:/usr/local/bin/docker-entrypoint.sh
```

I've also included the change to the `cassandra-passthrough` example that I needed this change for. User defined functions need to be enabled in Cassandra to test them but they can't be set from the docker-compose so we need to modify the `docker-entrypoint.sh` to do this.

The `docker-entrypoint.sh` is mostly the same as the [default](https://github.com/docker-library/cassandra/blob/master/3.11/docker-entrypoint.sh) but I added `enable_user_defined_functions` and `enable_scripted_user_defined_functions` to the `for yaml in \` loop.

I considered opening a PR to enable this on the official docker image but I found this [comment](https://github.com/docker-library/cassandra/pull/122#issuecomment-426774071) explaining why they don't want to enable this.

### DIFF
--- a/shotover-proxy/examples/cassandra-passthrough/docker-compose.yml
+++ b/shotover-proxy/examples/cassandra-passthrough/docker-compose.yml
@@ -8,3 +8,7 @@ services:
       MAX_HEAP_SIZE: "128M"
       MIN_HEAP_SIZE: "128M"
       HEAP_NEWSIZE: "24M"
+      CASSANDRA_ENABLE_SCRIPTED_USER_DEFINED_FUNCTIONS: "true"
+      CASSANDRA_ENABLE_USER_DEFINED_FUNCTIONS: "true"
+    volumes:
+      - ./docker-entrypoint.sh:/usr/local/bin/docker-entrypoint.sh
--- a/shotover-proxy/examples/cassandra-passthrough/docker-entrypoint.sh
+++ b/shotover-proxy/examples/cassandra-passthrough/docker-entrypoint.sh
@@ -1,6 +1,10 @@
 #!/bin/bash
 set -e
 
+###  https://github.com/docker-library/cassandra/blob/master/3.11/docker-entrypoint.sh
+###  This file has been copied from the above link and modified to include enable_user_defined_functions and enable_scripted_user_defined_functions to the for yaml in \ loop
+###  so extra cassandra config options can be modified from the docker-compose.yml
+
 # first arg is `-f` or `--some-option`
 # or there are no args
 if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
--- a/shotover-proxy/examples/cassandra-passthrough/docker-entrypoint.sh
+++ b/shotover-proxy/examples/cassandra-passthrough/docker-entrypoint.sh
@@ -1,0 +1,96 @@
+#!/bin/bash
+set -e
+
+# first arg is `-f` or `--some-option`
+# or there are no args
+if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
+	set -- cassandra -f "$@"
+fi
+
+# allow the container to be started with `--user`
+if [ "$1" = 'cassandra' -a "$(id -u)" = '0' ]; then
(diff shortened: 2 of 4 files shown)


shotover-proxy/examples/cassandra-passthrough/docker-entrypoint.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
```
benbromhead, line 1: this file looks like we've taken it from the docker image we pull in. Do you want to check whether we need to add anything to our NOTICES file from a license perspective?

benbromhead (approved):

shotover-proxy/examples/cassandra-passthrough/docker-entrypoint.sh
```diff
@@ -0,0 +1,96 @@
...
+set -e
```
rukai, line 2: Can you also add a comment at the top describing how this differs from the default file?

shotover-proxy/examples/cassandra-passthrough/docker-compose.yml
```diff
@@ -8,3 +8,7 @@ services:
       MAX_HEAP_SIZE: "128M"
       MIN_HEAP_SIZE: "128M"
       HEAP_NEWSIZE: "24M"
+      CASSANDRA_ENABLE_SCRIPTED_USER_DEFINED_FUNCTIONS: "true"
+      CASSANDRA_ENABLE_USER_DEFINED_FUNCTIONS: "true"
+    volumes:
+      - ./docker-entrypoint.sh:/usr/local/bin/docker-entrypoint.sh
```
rukai, line 14: Cant we just specify the path here relative to the default CWD instead of changing the CWD?

rukai (changes requested):

shotover-proxy/examples/cassandra-passthrough/docker-compose.yml
conorbros: I frequently run the examples from within the directory and I would assume users will attempt the same. Also if you just download the example files to run them the paths won't be right.

rukai: ah! good point

rukai: Wait, uh... how does the redis-passthrough work?
It is already doing the same volume thing.

shotover-proxy/examples/cassandra-passthrough/docker-entrypoint.sh
conorbros: @justinmclean

shotover-proxy/examples/cassandra-passthrough/docker-compose.yml
conorbros: ...interesting. I thought the volume mounting worked relative to the processes CWD not the directory of the `docker-compose.yml`.

conorbros: Reverted the changes.

shotover-proxy/examples/cassandra-passthrough/docker-entrypoint.sh
conorbros: Addressed this btw

rukai (approved):
This can be merged once you've resolved the licensing issues.
One day we might want to change this to directly provide the yaml file instead of generating it in bash.
But for now this is fine.

conorbros:
Sounds good. BTW it won't merge until that  conversation is resolved so it's safe to enable auto-merge.

shotover-proxy/examples/cassandra-passthrough/docker-entrypoint.sh
conorbros: Added the changes to LICENSE

PR #399, opened 2021-12-02
Cassandra functions int tests

conorbros:


rukai (approved):
Might be worth combining drop_function and create_function into a single function, as drop_function is pretty much a superset of create_function.

Im also happy to merge as is though.

rukai:
What did you change?

conorbros:
I removed creating the function in drop and just reused the one from the created test.

benbromhead (approved):

PR #402, opened 2021-12-03
Clean up BufferedChain API

conorbros:
- We never use the client details in `BufferedChain` so I have removed it
- Make an inner method private so it's clear you should use the other 2 only


rukai (approved):
LGTM

Note for the second reviewer:
Wrapper already has a client_details field, so if we ever want to use it we should just use it from there.

XA21X (approved):

Issue #270, opened 2021-10-13
bench CI explodes when a new benchmark is added.


rukai:
## To Reproduce
Make a PR that adds or renames a benchmark

thread 'main' panicked at 'Baseline 'master' must exist before comparison is allowed; try --save-baseline master', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/criterion-0.3.5/src/analysis/mod.rs:56:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b/library/std/src/panicking.rs:515:5
   1: std::panicking::begin_panic_fmt
             at /rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b/library/std/src/panicking.rs:457:5
   2: criterion::analysis::common
   3: redis_benches::redis
   4: redis_benches::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Please include any configuration for setting up services such as `docker-compose` that the `topology.yaml` depends on.

PR #403, opened 2021-12-03
Fix CI failures when a new benchmark is added

rukai:
closes https://github.com/shotover/shotover-proxy/issues/270

~~Waiting on https://github.com/bheisler/criterion.rs/pull/532 to be merged~~

The problem was caused because criterion expects a baseline to exist for every benchmark.
This PR solves the problem by making use of a new criterion feature (https://github.com/bheisler/criterion.rs/pull/532) that silently ignores missing baselines.

### DIFF
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -67,7 +67,7 @@ csv = "1.1.6"
 strum_macros = "0.23"
 
 [dev-dependencies]
-criterion = { version = "0.3", features = ["async_tokio", "html_reports"] }
+criterion = { git = "https://github.com/bheisler/criterion.rs", branch = "version-0.4", version = "0.3", features = ["async_tokio", "html_reports"] }
 redis = { version = "0.21.0", features = ["tokio-comp", "cluster"] }
 pcap = "0.9.0"
 pktparse = {version = "0.7.0", features = ["serde"]}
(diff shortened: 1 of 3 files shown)


rukai:
I suspect it will be a while before criterion 0.4 releases so lets land this PR with the git dependency.

shotover-proxy/Cargo.toml
```diff
@@ -63,7 +63,7 @@ csv = "1.1.6"
 strum_macros = "0.23"
 
 [dev-dependencies]
-criterion = { version = "0.3", features = ["async_tokio", "html_reports"] }
+criterion = { git = "https://github.com/bheisler/criterion.rs", branch = "version-0.4", version = "0.3", features = ["async_tokio", "html_reports"] }
```
benbromhead, line 66: Is it worth just waiting for the crate to get published or is this blocking something urgent?

rukai: We can work around the issue by temporarily disabling merge protection and then merging while CI is red.
Not a great habit to form though.

I'm not currently writing a benchmark writing benchmarks is something I expect people to be doing all the time.

conorbros: Happy to just use Git for now. I'll watch [this PR](https://github.com/bheisler/criterion.rs/pull/496) so we can upgrade once it's released.

conorbros (approved):

benbromhead (approved):

PR #405, opened 2021-12-06
Fix QueryTypeFilter on main chain

rukai:
Fix QueryTypeFilter and DebugReturner so that they always return the same number of messages that they receive.
The only reason this hasnt been a problem yet is because in production we only use QueryTypeFilter in an ignore tee branch where breaking this invariant has no consequences.

DebugReturner's Response::Message case could also cause this problem but thats only used internally for unit tests and this PR is large enough already so it should be investigated separately.

### DIFF
--- a/shotover-proxy/src/message/mod.rs
+++ b/shotover-proxy/src/message/mod.rs
@@ -1,6 +1,10 @@
 use crate::protocols::RawFrame;
 use bytes::Bytes;
+use cassandra_protocol::frame::frame_error::{AdditionalErrorInfo, ErrorBody};
 use cassandra_protocol::frame::frame_result::{ColSpec, ColType};
+use cassandra_protocol::frame::{
+    Direction, Flags, Frame as CassandraFrame, Opcode, Serialize as CassandraSerialize,
+};
 use cassandra_protocol::types::data_serialization_types::{
     decode_ascii, decode_bigint, decode_boolean, decode_double, decode_float, decode_inet,
     decode_int, decode_smallint, decode_tinyint, decode_varchar,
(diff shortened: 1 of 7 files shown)


shotover-proxy/src/message/mod.rs
```diff
@@ -72,6 +76,32 @@ impl Message {
             original: RawFrame::None,
         }
     }
+
+    pub fn to_filtered_reply(&self) -> Message {
+        Message {
+            details: MessageDetails::Unknown,
+            modified: true,
+            original: match &self.original {
+                RawFrame::Redis(_) => RawFrame::Redis(Frame::Null),
```
conorbros, line 85: Why doesn't redis get an error?

rukai:
The advantages of redis replying error:
* Consistent with cassandra which has no choice but to return error as there is no consistently correct response
* Gives an obvious indication to the user that the message was filtered.

The advantages of redis replying null:
* Its not actually an error but an expected case
* more performant?

So... I think it would be better as an error, ill change it tomorrow.
We can always change it back later if we find it has important performance implications.

conorbros:
Apologies I deleted the comment to think about it more, but yeah that sounds good.

shotover-proxy/tests/transforms/query_type_filter.rs
```diff
@@ -0,0 +1,30 @@
+use crate::helpers::ShotoverManager;
+use serial_test::serial;
+
+#[tokio::test(flavor = "multi_thread")]
+#[serial]
+async fn test_query_type_filter() {
```
conorbros, line 6: Can we try pipelining too?

shotover-proxy/src/message/mod.rs
```diff
@@ -72,6 +76,32 @@ impl Message {
             original: RawFrame::None,
         }
     }
+
+    pub fn to_filtered_reply(&self) -> Message {
```
conorbros, line 80: To me this is leaking a transform specific implementation into `Message`. We won't be using `to_filtered_reply` anywhere but the `QueryTypeFilter` transform so I think it should be there.

shotover-proxy/tests/transforms/query_type_filter.rs
conorbros: Assuming each pipeline message gets represented as a separate message in shotover

shotover-proxy/src/message/mod.rs
rukai: reasons for putting it in Message:
* Any transform that filters out messages can make use of it.
* The transform doesnt have to manually deal with each protocol type.

But I dont mind too much either way.

shotover-proxy/tests/transforms/query_type_filter.rs
rukai: you were correct!
I added a test case for this very useful case.

conorbros (approved):

benbromhead:
Happy to do this, but I would suggest maybe a "return_type_on_filter" config that can either be NULL, ERROR or NO_RETURN and make this user configurable

benbromhead:
Let's create a new issue to track making this configurable... I'm not going to block the ticket on this... but something we should definitely do

benbromhead (approved):

PR #406, opened 2021-12-07
Fix redis auth issue

rukai:
closes https://github.com/shotover/shotover-proxy/issues/385
Solves the problem by taking ben's suggested chain_to_clone solution in the tee transform.

The redis-cluster-dr example was extended to use authentication instead of creating an entirely new test to allow for testing this case without a large hit to test runtime.

### DIFF
--- a/shotover-proxy/src/transforms/load_balance.rs
+++ b/shotover-proxy/src/transforms/load_balance.rs
@@ -102,7 +102,7 @@ mod test {
             ),
         });
 
-        let chain = TransformChain::new(vec![transform], "test".to_string());
+        let mut chain = TransformChain::new(vec![transform], "test".to_string());
 
         for _ in 0..90 {
             let r = chain
@@ -110,6 +110,18 @@ mod test {
                 .process_request(Wrapper::new(Messages::new()), "test_client".to_string())
                 .await;
             assert!(r.is_ok());
+        }
+
+        match chain.chain.remove(0) {
+            Transforms::PoolConnections(p) => {
+                let guard = p.other_connections.lock().await;
+                assert_eq!(guard.len(), 3);
+                for bc in guard.iter() {
+                    let guard = bc.count.lock().await;
+                    assert_eq!(*guard, 30);
+                }
+            }
+            _ => panic!("whoops"),
         }
 
         Ok(())
--- a/shotover-proxy/src/transforms/load_balance.rs
+++ b/shotover-proxy/src/transforms/load_balance.rs
@@ -102,7 +102,7 @@ mod test {
             ),
         });
 
-        let mut chain = TransformChain::new(vec![transform], "test".to_string());
+        let chain = TransformChain::new(vec![transform], "test".to_string());
 
         for _ in 0..90 {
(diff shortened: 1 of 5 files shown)


benbromhead (approved):
High level LGTM - will need a nit based review :)

rukai:
@benbromhead I played with this some more and implementing the custom clone in BufferedChain seems way better. So I went ahead and pushed that change, does it look good to you?

shotover-proxy/src/transforms/load_balance.rs
```diff
@@ -118,7 +118,7 @@ mod test {
                 assert_eq!(guard.len(), 3);
                 for bc in guard.iter() {
                     let guard = bc.count.lock().await;
-                    assert_eq!(*guard, 30);
+                    assert_eq!(*guard, 0);
```
benbromhead, line 121: This now seems like we have changed the behaviour of load balance? The count of each balanced chain should be 30 (90 ops split amongst three connections).

In load balance we are deliberately sharing the sender handles across multiple chains

benbromhead: this could be solved with a get_handle method or something and using that in load_balance

rukai: image
I dont think the load balance implementation ever achieved something useful.
once self.active_connection is Some it can never be changed to another connection.

rukai: This transform is also not exposed in our public API yet.

rukai: I removed the assert from the test entirely:
* It would be confusing to have it `assert_eq(*guard, 0)` because that makes it seem like the correct behaviour
* Testing implementation details like this is not great

rukai: That said I am also happy to just completely revert this 2nd commit and keep all the changes localized to the Tee transform.

rukai: Ooooooh, is the idea behind ConnectionBalanceAndPool that it uses a new connection every time its cloned?

conorbros (approved):

shotover-proxy/src/transforms/load_balance.rs
rukai: I made this PR to document how ConnectionBalanceAndPool works https://github.com/shotover/shotover-proxy/pull/411

rukai:
I'm marking as draft because I want to investigate implementing the new BufferedChain clone functionality into a separate method to avoid breaking ConnectionBalanceAndPool

shotover-proxy/src/transforms/load_balance.rs
benbromhead: Let's revert and pull this out for another PR -> Let's just fix the broken Tee / Auth workflow issue

rukai:
Alright this is the last switcharoo, I promise.
Its functionally identical to the original implementation, just moves the `into_buffered_chain` calls from tee into a method on BufferedChain

PR #404, opened 2021-12-03
Cassandra native types tests

conorbros:
- changes to `Message` for parsing cassandra types (move to separate PR?)
- Tests for Cassandra native types (date, int, ascii, etc)
- Started tests for Cassandra collections: only have list at the moment and am not testing selects yet as the `ResultValue` needs to have stuff added for parsing collections.

PR #409, opened 2021-12-08
Value support for Cassandra types

conorbros:
Pulling some changes out of #404 

### Done

- Support all native and collection Cassandra types in Value
- Add todos for conversions not yet implemented.

### DIFF
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -54,7 +54,8 @@ halfbrown = "0.1.11"
 
 # Transform dependencies
 redis-protocol = "3.0.1"
-cassandra-protocol = { git = "https://github.com/krojew/cdrs-tokio" }
+#cassandra-protocol = { git = "https://github.com/krojew/cdrs-tokio" }
+cassandra-protocol = { git = "https://github.com/conorbros/cdrs-tokio", branch = "typed-collections" }
 rdkafka = "0.28"
 crc16 = "0.4.0"
 ordered-float = { version = "2.8.0", features = ["serde"] }
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -54,8 +54,7 @@ halfbrown = "0.1.11"
 
 # Transform dependencies
 redis-protocol = "3.0.1"
-#cassandra-protocol = { git = "https://github.com/krojew/cdrs-tokio" }
-cassandra-protocol = { git = "https://github.com/conorbros/cdrs-tokio", branch = "typed-collections" }
+cassandra-protocol = { git = "https://github.com/krojew/cdrs-tokio" }
 rdkafka = "0.28"
 crc16 = "0.4.0"
 ordered-float = { version = "2.8.0", features = ["serde"] }
--- a/shotover-proxy/src/message/mod.rs
+++ b/shotover-proxy/src/message/mod.rs
@@ -325,7 +325,6 @@ pub enum Value {
     None,
     #[serde(with = "my_bytes")]
     Bytes(Bytes),
-    Bigint(BigInt),
     Ascii(String),
     Strings(String),
     Integer(i64, IntSize),
@@ -389,7 +388,6 @@ impl From<&Frame> for Value {
 impl From<Value> for Frame {
     fn from(value: Value) -> Frame {
         match value {
-            Value::Bigint(_) => unimplemented!(),
             Value::NULL => Frame::Null,
             Value::None => unimplemented!(),
             Value::Bytes(b) => Frame::BulkString(b.to_vec()),
@@ -574,7 +572,7 @@ impl Value {
     fn create_element(element: CassandraType) -> Value {
         match element {
             CassandraType::Ascii(a) => Value::Ascii(a),
-            CassandraType::Bigint(b) => Value::Bigint(b),
+            CassandraType::Bigint(b) => Value::Integer(b, IntSize::I64),
             CassandraType::Blob(b) => Value::Bytes(b.into_vec().into()),
             CassandraType::Boolean(b) => Value::Boolean(b),
             CassandraType::Counter(c) => Value::Counter(c),
@@ -649,7 +647,6 @@ impl Value {
 
     pub fn into_str_bytes(self) -> Bytes {
         match self {
-            Value::Bigint(_) => unimplemented!(),
             Value::NULL => Bytes::from("".to_string()),
             Value::None => Bytes::from("".to_string()),
             Value::Bytes(b) => b,
@@ -683,7 +680,6 @@ impl Value {
 
     pub fn into_bytes(self) -> Bytes {
         match self {
-            Value::Bigint(_) => unimplemented!(),
             Value::NULL => Bytes::new(),
             Value::None => Bytes::new(),
             Value::Bytes(b) => b,
@@ -726,7 +722,6 @@ impl Value {
 impl From<Value> for cassandra_protocol::types::value::Bytes {
     fn from(value: Value) -> cassandra_protocol::types::value::Bytes {
         match value {
-            Value::Bigint(_) => unimplemented!(),
             Value::NULL => (-1).into(),
             Value::None => cassandra_protocol::types::value::Bytes::new(vec![]),
             Value::Bytes(b) => cassandra_protocol::types::value::Bytes::new(b.to_vec()),
--- a/shotover-proxy/src/message/mod.rs
+++ b/shotover-proxy/src/message/mod.rs
@@ -389,7 +389,7 @@ impl From<Value> for Frame {
     fn from(value: Value) -> Frame {
         match value {
             Value::NULL => Frame::Null,
-            Value::None => unimplemented!(),
+            Value::None => todo!(),
             Value::Bytes(b) => Frame::BulkString(b.to_vec()),
             Value::Strings(s) => Frame::SimpleString(s),
             Value::Integer(i, _) => Frame::Integer(i),
@@ -398,24 +398,24 @@ impl From<Value> for Frame {
             Value::Inet(i) => Frame::SimpleString(i.to_string()),
             Value::List(l) => Frame::Array(l.into_iter().map(|v| v.into()).collect()),
             Value::Rows(r) => Frame::Array(r.into_iter().map(|v| Value::List(v).into()).collect()),
-            Value::NamedRows(_) => unimplemented!(),
-            Value::Document(_) => unimplemented!(),
+            Value::NamedRows(_) => todo!(),
+            Value::Document(_) => todo!(),
             Value::FragmentedResponse(l) => Frame::Array(l.into_iter().map(|v| v.into()).collect()),
-            Value::Ascii(a) => Frame::SimpleString(a),
-            Value::Double(d) => Frame::SimpleString(d.to_string()),
-            Value::Set(s) => Frame::Array(s.into_iter().map(|v| v.into()).collect()),
-            Value::Map(_) => unimplemented!(),
-            Value::Varint(v) => Frame::SimpleString(v.to_string()),
-            Value::Decimal(d) => Frame::SimpleString(d.to_string()),
-            Value::Date(date) => Frame::Integer(date.into()),
-            Value::Timestamp(timestamp) => Frame::Integer(timestamp),
-            Value::Timeuuid(timeuuid) => Frame::SimpleString(timeuuid.to_hyphenated().to_string()),
-            Value::Varchar(v) => Frame::SimpleString(v),
-            Value::Uuid(uuid) => Frame::SimpleString(uuid.to_hyphenated().to_string()),
-            Value::Time(t) => Frame::Integer(t),
-            Value::Counter(c) => Frame::Integer(c),
-            Value::Tuple(_) => unimplemented!(),
-            Value::Udt(_) => unimplemented!(),
+            Value::Ascii(_a) => todo!(),
+            Value::Double(_d) => todo!(),
+            Value::Set(_s) => todo!(),
+            Value::Map(_) => todo!(),
+            Value::Varint(_v) => todo!(),
+            Value::Decimal(_d) => todo!(),
+            Value::Date(_date) => todo!(),
+            Value::Timestamp(_timestamp) => todo!(),
+            Value::Timeuuid(_timeuuid) => todo!(),
+            Value::Varchar(_v) => todo!(),
+            Value::Uuid(_uuid) => todo!(),
+            Value::Time(_t) => todo!(),
+            Value::Counter(_c) => todo!(),
+            Value::Tuple(_) => todo!(),
+            Value::Udt(_) => todo!(),
         }
     }
 }
@@ -512,7 +512,7 @@ impl Value {
 
                         Value::Udt(typed_udt)
                     } else {
-                        panic!("not a udt")
+                        panic!("Not a UDT. This indicates a bug in cassandra_protocol")
                     }
                 }
                 ColType::Tuple => {
@@ -529,13 +529,11 @@ impl Value {
                         let typed_tuple = Value::create_tuple(tuple);
                         Value::Tuple(typed_tuple)
                     } else {
-                        panic!("not a tuple") // TODO
+                        panic!("Not a Tuple. This indicates a bug in cassandra_protocol")
                     }
                 }
                 ColType::Custom => unimplemented!(),
                 ColType::Null => Value::NULL,
-                // TODO: process collection types based on ColTypeOption
-                // (https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v4.spec#L569)
             }
         } else {
             Value::NULL
@@ -551,7 +549,7 @@ impl Value {
 
             values
         } else {
-            panic!("not a udt");
+            panic!("Not a UDT. Only CassandraType::Udt should be passed to this method.");
         }
     }
 
@@ -565,7 +563,7 @@ impl Value {
 
             value_list
         } else {
-            panic!("this is not a map");
+            panic!("Not a Map. Only CassandraType::Map should be passed to this method");
         }
     }
 
@@ -612,7 +610,7 @@ impl Value {
 
                 value_list
             }
-            _ => panic!("this is not a list"), // TODO handle this better
+            _ => panic!("Not a List. Only CassandraType::List should be passed to this method."),
         }
     }
 
@@ -627,7 +625,9 @@ impl Value {
 
                 value_list
             }
-            _ => panic!("this is not a set"), // TODO handle this better
+            _ => panic!(
+                "Not a List or Set. Only CassandraType::List or CassandraType::Set should be passed to this method."
+            ),
         }
     }
 
@@ -641,7 +641,7 @@ impl Value {
 
                 value_list
             }
-            _ => panic!("this is not a tuple"),
+            _ => panic!("Not a Tuple. Only CassandraType::Tuple should be passed to this method."),
         }
     }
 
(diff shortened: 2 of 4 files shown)


shotover-proxy/src/message/mod.rs
```diff
@@ -345,26 +350,33 @@ impl Value {
...
                 ColType::Ascii => Value::Strings(decode_ascii(actual_bytes).unwrap()),
                 ColType::Bigint => Value::Integer(decode_bigint(actual_bytes).unwrap()),
-                ColType::Blob => Value::Bytes(Bytes::copy_from_slice(actual_bytes)),
                 ColType::Boolean => Value::Boolean(decode_boolean(actual_bytes).unwrap()),
-                ColType::Counter => Value::Integer(decode_int(actual_bytes).unwrap() as i64),
-                ColType::Decimal => unimplemented!("We dont have a decimal type yet"),
+                ColType::Counter | ColType::Int => {
+                    Value::Integer(decode_int(actual_bytes).unwrap() as i64)
+                }
+                ColType::Decimal => todo!(),
```
conorbros, line 357: Do we make another `Value` variant for Decimal or store it in the float one?

shotover-proxy/src/message/mod.rs
```diff
@@ -345,26 +350,33 @@ impl Value {
...
                 ColType::Double => Value::Float(decode_double(actual_bytes).unwrap()),
                 ColType::Float => Value::Float(decode_float(actual_bytes).unwrap() as f64),
-                ColType::Int => Value::Integer(decode_int(actual_bytes).unwrap() as i64),
-                ColType::Uuid => Value::Bytes(Bytes::copy_from_slice(actual_bytes)),
                 ColType::Varchar => Value::Strings(decode_varchar(actual_bytes).unwrap()),
-                ColType::Varint => unimplemented!("We dont have a varint type yet"),
-                ColType::Timeuuid => Value::Bytes(Bytes::copy_from_slice(actual_bytes)),
+                ColType::Varint => todo!(),
```
conorbros, line 361: Do we make another Value variant for Decimal or store it in the int one?

shotover-proxy/src/message/mod.rs
```diff
@@ -345,26 +350,33 @@ impl Value {
...
+                ColType::Timeuuid | ColType::Blob | ColType::Uuid => {
+                    Value::Bytes(Bytes::copy_from_slice(actual_bytes))
+                }
                 ColType::Inet => Value::Inet(decode_inet(actual_bytes).unwrap()),
-                ColType::Timestamp => Value::NULL,
-                ColType::Date => Value::NULL,
-                ColType::Time => Value::NULL,
+                ColType::Date | ColType::Timestamp | ColType::Time => Value::NULL, // TODO
```
conorbros, line 366: Store Timestamp in the `Bytes` variant?

conorbros, line 366: Date and Time are represented as integers in Cassandra, should they be stored in the Int variant or make a separate one

shotover-proxy/src/message/mod.rs
```diff
@@ -281,13 +285,12 @@ pub enum Value {
     NamedRows(Vec<HashMap<String, Value>>),
     Document(HashMap<String, Value>),
```
conorbros, line 286: Should `Document` and `Map` be merged?

shotover-proxy/src/message/mod.rs
rukai: Decimal can not be losslessly stored as a float.
So I think our only option is to create a Decimal variant.

shotover-proxy/src/message/mod.rs
rukai: If we stored them as ints I dont think we would have a way of telling that they were originally dates.

shotover-proxy/src/message/mod.rs
conorbros: In that case we can use https://docs.rs/bigdecimal/latest/bigdecimal/index.html

rukai:
I think the questions we really need to answer before trying to resolve these issues is:
* Does the data represented in Value need to be able to read without losing precision/accuracy?
* And a stronger requirement: Do we need to be able to roundtrip all data through `Value`?

shotover-proxy/src/message/mod.rs
benbromhead: That would make sense as a Document is a subset of Map

shotover-proxy/src/message/mod.rs
benbromhead: Yup we need a lossless Decimal type (probably a bigint as well)

shotover-proxy/src/message/mod.rs
benbromhead: Let's look at what the Java implementation of Varint is and see if there is a stdlib one in Rust or if we need to implement our own / import a lib

shotover-proxy/src/message/mod.rs
benbromhead: I'm happy for the discriminate on the enum to be the type tag here, but we probably need to make sure we have enough fidelity to represent the most precise version of this... and it's something easily convertible to other time formats

benbromhead:
To answer @rukai questions. Yes to both. 

We should be able to losslesly go from arbitrary Redis <=> Value and Cassandra <=> Value (thus Redis <=> Cassandra)

benbromhead:
This will result in some "less than optimal" decisions. E.g. storing certain types that use more memory than needed. E.g. if one datastores Int type is 32 bits and another is 64 bits.

For the moment, I would just continue down the path of conversion to a common Value type and using that as common intermediary. For some context, however, @Claude-at-Instaclustr and I have also had some discussions about a better way of representing internal types and presenting a nicer way to reason about them. Mainly using a resource descriptor framework (RDF triple). An RDF won't solve the above two questions, but it will provide a better way of reasoning about types... potentially. 

See https://ieeexplore.ieee.org/document/7226706 for some background reading.

Claude-at-Instaclustr:
A better link to the paper: https://arxiv.org/abs/1410.4977

shotover-proxy/src/message/mod.rs
conorbros: Will follow up with this in a later PR when I have a better idea of what parsing and storing a cassandra map into value will look like.

shotover-proxy/src/message/mod.rs
```diff
@@ -377,26 +400,39 @@ impl Value {
             match spec.col_type.id {
                 ColType::Ascii => Value::Strings(decode_ascii(actual_bytes).unwrap()),
                 ColType::Bigint => Value::Integer(decode_bigint(actual_bytes).unwrap()),
-                ColType::Blob => Value::Bytes(Bytes::copy_from_slice(actual_bytes)),
                 ColType::Boolean => Value::Boolean(decode_boolean(actual_bytes).unwrap()),
-                ColType::Counter => Value::Integer(decode_int(actual_bytes).unwrap() as i64),
-                ColType::Decimal => unimplemented!("We dont have a decimal type yet"),
+                ColType::Counter | ColType::Int => {
```
rukai, line 404: We need to be able to differentiate between Counter and Int in order to be lossless.
We can just create a Value::Counter for now and refactor once we add support for a second SQL-like db

shotover-proxy/src/message/mod.rs
```diff
@@ -377,26 +400,39 @@ impl Value {
...
-                ColType::Time => Value::NULL,
+                ColType::Date => Value::Date(decode_date(actual_bytes).unwrap()),
+                ColType::Timestamp => Value::Timestamp(decode_timestamp(actual_bytes).unwrap()),
+                ColType::Time => Value::Time(decode_time(actual_bytes).unwrap()),
                 ColType::Smallint => Value::Integer(decode_smallint(actual_bytes).unwrap() as i64),
                 ColType::Tinyint => Value::Integer(decode_tinyint(actual_bytes).unwrap() as i64),
-                // TODO: process collection types based on ColTypeOption
-                // (https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v4.spec#L569)
-                _ => Value::NULL,
+                ColType::Custom => Value::Strings(decode_custom(actual_bytes).unwrap()),
```
rukai, line 425: Collides with ColType::Varchar

shotover-proxy/src/message/mod.rs
```diff
@@ -434,27 +483,61 @@ impl Value {
                 IpAddr::V4(four) => Vec::from(four.octets()),
                 IpAddr::V6(six) => Vec::from(six.octets()),
             }),
-            _ => unimplemented!(),
+            Value::List(_) => todo!(),
```
rukai, line 486: Lets delete into_bytes instead.
Its unused and I have no idea what its intended use is.
In the current implementation there is no way to revert the data back as there is no tag to identify what the original type was.

rukai (changes requested):
Some initial feedback, I'll look into the BigInt and BigDecimal stuff deeper tomorrow

shotover-proxy/src/message/mod.rs
```diff
@@ -359,6 +373,15 @@ impl From<Value> for Frame {
             Value::NamedRows(_) => unimplemented!(),
             Value::Document(_) => unimplemented!(),
             Value::FragmentedResponse(l) => Frame::Array(l.into_iter().map(|v| v.into()).collect()),
+            Value::Set(_) => todo!(),
+            Value::Map(_) => todo!(),
+            Value::Varint(v) => Frame::SimpleString(v.to_string()),
```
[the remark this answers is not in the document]
conorbros: Once we have tests we will find out what parts don't work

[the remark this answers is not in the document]
rukai: In that case I guess I wont bother reviewing the conversions themselves I'll just make sure the types we are going between are correct.

shotover-proxy/src/message/mod.rs
```diff
@@ -313,13 +321,20 @@ pub enum Value {
     NamedRows(Vec<HashMap<String, Value>>),
     Document(HashMap<String, Value>),
     FragmentedResponse(Vec<Value>),
+    Set(Vec<Value>),
+    Map(Vec<(Value, Value)>),
+    Varint(num::BigInt),
```
rukai, line 326: lets add a `use num::BigInt`

rukai (changes requested):
Once all the feedback is addressed we can land this.

As you've said, in a follow up PR you can add tests for all functions here (Dont bother with into_str_bytes though as its implementation is currently nonsense)

conorbros:
I'll implement the collections stuff in this PR too because that's blocking my other Cassandra integration test PRs

shotover-proxy/src/message/mod.rs
```diff
@@ -377,26 +402,38 @@ impl Value {
             match spec.col_type.id {
                 ColType::Ascii => Value::Strings(decode_ascii(actual_bytes).unwrap()),
```
conorbros, line 403: Shouldn't Ascii have it's own field in Value? Same goes for other Cassandra types that we are currently putting in string. 

@benbromhead

rukai: yes.
The reason I didnt bring it up in the review was because it was a preexisting issue.
Although thinking about it, it does make sense to resolve it in this PR. (Although still happy for it to go in a follow up)

conorbros: I'll do it in this PR

conorbros:
Updates: 
 - added support for parsing collections into Value
 - changes to cdrs-tokio: https://github.com/conorbros/cdrs-tokio/commit/0353d89b6b549aa6e5ecdd53f28854e36ab182dd

shotover-proxy/src/message/mod.rs
```diff
@@ -313,23 +326,38 @@ pub enum Value {
...
     Inet(IpAddr),
     List(Vec<Value>),
     Rows(Vec<Vec<Value>>),
     NamedRows(Vec<HashMap<String, Value>>),
     Document(HashMap<String, Value>),
     FragmentedResponse(Vec<Value>),
+    Set(Vec<Value>),
+    Map(Vec<(Value, Value)>),
+    Varint(num::BigInt),
+    Bigint(num::BigInt),
```
rukai, line 344: a cassandra bigint appears to just be an i64
https://cassandra.apache.org/doc/latest/cassandra/cql/types.html
lol

Maybe instead of bigint, tinyint etc we should name the integer variants like:
```
I64(i64)
I32(i32)
I16(i16)
I8(i8)
```
Unless the names cassandra uses happen to be standard across many DBs...?

rukai: moved to the other PR

shotover-proxy/Cargo.toml
```diff
@@ -52,7 +52,9 @@ halfbrown = "0.1.11"
 
 # Transform dependencies
 redis-protocol = "3.0.1"
-cassandra-protocol = { git = "https://github.com/krojew/cdrs-tokio", rev="e1de5ffc30cf0e2cc13a657835f916fd4e61bf99" }
+#cassandra-protocol = { path = "/home/conor/src/cdrs-tokio/cassandra-protocol"}
+cassandra-protocol = { git = "https://github.com/conorbros/cdrs-tokio", branch = "typed-collections" }
```
rukai, line 56: we'll need to land this cdrs-tokio change before we can merge this PR

shotover-proxy/examples/cassandra-passthrough/topology.yaml
```diff
@@ -2,7 +2,7 @@
 sources:
   cassandra_prod:
     Cassandra:
-      query_processing: true
+      query_processing: false
```
rukai, line 5: why?

conorbros: Just a mistake left in while testing

shotover-proxy/src/message/mod.rs
conorbros: What collides?

shotover-proxy/src/message/mod.rs
```diff
@@ -321,6 +325,7 @@ pub enum Value {
     None,
     #[serde(with = "my_bytes")]
     Bytes(Bytes),
+    Bigint(BigInt),
```
rukai, line 328: is this different to Varint?

rukai: ah... maybe this got left in from a bad merge?

conorbros: Wouldn't we want to know whether it was a BigInt or VarInt when writing back

rukai: Are you referring to cassandra bigint? I believe we decided to represent that as Integer(_, IntSize::U64).
Cassandra Varint is shotover Varint.

conorbros: *IntSize::I64, but yes I remember that.

conorbros:
While rebasing I accidentally added `into_str_bytes` back so I have removed it in the most recent commit.

shotover-proxy/Cargo.toml
```diff
@@ -54,7 +54,8 @@ halfbrown = "0.1.11"
 
 # Transform dependencies
 redis-protocol = "3.0.1"
-cassandra-protocol = { git = "https://github.com/krojew/cdrs-tokio" }
+#cassandra-protocol = { git = "https://github.com/krojew/cdrs-tokio" }
+cassandra-protocol = { git = "https://github.com/conorbros/cdrs-tokio", branch = "typed-collections" }
```
rukai, line 58: is this an accident to?

conorbros: No, I had to put some changes into cdrs-tokio to fix the BigInt issue.

shotover-proxy/src/message/mod.rs
rukai: They both converted to Value::Strings which would make it impossible to unambiguously return trip.
Problem is resolved now though because ColType::Custom now is now `unimplemented!()`

shotover-proxy/Cargo.toml
rukai: ah ok, I was confused because it had the same branch name.
Dont forget to make a PR for it.

conorbros: Yep, just wanted to confirm it's the right approach before I made a PR.

shotover-proxy/src/message/mod.rs
```diff
@@ -397,19 +401,19 @@ impl From<Value> for Frame {
             Value::NamedRows(_) => unimplemented!(),
             Value::Document(_) => unimplemented!(),
             Value::FragmentedResponse(l) => Frame::Array(l.into_iter().map(|v| v.into()).collect()),
-            Value::Ascii(_) => unimplemented!(),
-            Value::Double(_) => unimplemented!(),
-            Value::Set(_) => unimplemented!(),
+            Value::Ascii(a) => Frame::SimpleString(a),
+            Value::Double(d) => Frame::SimpleString(d.to_string()),
```
rukai, line 405: double to string?

shotover-proxy/Cargo.toml
conorbros: https://github.com/krojew/cdrs-tokio/pull/73

shotover-proxy/src/message/mod.rs
```diff
@@ -447,27 +451,200 @@ impl Value {
...
+                }
+
+                value_list
+            }
+            _ => panic!("Not a List. Only CassandraType::List should be passed to this method."),
+        }
+    }
+
+    #[allow(clippy::mutable_key_type)]
+    fn create_set(collection: CassandraType) -> BTreeSet<Value> {
```
rukai, line 618: This should return Value.
Otherwise its super weird for `Value::create_set(set)` to not return a Value given that create_set belongs to Value

shotover-proxy/src/message/mod.rs
```diff
@@ -447,27 +451,200 @@ impl Value {
...
+            CassandraType::Tuple(_) => Value::Tuple(Value::create_tuple(element)),
+            CassandraType::Null => Value::NULL,
+        }
+    }
+
+    fn create_list(collection: CassandraType) -> Vec<Value> {
+        match collection {
+            CassandraType::List(collection) => {
+                let mut value_list = Vec::with_capacity(collection.len());
+                for element in collection.into_iter() {
```
rukai, line 607: This should use into_iter -> map -> collect.
Internally that will call Vec::with_capacity and is just as fast as what you have written here.

shotover-proxy/src/message/mod.rs
conorbros: Lol this seems obvious now that you pointed it out. The caller of this function just wraps it straight away

shotover-proxy/src/message/mod.rs
```diff
@@ -447,27 +451,200 @@ impl Value {
...
+                value_list.insert(Value::create_element(key), Value::create_element(value));
+            }
+
+            value_list
+        } else {
+            panic!("Not a Map. Only CassandraType::Map should be passed to this method");
+        }
+    }
+
+    fn create_element(element: CassandraType) -> Value {
```
rukai, line 570: Its unfortunate this duplicates a bunch of logic with build_value_from_cstar_col_type.
Is it not possible to turn the `spec: &ColSpec, data: &CBytes` into a `CassandraType` and call this instead?

shotover-proxy/src/message/mod.rs
```diff
@@ -426,45 +423,113 @@ impl Value {
     }
 
     pub fn build_value_from_cstar_col_type(spec: &ColSpec, data: &CBytes) -> Value {
-        if let Some(actual_bytes) = data.as_slice() {
```
conorbros, line 429: No longer need this check because cdrs-tokio will make the type CassandraType::Null

conorbros:
Addressed all of the feedback. The changes to `cdrs-tokio` can be found [here](https://github.com/conorbros/cdrs-tokio/tree/typed-collections)

shotover-proxy/src/message/mod.rs
```diff
@@ -426,45 +423,113 @@ impl Value {
...
+            CassandraType::Udt(_) => Value::create_udt(element),
+            CassandraType::Tuple(_) => Value::create_tuple(element),
+            CassandraType::Null => Value::NULL,
+        }
+    }
+
+    /// Create a `Value::Udt` from a `CassandraType::Udt`
+    fn create_udt(collection: CassandraType) -> Value {
+        if let CassandraType::Udt(udt) = collection {
+            let mut values = BTreeMap::new();
```
rukai, line 471: This can be written as:
image

Which preallocates the correct capacity.

And same for the other `BtreeMap`/`BTreeSet`s

conorbros: TIL you can use collect on sets and maps....thought it was just Vec

shotover-proxy/src/message/mod.rs
```diff
@@ -426,45 +423,112 @@ impl Value {
...
+            _ => panic!("Not a List. Only CassandraType::List should be passed to this method."),
+        }
+    }
+
+    /// Create a `Value::Set` from a `CassandraType::Set` or `CassandraType::List`
+    #[allow(clippy::mutable_key_type)]
+    fn create_set(collection: CassandraType) -> Value {
+        match collection {
+            CassandraType::List(collection) | CassandraType::Set(collection) => {
+                let values = collection.into_iter().map(|element| {Value::create_element(element)}).collect();
```
rukai, line 514: lol, the redundant braces are confusing clippy

shotover-proxy/src/message/mod.rs
```diff
@@ -426,45 +423,112 @@ impl Value {
...
+            CassandraType::Uuid(u) => Value::Uuid(u),
+            CassandraType::Varchar(v) => Value::Varchar(v),
+            CassandraType::Varint(v) => Value::Varint(v),
+            CassandraType::Timeuuid(t) => Value::Timeuuid(t),
+            CassandraType::Inet(i) => Value::Inet(i),
+            CassandraType::Date(d) => Value::Date(d),
+            CassandraType::Time(d) => Value::Time(d),
+            CassandraType::Smallint(d) => Value::Integer(d.into(), IntSize::I16),
+            CassandraType::Tinyint(d) => Value::Integer(d.into(), IntSize::I8),
+            CassandraType::List(_) => Value::create_list(element),
```
rukai, line 459: We could drastically simplify this by moving the logic from the functions directly into the match.

image

This removes the need for the panic and the logic is small enough to comfortably fit in the match

rukai (commented):
This looks fine to me.
I can approve once the cassandra-protocol changes land upstream.

One thing I did notice: for some reason `#[allow(clippy::mutable_key_type)]` isnt needed if we dont store `values` as an intermediate variable.
Up to you if you want to make use of that.

conorbros:
That is strange, I've removed it so we don't have an unnecessary `#[allow]`.

Opened a PR for the changes to cdrs-tokio: https://github.com/krojew/cdrs-tokio/pull/74

rukai (approved):
LGTM

PR #410, opened 2021-12-08
Add metrics to Redis Sink Single

conorbros:


shotover-proxy/src/transforms/redis/sink_single.rs
```diff
@@ -89,7 +98,10 @@ impl Transform for RedisSinkSingle {
 
         match outbound_framed_codec.next().fuse().await {
             Some(a) => a,
-            None => Err(anyhow!("couldnt get frame")),
+            None => {
+                counter!("redis_single_failed_request", 1, "chain" => self.chain_name.clone());
```
rukai, line 102: This doesnt seem to match up with when the cluster sink emits this counter

conorbros: Ah I see the cluster records specifically Redis error message not just a general error.

conorbros: Aren't those "shotover generated redis errors" just actual Redis errors being passed back? 

We aren't changing the type of Redis error or anything about the message.

rukai: ah, true!
So the functionality for the counter looks like: record every error the remote server sends back do not record any shotover side errors.

rukai (approved):
LGTM!

shotover-proxy/src/transforms/redis/sink_single.rs
```diff
@@ -35,20 +39,27 @@ pub struct RedisSinkSingle {
...
+            self.tls.clone(),
+            self.chain_name.clone(),
+        )
     }
 }
 
 impl RedisSinkSingle {
-    pub fn new(address: String, tls: Option<TlsConnector>) -> RedisSinkSingle {
+    pub fn new(address: String, tls: Option<TlsConnector>, chain_name: String) -> RedisSinkSingle {
+        register_counter!("redis_single_failed_request", Unit::Count, "chain" => chain_name.clone());
```
benbromhead, line 57: Naming wise we should call this "failed_requests" and add a tag "transform" => "redis_sink_single"

shotover-proxy/src/transforms/redis/sink_single.rs
```diff
@@ -88,7 +99,16 @@ impl Transform for RedisSinkSingle {
             .ok();
 
         match outbound_framed_codec.next().fuse().await {
-            Some(a) => a,
+            Some(a) => {
+                if let Ok(ref messages) = a {
+                    for message in messages {
+                        if let RawFrame::Redis(Frame::Error(_)) = message.original {
+                            counter!("redis_single_failed_request", 1, "chain" => self.chain_name.clone());
```
benbromhead, line 106: same here

benbromhead (changes requested):
Also clear out the commented out test code (no longer applicable btw)

shotover-proxy/src/transforms/redis/sink_single.rs
conorbros: Should redis cluster also have these changes?

benbromhead (approved):

rukai (approved):
Hmmm I wonder if we should remove the use of `CONTEXT_CHAIN_NAME` from sink_cluster and just do it the same way as sink_single.
Regardless lets this PR as is, and potentially clean that up later.

PR #411, opened 2021-12-08
Cleanup load balance

rukai:
Just renames variables to be more accurate and adds a comment explaining the intended use.

I actually cant think of how you would write a shotover config to make use of the "cloning = new connection" design, but that does seem to be the intended design.


conorbros (approved):

XA21X:
"Connections" between transforms? Or upstream connections? Maybe max-connections gives the wrong idea when going upstream isn't necessarily one to one? 🤔 

Just a random thought.

XA21X (approved):

PR #416, opened 2021-12-10
Get the query type for a message by a common get_query_type method

rukai:
Refactor towards getting RedisCache working.

Simplifies and unifies the implementation of various transforms

### DIFF
--- a/shotover-proxy/src/message/mod.rs
+++ b/shotover-proxy/src/message/mod.rs
@@ -104,6 +104,16 @@ impl Message {
             },
         }
     }
+
+    /// Gets the `QueryType` for a message.
+    /// First checks the `details` of the message and then falls back to the `original` message.
+    pub fn get_query_type(&self) -> QueryType {
+        if let MessageDetails::Query(query) = &self.details {
+            query.query_type.clone()
+        } else {
+            self.original.get_query_type()
+        }
+    }
 }
 
 #[derive(PartialEq, Debug, Clone)]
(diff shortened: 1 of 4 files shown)


shotover-proxy/src/message/mod.rs
```diff
@@ -104,6 +104,16 @@ impl Message {
             },
         }
     }
+
+    /// Gets the `QueryType` for a message.
+    /// First checks the `details` of the message and then falls back to the `original` message.
+    pub fn get_query_type(&self) -> QueryType {
+        if let MessageDetails::Query(query) = &self.details {
+            query.query_type.clone()
```
conorbros, line 112: We can avoid this clone by borrowing since we are only using the result for equality checks. It's only an enum though so not going to block on this.

conorbros (approved):

shotover-proxy/src/message/mod.rs
rukai: This is a very simple enum without any fields so cloning here should be as cheap as returning an i32.
I should probably implement Copy for QueryType to remove the `clone()` to make it clear that copying it is a cheap operation. AFAIK that would compile down to the same thing but it would just make it look less performance-scary.

benbromhead (approved):

shotover-proxy/src/message/mod.rs
rukai: I'll resolve this and merge, we can do such a change in a separate PR later.

rukai:
Just rebased and pushed

PR #417, opened 2021-12-10
Implement test_cassandra_redis_cache integration test

rukai:
The test currently demonstrates that the RedisCache transform can at least pass on messages to cassandra without issues.

Still need to test that the cache actually caches something.
We can do that by manually modifying the cache and checking we get the modified version back.
But that should go in a follow up PR.

### DIFF
--- a/shotover-proxy/examples/cassandra-redis-cache/docker-entrypoint.sh
+++ b/shotover-proxy/examples/cassandra-redis-cache/docker-entrypoint.sh
@@ -1,0 +1,100 @@
+#!/bin/bash
+set -e
+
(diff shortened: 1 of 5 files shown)


shotover-proxy/examples/cassandra-redis-cache/docker-entrypoint.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
```
conorbros, line 1: We aren't enabling user defined functions for this test so we don't need this.

conorbros (approved):

benbromhead (approved):

PR #418, opened 2021-12-10
Replace CoalesceBehaviour with individual fields in Coalesce.

rukai:
Using enums in configuration to make it impossible to represent invalid states is really powerful.
I think the `tls` field in RedisSink is a good example of where this works really well.

However in the case of the `max_behaviour` field in Coalesce I think the benefit tips the other way and we are better off with separate fields.
Conceptually we have two separate optional values that we want to configure: flush by timeout and flush by message count.
But we are currently combining both into one enum to avoid the invalid state where none of these values are provided.
This gives us a really awkward 3rd variant: CountOrWait that is both difficult to document and difficult to use.
The current CountOrWait implementation could be improved by replacing it with a `CountOrWait { count: usize, wait: u128 }`
But I think that would still be pretty confusing.

I propose we split the enum into two fields and maintain the same safety by adding an equivalent validation check.
This will make the API simpler to use and the documentation easier to understand.
Incidentally it makes the implementation a bit simpler too.

### DIFF
--- a/docs/src/transforms.md
+++ b/docs/src/transforms.md
@@ -64,21 +64,17 @@ Note: this will just pass the query to the remote node. No cluster discovery or 
 ### Coalesce
 
 This transform holds onto messages until some requirement is met and then sends them batched together.
+Validation will fail if none of the `flush_when_` fields are provided, as this would otherwise result in a Coalesce transform that never flushes.
 
 ```yaml
 - Coalesce:
-    max_behavior:
-      # Messages are held until the specified number of messages have been received.
-      Count: 2000
-      # alternatively:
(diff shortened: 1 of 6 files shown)


conorbros (approved):
LGTM. Only one pedantic comment: link rot is something we will have to be diligent about if we are going to provide links to documentation in error messages.

rukai:
Hmmm good point, if we end up doing this a lot might be worth investigating an automated solution for detecting dead links in the codebase.

benbromhead (approved):
CountOrWait always felt dirty to me... so lgtm!

PR #422, opened 2021-12-13
Update deps

rukai:
The cassandra_codec changes are a lot easier to review if you set the diff to ignore whitespace.
While making these changes I noticed that we were completely ignoring failure to parse the csql so I added the bare minimum handling (just logging the failure)

Not sure how best to untokenize the assignment id but `assignment.id.iter().map(|x| &x.value).join(".")` is my best guess.
I dont think its worth spending too much time thinking about, we probably need to completely replace this stuff with a cassandra specific parser anyway.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -96,9 +96,9 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.51"
+version = "0.1.52"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


rukai:
cargo deny failing on https://github.com/hsivonen/encoding_rs/issues/75

conorbros (approved):

benbromhead (approved):
LGTM - agreed I wouldn't spend to much time on the update assignment detokenisation right now as I dont think C* supports complex assignment qualification of column names anyway

benbromhead (approved):

PR #430, opened 2021-12-16
Redis cache doesnt blow up when the cache path is taken

rukai:
The integration test now actually takes the cache path.
This revealed the issue that we actually attempt to directly return the redis messages as cassandra messages, this naturally causes the decoder to error out.
In order to get the integration test passing I made a dummy redis-cache to cassandra message conversion that just ignores the redis message and generates an empty cassandra message.
This logic will be completely replaced later on but it will be really useful to have a test checked in that actually takes the cache path.

There is a separate testcase for queries against the PK and against some other field because they take different codepaths and the 2nd case currently sends an invalid query to redis, as can be seen by the output of `info!("Received reply from redis cache {:?}", message);`

After staring at the code for a while it seems we can only cache queries that have an equality check against a primary key.
Is that an intentional restriction or have we just not yet implemented handling for other cases?

### DIFF
--- a/shotover-proxy/src/transforms/redis/cache.rs
+++ b/shotover-proxy/src/transforms/redis/cache.rs
@@ -3,6 +3,7 @@ use std::collections::HashMap;
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
 use serde::Deserialize;
+use tracing::info;
 
 use crate::config::topology::TopicHolder;
 use crate::error::ChainResponse;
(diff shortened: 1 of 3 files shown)


shotover-proxy/src/transforms/redis/cache.rs
```diff
@@ -89,12 +94,29 @@ impl SimpleRedisCache {
             message.modified = true;
         }
 
-        self.cache_chain
+        let mut messages = self
```
conorbros, line 97: This errors if any message is an error?

Just checking with regards to making sure the `stream_id` always match up

conorbros (approved):

benbromhead (approved):

rukai:
Rebased to make sure CI is still green

PR #433, opened 2021-12-17
Add new value variants

conorbros:
changes pulled out from #409.

### DIFF
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -38,6 +38,9 @@ serde = { version = "1.0.111", features = ["derive"] }
 serde_json = "1.0"
 serde_yaml = "0.8.21"
 bincode = "1.3.1"
+num = { version = "0.4.0", features = ["serde"] }
+uuid = { version = "0.8.2", features = ["serde"]}
+bigdecimal = {version ="0.3.0", features = ["serde"]}
 
 #Observability
 metrics = "0.17.0"
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -54,9 +54,10 @@ halfbrown = "0.1.11"
 
 # Transform dependencies
 redis-protocol = "3.0.1"
-cassandra-protocol = { git = "https://github.com/krojew/cdrs-tokio", rev="e1de5ffc30cf0e2cc13a657835f916fd4e61bf99" }
+cassandra-protocol = { git = "https://github.com/conorbros/cdrs-tokio", branch = "typed-collections" }
 rdkafka = "0.28"
 crc16 = "0.4.0"
+ordered-float = { version = "2.8.0", features = ["serde"] }
 
 #Crypto
 sodiumoxide = "0.2.5"
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -54,7 +54,7 @@ halfbrown = "0.1.11"
 
 # Transform dependencies
 redis-protocol = "3.0.1"
-cassandra-protocol = { git = "https://github.com/conorbros/cdrs-tokio", branch = "typed-collections" }
+cassandra-protocol = { git = "https://github.com/krojew/cdrs-tokio" }
 rdkafka = "0.28"
 crc16 = "0.4.0"
 ordered-float = { version = "2.8.0", features = ["serde"] }
--- a/shotover-proxy/src/message/mod.rs
+++ b/shotover-proxy/src/message/mod.rs
@@ -20,7 +20,7 @@ use ordered_float::OrderedFloat;
 use redis_protocol::resp2::prelude::Frame;
 use serde::{Deserialize, Serialize};
 use sqlparser::ast::Statement;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 use std::{collections::HashMap, net::IpAddr};
 use uuid::Uuid;
 
@@ -315,7 +315,7 @@ pub enum QueryType {
     PubSubMessage,
 }
 
-#[derive(PartialEq, Debug, Clone, Serialize, Deserialize, Hash)]
+#[derive(PartialEq, Debug, Clone, Serialize, Deserialize, Hash, Eq, PartialOrd, Ord)]
 pub enum Value {
     NULL,
     None,
@@ -333,8 +333,8 @@ pub enum Value {
     NamedRows(Vec<BTreeMap<String, Value>>),
     Document(BTreeMap<String, Value>),
     FragmentedResponse(Vec<Value>),
-    Set(Vec<Value>),
-    Map(Vec<(Value, Value)>),
+    Set(BTreeSet<Value>),
+    Map(BTreeMap<Value, Value>),
     Varint(BigInt),
     Decimal(BigDecimal),
     Date(i32),
@@ -348,12 +348,12 @@ pub enum Value {
     Udt(BTreeMap<String, Value>),
 }
 
-#[derive(PartialEq, Debug, Clone, Serialize, Deserialize, Hash)]
+#[derive(PartialEq, Debug, Clone, Serialize, Deserialize, Hash, Eq, PartialOrd, Ord)]
 pub enum IntSize {
     I64, // BigInt
     I32, // Int
     I16, // Smallint
-    I8,  // tinyint
+    I8,  // Tinyint
 }
 
 impl From<Frame> for Value {
--- a/shotover-proxy/src/message/mod.rs
+++ b/shotover-proxy/src/message/mod.rs
@@ -15,6 +15,7 @@ use cassandra_protocol::{
         CBytes,
     },
 };
+use num::BigInt;
 use redis_protocol::resp2::prelude::Frame;
 use serde::{Deserialize, Serialize};
 use sqlparser::ast::Statement;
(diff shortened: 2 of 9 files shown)


shotover-proxy/src/message/mod.rs
```diff
@@ -313,23 +318,38 @@ pub enum Value {
...
     Inet(IpAddr),
     List(Vec<Value>),
     Rows(Vec<Vec<Value>>),
     NamedRows(Vec<HashMap<String, Value>>),
     Document(HashMap<String, Value>),
     FragmentedResponse(Vec<Value>),
+    Set(Vec<Value>),
+    Map(Vec<(Value, Value)>),
+    Varint(num::BigInt),
+    Bigint(num::BigInt),
```
rukai, line 336: a cassandra bigint appears to just be an i64
https://cassandra.apache.org/doc/latest/cassandra/cql/types.html
lol

Maybe instead of bigint, tinyint etc we should name the integer variants like:
```
I64(i64)
I32(i32)
I16(i16)
I8(i8)
```
Unless the names cassandra uses happen to be standard across many DBs...?

shotover-proxy/src/message/mod.rs
```diff
@@ -313,23 +318,38 @@ pub enum Value {
...
     Boolean(bool),
     Inet(IpAddr),
     List(Vec<Value>),
     Rows(Vec<Vec<Value>>),
     NamedRows(Vec<HashMap<String, Value>>),
     Document(HashMap<String, Value>),
     FragmentedResponse(Vec<Value>),
+    Set(Vec<Value>),
+    Map(Vec<(Value, Value)>),
+    Varint(num::BigInt),
```
rukai, line 335: lets add a `use num::BigInt`

shotover-proxy/src/message/mod.rs
rukai: ahh maybe what we want is something like:
```rust
enum Value {
Int(i64, IntSize)
}


enum IntSize {
    I64,
    I32,
    I16,
    I8,
}
```

That way we dont have to special case every int type?

conorbros: Yeah I think that's a good idea.

conorbros: Okay added those changes but put all the existing Integers as I32 since that's all we had before

shotover-proxy/src/message/mod.rs
```diff
@@ -313,27 +319,47 @@ pub enum Value {
...
+    Float(f32),
     Boolean(bool),
     Inet(IpAddr),
     List(Vec<Value>),
     Rows(Vec<Vec<Value>>),
     NamedRows(Vec<HashMap<String, Value>>),
     Document(HashMap<String, Value>),
     FragmentedResponse(Vec<Value>),
+    Set(Vec<Value>),
+    Map(Vec<(Value, Value)>),
```
rukai, line 335: Why arent we using a HashMap? are cassandra maps ordered?

conorbros: Yeah we can use a HashMap, I think I had it has that because Cassandra gives it raw back like that but we can put it in a HashMap instead.

rukai: Yeah I think a HashMap makes the most sense.
We can of course change it later if we realize we need something else.

Along the same line of thinking, we should also use HashSet for Set.

conorbros: Ah now I remember. We would need to implement `Hash` for all the types in Value.

shotover-proxy/src/message/mod.rs
```diff
@@ -313,27 +319,47 @@ pub enum Value {
...
+    Varint(BigInt),
+    Decimal(BigDecimal),
+    Date(i32),
```
rukai, line 338: Possibly we want something like:
```rust
DateInt(i32)
DateString(String)
```
And equivalent for time.
But what you have is good for now, lets wait and see what our use looks like.

shotover-proxy/src/message/mod.rs
conorbros: Doing something like that might mean we move "Collection" Values to other enum

rukai: Oh, were we going to replace Document with Map?

rukai: Oh I see, because the key is a Value instead of a String.

rukai: What would it take to impl Hash for Value?

rukai: you cant hash f64???
hmmm...

conorbros: Yeah that's why I ended up with the Vec's instead. Should we fine given we can trust that Cassandra will give us a valid map or set

rukai: Hmmm its tempting to do something like:
```rust
Map(Vec<KeyValue>)

struct KeyValue {
    key: Value,
    value: Value,
}
```

however ... the ordering of key and then value is pretty usual, so maybe we dont need the struct.
We can always rewrite to use a struct if we find the code that uses it would benefit from it.

In general though, the `Map(Vec<(Value, Value)>)` approach is fine by me.

shotover-proxy/src/protocols/cassandra_codec.rs
```diff
@@ -145,14 +145,14 @@ impl CassandraCodec {
                                         Value::NULL => to_int(-1_i32),
                                         Value::Bytes(x) => x.to_vec(),
                                         Value::Strings(x) => Vec::from(x.as_bytes()),
-                                        Value::Integer(x) => {
+                                        Value::Integer(x, _) => {
                                             let mut temp: Vec<u8> = Vec::new();
                                             let _ = temp.write_i64::<BigEndian>(*x).unwrap();
```
rukai, line 150: is this correct for integer sizes?

rukai (approved):
LGTM!

shotover-proxy/src/protocols/cassandra_codec.rs
conorbros: Added a commit properly writing the ints

shotover-proxy/src/message/mod.rs
benbromhead: this doesn't quite feel right to be honest from a Developer Experience perspective. 

I don't think it would be a ton of effort to implement hash for `Value`

shotover-proxy/src/message/mod.rs
```diff
@@ -313,27 +319,47 @@ pub enum Value {
...
+    Timestamp(i64),
+    Timeuuid(Uuid),
+    Varchar(String),
+    Uuid(Uuid),
+    Time(i64),
+    Counter(i64),
+    Tuple(Vec<Value>),
+    Udt(HashMap<String, Value>),
```
benbromhead, line 346: Are we doing this like field_name maps to value??

benbromhead (changes requested):
The hash map approach needs to be refined imho as it's a user facing API that transform authors will need to work with and the `Vec` approach makes it difficult to work with.

shotover-proxy/src/message/mod.rs
conorbros: Implemented Hash for Value:
- switched the `HashMap`'s to `BTreeMap` which implements the `Hash` trait
- wrapped the floats in [`OrderedFloat`](https://docs.rs/ordered-float/latest/ordered_float/struct.OrderedFloat.html)

rukai: Given that HashMap doesnt implement Hash BTreeMap seems like a reasonable compromise.

Set and Map variants need to be updated still.

shotover-proxy/Cargo.toml
```diff
@@ -52,9 +55,10 @@ halfbrown = "0.1.11"
 
 # Transform dependencies
 redis-protocol = "3.0.1"
-cassandra-protocol = { git = "https://github.com/krojew/cdrs-tokio", rev="e1de5ffc30cf0e2cc13a657835f916fd4e61bf99" }
+cassandra-protocol = { git = "https://github.com/conorbros/cdrs-tokio", branch = "typed-collections" }
```
rukai, line 58: Currently blocked on upstreaming this

shotover-proxy/src/message/mod.rs
conorbros: Yeah

shotover-proxy/src/message/mod.rs
conorbros: Fixed the rest of it. Should be good to resolve this now.

rukai (approved):

shotover-proxy/src/message/mod.rs
```diff
@@ -307,33 +315,53 @@ pub enum QueryType {
...
 impl From<Frame> for Value {
     fn from(f: Frame) -> Self {
-        // panic!("Aug 07 15:56:49.621  INFO instaproxy::transforms::tuneable_consistency_scatter: Response(QueryResponse { matching_query: None, original: Redis(BulkString([102, 111, 111])), result: Some(Strings("foo")), error: None })
-        // ");
-        //         panic!("This should be a bulk_string(byte array) and not be a string"); // I wonder if the bytes themselves need to get serialised differently....?
         match f {
             Frame::SimpleString(s) => Value::Strings(s),
             Frame::Error(e) => Value::Strings(e),
-            Frame::Integer(i) => Value::Integer(i),
+            Frame::Integer(i) => Value::Integer(i, IntSize::I32),
```
rukai, line 364: This should be IntSize::I64 as per https://redis.io/topics/protocol#resp-integers

benbromhead (approved):

PR #437, opened 2021-12-23
Cassandra TLS

conorbros:


conorbros:
Running into a strange issue. I can use Shotover with a TLS Cassandra source fine with `cqlsh` but `cassandra_cpp` does not work. I think it's using the wrong version of TLS (TLS v1.1?).

conorbros:
This PR is now ready for review. 

### Summary of the changes: 
- As discussed on slack, changed the benchmark runner to `ubuntu-18.04` and used the ubuntu install script so we can use `cassandra_cpp` in Cassandra TLS benchmarks, coming in a followup PR.
-  Added SSL keys/certs and the keystore and truststore formats that Cassandra uses. 
- The `cassandra-passthrough-tls` image is `bitnami/cassandra` instead of `library/cassandra`, this is because they have added the ability to enable client encryption through environment variables. I opened a PR to add the other environment variables that we need, so we could switch all our cassandra images to this and get rid of the modified `docker-entrypoint.sh` but I will leave that to a followup PR. 
- Changed the `SslAcceptor` to use `mozilla_intermediate` instead of `mozilla_modern_v5` because `cassandra_cpp` has been limited to use TLS 1.2 or below, as can be seen [here](https://github.com/datastax/cpp-driver/blob/90df2c9ca1aa184a746445698533c71f7f34a2e1/src/ssl/ssl_openssl_impl.cpp#L555). It seems the intermediate [configuration suits well as a general purpose server](https://wiki.mozilla.org/Security/Server_Side_TLS).


### Moving forward
- Add TLS benchmarks
- Move our all our Cassandra images over to the bitnami one.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -944,6 +945,7 @@ mod functions {
 
         let statement = stmt!("SELECT test_function_keyspace.my_function(x) FROM test_function_keyspace.test_function_table WHERE id=1;");
         let result = session.execute(&statement).wait().unwrap_err().to_string();
+        println!("{:?}", result);
```
rukai, line 948: println

shotover-proxy/tests/helpers/mod.rs
```diff
@@ -139,6 +141,32 @@ impl ShotoverManager {
         .unwrap()
     }
 
+    #[allow(unused)]
+    pub fn cassandra_connection_tls(
```
rukai, line 145: Shouldnt this go next to the `fn cassandra_connection` ?

conorbros: I've moved `fn cassandra_connection` to `ShotoverManger` in #512, so once that is merged they will be in the same place.

rukai: ah, alright, yeah I think its fine for both of them to go in ShotoverManager.
While its true that we could use cassandra_connection to connect directly to cassandra instead of through shotover, there is no case where we would want to do that without also having a shotover instance running anyway.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1310,6 +1311,38 @@ fn test_passthrough() {
...
+
+    let connection = shotover_manager.cassandra_connection_tls(
+        "127.0.0.1",
+        9043,
+        tls_config,
+        "cassandra",
+        "cassandra",
+    );
+
+    keyspace::test(&connection);
```
rukai, line 1337: Can we run a simple test case directly against cassandra to make sure that cassandra_connection_tls can connect to both cassandra and shotover?
This would protect against a possible case where both shotover and cassandra_connection_tls were doing something slightly wrong compared to actual cassandra.
I think `table::test` would be fast enough for this use case.

shotover-proxy/tests/helpers/mod.rs
```diff
@@ -139,6 +141,32 @@ impl ShotoverManager {
...
+        let ca_cert = read_to_string(tls_config.certificate_path).unwrap();
+        let mut ssl = Ssl::default();
+        Ssl::add_trusted_cert(&mut ssl, &ca_cert).unwrap();
+
+        for contact_point in contact_points.split(',') {
+            crate::helpers::wait_for_socket_to_open(contact_point, port);
+        }
+
+        let mut cluster = Cluster::default();
+        cluster.set_credentials(username, password).unwrap();
```
rukai, line 162: why are we introducing username and password as well?
Is it tied to TLS somehow?

conorbros: The docker-image has it enabled by default, and I thought to leave it on so we can test that as well.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
conorbros: Decided to just run a simple query against the `system.local` table, rather than implement teardown code for the `table::test` suite.

conorbros: Decided to just run a simple query against the `system.local` table, rather than implement teardown code for the `table::test` suite.

shotover-proxy/tests/helpers/mod.rs
rukai: Hmmm alright, sounds reasonable

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
rukai: nice!

shotover-proxy/tests/helpers/mod.rs
```diff
@@ -139,6 +141,33 @@ impl ShotoverManager {
...
+        contact_points: &str,
+        port: u16,
+        tls_config: TlsConfig,
+        username: &str,
+        password: &str,
+    ) -> Session {
+        let ca_cert = read_to_string(tls_config.certificate_path).unwrap();
+        let mut ssl = Ssl::default();
+        Ssl::add_trusted_cert(&mut ssl, &ca_cert).unwrap();
+        Ssl::set_verify_flags(&mut ssl, &[SslVerifyFlag::NONE]);
```
rukai, line 156: What is this for?

benbromhead (commented):
We also probably need to have a `connect` method on `TlsConnector` that will verify the hostname (I know we dont do it at the moment as redis never did it that way)

shotover-proxy/src/tls.rs
```diff
@@ -26,7 +26,7 @@ pub struct TlsAcceptor {
 
 impl TlsAcceptor {
     pub fn new(tls_config: TlsConfig) -> Result<TlsAcceptor> {
-        let mut builder = SslAcceptor::mozilla_modern_v5(SslMethod::tls())?;
+        let mut builder = SslAcceptor::mozilla_intermediate(SslMethod::tls())?;
```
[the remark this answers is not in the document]
benbromhead: lol thanks

rukai:
Are all the 6 key files used?

And how are they generated?
Might be worth checking in a shell script in the keys directory for future reference.
I think the reason I didnt do this for redis is there was already a shell script `gen-test-certs` in the redis repo for generating keys.

shotover-proxy/src/tls.rs
```diff
@@ -67,12 +67,16 @@ impl TlsConnector {
         })
     }
 
-    pub async fn connect(&self, tcp_stream: TcpStream) -> Result<SslStream<TcpStream>> {
+    pub async fn connect(
+        &self,
+        tcp_stream: TcpStream,
+        verify_hostname: bool,
```
rukai, line 73: Can we make setting this to false scarier?
Maybe a separate scary sounding method name?
Maybe a seperate constructor for TLSAcceptor?
Or maybe replace the bool with an enum with a scary sounding variant?

As is, its quite easy to accidentally drop an important security measure by copy-pasting the connect call from redis which doesnt give any context to the false value it uses.

conorbros:
Yeah, I'll remove `*.jks`.

And yes, the process is a little complicated. I'll put together a script for it.

shotover-proxy/src/transforms/cassandra/connection.rs
```diff
@@ -26,33 +27,36 @@ struct Request {
 
 #[derive(Clone, Derivative)]
 #[derivative(Debug)]
-pub struct CassandraConnection<C: Codec> {
+pub struct CassandraConnection {
     host: String,
     connection: Option<mpsc::UnboundedSender<Request>>,
```
rukai, line 32: This can no longer be None so we should remove the Option.

shotover-proxy/src/transforms/cassandra/connection.rs
```diff
@@ -26,33 +27,36 @@ struct Request {
...
-            codec,
-        }
-    }
+impl CassandraConnection {
+    pub async fn new<C: Codec + 'static>(
+        host: String,
+        codec: C,
+        mut tls: Option<TlsConnector>,
+    ) -> Result<Self> {
+        let tcp_stream: TcpStream = TcpStream::connect(host.clone()).await?;
```
rukai, line 41: we can do `&host` here instead of `host.clone()`

rukai (changes requested):
Mostly reviewed at this point.
Once you've addressed these up i'll go over it again.

shotover-proxy/src/tls.rs
```diff
@@ -80,6 +81,15 @@ impl TlsConnector {
 
         Ok(ssl_stream)
     }
+
+    pub async fn connect(&self, tcp_stream: TcpStream) -> Result<SslStream<TcpStream>> {
+        let ssl = self.connector.configure()?.into_ssl("localhost")?;
```
rukai, line 86: this into_ssl("localhost") looks suspicious.
https://docs.rs/openssl/latest/openssl/ssl/struct.ConnectConfiguration.html#method.into_ssl
From reading the docs, it seems this might break in a real non localhost scenario?

It wouldnt have been a problem on redis because hostname verification is disabled there.

If we find an issue and it needs a complex solution/investigation I would be happy to land what we have so far as it at least works in a subset of scenarios and get non-localhost connections working in a follow up PR.

shotover-proxy/src/transforms/cassandra/connection.rs
```diff
@@ -26,33 +27,37 @@ struct Request {
...
 
-    pub async fn connect(&mut self) -> Result<()> {
-        let socket: TcpStream = TcpStream::connect(self.host.clone()).await?;
-        let (read, write) = socket.into_split();
         let (out_tx, out_rx) = mpsc::unbounded_channel::<Request>();
         let (return_tx, return_rx) = mpsc::unbounded_channel::<Request>();
 
-        tokio::spawn(tx_process(write, out_rx, return_tx, self.codec.clone()).in_current_span());
+        if let Some(tls) = tls.as_mut() {
+            let tls_stream = tls.connect(tcp_stream).await.unwrap();
```
rukai, line 47: Should we return Err instead of unwrap here?
Currently we never return Err in this function despite being Result<()>

Lets make the redis equivalent call do the same as well.

shotover-proxy/src/transforms/util/cluster_connection_pool.rs
```diff
@@ -177,7 +177,10 @@ impl<C: Codec + 'static, A: Authenticator<T>, T: Token> ConnectionPool<C, A, T>
             .map_err(ConnectionError::IO)?;
 
         let mut connection = if let Some(tls) = &self.tls {
-            let tls_stream = tls.connect(stream).await.map_err(ConnectionError::TLS)?;
+            let tls_stream = tls
+                .connect_unverified_hostname(stream)
```
rukai, line 181: Its concerning that we are making this call in a supposedly generic `util`.
Should we raise an issue to make this cluster logic redis specific?

shotover-proxy/examples/cassandra-passthrough-tls/certs/gen_certs.sh
```diff
@@ -0,0 +1,16 @@
+
```
rukai, line 1: best to add bash or sh crunchbang

shotover-proxy/examples/cassandra-passthrough-tls/gen_certs.sh
```diff
@@ -0,0 +1,16 @@
+
```
rukai, line 1: Accidental duplicate?

shotover-proxy/src/transforms/util/cluster_connection_pool.rs
conorbros: I don't think so, the method is pretty explicit that it does not verify the hostname. I'd say it's as clear as the `verify_hostname` method itself on `openssl::ssl::ConnectConfiguration`.

rukai: right but you would have to check inside the cluster_connection_pool implementation to know that.
We can just leave at as is though, I dont mind.

conorbros: Oh true, I misinterpreted. Thought you were saying the `TlsConnector` was a generic util.

conorbros: When we look at Cassandra cluster support we can revisit that. We will find out whether we can have a generic cluster connection pool or if this needs to be a Redis specific one.

rukai (approved):
Nice, LGTM!

Lets investigate the unresolved comment, before merging.
But im happy to merge no matter the result and just raise an issue to resolve the problem if there is one.

shotover-proxy/src/tls.rs
conorbros: image

I'm not sure if that is the right path. Maybe we need to providing SSL connections with a hostname to connect to instead of an address.

rukai: were you able to reproduce a problem by connecting to an instance that isnt on local host?

conorbros: Yes, trying to connect to with hostname "notlocalhost" creates a problem.

conorbros: I'll open an issue regarding TLS configuration. I think we will want an option for validating the hostname as well.

benbromhead:
Sorry one last final question. Given we have the script to generate certs, I'm now hesitant to checkin certs to the repo (even self signed / self generated certs for testing) as they always have a habbit of making their way to live servers in the wild as folks grab them to "just make something work".

Should we just gen the certs as part of the build / CI process and not checkin any certs

rukai:
Oh I hadnt heard of incidents like that, but if thats a thing then generating the certs at runtime definitely makes sense to me.
We will want to do the same for redis as well, so we could either:
*  check in as is and raise an issue to fix both redis and cassandra.
*  fix cassandra now and raise an issue for just redis

I dont mind either way

benbromhead (approved):
lgtm

PR #441, opened 2022-01-06
Redis cache cleanup

rukai:
Implementation is now less brittle and easier to understand.

Anything touching ValueHelper will need to be rewritten but that will happen later when we figure out how we are going to serialize/deserialize the values.

### DIFF
--- a/shotover-proxy/src/transforms/redis/cache.rs
+++ b/shotover-proxy/src/transforms/redis/cache.rs
@@ -143,23 +143,23 @@ impl ValueHelper {
     }
 }
 
-fn append_seperator(command_builder: &mut Vec<u8>) {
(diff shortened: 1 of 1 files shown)


conorbros (approved):

benbromhead (approved):
LGTM - Though to be honest I didn't have the confidence in the correctness of the previous code, so I'm not reviewing this thoroughly from that perspective. 

I think we will probably want to write a number of additional edge case tests for the boundary conditions on building this cache query. 

We can build confidence as we go, via adding tests. This may also be a candidate for some quicktheories or fuzz style testing

rukai:
Ok, good to know.

Once we can actually cache a basic query i'll come back and write thorough tests for it.

PR #442, opened 2022-01-10
Update to redis-protocol 4.0

rukai:
The only change in 4.0.0 was to replace String with StrMut and Vec<u8> with BytesMut https://github.com/aembke/redis-protocol.rs/pull/17

I think the `Frame` fields should be using `Bytes`/`Str` instead of `BytesMut`/`StrMut`.
I raised an issue https://github.com/aembke/redis-protocol.rs/issues/18 for this.
In the mean time we should upgrade to ensure that when we reason about performance implications we are working from the current state of the library.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -48,9 +48,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.52"
+version = "1.0.53"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "84450d0b4a8bd1ba4144ce8ce718fbc5d071358b1e5384bace6536b3d1f2d5b3"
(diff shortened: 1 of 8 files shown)


rukai:
Looks like a fix to the raised issue will be available in a reasonable amount of time so lets just wait for that before proceeding.

benbromhead:
Thanks @rukai and I agree with the rationale here. I'll review once the lib has been upgraded

rukai:
Frame now uses Bytes instead of BytesMut so we can proceed with this.
The PR has been updated and is ready for review.

conorbros (approved):

rukai:
rebased to make sure its compatible with the recent changes

rukai:
Seems like the comment is accidentally hitting some markdown formatting, other than that pretty cool to see it confirming the performance improvements from using Bytes.

PR #443, opened 2022-01-12
RedisTimestampTagger cleanup

rukai:
Simplifies the implementation, improves performance and improves logging

Why am I fiddling with RedisTimestampTagger?
Its a good example of a transform performing message details manipulation so I wanted to properly understand it.
In the process of doing so I noticed a bunch of minor improvements.

Remember to ignore whitespace in diff.

### DIFF
--- a/shotover-proxy/src/transforms/redis/timestamp_tagging.rs
+++ b/shotover-proxy/src/transforms/redis/timestamp_tagging.rs
@@ -128,7 +128,11 @@ fn unwrap_response(qr: &mut QueryResponse) {
                         let since_the_epoch = start
                             .duration_since(UNIX_EPOCH)
                             .expect("Time went backwards");
-                        Value::Integer(since_the_epoch.as_secs() as i64 - i, IntSize::I32)
+                        let seconds = since_the_epoch.as_secs();
+                        if seconds.leading_ones() > 1 {
+                            panic!("Cannot convert u64 to i64 without overflow");
+                        }
+                        Value::Integer(seconds as i64 - i, IntSize::I64)
                     }
                     v => v,
                 };
--- a/shotover-proxy/src/transforms/redis/timestamp_tagging.rs
+++ b/shotover-proxy/src/transforms/redis/timestamp_tagging.rs
@@ -1,7 +1,7 @@
 use crate::error::ChainResponse;
 use crate::message::{ASTHolder, IntSize, MessageDetails, QueryMessage, QueryResponse, Value};
 use crate::transforms::{Transform, Transforms, Wrapper};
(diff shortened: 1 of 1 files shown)


conorbros:
I won't block this useful PR on this, but do we have any integration/unit tests for redis timestamp tagger?

shotover-proxy/src/transforms/redis/timestamp_tagging.rs
```diff
@@ -122,38 +114,40 @@ fn unwrap_response(qr: &mut QueryResponse) {
...
-                    hm.insert("timestamp".to_string(), v);
-                }
-                Value::Document(hm)
+                let processed_value = match v {
+                    Value::Integer(i, _) => {
+                        let start = SystemTime::now();
+                        let since_the_epoch = start
+                            .duration_since(UNIX_EPOCH)
+                            .expect("Time went backwards");
+                        Value::Integer(since_the_epoch.as_secs() as i64 - i, IntSize::I32)
```
conorbros, line 131: `as i64` with `IntSize::I32`?

conorbros (changes requested):

rukai:
yep, it is integration tested in https://github.com/shotover/shotover-proxy/blob/146ff73d49e3788aaa231f95aa339b1e5d0144b5/shotover-proxy/examples/redis-multi/topology.yaml

conorbros:
Does that test the functionality of the transform?

shotover-proxy/src/transforms/redis/timestamp_tagging.rs
rukai: There is nothing inherently wrong with converting to i64 and setting the size to IntSize::I32 if it was originally i32. That was the design remember.

However looking at this, `as_secs` returns u64 so we should use IntSize::I64 here.

rukai:
ah true, it doesnt test the timestamp functionality.
It does run our standard redis test suite against it though. So we can be sure that commands are still being passed through correctly.

conorbros (approved):

benbromhead:
So some explanation is probably worthwhile here. 

The timestamp tagger is built to work in conjunction with Consistent Scatter. If the timestamp tagger can get a timestamp from redis, it will inject it into the response meta data. Consistent Scatter will then be able to use that to resolve inconsistent reads from different redis replicas (e.g. it brings in the C* consistency model).

Given that scatter will just fall back to "what's the biggest response" tagging failures will be silent. So let's raise an issue to put in a few tests here. Doesn't need to be extensive as this is not remotely suitable for production, but it's a nice example to illustrate certain concepts.

benbromhead (approved):

PR #446, opened 2022-01-13
redis cache new int test case

rukai:
Adds a second RedisCache test case that actually manages to store values in redis.

This sets up for the next step which will be changing the stored redis values to a format that can be deserialized back into cassandra values.

### DIFF
--- a/shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
+++ b/shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
@@ -222,21 +222,24 @@ mod cache {
 
     use crate::cassandra_int_tests::{assert_query_result, run_query, ResultValue};
 
-    pub fn test(session: &Session, redis_connection: &mut redis::Connection) {
-        test_batch_insert(session, redis_connection);
-        test_simple(session, redis_connection);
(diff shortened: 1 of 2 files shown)


benbromhead (approved):
1 minor nit: Can you change the name of the parameter `session` for `test_batch_insert` to `cassandra_session` so it matches the other one and it's clear as mud

rukai:
nice improvement, done

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -280,6 +291,72 @@ mod cache {
         let result: Vec<String> = redis_connection.keys("*").unwrap();
         assert_eq!(result, ["dummy_key".to_string()]);
     }
+
+    pub fn test_simple(cassandra_session: &Session, redis_connection: &mut redis::Connection) {
```
conorbros, line 295: nit: why are these public?

conorbros (approved):

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
rukai: copy/paste :D
Fixed.

PR #450, opened 2022-01-13
Optimise Redis cache loop

conorbros:
Something I noticed while investigating #448. We can early return from this loop the first time we notice a `QueryType::Write` and not scan through the rest of the messages, and I added a comment to make the logic a little clearer.


rukai (approved):
LGTM, I cant imagine what the purpose of a count over a simple bool would have been

benbromhead (approved):

PR #453, opened 2022-01-17
Message details redesign

rukai:
Ready for reviewing the general approach.

All of these changes will need to be split up into smaller mergeable PRs.

### DIFF
--- a/docs/src/source-types.md
+++ b/docs/src/source-types.md
@@ -16,8 +16,6 @@ Cassandra:
   cassandra_ks:
     foo: "bar"
 
-  query_processing: false
-
   # The number of concurrent connections the source will accept.
   connection_limit: 1000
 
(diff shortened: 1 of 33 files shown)


shotover-proxy/src/protocols/cassandra_codec.rs
```diff
@@ -1,598 +1,416 @@
...
-            SQLValue::Number(v, false) => v.to_string(),
-            SQLValue::SingleQuotedString(v)
-            | SQLValue::NationalStringLiteral(v)
-            | SQLValue::HexStringLiteral(v) => v.to_string(),
-            SQLValue::Boolean(v) => v.to_string(),
-            _ => "NULL".to_string(),
-        }
+impl CassandraFrame {
+    pub fn get_query_type(&self) -> QueryType {
+        QueryType::Read
```
Claude-at-Instaclustr, line 48: Doesn't this step on the work to get the QueryType properly set?

rukai: yes. Once your QueryType PR is merged i'll make sure to maintain that functionality.

rukai:
Only thing left in this PR is removal of CQL enum to make CQL parsing infallible.
I'm going to close this PR because even if we want to make that change it will be easier to just do from scratch at this point.

PR #456, opened 2022-01-18
Cassandra map tests

conorbros:
Last set of tests for Cassandra collections


rukai:
I guess you need to bump the ram again?

conorbros:
Looks like it, must be the migrations. I don't think the data we are putting in is enough to cause the OOM errors.

benbromhead:
Given that we are creating a bunch of tables, I'm not suprised. C* has a base overhead of like 1-2MB of memory per table.
If we want to keep the C* memory useage low in testing, then we probably want to have one or two tables, with say a couple of hundred columns rather than a couple of hundred tables.

The table with a couple of hundred columns will only occupy a mb or two in memory

rukai:
If we delete the tables will the memory be reclaimed?

benbromhead:
> If we delete the tables will the memory be reclaimed?

Yes

rukai:
Having each test in its own table helps to keep the tests isolated.
So it sounds like deleting the tables might be the way to go.
Maybe we can come up with some nice abstraction for easily creating/deleting tables.

I'm happy to land this PR with just the memory increase, then we can do a follow up PR to properly fix the memory usage sometime before we next add more tests.

conorbros:
> Having each test in its own table helps to keep the tests isolated.

Yep this is why I had a separate table for each test. Deleting the tables should be simple enough given how the test modules are setup, so I will open a PR for that after this lands and see how low we can get the memory usage.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -751,11 +751,192 @@ mod collections {
...
+                    session,
+                    format!(
+                        "CREATE TABLE test_collections_keyspace.test_map_table_{}_{} (id int PRIMARY KEY, my_map frozen<map<int, {}<{}>>>);",
+                        i,
+                        j,
+                        get_type_str(*collection_col_type),
+                        get_type_str(*native_col_type)
+                    )
+                    .as_str(),
+                );
```
rukai, line 784: formatting

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -751,11 +751,192 @@ mod collections {
...
+                let query = format!(
+                    "SELECT my_map FROM test_collections_keyspace.test_map_table_{};",
+                    i
+                );
+
+                assert_query_result(
+                    session,
+                    query.as_str(),
+                    &[&[ResultValue::Map(vec![(
+                        ResultValue::Int(0),
```
rukai, line 869: Every test case has one key-value pair and  the key is always 0 should we be testing more complicated cases?

conorbros: Possibly? It's the same as the set and list tests as they only have one element.

rukai: ah true, we can consider that for a future improvement then

rukai (approved):

benbromhead (approved):

PR #457, opened 2022-01-19
Update deps

conorbros:
Updating our metrics deps


rukai:
Did you find out why Unit was removed?

conorbros:
https://github.com/metrics-rs/metrics/blob/main/metrics/RELEASES.md#handle-based-design

Yes, the functionality has been replaced by the `describe_*` macros.

rukai:
Should we be using the describe macro instead of register then?

conorbros:
Yeah I think so. Doesn't look like the output is different though

conorbros:
I think we would end up having to use it like this:
```rust
register_gauge!("shotover_available_connections", "source" => source_name.clone());
describe_gauge!(
    "shotover_available_connections",
    Unit::Count,
    "test description"
);
gauge!("shotover_available_connections", limit_connections.available_permits() as f64, "source" => source_name.clone());

```

rukai:
The describe doesnt implicitly register?

conorbros:
No it doesn't register it, it's now just for adding details to a metric like the unit type and a description. The output is the same as what we had but with an extra `"#HELP"` label. e.g. 
image


The `register` macros have been improved a lot it seems, because `gauge!` implicitly registers the metric if it wasn't already registered, it's got a bit of overhead. Now the `register` macros return a handle to a metric that you can use directly. According to the author of `metrics-rs`, using the macros to update a metric takes around 50-60ns, whereas the new handle based design takes ~5ns. It's essentially the same as working with the underlying data structures of the metrics/using atomic stores.

Moving forward: 
- I think we should take advantage of this easy performance win and store and use the metrics handles instead of the macros.
- We can use the `describe_*` macros too, just need to come up with some descriptions for our metrics.

rukai:
Thanks for the deep investigation!
Shall we land this PR as is and raise an issue containing your findings?

conorbros:
Yep that sounds good to me, the changes would be a bit much to include in this PR.

rukai (approved):

rukai:
Got some merge conflicts

XA21X (approved):

conorbros:
Hmm the `license/cla` check seems to be frozen.

rukai:
Can you raise an issue with your findings?

PR #461, opened 2022-01-20
Improve QueryTypeFilter unit test

rukai:
Cleanup inspired by the message details rewrite.

Main purpose is to replace this logic:
```rust
        let any = result.iter().find(|m| {
            if let MessageDetails::Response(qr) = &m.details {
                if let Some(qm) = &qr.matching_query {
                    return qm.query_type == QueryType::Read;
                }
            }

            false
        });

        assert_eq!(any, None);
```
Which is very confusing to read and not a very robust test.


Claude-at-Instaclustr (approved):
LGTM

conorbros (approved):

Issue #445, opened 2022-01-13
Redesign MessageDetails


rukai:
## TODO
Break down of PRs that need to be made to complete the rewrite.
The stage headings demonstrate dependencies between PR's, e.g. Stage 3 PRs are dependent on Stage 2 PRs
Many of these items already have a corresponding implementation in https://github.com/shotover/shotover-proxy/pull/453 so please refer there first. The item will say `(take from PR)` if the functionality is already implemented in the draft PR.

### Stage 1
- [x] #454
- [x] #462
- [x] #495
- [x] #505

### Stage 2
- [x] #476
- [x] #472
- [x] #473
- [x] #474
- [x] #475
- [ ] Implement MessageInner. Frame, CassandraFrame and new Message constructors. Dont use them yet though (take from PR) (Can be further split up if you want)

### Stage 3
- [ ] Draw the rest of the owl. We might be able to further split up by trying to have some transforms run on the old system and some on the new. But I think it would be better at this point to just swap everything over. (refer to the PR)

### Stage 4
- [ ] Remove message decoding flags from source config (pull from PR)

## Problems

The handling of MessageDetails is a bit of a mess currently:
* Modifying just the raw message or message details leaves the other out of sync. We should pick one to be the source of truth and then generate the other from it.
* Some transforms expect the message details to be already generated, others will trigger generation of message details when they need it. We should handle this in a uniform way.
* There is duplicated yaml configuration across the Cassandra source and RedisCache for determining the pks to store in the message details, we should improve configuration by having the transform specify the pks it needs provided in the message details.

## Current instances of modifying messages

* RedisClusterPortsRewrite - modifies RawFrame (of the response)
* RedisTimestampTagger - modifies MessageDetails::Query(QueryMessage { ast, .. }) and MessageDetails::Response(QueryResponse { response_meta, result, .. })
* RedisCache - doesnt do anything yet but will soon - prompted me to investigate this issue.

## Proposed Redesign

Generate each part of QueryMessage/QueryResponse on demand in the transforms.
The only way to modify the message is by modifying the RawFrame which is extended to include the CSQL ast.

Effects:
* Make transforms drastically more robust, one source of truth means transforms are always reading the message provided by the previous transform.
* By requesting each part individually it should improve performance over generating MessageDetails up front for every message
    + Only a subset of messages in a subset of transforms need access to a subset of MessageDetails
[...]

PR #464, opened 2022-01-21
Implement From between Value and SQLValue

rukai:
Moves the value_to_expr and expr_to_value functions into From impls.
This generally makes sense and is already done for RedisFrame.
But it also make progress towards https://github.com/shotover/shotover-proxy/issues/445 as we need access to this logic from transforms.

### DIFF
--- a/shotover-proxy/src/message/mod.rs
+++ b/shotover-proxy/src/message/mod.rs
@@ -18,6 +18,7 @@ use num::BigInt;
 use ordered_float::OrderedFloat;
 use serde::{Deserialize, Serialize};
 use sqlparser::ast::Statement;
+use sqlparser::ast::Value as SQLValue;
 use std::collections::{BTreeMap, BTreeSet};
 use std::{collections::HashMap, net::IpAddr};
 use uuid::Uuid;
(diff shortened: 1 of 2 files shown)


conorbros (approved):

shotover-proxy/src/message/mod.rs
```diff
@@ -354,6 +355,33 @@ pub enum IntSize {
     I8,  // Tinyint
 }
 
+impl From<&Value> for SQLValue {
+    fn from(v: &Value) -> Self {
+        match v {
+            Value::NULL => SQLValue::Null,
+            Value::Bytes(b) => SQLValue::SingleQuotedString(String::from_utf8(b.to_vec()).unwrap()), // TODO: this is definitely wrong
```
benbromhead, line 362: It might be worth fixing this here? Or given a likely change in what we do with C* AST ... we can punt down the road again

benbromhead (approved):

shotover-proxy/src/message/mod.rs
rukai: Yeah I would rather address this properly later on.

After all this change is equivalent to just making these methods public.
But calling them from CassandraCodec::expr_to_value was not ideal.

PR #468, opened 2022-01-24
Tee FailOnMismatch returns the correct number of messages

rukai:
It is an invariant of transforms that they return the same number of messages as they received.
The previous implementation failed to do that.


conorbros (approved):
LGTM. I won't block since on this fixes the issue, but we should add some tests to Tee that verify this (and the behaviour of it's other options).

benbromhead (approved):

Issue #337, opened 2021-11-02
Improve naming


rukai:
For the purposes of logging and metrics we use a name to refer to a specific transform and to specific transform chains.

We want to rewrite this naming system to ensure that the displayed names have the following properties:
* Names should be unique. If the topology contains multiple transforms of the same type then they should be named such that the user can tell which transform in the topology.yaml is responsible for that log or metric. Same goes for transform chains.
* Implementation-wise they should get the name in the same way e.g. passed by argument or read from a global.

Maybe some of these goals are impossible or we need to compromise somewhere, but this should be our goal at least.

rukai:
I have arrived at 2 possible designs to solve this problem.

## Derived names
```yaml
---
sources:
  - redis:
      name: "redis"
      listen_addr: "127.0.0.1:6379"
      connection_limit: 3000000
      chain:
        - Tee:
            behavior: Ignore
            buffer_size: 10000
            chain:
            - QueryTypeFilter:
                DenyList: [Read]
            - Coalesce:
                flush_when_buffered_message_count: 2000
                # Use an unreasonably large timeout here so that integration tests dont break on slow hardware or a performance regression
                flush_when_millis_since_last_flush: 1000000000
            - QueryCounter:
                name: "backup chain"
            - RedisSinkCluster:
                first_contact_points: [ "127.0.0.1:2120", "127.0.0.1:2121", "127.0.0.1:2122", "127.0.0.1:2123", "127.0.0.1:2124", "127.0.0.1:2125" ]
                connect_timeout_ms: 3000
        - QueryCounter:
            name: "primary chain"
        - RedisSinkCluster:
            first_contact_points: [ "127.0.0.1:2220", "127.0.0.1:2221", "127.0.0.1:2222", "127.0.0.1:2223", "127.0.0.1:2224", "127.0.0.1:2225" ]
            connect_timeout_ms: 3000
  - redis:
      name: "redis passthrough"
      listen_addr: "127.0.0.1:6380"
      connection_limit: 3000000
      chain:
        - RedisSinkCluster:
            first_contact_points: [ "127.0.0.1:2220", "127.0.0.1:2221", "127.0.0.1:2222", "127.0.0.1:2223", "127.0.0.1:2224", "127.0.0.1:2225" ]
            connect_timeout_ms: 3000
```
* The sources are named "redis" and "redis passthrough"
* The source chains are named "redis source" and "redis passthrough source"
* The tee chain is named "redis source -> Tee chain transform"
[...]

PR #469, opened 2022-01-24
Use metrics handles in transforms

conorbros:
- Using metrics handles in transforms is pretty simple. 
- Using handles instead of the macros in `TransformChain` and `Wrapper` will require more complex changes. Will see if it's worth re-architecting things to do this.

### DIFF
--- a/shotover-proxy/src/server.rs
+++ b/shotover-proxy/src/server.rs
@@ -101,7 +101,7 @@ pub struct TcpCodecListener<C: Codec> {
     message_count: u64,
 
     /// Metric recording shotover available connections
-    gauge: Gauge,
+    available_connections_gauge: Gauge,
 }
 
 impl<C: Codec + 'static> TcpCodecListener<C> {
@@ -116,9 +116,9 @@ impl<C: Codec + 'static> TcpCodecListener<C> {
         trigger_shutdown_rx: watch::Receiver<bool>,
         tls: Option<TlsAcceptor>,
     ) -> Self {
-        let gauge =
+        let available_connections_gauge =
             register_gauge!("shotover_available_connections", "source" => source_name.clone());
-        gauge.set(limit_connections.available_permits() as f64);
+        available_connections_gauge.set(limit_connections.available_permits() as f64);
 
         TcpCodecListener {
             chain,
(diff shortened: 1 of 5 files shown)


shotover-proxy/src/server.rs
```diff
@@ -99,6 +99,9 @@ pub struct TcpCodecListener<C: Codec> {
 
     /// Keep track of how many messages we have received so we can use it as a request id.
     message_count: u64,
+
+    /// Metric recording shotover available connections
+    gauge: Gauge,
```
rukai, line 104: Can we name the metrics properly rather than putting the name in a comment?
Something like: `available_connections_gauge: Gauge`

rukai (approved):
Improved type safety and performance, nice!

Happy to merge as is but some thoughts:
There are some cases where second handle is created with the same args as another handle, will that result in duplicate metrics entries? And if yes, is that undesirable?

I think the TransformChainMetrics struct doesnt really do much for us.
It does have the advantage that we dont have to write `#[derivative(Debug = "ignore")]` twice but I think we would just be better off with less indirection.

conorbros:
There doesn't appear to be any side effects re-registering a metric. Also I presume once we implement a solution for https://github.com/shotover/shotover-proxy/issues/337 we will use the transform names in the metrics so it won't be the case for long.

I think that's reasonable, I'll remove the `TransformChainMetrics` struct.

Yeah I think it can be closed given we are using metric handles wherever possible

conorbros:
Ok, I've addressed the feedback. 

I've realized that the `new_no_shared_state` function does nothing different to the `new` function? I will make a PR after this to remove it.

rukai (approved):
passed x2

benbromhead (approved):
It looks like ithis changes some of the metrics names

rukai:
oh, I missed that, we should avoid that where possible

conorbros:
Yep that wasn't the intention. Which one changed?

benbromhead:
> Yep that wasn't the intention. Which one changed?
> 
> Edit: @benbromhead Are you referring to the ones with the transform names? e.g.
> 
> ```rust
> let failed_requests = register_counter!("failed_requests", "chain" => chain_name.clone(), "transform" => "CassandraSinkSingle");
> ```
> 
> Those haven't changed, I just removed the method that returned the transform name and put it as a literal in the metric.

Ah gotcha... all g if nothing has changed

PR #477, opened 2022-01-25
Remove new no state method from TransformChain

conorbros:
As mentioned on https://github.com/shotover/shotover-proxy/pull/469: 

The `new_no_shared_state` function on `TransformChain` does nothing different to the `new` function so this PR removes it.


rukai (approved):
nice

benbromhead (approved):

PR #478, opened 2022-01-25
Fix formatting on bench comments

rukai:
grep will output a `--` between results which markdown interprets to mean the above lines are headings.
We can stop grep from outputting those by doing: `grep --no-group-separator`

However we are better off just adding ``` around the log output because it also keeps the monospaced spacing in tact.
So that is what this PR does.


rukai:
nice it works!

Guess we'll also have to investigate why these simple bench cases are so volatile...

conorbros (approved):

benbromhead:
TBH I have a feeling it's the infra that we run on which is githubs oversubscribed workers...

iirc running benches locally I get a lot less variance

benbromhead (approved):

PR #479, opened 2022-01-25
Cleanup RedisCodec

rukai:
This change is functionally equivalent but:
* is simpler
* allocates less
* is a step towards the message details rewrite storing the raw redis Bytes in the message to skip encoding in same cases.

### DIFF
--- a/shotover-proxy/src/protocols/redis_codec.rs
+++ b/shotover-proxy/src/protocols/redis_codec.rs
@@ -5,7 +5,7 @@ use crate::message::{
 use crate::protocols::Frame;
 use crate::protocols::RedisFrame;
 use anyhow::{anyhow, Result};
-use bytes::{Bytes, BytesMut};
+use bytes::{Buf, Bytes, BytesMut};
 use itertools::Itertools;
 use redis_protocol::resp2::prelude::decode_mut;
 use redis_protocol::resp2::prelude::encode_bytes;
@@ -25,6 +25,7 @@ pub enum DecodeType {
 pub struct RedisCodec {
     decode_type: DecodeType,
     enable_metadata: bool,
+    messages: Messages,
 }
 
 #[inline]
@@ -493,6 +494,7 @@ impl RedisCodec {
         RedisCodec {
             decode_type,
             enable_metadata: false,
+            messages: vec![],
         }
     }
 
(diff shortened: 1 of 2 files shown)


rukai:
This doesnt block this PR but:
I think we are using the codec API slightly wrong.
Our `Item` is set to `Vec<Message>` but I think we are supposed to use just a `Message`.

The decode docs at https://docs.rs/tokio-util/0.6.9/tokio_util/codec/trait.Decoder.html#tymethod.decode say:
> If the bytes look valid, but a frame isn’t fully available yet, then Ok(None) is returned. This indicates to the Framed instance that it needs to read some more bytes before calling this method again.

Which suggests that we are supposed to read in a single frame at a time.

I assume they have some performance reasons for doing things this way so will have to investigate some time.

benbromhead:
Yup we do this for perf reasons. 

It makes a significant different if we can try and read as many frames off the network buffer in one hit for pipelined redis requests.

shotover-proxy/src/protocols/redis_codec.rs
```diff
@@ -587,26 +544,30 @@ impl Decoder for RedisCodec {
...
-    fn decode(
-        &mut self,
-        src: &mut BytesMut,
-    ) -> std::result::Result<Option<Self::Item>, Self::Error> {
-        Ok(match self.decode_raw(src)? {
-            None => None,
-            Some(f) => Some(self.process_redis_bulk(f)?),
-        })
+    fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>> {
+        let mut messages = vec![];
```
benbromhead, line 548: doesn't this mean that we create a new vec and then allocate on each push in the loop. We do this for each new call to decode (which happens a lot)?

rukai: This should have the same amount of allocations as the previous implementation.
* Empty vecs dont allocate. - Once we push a single item on an empty vec it allocates for 8 items.
* Every time we called `std::mem::take` in the previous implementation that would return the previous vec and leave a new empty vec in its place. - equivalent to just creating an empty vec every time.

benbromhead (approved):

conorbros (approved):

rukai:
There is a test consistently hanging in --release ci on this branch.
I'll investigate whats going on.

rukai:
Marking as draft so I dont merge until I figure out whats going on. (It does pass sometimes apparently)

rukai:
The issue occurs on around ~90% of test runs.
I resolved the issue by more accurately carrying over the original logic
You can see the changes in this commit: https://github.com/shotover/shotover-proxy/pull/479/commits/06f95e606e2b15689dcd683e095a4c5ef0479c29

I discovered there is an existing problem on main but this problem is super rare occurring on something like 1-5% of test runs.
I raised an issue to investigate this problem separately https://github.com/shotover/shotover-proxy/issues/491

conorbros (approved):

PR #480, opened 2022-01-25
Rename RawFrame -> Frame

rukai:
Frame and RawFrame are semantically a bit different but they end up getting used in pretty much the same places so may as well arrive at Frame by converting RawFrame into it.


benbromhead (approved):
LGTM

rukai:
rebased to ensure it passes on main.

conorbros (approved):

Issue #459, opened 2022-01-20
Handle duplicate stream_id in unordered_cluster_connection_pool.rs


rukai:
This code currently assumes stream id is unique but that is definitely not the case, it is valid for stream id to always be the same value

Anyone can start on this now, as its an improvement even outside of the context of the message details rewrite.

conorbros:
I can take this if you aren't

rukai:
that would be great!

conorbros:
Closing as we no longer have `UnorderedClusterConnectionPool`

PR #481, opened 2022-01-25
OwnedUnorderedConnectionPool fixes

conorbros:
While it's perfectly valid for a `stream_id` to be the same as another `stream_id`, it is very unlikely that a client will be sending multiple messages with the same `stream_id` asynchronously. If they are, it would indicate that they don't care about the order that the messages are coming back in. What's more likely is that they will only reuse a `stream_id` if sending messages synchronously so we wouldn't have any messages with the same `stream_id` waiting for responses in Shotover at the same time. 

More info on the use of `stream_id` in the Cassandra protocol is [here](https://github.com/apache/cassandra/blob/ff4d63d392c89e27ca6e02e38457a5199a24f6a9/doc/native_protocol_v4.spec#L160).

The current implementation of `OwnedUnorderedConnectionPool` will lose messages if they have the same `stream_id` so I have switched over to using a multi-map. If the same `stream_id` is used, the order doesn't matter, we just need to make sure we don't lose any messages. 

TODO:
- https://github.com/havarnov/multimap/pull/33 wait for this to be upstreamed. This PR is pointing at my personal fork for the moment.

### DIFF
--- a/shotover-proxy/src/transforms/util/unordered_cluster_connection_pool.rs
+++ b/shotover-proxy/src/transforms/util/unordered_cluster_connection_pool.rs
@@ -3,9 +3,10 @@ use crate::server::CodecReadHalf;
 use crate::server::CodecWriteHalf;
 use crate::transforms::util::{Request, Response};
 use crate::{message::Message, server::Codec};
+
 use anyhow::{anyhow, Result};
 use futures::StreamExt;
-use halfbrown::HashMap;
+use multimap::MultiMap;
 use std::fmt;
 use std::fmt::Formatter;
 use tokio::net::tcp::{OwnedReadHalf, OwnedWriteHalf};
@@ -113,10 +114,10 @@ async fn rx_process<C: CodecReadHalf>(
     codec: C,
 ) -> Result<()> {
     let mut in_r = FramedRead::new(read, codec);
-    let mut return_channel_map: HashMap<i16, (tokio::sync::oneshot::Sender<Response>, Message)> =
-        HashMap::new();
+    let mut return_channel_map: MultiMap<i16, (tokio::sync::oneshot::Sender<Response>, Message)> =
+        MultiMap::new();
 
-    let mut return_message_map: HashMap<i16, Message> = HashMap::new();
+    let mut return_message_map: MultiMap<i16, Message> = MultiMap::new();
 
     loop {
         tokio::select! {
@@ -130,7 +131,7 @@ async fn rx_process<C: CodecReadHalf>(
                     Ok(req) => {
                         for m in req {
                             if let Frame::Cassandra(frame) = &m.original {
-                                match return_channel_map.remove(&frame.stream_id) {
+                                match return_channel_map.remove_first(&frame.stream_id) {
                                     None => {
                                         return_message_map.insert(frame.stream_id, m);
                                     },
@@ -154,7 +155,7 @@ async fn rx_process<C: CodecReadHalf>(
                 }
 
                 if let Request { messages: orig , return_chan: Some(chan), message_id: Some(id) } = original_request {
-                    match return_message_map.remove(&id) {
+                    match return_message_map.remove_first(&id) {
                         None => {
                             return_channel_map.insert(id, (chan, orig));
                         }
--- a/shotover-proxy/src/transforms/util/unordered_cluster_connection_pool.rs
+++ b/shotover-proxy/src/transforms/util/unordered_cluster_connection_pool.rs
@@ -121,11 +121,11 @@ async fn rx_process<C: CodecReadHalf>(
     loop {
         tokio::select! {
             Some(maybe_req) = in_r.next() => {
+                if return_message_map.len() > 1 || return_channel_map.len() > 1 {
+                    info!("message map {:?}", return_message_map);
+                    info!("channel map {:?}", return_channel_map);
+                }
 
-                        if return_message_map.len() > 1 || return_channel_map.len() > 1 {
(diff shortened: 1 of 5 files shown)


benbromhead:
I think we can ignore the bench noise as the changes don't touch the regressed tests.

shotover-proxy/src/transforms/util/unordered_cluster_connection_pool.rs
```diff
@@ -143,20 +144,18 @@ async fn rx_process<C: CodecReadHalf>(
...
                     }
                 }
             },
             Some(original_request) = return_rx.recv() => {
-            if return_message_map.len() > 1 || return_channel_map.len() > 1 {
-                info!("message map {:?}", return_message_map);
-                info!("channel map {:?}", return_channel_map);
-            }
+                if return_message_map.len() > 1 || return_channel_map.len() > 1 {
+                    info!("message map {:?}", return_message_map);
```
rukai, line 153: This or the other message map log is giving an absurd amount of log spam.

rukai: Oh its because the len of a MultiMap never goes down.

conorbros: I'll remove them.

rukai:
I think MultiMap is problematic as it effectively leaks memory as empty entries are never cleaned up.

In general im struggling to review this because I do not understand unordered_cluster_connection_pool.rs 
Im kind of hoping we can clear out a bunch of this complexity once we understand what its actually doing.

rukai:
I'm not impressed by the existing implementation of OwnedUnorderedConnectionPool.
* Its not generic and is instead designed around cassandra. - relies on cassandra's stream id field.
* While it generates a bunch of connections its pool functionality is poor and only one of these connections is ever actually used.
* it doesnt seem unordered to me?

Can we throw it away and instead have CassandraSinkSingle do the following:
* create a single connection to the cassandra server
* send every message at once
* wait till we receive every reply and then return the replies.

In the future we can reimplement connection pooling for cassandra sinks, but we can ensure it is properly tested and code reviewed.

conorbros:
I think that's the best way forward, I'll close this PR and open a new one for removing the connection pool.

benbromhead:
Unfortuantly when you call the method `call_next_transform` on a wrapper in a transform, we need to guarentee (well... try really really hard) that we return the response associated with that request. Which is why we do all this stuff with maps and why I suggested that @conorbros should look at multi-maps and cassandra stream_ids to solve this challenge.

As @rukai rightfully points out in https://github.com/shotover/shotover-proxy/issues/459, it is valid for the stream id to always be the same value. However this would usually be in the case of a synchronous client (one request, one response only ever on the wire). With this in mind, we would expect that even by reusing a stream id, we would only have to deal with one set of request/responses in flight.

This doesn't take into account the fact a client could misbehave and this is where the choice to use the multi-map came in. 

If a cassandra client sends multiple requests with the same stream_id it is perfectly valid from a protocol perspective despite the fact it is somewhat of a "yolo" and now you don't know what response is for what request. This approach also breaks our ability to meet the `call_next_transform` guarenetee, with a multi-map approach we can at least guarentee we don't loose the message.

Re-reading through the code, some of the weirdness can certainly go away. 

- The needless clone is due to the fact that ResponseDetails wants a copy of the originating request! Which now that I look at it, is very silly. Transform authors, if they need the original request while doing stuff with the response, can simply `clone` it themselves before calling `call_next_transform`.
- Connection pooling on a per host basis (e.g. multiple upstream tcp sessions to a single host, from a single shotover tcp incoming session), largely doesn't make sense on a broader model of shotover. We should just maintain 1 connection per upstream host per shotover tcp session by default. If the client wants to drive more connections, they should open more connections through shotover. Stuff like this exists in the codebase for redis, but it's probably worth removing as it never improved performance and could also break assumptions about pipelined request ordering. 

However some of the weirdess around unordered message processing, but still being able to match requests and responses for sensible clients via stream_id to provide ordering / consistency for transform authors still very much needs to happen.

> Can we throw it away and instead have CassandraSinkSingle do the following:
> 
> 1. if a connection doesnt exist, create a single connection to the cassandra server
> 2. send every message at once
> 3. wait till we receive every reply and then return the replies

We can do this approach, however there are some changes / conditions we would need to do, to support it:
- Ensure the order of the `ChainResponse` messages matches the received order which we read the requesets off the wire, this would need to be done via stream_id.  
- Change the cassandra codec to support reading multiple frames off the network buffer in one go (the same way that redis does)

Thanks for your work @conorbros, though the points raised by @rukai and some further reflection of the whole way we do cassandra stuff means this as a whole needs a rework. So my apologies for the narrow guidance I provided earlier!

rukai:
> Change the cassandra codec to support reading multiple frames off the network buffer in one go (the same way that redis does)

Its possible tokio-util might do this internally for us already.
Will need to investigate how tokio-util implements this before we commit one way or the other.

conorbros:
Moving forward, my plan is to:
- simplify `OwnedUnorderedConnectionPool` to just be a single `CassandraConnection`. Keep most of the code that matches up responses using the `stream_id`. 
- Make `CassandraSinkSingle` have one of `CassandraConnection` similar to the way that `RedisSinkSingle` does.
- Investigate/fix the memory leak in the `MultiMap`. 

After that I'll investigate changing the Cassandra codec to support reading multiple frames off the network buffer in one go.

conorbros:
Closing in favour of #483

PR #482, opened 2022-01-27
Rename message::Value

rukai:
I chose ShotoverValue but maybe something like GenericValue or MessageValue could work?

### DIFF
--- a/shotover-proxy/src/message/mod.rs
+++ b/shotover-proxy/src/message/mod.rs
@@ -141,7 +141,7 @@ pub struct RawMessage {
 pub enum ASTHolder {
     // Statement is boxed because Statement takes up a lot more stack space than Value.
     SQL(Box<Statement>),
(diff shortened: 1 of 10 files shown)


conorbros (approved):
I prefer `MessageValue` because I imagine a transform developer interacting with `Message` and `MessageValue` better than `Message` and `ShotoverValue` but isn't a big deal so LGTM.

rukai:
Yeah, after thinking about it, I agree, MessageValue does sound better.
ill swap to that.

benbromhead (approved):

benbromhead (commented):
I'm ok with MessageValue, but something like DatastoreValue is more descriptive (e.g. it is a value from the datastore)

rukai:
It is currently used in cases the value didnt come from a datastore, so atm MessageValue makes more sense.
Although we might want to remove those cases in the future and then DatastoreValue would end up a better name.

PR #484, opened 2022-01-27
Re-enable test_protect_kms_transform

### LINKED ISSUE #316: Reenable test_protect_kms_transform
We had to disable this in https://github.com/shotover/shotover-proxy/pull/315 due to making the repo public preventing the use of github secrets.

Maybe we can mock out the AWS API like https://github.com/spulec/moto
Or maybe theres a way to safely distribute the key or a lower access key.

Claude-at-Instaclustr:
Adds moto as an AWS mock

fixes #316

### DIFF
--- a/shotover-proxy/src/transforms/protect/key_management.rs
+++ b/shotover-proxy/src/transforms/protect/key_management.rs
@@ -54,10 +54,10 @@ impl KeyManagerConfig {
                 client: {
                     let region_obj = match endpoint {
                         Some(x) => Region::Custom {
-                            name: Region::from_str(&*region)?.name().to_string(),
+                            name: Region::from_str(&region)?.name().to_string(),
                             endpoint: x,
                         },
-                        _ => Region::from_str(&*region)?,
+                        _ => Region::from_str(&region)?,
                     };
                     KmsClient::new(region_obj)
                 },
--- a/shotover-proxy/src/transforms/protect/key_management.rs
+++ b/shotover-proxy/src/transforms/protect/key_management.rs
@@ -31,6 +31,7 @@ pub enum KeyManagerConfig {
         key_spec: Option<String>,
         number_of_bytes: Option<i64>,
         grant_tokens: Option<Vec<String>>,
+        endpoint : Option<String>,
     },
     Local {
         kek: Key,
@@ -48,8 +49,21 @@ impl KeyManagerConfig {
                 key_spec,
                 number_of_bytes,
                 grant_tokens,
+                endpoint,
             } => Ok(KeyManager::AWSKms(AWSKeyManagement {
-                client: KmsClient::new(Region::from_str(region.as_str())?),
+                client: {
+                    let region_obj = match endpoint {
+                        Some(x) => {
+                            Region::Custom {
+                                name: Region::from_str(&*region)?.name().to_string(),
+                                endpoint: x,
+                            }
+                        },
+                        _ => Region::from_str(&*region)?,
+
+                    };
+                    KmsClient::new(region_obj )
+                },
                 cmk_id,
                 encryption_context,
                 key_spec,
--- a/shotover-proxy/src/transforms/protect/key_management.rs
+++ b/shotover-proxy/src/transforms/protect/key_management.rs
@@ -31,7 +31,7 @@ pub enum KeyManagerConfig {
         key_spec: Option<String>,
         number_of_bytes: Option<i64>,
         grant_tokens: Option<Vec<String>>,
-        endpoint : Option<String>,
+        endpoint: Option<String>,
     },
     Local {
         kek: Key,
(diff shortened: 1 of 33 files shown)


shotover-proxy/src/transforms/protect/mod.rs
```diff
@@ -421,11 +422,38 @@ mod protect_transform_tests {
         panic!()
     }
 
-    //#[tokio::test(flavor = "multi_thread")]
+    /// Creates a new AWS key and returns the key ARN
+    async fn create_aws_key() -> Result<String> {
+        let client = reqwest::Client::new();
+
+        let res = client.post("http://localhost:5000")
```
conorbros, line 429: Can use a [raw string literal](https://doc.rust-lang.org/reference/tokens.html#raw-string-literals) here

shotover-proxy/src/transforms/protect/mod.rs
```diff
@@ -421,11 +422,38 @@ mod protect_transform_tests {
...
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
     //#[ignore] // reason: requires AWS credentials
-    #[allow(unused)]
+    //#[allow(unused)]
     // Had to disable this when the repo went public as we dont have access to github secrets anymore
     async fn test_protect_kms_transform() -> Result<()> {
+        let _compose = DockerCompose::new("tests/transforms/docker-compose-moto.yml")
+            .wait_for("Press CTRL\\+C to quit");
```
conorbros, line 453: Can use raw string literal here too.

shotover-proxy/src/transforms/protect/key_management.rs
```diff
@@ -48,8 +49,18 @@ impl KeyManagerConfig {
...
+                endpoint,
             } => Ok(KeyManager::AWSKms(AWSKeyManagement {
-                client: KmsClient::new(Region::from_str(region.as_str())?),
+                client: {
+                    let region_obj = match endpoint {
+                        Some(x) => Region::Custom {
+                            name: Region::from_str(&*region)?.name().to_string(),
+                            endpoint: x,
+                        },
+                        _ => Region::from_str(&*region)?,
```
conorbros, line 60: Borrow and deference? We only need the borrow.

shotover-proxy/src/transforms/protect/mod.rs
```diff
@@ -421,11 +422,38 @@ mod protect_transform_tests {
...
+        Ok(parsed["KeyMetadata"]["Arn"].as_str().unwrap().to_string())
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
     //#[ignore] // reason: requires AWS credentials
-    #[allow(unused)]
+    //#[allow(unused)]
     // Had to disable this when the repo went public as we dont have access to github secrets anymore
     async fn test_protect_kms_transform() -> Result<()> {
+        let _compose = DockerCompose::new("tests/transforms/docker-compose-moto.yml")
```
conorbros, line 452: This is an integration test, so it should go with the other integration tests. You would typically only put unit tests at the bottom of a file.

shotover-proxy/src/transforms/protect/mod.rs
```diff
@@ -421,11 +422,38 @@ mod protect_transform_tests {
...
+
+        let result = res.text().await?;
+        let parsed: serde_json::Value = serde_json::from_str(result.as_str())?;
+        Ok(parsed["KeyMetadata"]["Arn"].as_str().unwrap().to_string())
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
     //#[ignore] // reason: requires AWS credentials
-    #[allow(unused)]
+    //#[allow(unused)]
```
conorbros, line 449: Should delete this commented out stuff if we don't need it anymore.

conorbros (changes requested):

conorbros:
Looks like you need to run `cargo fmt`

shotover-proxy/src/transforms/protect/mod.rs
Claude-at-Instaclustr: I did not place the test, just fixed it.

shotover-proxy/src/transforms/protect/key_management.rs
Claude-at-Instaclustr: done

shotover-proxy/src/transforms/protect/mod.rs
Claude-at-Instaclustr: done

shotover-proxy/src/transforms/protect/mod.rs
rukai: Yeah I wouldnt call this an integration test unless we actually run full shotover against a cassandra instance like our other integration tests do.
Also its already here, moving it in this PR would mess up the diff.

rukai (approved):
If rusoto has an API for creating the key we should use that instead, but I dont consider that a blocker because we will want to replace rusoto with amazons official crate at some point.

Nice work getting this test running again!

shotover-proxy/src/transforms/protect/mod.rs
conorbros: Yep, I was incorrect. The `DockerCompose` threw me off.

conorbros (approved):
LGTM!

PR #487, opened 2022-01-27
Loopback transform directly returns the message

### LINKED ISSUE #472: Rewrite the loopback transform to directly return the message
Rewrite the loopback transform such that it just directly returns the messages without processing.
Will also need to remove it from the public API as this will perform poorly in real world scenarios.
(take from https://github.com/shotover/shotover-proxy/pull/453)

conorbros:
closes #472

### DIFF
--- a/shotover-proxy/src/transforms/mod.rs
+++ b/shotover-proxy/src/transforms/mod.rs
@@ -25,6 +25,7 @@ use crate::transforms::distributed::consistent_scatter::{
 use crate::transforms::filter::{QueryTypeFilter, QueryTypeFilterConfig};
 use crate::transforms::kafka_sink::{KafkaSink, KafkaSinkConfig};
 use crate::transforms::load_balance::ConnectionBalanceAndPool;
+#[cfg(test)]
 use crate::transforms::loopback::Loopback;
 use crate::transforms::null::Null;
 use crate::transforms::parallel_map::{ParallelMap, ParallelMapConfig};
@@ -74,6 +75,7 @@ pub enum Transforms {
     RedisCache(SimpleRedisCache),
     Tee(Tee),
     Null(Null),
+    #[cfg(test)]
     Loopback(Loopback),
     Protect(Protect),
     ConsistentScatter(ConsistentScatter),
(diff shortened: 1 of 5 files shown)


shotover-proxy/src/transforms/protect/mod.rs
```diff
@@ -269,7 +269,9 @@ mod protect_transform_tests {
     use crate::transforms::protect::ProtectConfig;
     use crate::transforms::{Transform, Transforms, Wrapper};
 
-    #[tokio::test(flavor = "multi_thread")]
+    //#[tokio::test(flavor = "multi_thread")]
+    #[allow(unused)]
+    #[ignore]
```
rukai, line 274: Can you remove these changes please?
Dont want merge conflicts with https://github.com/shotover/shotover-proxy/pull/484

conorbros: I'll rebase now that the test is fixed.

rukai (approved):
lgtm

conorbros:
Hmm, it's failing on the new protect tests.

shotover-proxy/src/transforms/mod.rs
```diff
@@ -105,6 +107,7 @@ impl Transforms {
             Transforms::Tee(m) => m.transform(message_wrapper).await,
             Transforms::DebugPrinter(p) => p.transform(message_wrapper).await,
             Transforms::Null(n) => n.transform(message_wrapper).await,
+            #[cfg(test)]
```
rukai, line 110: Oh wait, cant we get away with just adding the `#[cfg(test)]` to the `TransformsConfig::get_transforms` Loopback arm?

conorbros: Yes. Though only compiling/including what we need in our build would be a good habit to follow.

rukai: Is the advantage of that a smaller binary size?
I think in general we want to keep our test and production binaries as close as possible.
But in this particular case im not super worried either way, so just resolve this if you'd rather leave it as is.

conorbros: Yes I was thinking smaller binary size.

benbromhead (approved):

PR #486, opened 2022-01-27
Cassandra prepared statements int tests

conorbros:

PR #493, opened 2022-01-31
NO_METADATA CassandraCodec fix

conorbros:
Current implementation assumes there will always be metadata, we should first check if the `NO_METADATA` flag was included before trying to use `col_spec` to convert the `row_content` into a `MessageValue`.

There is also a PR <s>open</s> merged on cdrs-tokio that fixes another issue associated with this: https://github.com/krojew/cdrs-tokio/pull/76

The combination of these PRs will fix the bug discovered on https://github.com/shotover/shotover-proxy/pull/486.

### DIFF
--- a/shotover-proxy/src/protocols/cassandra_codec.rs
+++ b/shotover-proxy/src/protocols/cassandra_codec.rs
@@ -512,25 +512,42 @@ impl CassandraCodec {
         match frame.response_body().unwrap() {
             ResponseBody::Error(e) => error = Some(MessageValue::Strings(e.message)),
             ResponseBody::Result(ResResultBody::Rows(rows)) => {
-                let converted_rows = rows
-                    .rows_content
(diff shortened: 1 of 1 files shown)


rukai:
So when is the NO_METADATA flag set?
I think this could cause problems for some transforms though I would have to check.

conorbros:
So far, I've found that when clients execute a prepared statement they set the `NO_METADATA` flag as they will already know the resulting metadata of the query they are about to get a response to.

rukai (approved):
So currently we have different parts of shotover reading and writing to Document, NamedRows and Rows and they are all incompatible with each other.

I'm happy to land this as is as its more correct than the previous implementation, however this incompatibility is an issue in this area that I think we should properly resolve once we finish the message details rewrite.

benbromhead (approved):

Issue #462, opened 2022-01-20
Cleanup test_protect_kms_transform and test_protect_transform


rukai:
Prereq https://github.com/shotover/shotover-proxy/issues/316

Rewrite test_protect_kms_transform and test_protect_transform

Getting the keys setup on CI and cleaning up the tests can be done before the message details rewrite is approved.

However once it is approved we will want to remove dependencies on QueryResponse stuff in a follow up PR.

PR #494, opened 2022-01-31
Implement Protect integration tests

rukai:
closes https://github.com/shotover/shotover-proxy/issues/462

The motivation here is these integration tests will serve as replacements for the Protect unittests as those unittests are filled with implementation details of the old implementation.

I put the new tests in with the cassandra int tests as that is the only protocol that attempts to support protect at the moment.
When new protocols support it they should get their own int tests for it.

I added a new `protect` feature to allow us to write integration tests for protect while not including it in the public shotover API.
`#[cfg(test)]` does not allow this as it is only enabled during unit tests not integration tests.

I manually verified that the integration test tests the encryption code path as far as the current Protect implementation will go, but hopefully the message details rewrite will remove a lot of broken stuff and allow this test to actually test that encryption/decryption occurs.

### DIFF
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -8,6 +8,10 @@ license = "Apache-2.0"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[features]
+# Include WIP alpha transforms in the public API
+alpha-transforms = []
+
 [dependencies]
 tokio = { version = "1.14.0", features = ["full", "macros"] }
 tokio-util = { version = "0.6.1", features = ["full"] }
--- a/shotover-proxy/examples/cassandra-protect-aws/docker-entrypoint.sh
+++ b/shotover-proxy/examples/cassandra-protect-aws/docker-entrypoint.sh
@@ -1,0 +1,100 @@
+#!/bin/bash
+set -e
+
+###  https://github.com/docker-library/cassandra/blob/master/3.11/docker-entrypoint.sh
+###  This file has been copied from the above link and modified to include enable_user_defined_functions and enable_scripted_user_defined_functions to the for yaml in \ loop
+###  so extra cassandra config options can be modified from the docker-compose.yml
+
+# first arg is `-f` or `--some-option`
+# or there are no args
+if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
+	set -- cassandra -f "$@"
+fi
+
+# allow the container to be started with `--user`
+if [ "$1" = 'cassandra' -a "$(id -u)" = '0' ]; then
+	find "$CASSANDRA_CONF" /var/lib/cassandra /var/log/cassandra \
+		\! -user cassandra -exec chown cassandra '{}' +
+	exec gosu cassandra "$BASH_SOURCE" "$@"
+fi
+
+_ip_address() {
+	# scrape the first non-localhost IP address of the container
+	# in Swarm Mode, we often get two IPs -- the container IP, and the (shared) VIP, and the container IP should always be first
+	ip address | awk '
+		$1 != "inet" { next } # only lines with ip addresses
+		$NF == "lo" { next } # skip loopback devices
+		$2 ~ /^127[.]/ { next } # skip loopback addresses
+		$2 ~ /^169[.]254[.]/ { next } # skip link-local addresses
+		{
+			gsub(/\/.+$/, "", $2)
+			print $2
+			exit
+		}
+	'
+}
+
+# "sed -i", but without "mv" (which doesn't work on a bind-mounted file, for example)
+_sed-in-place() {
+	local filename="$1"; shift
+	local tempFile
+	tempFile="$(mktemp)"
+	sed "$@" "$filename" > "$tempFile"
+	cat "$tempFile" > "$filename"
+	rm "$tempFile"
+}
+
+if [ "$1" = 'cassandra' ]; then
+	: ${CASSANDRA_RPC_ADDRESS='0.0.0.0'}
+
+	: ${CASSANDRA_LISTEN_ADDRESS='auto'}
+	if [ "$CASSANDRA_LISTEN_ADDRESS" = 'auto' ]; then
+		CASSANDRA_LISTEN_ADDRESS="$(_ip_address)"
+	fi
+
+	: ${CASSANDRA_BROADCAST_ADDRESS="$CASSANDRA_LISTEN_ADDRESS"}
+
+	if [ "$CASSANDRA_BROADCAST_ADDRESS" = 'auto' ]; then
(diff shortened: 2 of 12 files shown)


conorbros:
Why are we not including the Protect transform in the public API?

shotover-proxy/examples/cassandra-protect-local/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+services:
+  cassandra-one:
+    image: library/cassandra:3.11.10
+    ports:
+      - "9043:9042"
+    environment:
+      MAX_HEAP_SIZE: "400M"
+      MIN_HEAP_SIZE: "400M"
+      HEAP_NEWSIZE: "48M"
+      CASSANDRA_ENABLE_SCRIPTED_USER_DEFINED_FUNCTIONS: "true"
```
conorbros, line 11: Don't need this if you aren't using any UDFs

shotover-proxy/examples/cassandra-protect-aws/docker-entrypoint.sh
```diff
@@ -0,0 +1,100 @@
...
+	if [ "$CASSANDRA_LISTEN_ADDRESS" = 'auto' ]; then
+		CASSANDRA_LISTEN_ADDRESS="$(_ip_address)"
+	fi
+
+	: ${CASSANDRA_BROADCAST_ADDRESS="$CASSANDRA_LISTEN_ADDRESS"}
+
+	if [ "$CASSANDRA_BROADCAST_ADDRESS" = 'auto' ]; then
+		CASSANDRA_BROADCAST_ADDRESS="$(_ip_address)"
+	fi
+	: ${CASSANDRA_BROADCAST_RPC_ADDRESS:=$CASSANDRA_BROADCAST_ADDRESS}
```
conorbros, line 60: Don't need this script if we aren't using any `CASSANDRA_*` env variables.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1140,6 +1140,52 @@ mod cache {
     }
 }
 
+#[cfg(feature = "protect")]
+mod protect {
```
conorbros, line 1144: This could just be a function since it only has one test in it, unless there is more tests you are planning to add.

shotover-proxy/Cargo.toml
```diff
@@ -8,6 +8,10 @@ license = "Apache-2.0"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[features]
+# Include the WIP Protect transform in the public API
+protect = []
```
conorbros, line 13: I'm thinking this should be called `int-test` or similar so it's clear it has no purpose other than integration testing. Unless it's not just for integration testing?

rukai: In theory it would allow users who are willing to compile from source to try out new unreleased features.
Maybe we should call the feature `alpha-transforms` to allow for that case while also avoiding having lots of different features for each alpha transform floating around.

> Why are we not including the Protect transform in the public API?

New alpha transforms should not be public because:
* We dont want non-functional features to be available to users.
* We dont want to commit to a stable API yet.

shotover-proxy/examples/cassandra-protect-local/docker-compose.yml
rukai: I would like to run the full test suite in these tests eventually, but for now i've left the rest of the test suite out to make developing Protect easier.
Maybe I should just add the rest of the test suite now though?

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
rukai: Having it as a module makes it easier to feature flag both the function and imports.

shotover-proxy/Cargo.toml
conorbros: That makes sense to me. I agree with the name suggestion `alpha-transforms`.

shotover-proxy/examples/cassandra-protect-local/docker-compose.yml
conorbros: Sure, might as well include in this PR.

rukai: alright, added

conorbros (approved):

shotover-proxy/examples/cassandra-protect-local/topology.yaml
```diff
@@ -0,0 +1,64 @@
...
+          - pk
+        test.clustering:
+          - pk
+          - clustering
+chain_config:
+  main_chain:
+    - Protect:
+        key_manager:
+          Local:
+            kek:
```
benbromhead, line 20: I think yaml allows us to put the raw bytes in like a json array format which is a bit easier on the eyes `[ 30, 223, 12 ... ]`

benbromhead (approved):
One minor nit on the cassandra protect example yaml

shotover-proxy/examples/cassandra-protect-local/topology.yaml
rukai: Personally I would just leave it as is otherwise we have a 160 character wide line.
@conorbros which would you pick?

conorbros: I saw that and thought there would be a better way to do it (perhaps formatting it as hex?) but decided to leave it to not block this PR and get the tests in.

rukai: Alright, I kept it on one line and hexed the values.

conorbros: I meant as a long hex string.

conorbros: Not sure if it's possible to do that, but it seems strange to have to a supply a key byte by byte. I'd imagine you'd have it as a hex or base64 encoded string instead.

rukai: oh is it possible to write an array of integers as a hex string in yaml?

conorbros: I was thinking we read it as an encoded string and then convert into the bytes format that `sodiumoxide` expects.

rukai: ah right, I see what you mean.
That would certainly be nicer, but as you said earlier, lets address that later.

conorbros (approved):

shotover-proxy/examples/cassandra-protect-local/topology.yaml
conorbros: Opened an issue for it: #510

PR #497, opened 2022-02-01
CassandraConnection cleanup

conorbros:
### CassandraConnection changes
- Moved `Request` struct to be private to `CassandraConnection`. This means that the `Request` used for `ConnectionPool` doesn't need to have an `Option<i16>` field for the message id.
- added a `send` method. Transforms can just give the connection pool their message and return channel instead of having to interact with the implementation of `CassandraConnection`.

### Other changes
- Renamed `messages` in `Request` to `message` because it only holds one message.
- Changed `Response` from a tuple to a struct. This was used lots and it was confusing. Sometimes it was used without the `Response` type and instead another tuple was declared.

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/connection.rs
+++ b/shotover-proxy/src/transforms/cassandra/connection.rs
@@ -1,7 +1,7 @@
 use crate::protocols::Frame;
 use crate::server::CodecReadHalf;
 use crate::server::CodecWriteHalf;
-use crate::transforms::util::{Request, Response};
+use crate::transforms::util::Response;
 use crate::{message::Message, server::Codec};
 
 use anyhow::{anyhow, Result};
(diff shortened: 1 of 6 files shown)


shotover-proxy/src/transforms/cassandra/connection.rs
```diff
@@ -36,26 +45,45 @@ impl<C: Codec + 'static> CassandraConnection<C> {
...
         self.connection = Some(out_tx);
         Ok(())
     }
+
+    /// Send a `Message` to this `CassandraConnection` and expect a response on `return_chan`
+    pub fn send(&self, message: Message, return_chan: oneshot::Sender<Response>) -> Result<()> {
+        let message_id = if let Frame::Cassandra(frame) = &message.original {
+            frame.stream_id
+        } else {
+            info!("no cassandra frame found");
```
rukai, line 63: I think returning Err should be enough, no need to also info log it.

conorbros:
My PR on multi-map hasn't been merged yet so I haven't included that in this PR. I can point it at my fork or wait until it's merged and then switch over the `HashMap`?

rukai:
Yeah HashMap can go in a follow up PR.
Keeps this PR more manageable too.

rukai (approved):
Some really nice improvements here!

benbromhead (approved):

PR #502, opened 2022-02-02
Fix ShotoverProcess timeout

rukai:
It is possible on very slow machines for the building part of `cargo run` to cause `test_shotover_responds_sigterm` to timeout due to spending the entire timeout time building shotover instead of running it.

I am confident that we dont need a timeout on `cargo build`, so this PR first runs that to ensure shotover is completely built before running `cargo run`.

This should fix the CI failures I am getting in https://github.com/shotover/shotover-proxy/pull/494
Even if it doesnt, then its still a useful protection against a possible failure case.


conorbros (approved):
Nice! Very glad to see this fixed.

benbromhead (approved):

Issue #208, opened 2021-09-22
Implement system.peers table port / IP address rewriting for Cassandra


benbromhead:
The `system.peers` table in Cassandra describes the nodes in the cluster. Cassandra drivers will query this to find other nodes. We need to be able to rewrite the port information in this command to point to the shotover port rather than the direct cassandra port. 

We will also need to support rewriting the IP addresses to refer to the shotover instance for cases where we want all traffic going through shotover. This will either be just rewriting the IP address and port combinations, or to rewrite the response entirely to have shotover pretend to be a single node cassandra cluster.

We will need to look at cassandra driver behavior to see what the best approach is

PR #507, opened 2022-02-03
Cassandra peers

conorbros:
Implements the Cassandra peers rewriting transform


rukai:
Using regex like that will result in false positives.
Also there is a bunch of things that are a bit wonky given shotovers current design, but resolving them doesnt make sense given the upcoming message details redesign.

Are you fine with putting this on hold until the message details redesign is finished or maybe at least until https://github.com/shotover/shotover-proxy/pull/508 lands

conorbros:
I can leave it on hold until the message details rewrite.

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
```diff
@@ -0,0 +1,323 @@
...
+                }
+            }
+        } else {
+            panic!(
+                "Expected CassandraOperation::Result(CassandraResult::Rows), got {:?}",
+                frame
+            );
+        }
+    } else {
+        panic!("Expected Frame::Cassandra, got {:?}", message.original);
```
rukai, line 126: I've been thinking about this a bit.
In general transforms are supposed to handle any message type but there are some transforms that only make sense with one message type (like this one) and would indicate a user configuration error if they encountered another message type.
A panic here is no good because it is possible for the user to hook a Redis Source up to a CassandraPeersRewrite and we shouldn't panic even under bad configuration.
We could return an Err here but that still allows the system to run under bad configuration.
The best solution here is to add a system to our validation system that allows transforms to define what message types they are compatible with.
Then we can fail to start shotover on a bad configuration.

However the transform invariants currently don't specify whether a transform instance will always receive messages of the same type.
If we cant think of a use case for multiple message types going down the same chain then we should:
* explicitly specify in the invariants that transforms will always receive messages of the same type and that transforms should never allow messages of mixed types to flow down the chain. (They can completely replace messages of type A with type B though)
* implement the allowed message type configuration validation
This should of course go in a separate issue so we can address it separately to this PR

For now I think the current implementation that panics is the best solution here.

shotover-proxy/src/transforms/mod.rs
```diff
@@ -217,6 +222,7 @@ pub enum TransformsConfig {
     CassandraSinkSingle(CassandraSinkSingleConfig),
     RedisSinkSingle(RedisSinkSingleConfig),
     KafkaSink(KafkaSinkConfig),
+    CassandraPeersRewrite(CassandraPeersRewriteConfig),
```
rukai, line 225: We either need to add `#[cfg(feature = "alpha-transforms")]` or document the functionality in transforms.md

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
conorbros: The panic here is because we should never be passing a Redis message into `rewrite_port` or `emulate_single` node unless the `is_system_peers` function is broken.

rukai: Oh, true!

Maybe include that in the panic message, and ill dump my above thoughts in an issue.

shotover-proxy/src/transforms/mod.rs
conorbros: I've got a docs and example ready in this branch: https://github.com/conorbros/shotover-proxy/tree/cassandra-docs

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
conorbros: Yep, I will make it more obvious what is happening there.

shotover-proxy/tests/test-topologies/cassandra-peers-rewrite/topology.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - CassandraPeersRewrite:
+        rewrite:
+          Port: 9044
+    - CassandraSinkSingle:
+        result_processing: false
+        remote_address: "172.16.1.2:9042"
+  peers_rewrite_emulate_single_node:
+    - CassandraPeersRewrite:
+        rewrite:
+          EmulateSingleNode
```
conorbros, line 56: Update: changed the config to be an enum instead of optional field and a boolean

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
```diff
@@ -0,0 +1,323 @@
...
+        }
+    }
+
+    false
+}
+
+/// Emulate a single node by removing the rows from a query to system.peers(_v2)
+/// Only Cassandra queries to the `system.peers` table found via the `is_system_peers(_v2)` functions should be passed to this
+fn emulate_single_node(message: &mut Message) {
+    if let Frame::Cassandra(ref mut frame) = message.original {
```
rukai, line 81: `ref mut frame` syntax was superceded by `&mut message.original` syntax.
So it is preferred to write as:
`if let Frame::Cassandra(frame) = &mut message.original {`

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1373,3 +1374,38 @@ fn test_cassandra_protect_transform_aws() {
...
+        &normal_connection,
+        "SELECT native_port FROM system.peers_v2;",
+        &[&[ResultValue::Int(9042)]],
+    );
+
+    assert_query_result(
+        &rewrite_port_connection,
+        "SELECT native_port FROM system.peers_v2;",
+        &[&[ResultValue::Int(9044)]],
+    );
```
rukai, line 1404: We should also have a case with multiple columns, maybe `SELECT * FROM system.peers_v2;`, to ensure that our logic for modifying just the native_port is correct.

rukai (changes requested):
Still need to go over the unit tests and ensure that the transform actually solves the problem it set out to solve, but heres some feedback before I get to that.

shotover-proxy/tests/test-topologies/cassandra-peers-rewrite/topology.yaml
```diff
@@ -0,0 +1,66 @@
+---
+sources:
+  cassandra_prod_1:
+    Cassandra:
+      query_processing: false
+      listen_addr: "127.0.0.1:9043"
+      cassandra_ks:
+        system.local:
```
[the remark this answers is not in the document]
conorbros: That field is not optional, I'll raise a ticket to make it so.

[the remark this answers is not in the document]
rukai: ah I just mean make the map empty

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1373,3 +1374,38 @@ fn test_cassandra_protect_transform_aws() {
...
+    let _docker_compose =
+        DockerCompose::new("tests/test-topologies/cassandra-peers-rewrite/docker-compose.yml")
+            .wait_for_n_t("Startup complete", 2, 90);
+
+    let shotover_manager = ShotoverManager::from_topology_file(
+        "tests/test-topologies/cassandra-peers-rewrite/topology.yaml",
+    );
+
+    let normal_connection = shotover_manager.cassandra_connection("127.0.0.1", 9043);
+    let rewrite_port_connection = shotover_manager.cassandra_connection("127.0.0.1", 9044);
```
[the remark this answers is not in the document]
conorbros: I don't think it's worth the extra runtime the run the entire test suite against them both. Do you think running just the table test suite would be enough?

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
conorbros: I've added a couple extra columns, I didn't want to use `*` as the `tokens` column would make the expected results very large.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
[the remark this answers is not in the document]
rukai: Yeah just the table test suite would be a good compromise

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
rukai: Thats fine, can we also keep the original test case though?
1 returned value is a nice edge case to test.

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -4,7 +4,7 @@ use ordered_float::OrderedFloat;
 mod basic_driver_tests;
 
 #[derive(Debug, Clone, PartialOrd, PartialEq, Eq, Ord)]
-enum ResultValue {
+pub enum ResultValue {
```
rukai, line 7: and this one too

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
conorbros: Sure, I'll add it back.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1389,31 +1388,50 @@ fn test_cassandra_peers_rewrite() {
     );
 
     let normal_connection = shotover_manager.cassandra_connection("127.0.0.1", 9043);
+
     let rewrite_port_connection = shotover_manager.cassandra_connection("127.0.0.1", 9044);
+    keyspace::test(&rewrite_port_connection); // run some basic tests to confirm it works as normal
```
rukai, line 1393: why keyspace instead of table?

rukai: oh uh... fair enough.
Thats a reasonable compromise for now then.

rukai:
So what is the actual use case of this transform?
My current understanding is:

## RewriteConfig::Port
* There is a shotover instance in front of every instance of cassandra
    + The paired cassandra and shotover instances are on the same IP address
    + Shotover is always on a specific port
    + Cassandra is always on a specific port
* We want to force the application to always send messages to shotover instead of going directly to cassandra instances.

## RewriteConfig::EmulateSingleNode
I'm not sure what this is for?
I guess its to do with https://github.com/shotover/shotover-proxy/issues/208#issuecomment-925482470
But it feels like only part of a solution, dont we now need to handle cassandra routing in shotover itself?

Is this correct? Anything I missed?

Is it possible to write integration tests that actually test these use cases?

conorbros:
Hmm true, it seems the port rewriting would handle that. BTW Cassandra will handle the routing in that case, it will forward requests to the right node.

rukai:
> BTW Cassandra will handle the routing in that case though, it will forward requests to the right node.

Oh I see. Thats different to redis where the node reports that you should requery against a different node via `MOVED`.

So what purpose does the cassandra peers table serve?
Is the idea that the client should hit a random node each time and then that node will fetch the actual data from the correct node for you?

conorbros:
No, the peers table provides you will a least of peers and their tokens, if you have the token information you can hit the correct node to get the data faster. If you don't have the tokens, you can hit any node and that node will request the data from the node that has it, retrieve it and then send it back to you.

That said, I'm not sure why you would want to emulate a single node instead of just using the port rewrite, since you would lose the ability to do token aware routing. I think Ben intended it to be used with a case where Shotover is maintaining connections to a cluster of nodes, not just running as sidecars to a single node each.

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
```diff
@@ -0,0 +1,321 @@
...
+            warnings: vec![],
+        };
+
+        let bytes = frame.encode_with(Compression::None).unwrap();
+
+        Message {
+            details: MessageDetails::Unknown,
+            return_to_sender: false,
+            modified: false,
+            original: Frame::Cassandra(CassandraFrame::from_bytes(bytes.into()).unwrap()),
```
rukai, line 179: Can we just directly use a CassandraFrame here without round tripping through bytes?

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
```diff
@@ -0,0 +1,321 @@
...
+            stream_id: 0,
+            opcode: Opcode::Query,
+            tracing_id: None,
+            body: body.serialize_to_vec(),
+            warnings: vec![],
+        };
+
+        let bytes = frame.encode_with(Compression::None).unwrap();
+
+        Message {
```
rukai, line 175: Can we use Message::from_frame here?
Manual construction of Message will be made impossible by the message details rewrite.

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
```diff
@@ -0,0 +1,321 @@
...
+            details: MessageDetails::Unknown,
+            return_to_sender: false,
+            modified: false,
+            original,
+        }
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_is_system_peers() {
```
rukai, line 220: Why are these async? I dont see any awaits

conorbros:
Yeah, I believe that was the plan all along, I think I misinterpreted Ben's comment thinking the emulate single node feature was supposed to be used with `CassandraSinkSingle`. I'll remove it from this PR and we can add it back in when we add Cassandra clustering support.

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
conorbros: There used to be async there but I forgot to remove it.

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
conorbros: I added a method to `CQL` to make constructing these simpler.

rukai:
Alright, sounds good to me

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
```diff
@@ -0,0 +1,244 @@
...
+                return namespace[0] == "system" && namespace[1] == "peers_v2";
+            }
+        }
+    }
+
+    false
+}
+
+/// Rewrite the `native_port` field in the results from a query to `system.peers_v2` table
+/// Only Cassandra queries to the `system.peers` table found via the `is_system_peers(_v2)` functions should be passed to this
```
rukai, line 65: is_system_peers_v2 does not exist?

rukai (approved):
LGTM!

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
conorbros: Ah yep, it used to exist.

benbromhead:
> So instead we have a single shotover instance through which everything is routed for this application.
> 
> As is this would "just work" due to cassandra internally rerouting to other nodes. But I dont see why we couldnt implement token aware routing within shotover to get back that lost performance?

This

benbromhead:
LGTM

benbromhead (approved):

PR #511, opened 2022-02-07
Remove duplicated docker-entrypoint script

conorbros:
Need to modify the script in #437 and figured it makes sense to only have one version of this script floating around the repo.


rukai (approved):
Thanks for doing this!

Maybe we should one day split the examples into `examples/cassandra/`, `examples/redis/`, `examples/multi-protocol/`, but we should probably leave that until we have more examples (and therefore a better idea of how to organize them)

benbromhead (approved):

PR #512, opened 2022-02-08
Cassandra benchmark

conorbros:
- Moved the `cassandra_connection` helper to a method on `ShotoverManager`. 
- Extended the `BenchResources` struct to work with Cassandra/Shotover combinations as well as Redis/Shotover. 
- Removed unneeded `move` closure in the redis benches.
- Changed the ports for the cassandra passthrough example, we should try and keep Cassandra running on it's default port and run Shotover on a different one. Makes easier to reason about what you are connecting to when testing things. 
- Added a benchmark for `cassandra-passthrough` example, I intended to add another one in #437.


rukai:
Looks like we lost the CI changes..?

conorbros:
They were on the other PR.

rukai:
Oh right, this one https://github.com/shotover/shotover-proxy/pull/437/files
The changes are pretty minor so I guess you can just duplicate it across both PR's and then rebase which other one is merged 2nd.
Otherwise pull them out into another PR because I suspect the TLS changes will take a lot more investigation.

conorbros:
Alright, I'll pulled them out into this one. I'll just confirm the CI works and then we can get this reviewed.

shotover-proxy/benches/mod.rs
```diff
@@ -0,0 +1,58 @@
+use cassandra_cpp::{stmt, Session};
+pub use test_helpers::docker_compose::DockerCompose;
+
+#[path = "../tests/helpers/mod.rs"]
+mod helpers;
+pub use helpers::ShotoverManager;
+
+pub struct BenchResources {
```
rukai, line 8: I feel like combining redis and cassandra bench resources into a single struct just complicates things.
As is, despite having fields for both cassandra and redis connections the API does not allow both to be used at the same time.
I think we would be better off with:
```rust
// defined in redis_benches.rs
BenchResources {
    _compose: DockerCompose,
    _shotover_manager: ShotoverManager,
    pub connection: redis::Connection,
}

// defined in cassandra_benches.rs
BenchResources {
    _compose: DockerCompose,
    _shotover_manager: ShotoverManager,
    pub connection: Session,
}
```

Unless theres some advantage im missing?

shotover-proxy/examples/cassandra-passthrough/docker-compose.yml
```diff
@@ -3,7 +3,7 @@ services:
   cassandra-one:
     image: library/cassandra:3.11.10
     ports:
-      - "9043:9042"
```
rukai, line 6: Can you move this into a separate PR?
It might be the correct approach but it will need some consideration and we will want to do the same for redis.

shotover-proxy/benches/mod.rs
conorbros: It was an attempt to de-duplicate the code given it's a very simple struct that performs the exact same function in both cases, enable the ability to extend it in cases where we want to benchmark both connections at the same time and add extra functionality to `BenchResources`.

However I don't feel strongly about this and perhaps it's over-engineering as you've pointed out so I'll revert it.

shotover-proxy/benches/cassandra_benches.rs
```diff
@@ -1,3 +1,66 @@
...
+pub struct BenchResources {
+    _compose: DockerCompose,
+    _shotover_manager: ShotoverManager,
+    connection: Session,
+}
+
+impl BenchResources {
+    #[allow(unused)]
+    pub fn new(shotover_manager: ShotoverManager, compose: DockerCompose) -> Self {
+        let connection = shotover_manager.cassandra_connection("127.0.0.1", 9043);
```
rukai, line 50: we are benching directly against the DB without shotover!
Shotover is currently on port 9042.

conorbros: Oops, fixed!

rukai (approved):
LGTM

Thanks for this PR and for figuring out the problem with our integration benches.

Having to do `handles[0].cassandra_connection` bothers me a little but not enough to block this PR.

benbromhead (approved):

PR #513, opened 2022-02-08
Add empty cassandra_benches file

conorbros:
<s>- Move the init code out of the actual bench-marking code</s>
<s>- Move each of the Redis benchmarks to functions so they can be run individually without having to run other benchmark's init code beforehand.</s>

- add cassandra_benches.rs file so CI on #512 works

### DIFF
--- a/shotover-proxy/benches/redis_benches.rs
+++ b/shotover-proxy/benches/redis_benches.rs
@@ -8,7 +8,7 @@ use helpers::ShotoverManager;
 fn redis_multi(c: &mut Criterion) {
     let mut group = c.benchmark_group("redis");
     group.throughput(criterion::Throughput::Elements(1));
-    group.noise_threshold(2.0);
+    group.noise_threshold(0.2);
 
     {
         let mut state = {
@@ -33,7 +33,7 @@ fn redis_multi(c: &mut Criterion) {
 fn redis_cluster(c: &mut Criterion) {
     let mut group = c.benchmark_group("redis");
     group.throughput(criterion::Throughput::Elements(1));
-    group.noise_threshold(2.0);
+    group.noise_threshold(0.2);
 
     {
         let mut state = {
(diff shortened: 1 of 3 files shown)


shotover-proxy/benches/redis_benches.rs
```diff
@@ -5,111 +5,140 @@ use test_helpers::docker_compose::DockerCompose;
 mod helpers;
 use helpers::ShotoverManager;
 
-fn redis(c: &mut Criterion) {
+fn redis_multi(c: &mut Criterion) {
     let mut group = c.benchmark_group("redis");
     group.throughput(criterion::Throughput::Elements(1));
     group.noise_threshold(2.0);
```
rukai, line 11: If this helps the noise issue then we should drop the noise_threshold down to a suitable value

conorbros: Yeah, I'm thinking around 5-10%?

rukai: Looking at https://github.com/shotover/shotover-proxy/runs/5104088154?check_suite_focus=true 20% looks about right.
We can bump it up another 5-10% if we get false positives.

conorbros: Alright, changed to 20%.

rukai:
I tested locally and this doesnt fix the initialization problem.
You can clearly see this by running something like `cargo bench null` and seeing all the shotover initialization log spam.

rukai:
You've made a good discovery that the noise issue is made drastically worse by my init hack, but in order to land this PR we should not regress on running individual benchmarks, instead we should add the ability to specify benchmark init logic to criterion.

conorbros:
I think that since the noisy benchmarks don't give us any meaningful data, we should land this and open a ticket for adding that functionality to criterion.

Commenting out benchmarks to run individually is not ideal, but I think it's better than not getting anything valuable from the benchmarks.

rukai:
Criterion is undermaintained and just opening a ticket wont get us anywhere.
I'm happy for this PR to revert the init hack if someone immediately starts working on implementing init functionality in criterion. I guess I'll be that person and start working on it now.

conorbros:
Thanks! I'll move back to the original setup and then wait until that your upcoming PR to criterion is merged.

rukai (approved):

benbromhead (approved):

PR #514, opened 2022-02-08
Fix redis benchmark initialization

rukai:
Uses a fork of criterion that allows us to provide an Fn that will generate our state only when the bench is actually run.
This gives us the best of both worlds: improve noise in the redis integration benches while allowing us to continue running benches individually.

In order to land this PR in a reasonable amount of time I suggest that we create a shotover/criterion repo to host this fork until it reaches upstream.

I have created a PR to review the temporary fork https://github.com/shotover/criterion.rs/pull/1

### DIFF
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -68,7 +68,7 @@ csv = "1.1.6"
 strum_macros = "0.23"
 
 [dev-dependencies]
-criterion = { git = "https://github.com/bheisler/criterion.rs", branch = "version-0.4", version = "0.3", features = ["async_tokio", "html_reports"] }
+criterion = { git = "https://github.com/shotover/criterion.rs", branch = "version-0.4", version = "0.3", features = ["async_tokio", "html_reports"] }
 redis = { version = "0.21.0", features = ["tokio-comp", "cluster"] }
 pcap = "0.9.0"
 pktparse = {version = "0.7.0", features = ["serde"]}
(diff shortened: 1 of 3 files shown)


conorbros:
Nice, I'll close #513 in favour of this.

rukai:
You'll still need to check in your empty Cassandra file

conorbros (approved):
Are you planning to open a PR on criterion or just use the fork for now?

rukai:
I will work on both once I catch up on reviews.
I suspect it will be a bit tricky to get my criterion changes into an upstreamable state.

benbromhead (approved):

shotover-proxy/Cargo.toml
```diff
@@ -68,7 +68,7 @@ csv = "1.1.6"
 strum_macros = "0.23"
 
 [dev-dependencies]
-criterion = { git = "https://github.com/bheisler/criterion.rs", branch = "version-0.4", version = "0.3", features = ["async_tokio", "html_reports"] }
+criterion = { git = "https://github.com/rukai/criterion.rs", branch = "bench_with_input_fn", version = "0.3", features = ["async_tokio", "html_reports"] }
```
conorbros, line 71: Switch this to the Shotover repo and then we can merge

rukai:
Realized I forgot to lower the noise level to 20%, so I did that now.

rukai:
Given those benchmarks that reported regressions were not touched in this PR and the regression did not occur in a rerun I will put it down to a very rare slow down on the CI VMs.

PR #521, opened 2022-02-14
Add some more cassandra integration benches

rukai:
These extra test cases should be useful for the message details rewrite.
I didnt bother to add the actual cache path for redis-cache as it doesnt work properly yet and the condition for caching is planned to change slightly.
protect-local encrypted case is taken from the integration test

### DIFF
--- a/shotover-proxy/benches/cassandra_benches.rs
+++ b/shotover-proxy/benches/cassandra_benches.rs
@@ -1,4 +1,4 @@
-use cassandra_cpp::{stmt, Session};
+use cassandra_cpp::{stmt, Session, Statement};
 use criterion::{criterion_group, criterion_main, Criterion};
 use test_helpers::docker_compose::DockerCompose;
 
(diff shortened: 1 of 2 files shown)


shotover-proxy/benches/cassandra_benches.rs
```diff
@@ -1,22 +1,75 @@
...
+                    state.connection.execute(&query.statement).wait().unwrap();
+                })
+            },
+        );
 
-    {
+        // Benches the case where the message does not meet the criteria for caching
         group.bench_with_input(
-            "passthrough",
+            format!("redis_cache_{}_uncached", query.name),
```
conorbros, line 54: So this is the same as the passthrough one for now, since Redis cache isn't working?

shotover-proxy/benches/cassandra_benches.rs
```diff
@@ -1,22 +1,75 @@
...
+                "INSERT INTO benchmark_keyspace.table_1 (id, x, name) VALUES (1, 11, 'foo');"
+            ),
+        },
+        Query {
+            name: "select",
+            statement: stmt!("SELECT id, x, name FROM benchmark_keyspace.table_1;"),
+        },
+    ];
+    for query in queries {
+        // Benches the case where the message does not meet the criteria for encryption
```
conorbros, line 32: If it's not being encrypted, wouldn't it just be the same as the passthrough one?

rukai: the point of these benches for redis cache and protect is to prove that when we dont hit the path with special logic it is as close as possible to passthrough.

shotover-proxy/benches/cassandra_benches.rs
```diff
@@ -1,22 +1,75 @@
...
+            statement: stmt!(
+                "INSERT INTO benchmark_keyspace.table_1 (id, x, name) VALUES (1, 11, 'foo');"
+            ),
+        },
+        Query {
+            name: "select",
+            statement: stmt!("SELECT id, x, name FROM benchmark_keyspace.table_1;"),
+        },
+    ];
+    for query in queries {
```
conorbros, line 31: It's a shame we need to startup the docker containers again when this loop iterates. Any way to avoid that? This will really bump up our test runtime.

rukai: if you compare the CI runs from this PR to another PR it bumps it up from 32mins to 34mins.
Not ideal but not world ending.

That said any ci runtime win we can get is welcome.
There are ways to improve the situation but they will be a significant investment and we should not let it stop us adding benches in the meantime.
In fact the more benches we add in the meantime will reveal how we should organize things to reduce runtime.

benbromhead (approved):

conorbros (approved):

conorbros (approved):

conorbros (approved):

Issue #476, opened 2022-01-25
Rewrite the Protect transform to use message.original 


rukai:
prereq https://github.com/shotover/shotover-proxy/issues/462

Rewrite the Protect transform such that it uses message.original instead of message.details. 
(take from https://github.com/shotover/shotover-proxy/pull/453)

PR #526, opened 2022-02-15
rewrite protect to use message.original

rukai:
closes https://github.com/shotover/shotover-proxy/issues/476

Currently the protect transform serializes the encrypted values to bincode and then stores the result in a MessageValue::Bytes.
The MessageValue::Bytes fails to round trip through cassandra because the current cassandra AST lacks a way to represent binary blobs.
Previously that didnt cause any problems because the query_values message details never got assembled back into the outgoing message.
But as of this PR this functionality actually works so this PR reverts the serialization to use serde instead of bincode as we can easily round trip through a string value.

Once we get a proper cql ast we can revert back to bincode and MessageValue::Bytes.

### DIFF
--- a/shotover-proxy/src/transforms/protect/mod.rs
+++ b/shotover-proxy/src/transforms/protect/mod.rs
@@ -1,20 +1,20 @@
-use std::collections::HashMap;
-
+use crate::error::ChainResponse;
+use crate::frame::{CassandraFrame, CassandraOperation, CassandraResult, Frame, CQL};
+use crate::message::MessageValue;
+use crate::transforms::protect::key_management::{KeyManager, KeyManagerConfig};
+use crate::transforms::{Transform, Transforms, Wrapper};
 use anyhow::anyhow;
 use anyhow::Result;
 use async_trait::async_trait;
 use bytes::Bytes;
+use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 use sodiumoxide::crypto::secretbox;
 use sodiumoxide::crypto::secretbox::{Key, Nonce};
+use sqlparser::ast::{Assignment, Expr, Ident, Query, SetExpr, Statement, Value as SQLValue};
+use std::borrow::BorrowMut;
+use std::collections::HashMap;
 use tracing::warn;
-
-use crate::error::ChainResponse;
-use crate::message::MessageValue;
-use crate::message::MessageValue::Rows;
-use crate::message::{MessageDetails, QueryMessage, QueryResponse, QueryType};
-use crate::transforms::protect::key_management::{KeyManager, KeyManagerConfig};
-use crate::transforms::{Transform, Transforms, Wrapper};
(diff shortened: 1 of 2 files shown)


benbromhead (approved):

shotover-proxy/src/transforms/protect/mod.rs
```diff
@@ -68,23 +68,23 @@ fn decrypt(ciphertext: Vec<u8>, nonce: Nonce, sym_key: &Key) -> Result<MessageVa
...
         secretbox::open(&ciphertext, &nonce, sym_key).map_err(|_| anyhow!("couldn't open box"))?;
     //TODO make error handing better here - failure here indicates a authenticity failure
-    // let decrypted_value: Value =
-    //     serde_json::from_slice(decrypted_bytes.as_slice()).map_err(|_| anyhow!("couldn't open box"))?;
-    let decrypted_value: MessageValue =
-        bincode::deserialize(&decrypted_bytes).map_err(|_| anyhow!("couldn't open box"))?;
+    let decrypted_value: MessageValue = serde_json::from_slice(decrypted_bytes.as_slice())
+        .map_err(|_| anyhow!("couldn't open box"))?;
+    // let decrypted_value: MessageValue =
+    //     bincode::deserialize(&decrypted_bytes).map_err(|_| anyhow!("couldn't open box"))?;
```
conorbros, line 74: Is the intention all this commented out code will be replaced when the Cassandra AST is ready?

rukai: yeah

shotover-proxy/src/transforms/protect/mod.rs
```diff
@@ -156,25 +156,77 @@ impl ProtectConfig {
...
     async fn transform<'a>(&'a mut self, mut message_wrapper: Wrapper<'a>) -> ChainResponse {
+        // encrypt the values included in any INSERT or UPDATE queries
         for message in message_wrapper.messages.iter_mut() {
-            if let MessageDetails::Query(qm) = &mut message.details {
-                // Encrypt the writes
-                if QueryType::Write == qm.query_type && qm.namespace.len() == 2 {
-                    if let Some((_, tables)) =
-                        self.keyspace_table_columns.get_key_value(&qm.namespace[0])
+            if let Some(namespace) = message.namespace() {
+                if namespace.len() == 2 {
```
conorbros, line 211: Won't a namespace always have the keyspace and column? Once we fix the `USE keyspace` handling of course.

conorbros (approved):

shotover-proxy/src/transforms/protect/mod.rs
rukai: I'm just maintaining the existing functionality, I think when we implement proper `USE keyspace` we can go through and make sure the namespace implementation and usage is sensible.

PR #527, opened 2022-02-16
Rewrite QueryCounter to use message.original

rukai:
Maintains the same format as the old QueryCounter but also fixes two bugs in the redis implementation:
* redis `query` value is now a nice value like "SET" instead of "BULKSTRING([71, 69, 84])"
* redis query counter actually works now - I didnt investigate too deeply but probably `ASTHolder::Commands(List(commands))` was not the correct way to get message details for redis.

### DIFF
--- a/shotover-proxy/src/transforms/query_counter.rs
+++ b/shotover-proxy/src/transforms/query_counter.rs
@@ -1,6 +1,7 @@
 use crate::error::ChainResponse;
-use crate::message::MessageValue::List;
-use crate::message::{ASTHolder, MessageDetails, QueryMessage};
+use crate::frame::cassandra::{CassandraFrame, CassandraOperation, CQL};
+use crate::frame::Frame;
+use crate::frame::RedisFrame;
 use crate::transforms::{Transform, Transforms, Wrapper};
 use anyhow::Result;
 use async_trait::async_trait;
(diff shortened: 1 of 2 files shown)


benbromhead (approved):

shotover-proxy/src/transforms/query_counter.rs
```diff
@@ -46,24 +54,45 @@ impl Transform for QueryCounter {
...
+                Frame::None => {
+                    counter!("query_count", 1, "name" => self.counter_name.clone(), "query" => "unknown", "type" => "none")
+                }
             }
         }
 
         message_wrapper.call_next_transform().await
     }
 }
+fn get_redis_query_type(frame: &RedisFrame) -> Option<String> {
```
conorbros, line 77: Should this be using something similar to `Message::get_query_type` and returning an enum of `QueryType` instead? That way developers using the metrics will have all query types in the same format and not have to handle different ones for Cassandra and Redis. e.g. 

Redis `"SET"` & Cassandra `"INSERT"` or `"UPDATE"` -> `QueryType::Write`
Redis `"GET"` & Cassandra `"SELECT"` -> `QueryType::Read`

That said, won't block this PR on it as it's just progress towards the `MessageDetails` rewrite and didn't change what was already implemented.

shotover-proxy/src/transforms/query_counter.rs
```diff
@@ -28,12 +29,19 @@ impl QueryCounter {
...
+                    operation:
+                        CassandraOperation::Query {
+                            query: CQL::Parsed(query),
+                            ..
+                        },
+                    ..
+                }) => {
+                    for statement in query {
+                        let query_type = match statement {
                             Statement::Query(_) => "SELECT",
```
conorbros, line 45: Why not use something like the `QueryType` enum? Do we want this metric to be specific or more generic (i.e. "write" or "read")?

rukai: I think being more specific is useful.
But theres nothing stopping us from from having a QueryType metric in another transform.
You can raise an issue if you want to make a design discussion about the kinds of metrics we report.

conorbros (approved):

PR #528, opened 2022-02-16
Remove Cassandra cluster tests

conorbros:
<s>
- Always run Shotover on 9043 and Cassandra on it's default port of 9042 so it's easier to reason about what you are connecting to
</s>
- Remove the cluster test as it does not properly test cluster functionality. Currently the driver makes the first connection to Shotover and then establishes the rest of the connections directly to Cassandra


rukai:
I'm not sure if deleting the cluster test is the right call as it at least has some protection against regressions.
At a minimum we could add a comment to the test describing the issue.

We could also go one step further.
Ultimately it would be good for our test to be able to assert that all our messages are going through shotover.
Is there a way we could assert the opposite of that, to demonstrate the brokenness and remind us to do the proper assertion when we fix the issue?
But that could end up being more complicated than its worth for now.

conorbros:
At the moment it tests nothing different than the passthrough tests, some messages may go through Shotover, some may not. I don't see what regressions that protects us against. And as an example, it's confusing because if you want to use Shotover as a sidecar to Cassandra nodes you need to be using the rewrite transform.

rukai:
Ahh I thought we were removing the test because the functionality was broken, but now it sounds like the functionality works but the test is just wrong.

What if we combined test_cassandra_peers_rewrite and test_cluster into a single integration test.

* 3 node cassandra cluster with 3 corresponding shotover instances 
* topology yamls configured with CassandraPeersRewrite
* current assertions we already have in test_cassandra_peers_rewrite
* include a chain that modifies the cassandra messages such that we can assert they are going through shotover.

conorbros:
Yeah, I agree that is the best path moving forward. I think we can move forward with this PR as is though, as it fixes the ports for Cassandra and removes a redundant test.

rukai:
What is the recommended way for users to setup shotover and db ports?
I can see arguments for both...

If shotover gets the default port then the application has no knowledge of shotover existing.
So this is suitable when all applications connect through shotover.

If the db gets the default port then existing applications can continue to connect directly to the db.
So this is suitable when only some applications connect through shotover.

So I guess the end user could pick either setup depending on their needs.

So in terms of demonstration of functionality either could work although I would probably lean towards shotover getting the default port as that is the idealized use case of shotover.
But it sounds like the reason you want to make this change is because its easier to reason about the integration tests if the db gets the default port.
I dont feel strongly either way in terms of intuitiveness for integration tests, but if it seems more intuitive to you then its probably more intuitive to others too.
So im happy to go with this change.

That said if we swap cassandra over then we need to swap redis over as well, otherwise that would definitely be unintuitive.
We should do it in the same PR because these kinds of changes are easy to bulk review and it makes the swap over atomic.

rukai:
I think you broke those benchmarks by not changing their ports :laughing:

conorbros:
Benchmarks need a "too good to be true" warning :laughing:

shotover-proxy/examples/redis-tls/topology.yaml
```diff
@@ -2,10 +2,10 @@
...
     Redis:
-      listen_addr: "127.0.0.1:6379"
+      listen_addr: "127.0.0.1:6380"
   redis_prod_tls:
     Redis:
-      listen_addr: "127.0.0.1:6380"
+      listen_addr: "127.0.0.1:6381"
       tls:
         certificate_authority_path: "examples/redis-tls/tls_keys/ca.crt"
         certificate_path: "examples/redis-tls/tls_keys/redis.crt"
```
rukai, line 11: The sink address isnt 6379.

rukai:
Looks like theres a whole bunch of clusters going 6379 -> 2000 -> 6379
Shouldnt that look more like: 6379 -> 6379 -> 2000 ?

Theres also a lot of `listen_addr: "127.0.0.1:6379"` used in scenarios where there is no real DB to hit.
I think that this PR should set them to 6380 for consistency with the other cases.
But maybe these scenarios is an argument that this PR is going in the wrong direction and we should just always give the shotover the default db port.

conorbros:
Yeah at this point I'm happy to leave it. I've realised what I thought was an inconsequential change is bigger than I thought and not worth the time it will take to get merged.

rukai:
In that case you can trim this PR down to just the test_cluster deletion.

rukai (approved):
lgtm

benbromhead (approved):
lgtm. Once we have C* cluster support this will (or something like it) will get added back in?

rukai:
Yeah, see https://github.com/shotover/shotover-proxy/pull/528#issuecomment-1041074936

PR #531, opened 2022-02-17
Delete KafkaSink

rukai:
shotovers Kafka support currently doesnt work (at least I assume so because there are no runnable examples or integration tests)
And we have no plans to support kafka any time soon.
So rather than doing a thorough investigation and implementation I opted to just stub out the methods it needs on the message struct.
Then we can properly investigate and implement this functionality when we add kafka support.

An alternative could be to just completely delete kafka_sink.rs but I think keeping this at least gives a hint at what plans were for kafka support.


shotover-proxy/src/message/mod.rs
```diff
@@ -102,6 +102,14 @@ impl Message {
...
     }
 
+    pub fn get_namespaced_primary_key(&self) -> Option<String> {
+        unimplemented!()
+    }
+
+    pub fn query_values(&self) -> Option<Vec<String>> {
+        unimplemented!()
+    }
+
```
conorbros, line 112: What will these be returning?

conorbros:
I'm leaning towards removing it until we properly support it.

shotover-proxy/src/message/mod.rs
rukai: I added some comments to explain what they did.

rukai:
Yeah, I dont mind either way, @benbromhead thoughts on deleting it?

rukai:
Alright, deleted, ready for review

PR #532, opened 2022-02-17
Replace message.message_details with message.frame()

rukai:
I think it would be easiest to just land the rest of the changes at this point but im more than happy to pull anything out that you think should go in separately.

## message/mod.rs
The most important changes are in message/mod.rs
There we implement the new MessageInner abstraction which manages the message internal state in an efficient way.
Initially all incoming messages are `MessageInner::RawBytes`.
A RawBytes is very cheap to create and to send as its just the raw bytes of the message.

Then when a transform needs to access message data it calls `message.frame()` which internally converts the message into a `MessageInner::Parsed` which contains the parsed message and the raw bytes of the message.
This is expensive to create but is still cheap to send.

Then if a transform needs to modify a message it modifies the `&mut Frame` returned by `message.frame()` and then calls `message.invalidate_cache()` which converts the message into a `MessageInner::Modified` deleting the internally stored raw bytes of the message.
Doing this is cheap but results in a message that is expensive to send as the message bytes must be rebuilt from the frame at send time.

## Transforms
Some minor adjustments to use the new Message API.
There should not be any logic changes here.

## Codecs

The codec changes are mostly just deleting the old message details -> message and message -> message details logic.
That has been superceded by calling Frame::from_bytes() to create a frame that just holds bytes at receive time.
Then at send time the codec calls message.into_encodable() which returns either a Bytes which can be directly sent or a Frame which the codec converts to bytes before sending.

## Config

Cassandra source and sink configs are drastically simplified.
We no longer need to toggle on/off message details generation/processing (query_processing)
We also no longer need to configure which fields should be included in the message details. (cassandra_ks)

### DIFF
--- a/shotover-proxy/src/codec/cassandra.rs
+++ b/shotover-proxy/src/codec/cassandra.rs
@@ -95,7 +95,7 @@ impl Decoder for CassandraCodec {
                     version: Version::V4,
                     stream_id: 0,
                     operation: CassandraOperation::Error(ErrorBody {
-                        error_code: 0xA,
+                        error_code: 0xA, // https://github.com/apache/cassandra/blob/adf2f4c83a2766ef8ebd20b35b49df50957bdf5e/doc/native_protocol_v4.spec#L1053
                         message: "Invalid or unsupported protocol version".into(),
                         additional_info: AdditionalErrorInfo::Server,
                     }),
@@ -120,7 +120,7 @@ impl Encoder<Messages> for CassandraCodec {
     ) -> std::result::Result<(), Self::Error> {
         for m in item {
             // TODO: always check if cassandra message
-            match m.into_encodable() {
+            match m.into_encodable(MessageType::Cassandra)? {
                 Encodable::Bytes(bytes) => dst.extend_from_slice(&bytes),
                 Encodable::Frame(frame) => self.encode_raw(frame.into_cassandra().unwrap(), dst),
             }
--- a/shotover-proxy/src/codec/cassandra.rs
+++ b/shotover-proxy/src/codec/cassandra.rs
@@ -1,46 +1,17 @@
-use std::borrow::{Borrow, BorrowMut};
-use std::collections::HashMap;
-use std::ops::DerefMut;
-
-use crate::frame::CassandraFrame;
-use anyhow::{anyhow, Result};
-
+use crate::frame::cassandra::CassandraOperation;
+use crate::frame::{CassandraFrame, Frame, MessageType};
+use crate::message::{Encodable, Message, Messages};
+use anyhow::anyhow;
 use bytes::{BufMut, BytesMut};
 use cassandra_protocol::compression::Compression;
-use cassandra_protocol::consistency::Consistency;
 use cassandra_protocol::frame::frame_error::{AdditionalErrorInfo, ErrorBody};
-
-use cassandra_protocol::frame::frame_result::{
-    ColSpec, ColType, ColTypeOption, RowsMetadata, RowsMetadataFlags, TableSpec,
-};
 use cassandra_protocol::frame::{Frame as RawCassandraFrame, ParseFrameError, Version};
-use cassandra_protocol::query::QueryValues;
-use cassandra_protocol::types::value::Value as CValue;
-
-use itertools::Itertools;
-use sqlparser::ast::Expr::{BinaryOp, Identifier};
-use sqlparser::ast::Statement::{Delete, Insert, Update};
-use sqlparser::ast::{
-    BinaryOperator, Expr, ObjectName, Select, SelectItem, SetExpr, Statement, TableFactor,
-    Value as SQLValue,
-};
-use sqlparser::dialect::GenericDialect;
-use sqlparser::parser::Parser;
 use tokio_util::codec::{Decoder, Encoder};
-use tracing::{debug, error, info, warn};
-
-use crate::frame::cassandra::{CassandraOperation, CassandraResult, CQL};
-use crate::frame::Frame;
-use crate::message::{
-    ASTHolder, Message, MessageDetails, MessageValue, Messages, QueryMessage, QueryResponse,
-    QueryType,
-};
+use tracing::{debug, info};
 
 #[derive(Debug, Clone)]
 pub struct CassandraCodec {
     compressor: Compression,
-    pk_col_map: HashMap<String, Vec<String>>,
-    bypass: bool,
     /// if force_close is Some then the connection will be closed the next time the
     /// system attempts to read data from it.  This is used in protocol errors where we
     /// need to return a message to the client so we can not immediately close the connection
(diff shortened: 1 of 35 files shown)


conorbros:
Are you planning on doing anything differently?

rukai:
Oh uhhh... I did mean to go over this PR again today, but ended up working on other things instead.
I'll do that now.
I dont currently know of anything that needs to be done differently but i'll check to make sure.

shotover-proxy/src/message/mod.rs
```diff
@@ -20,325 +17,233 @@ use itertools::Itertools;
...
+                    error_code: 0,
+                    message: "Message was filtered out by shotover".into(),
+                    additional_info: AdditionalErrorInfo::Server,
                 }),
-                Frame::None => Frame::None,
-            },
-            return_to_sender: false,
-            meta_timestamp: None,
-        }
+                tracing_id: None,
```
conorbros, line 124: I think the `tracing_id` should be carried over from the existing frame.

shotover-proxy/src/message/mod.rs
```diff
@@ -20,325 +17,233 @@ use itertools::Itertools;
...
-                Frame::Cassandra(CassandraFrame {
-                    version: frame.version,
-                    stream_id: frame.stream_id,
-                    tracing_id: None,
-                    warnings: frame.warnings.clone(),
-                    operation: body,
-                })
-            }
+                }),
+                tracing_id: None,
```
conorbros, line 153: Same re `tracing_id`

shotover-proxy/src/transforms/distributed/consistent_scatter.rs
```diff
@@ -1,5 +1,6 @@
 use crate::config::topology::TopicHolder;
 use crate::error::ChainResponse;
+
```
conorbros, line 3: Unnecessary newline

shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -18,11 +18,11 @@ use tokio::time::timeout;
 use tokio::time::Duration;
 use tracing::{debug, error, info, trace, warn};
```
conorbros, line 20: Remove this for formatting

shotover-proxy/src/transforms/redis/sink_single.rs
```diff
@@ -11,9 +10,10 @@ use tokio::net::TcpStream;
 use tokio_stream::StreamExt;
 use tokio_util::codec::Framed;
```
conorbros, line 12: Remove lines between imports for formatting

conorbros (changes requested):

conorbros:
Oops, just saw your comment after completing my review. I'll go over it again if you change anything.

shotover-proxy/src/codec/cassandra.rs
```diff
@@ -50,463 +21,22 @@ pub struct CassandraCodec {
     force_close: Option<String>,
 }
 
-pub(crate) struct ParsedCassandraQueryString {
-    namespace: Option<Vec<String>>,
-    pub(crate) colmap: Option<HashMap<String, MessageValue>>,
-    projection: Option<Vec<String>>,
-    primary_key: HashMap<String, MessageValue>,
-    pub(crate) ast: Option<Statement>,
+impl Default for CassandraCodec {
```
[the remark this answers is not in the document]
rukai: its a clippy lint :cry:

shotover-proxy/src/frame/mod.rs
```diff
@@ -15,31 +20,51 @@ pub enum Frame {
...
+    pub fn from_bytes(bytes: Bytes, message_type: MessageType) -> Result<Self> {
+        match message_type {
+            MessageType::Cassandra => CassandraFrame::from_bytes(bytes).map(Frame::Cassandra),
+            MessageType::Redis => redis_protocol::resp2::decode::decode(&bytes)
+                .map(|x| Frame::Redis(x.unwrap().0))
+                .map_err(|e| anyhow!("{e:?}")),
+        }
+    }
+
+    fn name(&self) -> &'static str {
```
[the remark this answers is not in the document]
rukai: that would log the entire frame.
We only want to log the name of the frame type.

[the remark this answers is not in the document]
conorbros: True! I only tested with the None variant.

shotover-proxy/src/message/mod.rs
```diff
@@ -20,325 +17,233 @@ use itertools::Itertools;
...
-                    }),
-                    tracing_id: None,
-                    warnings: vec![],
+        // TODO: clear message details cache fields if we ever add any
+
+        self.inner = self.inner.take().map(|x| x.invalidate_cache());
+    }
+
+    // TODO: this could be optimized to avoid parsing the cassandra sql
+    pub fn to_filtered_reply(&mut self) -> Message {
```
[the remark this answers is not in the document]
rukai: The existing filter transform filters by QueryType, there could be future transforms that filter on various other criteria.

I'll just leave it as is for now, we can easily move it later if we want.

shotover-proxy/src/message/mod.rs
```diff
@@ -20,325 +17,233 @@ use itertools::Itertools;
...
+        // TODO: set_filtered_reply
+        Message::from_frame(match self.frame().unwrap() {
+            Frame::Redis(_) => Frame::Redis(RedisFrame::Error(
+                "ERR Message was filtered out by shotover".into(),
```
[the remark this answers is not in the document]
rukai: image
Based on these search results im concluding that it doesnt.
The redis protocol specifies errors via a "-" prefix.
The "ERR" is a higher level concept that isnt part of the RESP protocol

shotover-proxy/src/message/mod.rs
```diff
@@ -20,325 +17,233 @@ use itertools::Itertools;
...
+            )),
+            Frame::Cassandra(frame) => Frame::Cassandra(CassandraFrame {
+                version: frame.version,
+                stream_id: frame.stream_id,
+                operation: CassandraOperation::Error(ErrorBody {
+                    error_code: 0,
```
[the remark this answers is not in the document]
rukai: Well that value was picked at random.
Looking through the options, maybe:
    0x2100    Unauthorized: The logged user doesn't have the right to perform
              the query.
or:
    0x2300    Config_error: The query is invalid because of some configuration issue
    
Could be better?
But I dont think it matters a whole lot.
For now I'll just set to 0x0

shotover-proxy/src/message/mod.rs
```diff
@@ -20,325 +20,265 @@ use itertools::Itertools;
...
-    /// First checks the `details` of the message and then falls back to the `original` message.
-    pub fn get_query_type(&self) -> QueryType {
-        if let MessageDetails::Query(query) = &self.details {
-            query.query_type.clone()
-        } else {
-            self.original.get_query_type()
+    pub fn get_query_type(&mut self) -> QueryType {
+        match self.frame() {
+            Some(Frame::Cassandra(cassandra)) => cassandra.get_query_type(),
+            Some(Frame::Redis(redis)) => redis_query_type(redis), // free-standing function as we cant define methods on RedisFrame
```
conorbros, line 170: Do you think it's worth writing an abstraction similar to `CassandraFrame` for `RedisFrame`? Then we could have a trait that has all the methods that we need for a frame to be used in message.

Though at this point I think we should just get this merged. Would be good to consider for later if we end up needing more frame types added.

rukai: Assuming you mean write a wrapper around redis-protocol types and call it RedisFrame, I cant see that being necessary.
Cassandra has a lot more going on and needed more complicated frame for shotover transforms to work with nicely.
Redis is a lot simpler and so the RedisFrame provided by redis-protocol should be fine.

shotover-proxy/src/transforms/protect/mod.rs
```diff
@@ -227,33 +228,38 @@ impl Transform for Protect {
...
-            }) = &mut response.original
+            })) = response.frame()
             {
                 if let Some(namespace) = request.namespace() {
-                    if let Frame::Cassandra(CassandraFrame {
+                    if let Some(Frame::Cassandra(CassandraFrame {
                         operation: CassandraOperation::Query { query, .. },
                         ..
-                    }) = &mut request.original
+                    })) = &mut request.frame()
```
conorbros, line 262: Do we need this `&mut`?

shotover-proxy/src/transforms/redis/cache.rs
```diff
@@ -454,10 +453,10 @@ impl Transform for SimpleRedisCache {
         let mut updates = false;
 
         for m in &mut message_wrapper.messages {
-            if let Frame::Cassandra(CassandraFrame {
+            if let Some(Frame::Cassandra(CassandraFrame {
                 operation: CassandraOperation::Query { .. },
                 ..
-            }) = &m.original
+            })) = &m.frame()
```
conorbros, line 459: Same with this reference

shotover-proxy/src/transforms/protect/mod.rs
rukai: Thanks, cleaned up a whole bunch of those

shotover-proxy/src/message/mod.rs
```diff
@@ -20,325 +20,265 @@ use itertools::Itertools;
...
-            let mut buffer = String::new();
-            let f = self.namespace.join(".");
-            buffer.push_str(f.as_str());
-            buffer.push('.');
-            buffer.push_str(serde_json::to_string(&pk).unwrap().as_str());
-            buffer
-        })
-    }
-}
+// TODO: When im finished, evaluate if MessageInner as an enum brings value over just a struct of Options
```
benbromhead, line 230: did we do this ? Doesn't seem worthwhile compared to the semantic value the current enum brings

shotover-proxy/src/message/mod.rs
```diff
@@ -20,325 +20,265 @@ use itertools::Itertools;
...
-        matching: Option<QueryMessage>,
-        result: Option<MessageValue>,
-        error: Option<MessageValue>,
-    ) -> Self {
-        QueryResponse {
-            matching_query: matching,
-            result,
-            error,
-            response_meta: None,
+impl MessageInner {
```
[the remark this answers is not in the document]
rukai: The transform author shouldnt know that MessageInner exists.
I should make sure the performance implications of frame() and invalidate_cache() are documented though.

shotover-proxy/src/message/mod.rs
rukai: Having a struct of Options would allow us to make `inner: Option<MessageInner>` into `inner: MessageInner`.

This could almost be an improvement however because of the MessageType needed in RawBytes representing MessageInner as a struct of Options would be quite problematic.

In conclusion both options are kinda bleh, thats the real world for you I guess.

I've deleted the comment, lets just leave MessageInner as it is.

rukai:
All the existing feedback should be addressed now.

shotover-proxy/tests/runner/observability_int_tests.rs
```diff
@@ -79,20 +79,20 @@ async fn test_metrics() {
         .arg(42)
         .query_async::<_, ()>(&mut connection)
         .await
-        .unwrap();
+        .unwrap_err();
```
conorbros, line 82: Why are these unwrap_err now?

rukai: Because Null transform now returns an error.

benbromhead (approved):
lets get this merged and build / improve on it

conorbros (approved):

rukai:
Thanks for the reviews!

PR #539, opened 2022-02-21
Cassandra TLS benchmarks

conorbros:
- Add TLS benchmarks.
- Change new methods of `BenchResources` to accept strings for topology and docker-compose files instead to reduce duplicated code. 
- Add `new_tls` method and move setup code to another method so it can be reused.


rukai (approved):
nice improvement!

benbromhead (approved):

PR #540, opened 2022-02-21
Switch Cassandra Docker image to bitnami

conorbros:
Switch all our usages of Cassandra docker images to bitnami because:
- The maintainers are more open to PRs and have merged my PR's adding new features that we needed.
- Removes the need for the `cassandra-docker-entrypoint.sh` hack because the environment variables have been added.


shotover-proxy/tests/helpers/mod.rs
```diff
@@ -145,12 +145,19 @@ impl ShotoverManager {
     }
 
     #[allow(unused)]
-    pub fn cassandra_connection(&self, contact_points: &str, port: u16) -> Session {
+    pub fn cassandra_connection(
+        &self,
+        contact_points: &str,
+        port: u16,
+        username: &str,
+        password: &str,
```
rukai, line 153: Can we just use hardcode the username/password instead of making it an arg?
Or do you plan to use a custom one somewhere?

conorbros: Nope, good idea.

rukai:
dont forget to run cargo test with `--all-features`

rukai:
oh huh it passes locally for me as well, even with `--all-features`

conorbros:
Very strange why only that one times out.

rukai:
Did you solve it or get lucky?

conorbros:
I bumped the timeout to 100 seconds. I'll try again with 90 seconds.

rukai:
We can remove this part from the license now:
```
The file shotover-proxy/example-configs/cassandra-passthrough/docker-entrypoint.sh originally came from <https://github.com/docker-library/cassandra> is copyright 2021 Docker, Inc and provided under the Apache v2 license; modified by Instacluster, Inc.
```

rukai:
Oh oops, I also just noticed we forgot to remove:
```
    volumes:
      - ../cassandra-docker-entrypoint.sh:/usr/local/bin/docker-entrypoint.sh
 ```

conorbros:
True! Removed them aswell.

rukai (approved):
LGTM.
Thanks for doing this, its a nice cleanup!

rukai:
looks like having two cassandra instances needs a larger timeout.
Shall we bump the global timeout to 120s?

Unfortunate that our int tests take so long to spin up so if you have ideas on how to speed it up that would be great, otherwise we just have to bump it again :/

conorbros:
Seems 100 seconds should do it. A 10 second bump isn't too bad

XA21X (approved):

PR #542, opened 2022-02-22
Add cqlparser crate implemented by nom

rukai:
So far only implements select and even then still doesnt have full coverage of select.
However it does implement enough demonstrate the feasibility of writing a parser like this.

## Some background on parsers
This PR uses [nom](https://github.com/Geal/nom) which is a parser combinator library for rust.
That means it provides a whole bunch of composable functions to easily construct your own parser.
nom is designed around binary protocols but also happens to work well enough for simple programming languages. (I'm considering CQL a simple programming language)
Parser combinators are known to fall short when building a full general purpose programming language because you want more precise control over parsing mostly for error generating reasonable error messages.
However for our use case I think parser combinators should be more than enough.

## Error reporting
I am proposing that when given a bad query we just fail parsing with a generic error message.
I do like me some useful errors but in this case for simpler implementation and better performance I think it makes sense for shotover to just say "query {} failed to parse" and leave it at that.
We can of course reevaluate this in the future.
Especially once we have something in place, it will be easier to evaluate the affect on performance of improving the diagnostics.
If we want to go all out in terms of error reporting we can use something like https://github.com/zesterer/chumsky instead of nom

## Bytes
I attempted to rewrite the parser in its current state to store Strings as Bytes however this unexpectedly reduced performance by 4x. There might be a way to get parsing with Bytes working performantly but for now I would focus on just using allocating types and then we can investigate a rewrite later if it turns up in benchmarks.

## Just a draft
This PR is not to say that this is the way we should do things, its just a draft and one that I only spent a few days on. So im happy for this to go in a completely different direction if we find a better direction.

If we decide this is a good direction and we want to start collaborating on this code then we can land as is and start iterating on it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -447,6 +447,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "cqlparser"
+version = "0.1.0"
+dependencies = [
+ "criterion",
+ "nom 7.1.0",
+]
+
+[[package]]
 name = "crc16"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 members = [
     "shotover-proxy",
     "test-helpers",
+    "cqlparser"
 ]
 
 # https://deterministic.space/high-performance-rust.html
--- a/cqlparser/Cargo.toml
+++ b/cqlparser/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "cqlparser"
+version = "0.1.0"
+edition = "2021"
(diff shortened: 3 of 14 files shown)


conorbros:
Nice! But I'd put it in a separate crate. Someone else will definitely find it useful and it's pretty decoupled from Shotover.

rukai:
>Nice! But I'd put it in a separate crate/repo. Someone else will definitely find it useful and it's pretty decoupled from Shotover.

That is definitely an endgoal!
I think we can iterate on using it in shotover faster if we keep it in the same repo for now and then move to a separate repo when its "finished"

rukai:
This is now being tracked by https://github.com/shotover/shotover-proxy/pull/542

Issue #517, opened 2022-02-11
Connection state


conorbros:
Internally Cassandra stores state related to a particular connection, such as the keyspace sent by `USE my_keyspace` query so that it knows that subsequent queries from the connection are to use `my_keyspace`. Shotover should do something similar so we know what table is being queried. The `CassandraPeersRewrite` transform will be possible to break without this functionality. You could send `USE system`, and then `SELECT native_port from peers_v2;` and Shotover would not know to rewrite the port. I'd assume the caching transform could be broken in a similar way.

Ideally this would be invisible to transform developers, if they call `Message::namespace`, Shotover should determine the namespace using the connection state and frame.

PR #545, opened 2022-02-24
Connection state

conorbros:
Initial implementation of a fix for #517. This makes use of a jarring hack because the `sqlparser` crate does not support `USE` statements yet. I don't expect this to be merged as is, I think we should wait until the CQL parser is ready. I am just opening this PR for feedback on the general approach.

### DIFF
--- a/shotover-proxy/src/message/connection_state.rs
+++ b/shotover-proxy/src/message/connection_state.rs
@@ -1,0 +1,11 @@
+#[derive(Default, PartialEq, Clone, Debug)]
+pub struct CassandraConnectionState {
+    pub used_keyspace: Option<String>,
+}
+
+#[derive(PartialEq, Clone, Debug)]
+pub enum ConnectionState {
+    Cassandra(CassandraConnectionState),
+    Redis,
+    Unknown,
+}
--- a/shotover-proxy/src/message/mod.rs
+++ b/shotover-proxy/src/message/mod.rs
@@ -4,6 +4,7 @@ use crate::frame::{
     cassandra::{to_cassandra_type, CassandraMetadata, CassandraOperation},
 };
 use crate::frame::{CassandraFrame, Frame, MessageType, RedisFrame};
+use crate::message::connection_state::{CassandraConnectionState, ConnectionState};
 use anyhow::{anyhow, Result};
 use bigdecimal::BigDecimal;
 use bytes::{Buf, Bytes};
@@ -29,6 +30,8 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::net::IpAddr;
 use std::num::NonZeroU32;
 use uuid::Uuid;
+
+pub mod connection_state;
 
 enum Metadata {
     Cassandra(CassandraMetadata),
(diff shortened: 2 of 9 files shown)


shotover-proxy/src/server.rs
```diff
@@ -459,6 +469,45 @@ impl<C: Codec + 'static> Handler<C> {
 
             debug!("Received raw message {:?}", messages);
 
+            {
+                let mut already_modified = false;
```
conorbros, line 473: Will clean this up and add it as a method on `ConnectionState` instead of directly in `server.rs`. I will also add a flag to transforms as to whether they need `ConnectionState` and then we can opt to not run this check if we don't need it.

shotover-proxy/src/message/mod.rs
```diff
@@ -40,6 +43,9 @@ pub struct Message {
     // TODO: Not a fan of this field and we could get rid of it by making TimestampTagger an implicit part of ConsistentScatter
     // This metadata field is only used for communication between transforms and should not be touched by sinks or sources
     pub meta_timestamp: Option<i64>,
+
+    /// State of connection that sent or received this message
+    pub(crate) connection_state: ConnectionState,
```
conorbros, line 48: Since the `Message` itself won't be modifying this field, we could probably have this as a reference to the server's connection state.

shotover-proxy/src/message/connection_state.rs
```diff
@@ -0,0 +1,11 @@
+#[derive(Default, PartialEq, Clone, Debug)]
+pub struct CassandraConnectionState {
+    pub used_keyspace: Option<String>,
+}
+
+#[derive(PartialEq, Clone, Debug)]
+pub enum ConnectionState {
```
conorbros, line 7: These are here to show an example of how we would add further variants of `ConnectionState`

shotover-proxy/src/server.rs
```diff
@@ -459,6 +469,45 @@ impl<C: Codec + 'static> Handler<C> {
...
+
+                for i in 0..messages.len() {
+                    if let Frame::Cassandra(cf) = &messages[i].original {
```
conorbros, line 476: I'm going to modify this to only parse the Opcode, and only do a full parse if the query is a `USE` statement

rukai: You wont be able to tell if its a USE just from the opcode...?

conorbros: True, we will have to check the opcode is a Query, and then parse the first bit of the CQL to determine that.

rukai: Honestly I would just parse it the normal way at that point.

rukai: Maybe we need some kind of toggle for this feature or detect if the transforms used need this set.

rukai:
Can you put a `USE` in a prepared query? :thinking:

shotover-proxy/src/server.rs
conorbros: Yeah, we did discuss that. A flag for if a transform is using the `Message::namespace` method is probably the way to go about it.

conorbros:
Closing as stale

Issue #543, opened 2022-02-23
Cassandra Request throttling


benbromhead:
## Is your feature request related to a problem? Please describe.
A number of Instaclustr customers are interested in being able to enforce a global request throttling for Cassandra clusters. Database traffic should be able to be throttled on per request (per query) basis.  

## Describe the solution you'd like
Shotover should be able to enforce only a certain number of requests/queries per second globally across all connections. If the number of requests entering shotover exceeds the configured value, an error is returned to the client (similar to Cassandras own backpressure mechanism).

benbromhead:
**Critical requirements**
- Support request throttling across all connections for a given transform chain
- Support proper handling of the cluster discovery process client drivers do (e.g. rewrite peer table queries)
- Support throwing Overloaded errors back to the client driver https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v3.spec#L968

**Nice to have requirements, but not important**
- C* cluster support (shotover can route to different upstream nodes)
- Different traffic quotas for different connection attributes (e.g. different C* users have different quotas)
- Shotover maintains a global view of total request traffic across a fleet of shotover proxies (e.g. a cluster wide request quota)

The preferred testing model will be shotover running in standalone proxy mode, the end production deployment model will be colocated with each Cassandra node

benbromhead:
For the moment...

**Out of scope**
- Including queries inside a single batch query

benbromhead:
Some useful background reading on the existing back pressure mechanisms in C* -> https://cassandra.apache.org/_/blog/Improving-Apache-Cassandras-Front-Door-and-Backpressure.html 

We will likely use the built in protocol back pressure messages to get clients to back off.

conorbros:
TODO:
- [ ] Support request throttling across all connections for a given transform chain
- [ ] https://github.com/shotover/shotover-proxy/issues/546
- [ ] Including queries inside a single batch query

PR #547, opened 2022-02-25
Cassandra request throttling

conorbros:


shotover-proxy/tests/test-topologies/cassandra-request-throttling.yaml
```diff
@@ -0,0 +1,24 @@
+---
+sources:
+  cassandra_prod:
+    Cassandra:
+      query_processing: true
+      listen_addr: "127.0.0.1:9042"
+      max_requests_per_second: 20
+      cassandra_ks:
```
rukai, line 8: not needed anymore

shotover-proxy/tests/test-topologies/cassandra-request-throttling.yaml
```diff
@@ -0,0 +1,24 @@
+---
+sources:
+  cassandra_prod:
+    Cassandra:
+      query_processing: true
```
rukai, line 5: not needed anymore

shotover-proxy/tests/test-topologies/cassandra-request-throttling.yaml
```diff
@@ -0,0 +1,24 @@
...
+          - key
+        test.simple:
+          - pk
+        test.clustering:
+          - pk
+          - clustering
+chain_config:
+  main_chain:
+    - CassandraSinkSingle:
+        result_processing: true
```
rukai, line 19: not needed anymore

shotover-proxy/src/server.rs
```diff
@@ -401,7 +414,11 @@ impl<C: Codec + 'static> Handler<C> {
     ///
     /// When the shutdown signal is received, the connection is processed until
     /// it reaches a safe state, at which point it is terminated.
-    pub async fn run(&mut self, stream: TcpStream) -> Result<()> {
+    pub async fn run(
+        &mut self,
+        stream: TcpStream,
+        limiter: Arc<Option<RateLimiter<NotKeyed, InMemoryState, DefaultClock, NoOpMiddleware>>>,
```
rukai, line 420: We should use an Option<Arc<T>> instead of Option<Arc<T>>.
That way it doesnt need to allocate when None.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1458,3 +1460,60 @@ fn test_cassandra_peers_rewrite() {
...
+            } else {
+                panic!(
+                    "wrong error returned, got {:?}, expected SERVER_OVERLOADED",
+                    e
+                )
+            }
+            false
+        }
+    });
+    assert!(results.len() < 22 && results.len() > 18);
```
rukai, line 1518: how did you arrive at the values of 22 and 18?

shotover-proxy/src/server.rs
```diff
@@ -461,6 +478,21 @@ impl<C: Codec + 'static> Handler<C> {
 
             debug!("client details: {:?}", &self.client_details);
 
+            if let Some(limiter) = limiter.as_ref() {
+                messages = messages
+                    .into_iter()
```
rukai, line 483: No need to recreate the vec we should be able to just `for message in &mut messages`

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
conorbros: Running it a few times and seeing the variance.

shotover-proxy/src/message/mod.rs
```diff
@@ -232,6 +232,32 @@ impl Message {
         self.invalidate_cache();
     }
 
+    pub fn set_backpressure(&mut self) {
+        *self = Message::from_frame(match &self.frame().unwrap() {
```
rukai, line 236: Dont unwrap here either fail silently or log an error.

Also the current implementation of frame() is too expensive to call here.
There isnt really an alternative though.
We will have to either:
* rewrite the cassandra frame to not parse the CQL
* add a seperate method that just returns frame metadata that we can call instead of frame()
  + this should be even faster than the first suggestion so probably the best solution.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
rukai: If it reached 22 and 18 maybe we should bump it to 25 and 15?

Also where are these inaccuracies in the rate limiter coming from?

conorbros: From Cassandra not taking the exact same amount of time to respond to each request.

shotover-proxy/src/message/mod.rs
conorbros: Okay, I added the `fn metadata` method to `Message` in my latest commit. I've tried to make it extendable and so we can add other metadata types in the future, let me know if we would rather it just handle the Cassandra ones for now.

shotover-proxy/src/message/mod.rs
```diff
@@ -232,6 +241,56 @@ impl Message {
         self.invalidate_cache();
     }
 
+    /// Get metadata for this `Message`
+    fn metadata(&mut self) -> Result<Metadata> {
+        let inner = self.inner.take().unwrap();
```
rukai, line 246: take here is quite scary.
This happens to work because we always destroy the message after calling metadata but we should be able to implement metadata without taking (and leaving as None) the self.inner

shotover-proxy/src/message/mod.rs
```diff
@@ -25,6 +28,12 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::net::IpAddr;
 use uuid::Uuid;
 
+enum Metadata {
+    Cassandra(CassandraMetadata),
+    Redis,
+    Unknown,
```
rukai, line 34: Can we call this None to be consistent?

shotover-proxy/src/message/mod.rs
```diff
@@ -232,6 +241,56 @@ impl Message {
...
+                    error_code: 0x1001,
+                    message: "".into(),
+                    additional_info: AdditionalErrorInfo::Overloaded,
+                });
+
+                Frame::Cassandra(CassandraFrame {
+                    version: metadata.version,
+                    stream_id: metadata.stream_id,
+                    tracing_id: metadata.tracing_id,
+                    warnings: metadata.warnings,
```
rukai, line 280: I'm not sure it makes sense to return the warnings back to the client?
If we can remove it here then we can also remove it from metadata.

rukai: image
That sounds like:
1. the client wont send warnings anyway
2. even if it did it wouldnt make sense to send those warnings back.

shotover-proxy/src/message/mod.rs
conorbros: How can we parse the bytes without cloning them for the `MessageInner::RawBytes` variant?

shotover-proxy/src/message/mod.rs
conorbros: Yep, good spot. I'll fix that up.

shotover-proxy/src/message/mod.rs
rukai: Just pass in an &[u8]

rukai: this works for me:
image

shotover-proxy/src/message/mod.rs
conorbros: Maybe we should leave it in the metadata struct since it technically is part of a Cassandra message's metadata, but remove it when we are setting a message to backpressure?

rukai: yeah thats fine for now.

conorbros: Had to remove it because it's unread, but I added a comment to the struct.

benbromhead:
Would we not construct the throttle mechanism within a transform?

conorbros:
I think its best at the source because it saves Shotover from doing any unnecessary work parsing messages that will later be back-pressured and means we don't need to use some global state to share the limiter between chains.

I also think it's best to consider request throttling as a responsibility of the source anyhow. Putting it in the chain could cause all sorts of mess like metrics/logging transforms being wrong if they end up logging something that was later back-pressured.

benbromhead:
> I think its best at the source because it saves Shotover from doing any unnecessary work parsing messages that will later be back-pressured and means we don't need to use some global state to share the limiter between chains.

I don't think the overhead of parsing a message will be dramatic given that messages that get allowed through need to get parsed anyway. So I wouldn't worry too much from a perf perspective. As for global state, reading through governer docs if we wanted the rate limit to be global across connections from a single source we would just wrap it in an `Arc<>`

> I also think it's best to consider request throttling as a responsibility of the source anyhow. Putting it in the chain could cause all sorts of mess like metrics/logging transforms being wrong if they end up logging something that was later back-pressured.

It could, however putting throttling first in the chain would avoid that, and folks may want to do something with all requests before throttling (e.g. audit all attempted queries).

> 
> What were your thoughts around why it should be a transform?

This fits as a transform as it changes the behavior and responds to messages on a request / response basis which is ideal for a transform. Ideally the source shouldn't need to touch messages other than to deal with codec or for protocol specific edge cases. 

As a transform it would also then be applicable to redis!

rukai:
While putting a throttle transform later in the chain may indicate a bug there are definitely legitimate use cases for having another transform before it.

Its true that some users may be confused that they get worse performance if they put the throttle transform after a transform that parses the message.
But as long as they put the throttle transform first in the chain the performance should be just as good as if the functionality was implemented into the source.
We might need to explicitly document which transforms trigger parsing of the messages under what conditions, but lets measure the impact message parsing has before complicating our docs with that.

So currently i'm definitely for having it in a transform.
Unless we discover some large nonobvious performance issue with it.

shotover-proxy/Cargo.toml
```diff
@@ -33,6 +33,7 @@ pin-project-lite = "0.2"
 tokio-openssl = "0.6.2"
 openssl = { version = "0.10.36", features = ["vendored"] }
 async-recursion = "1.0"
+governor = "0.4.2"
```
rukai, line 36: Can any of these default features be disabled? https://github.com/antifuchs/governor/blob/6eb1060686bb5e268ec27ea3a8c267338769a9ff/governor/Cargo.toml#L38

rukai: It compiles for me with: `governor = { version = "0.4.2", default-features = false, features = ["std", "jitter"] }`
We will need to check if dashmap or quanta were needed for something important.

shotover-proxy/src/message/mod.rs
```diff
@@ -232,6 +241,55 @@ impl Message {
         self.invalidate_cache();
     }
 
+    /// Get metadata for this `Message`
+    fn metadata(&mut self) -> Result<Metadata> {
```
rukai, line 245: I think we can replace `&mut` -> `&`

shotover-proxy/src/transforms/throttling.rs
```diff
@@ -0,0 +1,73 @@
...
+
+#[derive(Clone)]
+pub struct RequestThrottling {
+    limiter: Arc<RateLimiter<NotKeyed, InMemoryState, DefaultClock, NoOpMiddleware>>,
+}
+
+#[async_trait]
+impl Transform for RequestThrottling {
+    async fn transform<'a>(&'a mut self, mut message_wrapper: Wrapper<'a>) -> ChainResponse {
+        let mut backpressured_messages: Vec<crate::message::Message> = vec![];
```
rukai, line 41: use crate::message::Message

shotover-proxy/src/message/mod.rs
```diff
@@ -232,6 +241,55 @@ impl Message {
...
+                    operation: body,
+                })
+            }
+            Metadata::Redis => {
+                unimplemented!()
+            }
+            Metadata::None => Frame::None,
+        });
+
+        self.return_to_sender = true;
```
rukai, line 289: we dont need this anymore

shotover-proxy/src/server.rs
```diff
@@ -353,23 +353,18 @@ fn spawn_read_write_tasks<
     tx: W,
```
rukai, line 353: Are any of the changes to this file still needed?

shotover-proxy/src/transforms/throttling.rs
```diff
@@ -0,0 +1,73 @@
...
+use std::sync::Arc;
+
+#[derive(Deserialize, Debug, Clone)]
+pub struct RequestThrottlingConfig {
+    pub max_requests_per_second: NonZeroU32,
+}
+
+impl RequestThrottlingConfig {
+    pub async fn get_source(&self) -> Result<Transforms> {
+        tracing::info!("here");
```
rukai, line 24: stray trace

shotover-proxy/src/transforms/throttling.rs
```diff
@@ -0,0 +1,73 @@
...
+
+#[async_trait]
+impl Transform for RequestThrottling {
+    async fn transform<'a>(&'a mut self, mut message_wrapper: Wrapper<'a>) -> ChainResponse {
+        let mut backpressured_messages: Vec<crate::message::Message> = vec![];
+
+        // find indexes of messages that need to be backpressured
+        let remove_indexes = message_wrapper
+            .messages
+            .iter_mut()
```
rukai, line 46: iter_mut -> iter

shotover-proxy/src/transforms/throttling.rs
```diff
@@ -0,0 +1,73 @@
...
+            .collect::<Vec<_>>();
+
+        // remove messsages and set to backpressure
+        for i in &remove_indexes {
+            let mut message = message_wrapper.messages.remove(*i);
+            message.set_backpressure()?;
+            backpressured_messages.push(message);
+        }
+
+        let mut responses = message_wrapper.call_next_transform().await?;
```
rukai, line 64: we should skip this if every message got rate limited.

shotover-proxy/src/transforms/throttling.rs
```diff
@@ -0,0 +1,73 @@
...
+
+        // reinsert into responses
+        for i in remove_indexes.into_iter() {
+            responses.insert(i, backpressured_messages.pop().unwrap());
```
rukai, line 68: pop takes from the end but the indices are going forwards!

shotover-proxy/src/transforms/throttling.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+#[derive(Clone)]
+pub struct RequestThrottling {
+    limiter: Arc<RateLimiter<NotKeyed, InMemoryState, DefaultClock, NoOpMiddleware>>,
+}
+
+#[async_trait]
+impl Transform for RequestThrottling {
```
rukai, line 39: I came up with a way of writing this method that should help reduce allocations in 2 ways:
1. We only allocate one vec per call now.
2. The iterator logic should preallocate the vec at the correct length without resizes. (The itererator chaining is a little complicated so its possible rust might not be able to perform this optimization here, would have to check the assembly if we needed to know for sure)


```rust
#[async_trait]
impl Transform for RequestThrottling {
    async fn transform<'a>(&'a mut self, mut message_wrapper: Wrapper<'a>) -> ChainResponse {
        // extract throttled message from the message_wrapper
        let throttled_messages: Vec<(Message, usize)> = (0..message_wrapper.messages.len())
            .into_iter()
            .filter(|_| self.limiter.check().is_err())
            .map(|i| {
                let message = message_wrapper.messages.remove(i);
                (message, i)
            })
            .collect();

        let mut responses = message_wrapper.call_next_transform().await?;

        // reinsert into responses
        for (mut message, i) in throttled_messages {
            message.set_backpressure()?;
            responses.insert(i, message);
        }

        Ok(responses)
    }
}
```

conorbros: Thanks!

rukai:
The clippy lint is a FP, we'll need to add `#[allow(clippy::needless_collect)]`

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1442,3 +1444,59 @@ fn test_cassandra_peers_rewrite() {
...
+#[serial]
+async fn test_cassandra_request_throttling() {
+    let _docker_compose =
+        DockerCompose::new("example-configs/cassandra-passthrough/docker-compose.yml");
+
+    let shotover_manager =
+        ShotoverManager::from_topology_file("tests/test-configs/cassandra-request-throttling.yaml");
+
+    let connection = shotover_manager.cassandra_connection("127.0.0.1", 9042);
+    std::thread::sleep(std::time::Duration::from_secs(1));
```
rukai, line 1458: what is this sleep for?

conorbros: When the client connects it sends a bunch of requests to gather information about the cluster, we need to make sure we don't trigger the rate limiter with the second connection.

conorbros:
Bit of information on the governor library used:

- Uses [generic cell rate algorithm](https://en.wikipedia.org/wiki/Generic_cell_rate_algorithm).
- Doesn't use any locks internally. All state is stored in a single `AtomicU64` integer.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1442,3 +1444,59 @@ fn test_cassandra_peers_rewrite() {
...
+
+    futures = vec![];
+    // only around half of these should be let through the request throttling
+    for _ in 0..20 {
+        futures.push(connection.execute(&statement));
+        futures.push(connection_2.execute(&statement));
+    }
+    let mut results = join_all(futures).await;
+    results.retain(|result| match result {
```
rukai, line 1484: We can simplify this like:
```rust
    results.retain(|result| match result {
        Ok(_) => true,
        Err(Error(
            ErrorKind::CassErrorResult(cassandra_cpp::CassErrorCode::SERVER_OVERLOADED, ..),
            _,
        )) => false,
        Err(e) => panic!(
            "wrong error returned, got {:?}, expected SERVER_OVERLOADED",
            e
        ),
    });
```

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
rukai: Oh nice. That should definitely be added as a comment.

conorbros: It's a comment underneath, but I'll add it the first instance of sleep too.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
rukai: I can understand needing 20 < results.len() < 22

But I dont understand why we need 18 < results.len() < 22
If the limit is 20 then I dont see any way to time the messages such that only 18 get through.
So either:
* the test should be changed to 20 < results.len() < 22
* there is a bug in the governor crate or we dont understand how the governor crate works

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1442,3 +1444,51 @@ fn test_cassandra_peers_rewrite() {
...
+        Err(Error(
+            ErrorKind::CassErrorResult(cassandra_cpp::CassErrorCode::SERVER_OVERLOADED, ..),
+            _,
+        )) => false,
+        Err(e) => panic!(
+            "wrong error returned, got {:?}, expected SERVER_OVERLOADED",
+            e
+        ),
+    });
+    assert!(results.len() < 22 && results.len() > 20);
```
rukai, line 1493: We should include the actual value in the assert message

rukai: Could also be informative to include the whole `"{results:?}" ` to better understand whats going on.

rukai (approved):
I'm happy with this!
Lets land and then benchmark + do further testing.

conorbros:
Sounds good, I've got a list of TODOs on here: https://github.com/shotover/shotover-proxy/issues/543

benbromhead (approved):
heck yeah!

Issue #544, opened 2022-02-24
Benchmarks revamp


rukai:
The current benchmark system is alright.
It can currently be used by developers to get a better understanding of how the system performs in general and the effects of their changes if they specifically think to run benches.
We have some level of automated regression catching, with regressions of over 30% on CI triggering a warning.

However there is a lot of room for improvements still.
Here are some use cases I would like to fill that the current system cannot meet:
* Benches that send many messages concurrently
   + currently all our benchmarks send requests sequentially which does not reflect production use at all.
   + We have been working around this by using redis-bench but I dont think we have anything like that for cassandra.
   + This will be super important for our cassandra throttling work.
* Displaying per transform benchmarks to users so they understand the implications of the configured transforms.
    + Will need a clever way of describing the affect the MessageInner state has before and after the transform.
* Track benchmark results over time for developers so they can see trends in performance.
    + The rust-lang benchmarks do this incredibly well - https://perf.rust-lang.org/
* Better automated regression catching, can we do better than 30% variance?

rukai:
Windsock was implemented to solve #2 and has been quite successful.
#1 we ended up moving to codspeed which isnt what we described but improved the situation considerably.

PR #549, opened 2022-02-28
Add cassandra_bench example

rukai:
Abstracts over [latte](https://github.com/pkolaczk/latte) to allow simply running a bench against shotover at a specific rate/OPS.
It outputs a comparison between directly hitting cassandra and going through shotover.
It is a prototype/experiment towards the proposed 2nd system given in https://github.com/shotover/shotover-proxy/issues/544

To use run as: `cargo run --release --example cassandra_bench -- --config-dir example-configs/cassandra-passthrough -r 10000`
On my machine that gave me:
image

On my machine, response time was drastically better with a lowered rate of
To use run as: `cargo run --release --example cassandra_bench -- --config-dir example-configs/cassandra-passthrough -r 10000`
image

You will need to experiment to see what sort of rates give you insightful results on your own machine.

If we think this is useful we can try and review + land with the idea being to replace with it a better solution later on.
Alternatively we can just set the PR to draft and use as a reference for a better solution.


shotover-proxy/examples/cassandra_bench_at_rate.rs
```diff
@@ -0,0 +1,93 @@
...
+struct Latte {
+    rate: String,
+}
+
+impl Latte {
+    fn new(rate: String) -> Latte {
+        test_helpers::docker_compose::run_command(
+            "cargo",
+            &[
+                "install",
```
conorbros, line 50: Can we just add this to our dependencies list and install manually instead of doing this instead the benchmarks?

rukai: Well we have to install a specific version of latte as it has no stability gaurantees.
And I guess we could document to install this specific version of latte....

But I think in general why list a dependency when we can have things "just work"?

If you are worried about the time it takes to rerun this every time, once its installed it only takes 0.4s to rerun on my machine.
image

shotover-proxy/examples/cassandra_bench.rs
```diff
@@ -0,0 +1,112 @@
...
+                "--rev",
+                "5f54a4128b76118446de7ccb29223e53ac4e9208",
+            ],
+        )
+        .unwrap();
+        Latte { rate }
+    }
+
+    fn bench(&self, name: &str, address: &str) {
+        std::process::Command::new("latte")
```
benbromhead, line 81: was it easier to shell out to the bin rather than embedding the app as a dev / bench lib?

rukai: Its only available as a bin, otherwise I would have used it as a lib

rukai:
Discussed with connor and will mark as draft and investigate a proper benchmarking solution later on.
For now feel free to checkout the branch and use it as is if you think it would be useful.

rukai:
This was somehow broken by https://github.com/shotover/shotover-proxy/commit/441316bdecda8e2530c0333c01012a6dcdd153f9

rukai:
I made some changes to allow running the cassandra-passthrough-parse-request/response configs from the bencher.

I had to change the process runner to run with --all-features to be able to use the DebugForceParse transform

@conorbros you should rereview

.github/workflows/test.yml
```diff
@@ -55,6 +55,9 @@ jobs:
       run: cargo test ${{ matrix.cargo_profile }} --all-features -- --include-ignored --show-output
       # can not run tests if cpp driver not installed
       if: ${{ matrix.runner == 'ubuntu-18.04' }}
+    - name: Ensure that custom benches run
+      run: cargo run --release --example cassandra_bench -- --config-dir example-configs/cassandra-passthrough --rate 1000
```
conorbros, line 59: Does this just check that it still runs?

conorbros:
I think the work here is great and we will find it very useful but I'm not convinced that it should live in the Shotover repo. I anticipate us using the examples directory to demonstrate writing 3rd party transforms in the future and this doesn't align with that.

.github/workflows/test.yml
rukai: yeah

rukai:
3rd party transforms are a long way off and a lot of things will need to be shuffled around for that. A lot.
Unless a separate repo brings us some value now then I think its better to wait till we know what we want for 3rd party transforms.

conorbros:
Okay sounds good to me.

conorbros (approved):

benbromhead (approved):
lgtm

PR #552, opened 2022-03-01
Automate DockerCompose wait logic

rukai:
Improvements:
* We tend to reuse docker-compose files a lot. Sometimes bugs get introduced due to the current duplication of wait logic:
    + this instance is correct: https://github.com/shotover/shotover-proxy/blob/2dfca49d571371b642cdf989a31950be7397f248/shotover-proxy/benches/redis_benches.rs#L16
    + this instance is a bug: https://github.com/shotover/shotover-proxy/blob/e19c884c9acde30bd6d90f1f6d4111b6394ddef5/shotover-proxy/tests/redis_int_tests/basic_driver_tests.rs#L1317
* I regularly swap between existing docker-compose.yml files in tests without changing the call to `.wait_for_*` then get confused when I hit timing issues. Automatically using the correct wait logic is a huge improvement here.
* Panicking on a new docker-compose.yml file that hasnt had wait logic added yet is better than silently continuing.
* Enables flexible use of docker-compose paths.
    + It will make this code that is already checked in more robust https://github.com/shotover/shotover-proxy/blob/9e500314d2e35b914e4278150231c254a21111c3/shotover-proxy/benches/cassandra_benches.rs#L164
    + It will enable me to make this PR take an arbitrary example folder https://github.com/shotover/shotover-proxy/pull/549


test-helpers/src/docker_compose.rs
```diff
@@ -81,38 +86,43 @@ impl DockerCompose {
...
-    /// * `log_text` - A regular expression defining the text to find in the docker-container log
-    /// output.
-    /// * `count` - The number of times the regular expression should be found.
-    ///
-    /// # Panics
-    /// * If `count` occurrences of `log_text` is not found in the log within 30 seconds.
-    ///
-    pub fn wait_for_n(self, log_text: &str, count: usize) -> Self {
-        self.wait_for_n_t(log_text, count, 30)
+    fn wait_for_containers_to_startup(&self) {
```
benbromhead, line 91: I'm not a huge fan of this approach as it feels like embedded config. I won't block on this, but we'll probably need a config/file based solution at some point further

rukai: IMO using an external config file wouldnt give us much here.
The expected workflow is:
1. developer calls DockerCompose::new on a docker-compose.yml that hasnt been configured yet.
2. the test panics with `"Unknown compose file {path} Please implement waiting logic for it here."` and a file + line number in the stacktrace.
3. Developer jumps to the file + line number and adds a call to wait_for_log.

Adding a config file would result in an extra layer of complexity + prevent alternative wait methods if wait_for_log ends up being unsuitable for some case.

That said once we've used this for a while we might uncover reasons why a config solution would be better :shrug:

conorbros (approved):

PR #553, opened 2022-03-03
Cassandra peers IP address rewriting

conorbros:


rukai:
Need to update transforms.md

docs/src/transforms.md
```diff
@@ -59,12 +59,15 @@ Note: this will just pass the query to the remote node. No cluster discovery or
 
 ### CassandraPeersRewrite
 
-This transform should be used with the `CassandraSinkSingle` transform. It will write over the ports of the peers returned by queries to the `system.peers_v2` table in Cassandra with a user supplied value (typically the port that Shotover is listening on so Cassandra drivers will connect to Shotover instead of the Cassandra nodes themselves).
+This transform should be used with the `CassandraSinkSingle` transform. It will write over the IP addresses and ports of the peers returned by queries to the `system.peers` or `system.peers_v2` table in Cassandra with a user supplied value (typically the IP address or port that Shotover is listening on so Cassandra drivers will connect to Shotover instead of the Cassandra nodes themselves).
 
 ```yaml
 - CassandraPeersRewrite:
     # rewrite the peer ports to 9043
```
rukai, line 66: We should denote optional fields the same way the existing docs does: prefix the comment with "When this field is provided ..."

rukai:
Can we add an integration test for example-configs/cassandra-rewrite-peers?
This will prevent the example from breaking and also integration tests CassandraPeersRewrite in the case where it has ip address rewrite disabled.

docs/src/transforms.md
```diff
@@ -59,12 +59,15 @@ Note: this will just pass the query to the remote node. No cluster discovery or
 
 ### CassandraPeersRewrite
 
-This transform should be used with the `CassandraSinkSingle` transform. It will write over the ports of the peers returned by queries to the `system.peers_v2` table in Cassandra with a user supplied value (typically the port that Shotover is listening on so Cassandra drivers will connect to Shotover instead of the Cassandra nodes themselves).
+This transform should be used with the `CassandraSinkSingle` transform. It will write over the IP addresses and ports of the peers returned by queries to the `system.peers` or `system.peers_v2` table in Cassandra with a user supplied value (typically the IP address or port that Shotover is listening on so Cassandra drivers will connect to Shotover instead of the Cassandra nodes themselves).
```
rukai, line 62: I dont know what its supposed to do, but the implementation currently only rewrites system.peers_v2

conorbros: Oh true, yeah it should be rewriting both.

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
```diff
@@ -67,15 +77,62 @@ fn rewrite_port(message: &mut Message, new_port: u32) {
         if let CassandraOperation::Result(CassandraResult::Rows { value, metadata }) =
             &mut frame.operation
         {
-            let port_column_index = metadata
-                .col_specs
+            let columns = ["native_port", "rpc_port", "broadcast_port", "peer_port"];
```
rukai, line 80: Are these columns documented anywhere?
Should we really be overwriting all of them?

Ditto for the ip columns.

conorbros: I looked through the code of `cdrs-tokio` to see what is used.

conorbros:
For testing `example-configs/cassandra-rewrite-peers` we need to a release so I can use the published Docker image.

rukai:
> For testing example-configs/cassandra-rewrite-peers we need to a release so I can use the published Docker image.

Alright lets not block on that then.

conorbros:
Sounds good to me, I'll add those after the release.

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
```diff
@@ -67,15 +77,62 @@ fn rewrite_port(message: &mut Message, new_port: u32) {
...
+                .map(|col| {
+                    metadata
+                        .col_specs
+                        .iter()
+                        .position(|wanted_col| &wanted_col.name.as_str() == col)
+                })
+                .flatten()
+                .collect::<Vec<_>>();
+
+            if !port_column_indices.is_empty() {
```
rukai, line 93: This if statement is just for performance reasons right?
Can we have the same effect by just making `for i in &port_column_indice` the outermost loop?

conorbros: We can do this: 
```rust
if let MessageValue::Rows(rows) = &mut *value {
    for i in &port_column_indices {
        for row in rows.iter_mut() {
            row[*i] = MessageValue::Integer(new_port as i64, IntSize::I32);
        }
    }
    message.invalidate_cache();
}
```

But we can't move it to the outer-most for borrow checker reasons on `&mut *value`

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
rukai: Looking at the output of these queries I think peer (system.peers_v2) and rpc (system.peers) are the only ones we should be modifying.

## peers_v2
```
 peer       | peer_port | native_address | native_port | preferred_ip | preferred_port | schema_version                       | release_version | host_id                              | data_center | rack
------------+-----------+----------------+-------------+--------------+----------------+--------------------------------------+-----------------+--------------------------------------+-------------+------
 172.16.1.3 |      7000 |     172.16.1.3 |        9042 |         null |           null | 2207c2a9-f598-3971-986b-2926e09e239d |           4.0.2 | 6f15ed45-256c-4ae7-8b8b-46b99726393f |        Mars | West
```
## peers
```
 peer       | preferred_ip | rpc_address | schema_version                       | release_version | host_id                              | data_center | rack
------------+--------------+-------------+--------------------------------------+-----------------+--------------------------------------+-------------+------
 172.16.1.3 |         null |  172.16.1.3 | 2207c2a9-f598-3971-986b-2926e09e239d |           4.0.2 | 6f15ed45-256c-4ae7-8b8b-46b99726393f |        Mars | West
```

rpc was replaced with native https://community.datastax.com/questions/4511/is-it-necessary-to-configure-rpc-address-in-cassan.html
And native is the only one giving results that match up with the values we want to replace. (172.16.1.3, 9042)

If we do add support for system.peers then we should add an integration test case for it.

conorbros: I'll see what columns other drivers are using. We would want to stay compatible across Cassandra versions.

rukai:
This needs a rebase.

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
```diff
@@ -67,15 +78,60 @@ fn rewrite_port(message: &mut Message, new_port: u32) {
...
                 .iter()
-                .position(|col| col.name.as_str() == "native_port");
+                .map(|col| {
+                    metadata
+                        .col_specs
+                        .iter()
+                        .position(|wanted_col| &wanted_col.name.as_str() == col)
+                })
+                .flatten()
+                .collect::<Vec<_>>();
```
rukai, line 92: we dont need to collect here

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
```diff
@@ -67,15 +78,60 @@ fn rewrite_port(message: &mut Message, new_port: u32) {
...
+            let ip_column_indices = columns
+                .iter()
+                .map(|col| {
+                    metadata
+                        .col_specs
+                        .iter()
+                        .position(|wanted_col| &wanted_col.name.as_str() == col)
+                })
+                .flatten()
+                .collect::<Vec<_>>();
```
rukai, line 127: we dont need to collect here

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
```diff
@@ -67,15 +78,60 @@ fn rewrite_port(message: &mut Message, new_port: u32) {
...
 
-            if let Some(i) = port_column_index {
+            if let MessageValue::Rows(rows) = &mut *value {
+                for i in &port_column_indices {
+                    for row in rows.iter_mut() {
+                        row[*i] = MessageValue::Integer(new_port as i64, IntSize::I32);
+                    }
+                }
+                message.invalidate_cache();
```
rukai, line 100: Maybe just stick a single invalidate_cache call here?
image
If the system.peers is behaving correctly then it should be fine to just always invalidate the cache on its responses.

As its done currently we have some level of protection to only invalidate cache when a modification occurs so at first glance it looks like a possible bug that we dont have full protection here.
So I think its better to just go the other way and always invalidate the cache.

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
conorbros: Did you mean use `filter_map` instead?

rukai: I mean like 
image

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
rukai: because shotover cant speak the protocols that these other ports/addresses are using?

rukai: >  If the drivers are requesting them, we should modify to ensure any communication goes through Shotover.

Oh, I guess if the drive is requesting them then they might have some use in client/server communication?
What exactly is the driver doing with the peer_port?

conorbros: It uses it to connect to the node AFAIK. I'm not sure what other protocol they would be using to communicate?

rukai: The way this is worded makes it sound like a seperate protocol to me https://docs.datastax.com/en/cassandra-oss/3.x/cassandra/architecture/archGossipAbout.html

conorbros: That's internode communication, not client/server.

rukai: Hmmm we are definitely talking past each other here.
What I have been trying to say is that I think the other ports/addresses listed by system.peers are for internode communication not client/server communication.

rukai: https://cassandra.apache.org/doc/latest/cassandra/faq/#what-ports gossip is on port 7000 and the peer_port contains 7000

benbromhead: so to quickly clarify 
`peers.peer` is the address that internode comms should use for the remote peer
`peers.preferred_address` is the advertised or try first address/port for the internode protocol (some environments like AWS where you might have a natted public ip)
`peers.rpc_address` is the address that client port binds to

`peers_v2.peer` is the address that internode comms should use for the remote peer
`peers_v2.peer_port` is the port that internode comms should use for the remote peer
`peers_v2.preferred_ip` is the advertised or try first address for the internode protocol (some environments like AWS where you might have a natted public ip)
`peers_v2.preferred_port` is the advertised or try first port for the internode protocol (some environments like AWS or k8s where you might have a natted public ip/port forwarding)
`peers_v2.native_address` is the address that the remote peer listens for client connections
`peers_v2.native_port` is the port that the remote peer listens for client connections on

From a driver perspective, we only want to rewrite `peers_v2.native_address`, `peers_v2.native_port` and `peers.rpc_address`

The driver will fetch things like `peers.peer`, `peers_v2.peer` and `peers_v2.peer_port` to uniquely identify a node (this is how C* identifies them with cmd line tools like nodetool ... even though they have uuids). 

For system.local there are even more things in there, but more or less the driver will only care about the following:
`rpc_address` - is the address that the current connected node listens for client connections on (this should be the address we are talking to upstream)
`rpc_port` - is the port that the current connected node listens for client connections on (this should be the port we are talking to upstream).

**Why is the terminology different for peers, peers_v2 and local (why no local_v2)?** 
With local, the changes to the table required to support identifying the node by address and port where minimal, as was the ability to support different client list ports (they just needed to add columns). Clients that didn't support the new format wouldn't query for those new columns. 

**Internode/Gossip/Native/RPC/JMX oh my!**

- Internode is the catch all term for nodes talking to each other via (normally) port 7000. This includes all control messages, as well as data requests from a coordinator node to a replica.
- Gossip is the concensus mechanism that C* uses for discovery, gossip occurs over the internode protocol (e.g. port 7000)
- Native protocol is the broad term given to the new (since 2.0 iirc) client protocol, right now this is the only protocol that shotover supports. This is the thing defined by `doc/native_protocol_v5.spec` in the C* repo
- RPC usually refers to the old RPC Thrift based client protocol that Cassandra used to use
- JMX is a java based RPC mechanism that C* uses for node management (we really dont need to worry about it, but I included it for completeness sake).

benbromhead: This snipped from the Java driver is useful in reasoning around this -> https://github.com/datastax/java-driver/blob/ef56d561d97adcae48e0e6e8807f334aedc0d783/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java#L493-L533

benbromhead (approved):
lgtm

PR #554, opened 2022-03-03
Remove connection balance

conorbros:
Isn't tested or documented so I have removed it. Can add back when we want to properly support it.


rukai:
While I'll take any excuse to delete code, I'll defer to ben on this one.
A reason to keep it is to demonstrate future direction for shotover.
Its also not in the public API so its not causing too much problems leaving it around.

benbromhead:
Let's leave it for the moment and create a ticket to document.

The other approach would be to start having a folder / example dir or something that includes example / tutorial transforms that are not necessarily useful as fully baked transforms

PR #559, opened 2022-03-08
Add instruction counter to benches

rukai:
Unfortunately it seems unusable at the moment.
I am getting instruction counts of ~7000 per message sent which sounds way too low. It must be failing to count instructions that occur in other threads.
I raised https://github.com/jbreitbart/criterion-perf-events/issues/19

I'll leave this around as a draft in case we want to refer back to it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -25,7 +25,7 @@ checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
 dependencies = [
  "getrandom",
  "once_cell",
- "version_check",
+ "version_check 0.9.4",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


rukai:
I was suggested to use the `inerhit()` method on the builder which certainly improves things.
I'm now getting values around ~200,000.
But there are still some strange results such as passthrough_tls_select being 87,000 while passthrough_select is 198,000.

So looks like this might be doable but will require some investigation.

rukai:
https://github.com/actions/virtual-environments/issues/4974
https://github.community/t/getting-access-to-perf-events-on-github-hosted-ubuntu-latest/149164/4
https://stackoverflow.com/questions/60988282/perf-alternate-performance-management-tools-that-can-run-on-google-cloud-platf

I found these which show that getting instruction count on GA VMs will be impossible.
The only solution would be providing our own hardware or maybe there is some cloud provider that supports it.

benbromhead:
This will probably be something that folks can run locally on their own machines and not worth making part of CI. I would see generaly perf testing we do elsewhere as a "good enough" smell vs instruction count.

rukai:
There could be some value in instruction counting locally, but the main goal was to make CI bench results stable.
benchmarking on a local machine should be stable enough that instruction counts arent as useful.

I'm going to close this, our benchmarks need a lot of infrastructure improvements but we'll have to go a different direction.

PR #561, opened 2022-03-10
Bump up Cassandra throttling test threshold

conorbros:


rukai:
Here is an example of where this failed: https://github.com/shotover/shotover-proxy/runs/5491095274?check_suite_focus=true

rukai (approved):

XA21X (approved):
LGTM, but it would be slightly easier to read like:

`20 < len && len <= 25`

PR #562, opened 2022-03-11
Count BATCH queries for request throttling

conorbros:


shotover-proxy/src/frame/cassandra.rs
```diff
@@ -63,6 +82,18 @@ impl CassandraFrame {
...
 
+    // Count the amonut of queries in this `CassandraFrame`, this will either be the count of all queries in a BATCH statement or 1 for all other queries
+    pub(crate) fn get_query_count(&self) -> Result<NonZeroU32> {
+        Ok(match &self.operation {
+            CassandraOperation::Batch(bytes) => {
+                let short_bytes = &bytes[1..3];
+                let short = u16::from_be_bytes(short_bytes.try_into()?);
+                NonZeroU32::new(short.into()).unwrap()
+            }
+            _ => nonzero!(1u32),
```
rukai, line 93: this seems wrong.
operations other than Batch and Query should have count 0.
And is it gauranteed that query will always contain exactly one query.
I dont see that specified in the spec anywhere.

rukai: After discussion with conor:
* The name is misleading the desired functionality is to count messages and batch should be considered as each query being multiple sub messages.
* By sending messages like `INSERT INTO foo.foo (field) VALUES (1); INSERT INTO foo.foo (field) VALUES (2);` and `SELECT * FROM system.peers; SELECT * FROM system.peers;` I can see that these get split and sent in seperate messages.
    + It might be worth checking what happens if we manually construct a message with multiple queries in a single QUERY message but I cant imagine it working as there would be no way for the protocol to properly respond to such a message.

rukai: Ended up hacking shotover to send my cursed query:
image
And I can confirm that cassandra rejects it.

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -43,6 +46,22 @@ pub(crate) fn metadata(bytes: &[u8]) -> Result<CassandraMetadata> {
     })
 }
 
+/// Count queries only from an unparsed Cassandra frame
+pub(crate) fn get_message_count(bytes: &[u8]) -> Result<NonZeroU32> {
+    let frame = RawCassandraFrame::from_buffer(bytes, Compression::None)
```
rukai, line 51: Now that we have to actually allocate for the message body its not completely clear that these sorts of speed hacks are still worth it.
If we allocate here and then later on call frame() we will be allocating for the body twice.

Possibly we should have a way to get the body without allocation.
Possibly we should rewrite CQL to only parse when needed.
Happy to land this as is at the moment but eventually we will need to measure this kind of thing and figure out how far we should go with these kinds of speed hacks.

shotover-proxy/src/transforms/throttling.rs
```diff
@@ -43,10 +43,17 @@ impl Transform for RequestThrottling {
...
             .into_iter()
             .rev()
-            .filter(|_| self.limiter.check().is_err())
-            .map(|i| {
-                let message = message_wrapper.messages.remove(i);
-                (message, i)
+            .filter_map(|i| {
+                if self
+                    .limiter
+                    .check_n(message_wrapper.messages[i].message_count().ok()?)
```
rukai, line 49: Should this be configurable? Or does it make sense to just always enable this logic?
I'm leaning towards just always enable it, but it something we should think about.

shotover-proxy/src/frame/cassandra.rs
conorbros: I'm happy to do it in this PR.

shotover-proxy/src/transforms/throttling.rs
conorbros: I think enable it too, the transform is specifically "per query throttling".

shotover-proxy/src/frame/cassandra.rs
rukai: I think writing some benchmarks that just hit the RequestThrottling transform (in a prereq PR) would be a good place to start.
That would help evaluate the performance impact of each approach.

conorbros: Opened a PR for benchmarks: #565

benbromhead (approved):
lgtm

shotover-proxy/src/frame/cassandra.rs
conorbros: Resolving as benchmarks PR has been merged.

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -63,6 +90,14 @@ impl CassandraFrame {
         }
     }
 
+    // Count the amonut of queries in this `CassandraFrame`, this will either be the count of all queries in a BATCH statement or 1 for all other queries
```
rukai, line 93: amonut -> amount

rukai (approved):
Nice!
`transform/cassandra_request_throttling_unparsed` shows identical performance between this PR and main.

rukai:
@conorbros I just noticed a bug in this PR
Its valid for a batch to contain 0 statements which would cause a panic as we turn it into a NonZeroU32.
Maybe the best way to handle this is to just do `short.max(1)`?
Can you make a follow up PR to fix it?

PR #565, opened 2022-03-14
Cassandra Request Throttling Benches

conorbros:


shotover-proxy/example-configs/cassandra-request-throttling/topology.yaml
```diff
@@ -0,0 +1,15 @@
+---
+sources:
+  cassandra_prod:
+    Cassandra:
+      listen_addr: "127.0.0.1:9042"
+chain_config:
+  main_chain:
+    - RequestThrottling:
+        max_requests_per_second: 40000
```
rukai, line 9: Hmmmmm...
I think the only way to get consistent results from this benchmark would be to always allow messages through (some absurdly high number) or never allow messages through (value of 1 is close enough)
Otherwise we'll get all sorts of weird performance implications from reaching or not reaching the max on different machines and at different points of the 3 second benchmark.

conorbros: Yes, I thought 40k would be enough to allow all through.

rukai: ah, lets add a comment explaining that and maybe bump it up a bit more, should be safe at 100,000,000

rukai: Should we also have a second benchmark that lets 1 message through?

conorbros: I don't think 1 would work because the driver sends a bunch of startup messages. 

I'm starting to think we should set a minimum allowed value for the request throttling because it can break things if you set it too low.

rukai: > I don't think 1 would work because the driver sends a bunch of startup messages.

alright, the single bench we have now should be enough.

> I'm starting to think we should set a minimum allowed value for the request throttling because it can break things if you set it too low.

That makes sense to me.
Picking an appropriate value could be tricky though.
We need plenty of buffer in case the handshake changes between cassandra releases.

conorbros: After a bit of testing, the lowest value I could get to work was 15, so I think 20 is a reasonable minimum.

rukai: I would go to 50.
I think its better to be on the safe side here.

rukai (approved):
lgtm

rukai:
Sounds like 100s waiting for 2 cassandra instances isnt enough?

benbromhead (approved):
lgtm

PR #567, opened 2022-03-14
Minimum value for Request Throttling config

conorbros:
As per discussion on #565

### DIFF
--- a/shotover-proxy/src/transforms/throttling.rs
+++ b/shotover-proxy/src/transforms/throttling.rs
@@ -3,7 +3,7 @@ use crate::{
     message::Message,
     transforms::{Transform, Transforms, Wrapper},
 };
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use async_trait::async_trait;
 use governor::{
     clock::DefaultClock,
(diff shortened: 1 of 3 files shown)


shotover-proxy/src/transforms/throttling.rs
```diff
@@ -22,6 +23,12 @@ pub struct RequestThrottlingConfig {
 
 impl RequestThrottlingConfig {
     pub async fn get_source(&self) -> Result<Transforms> {
+        if self.max_requests_per_second < nonzero!(50u32) {
```
rukai, line 26: This should be written as a validate() implementation.
Failures in validation give better information to the user. e.g. Includes the chain name in the error and splits the error across multiple lines to make it more readable.

shotover-proxy/src/transforms/throttling.rs
```diff
@@ -66,4 +69,63 @@ impl Transform for RequestThrottling {
 
         Ok(responses)
     }
+
+    fn validate(&self) -> Vec<String> {
+        if self.max_requests_per_second < nonzero!(50u32) {
+            vec![
+                "RequestThrottling".into(),
```
rukai, line 76: `:` after the transform name

rukai (approved):
lgtm

benbromhead:
This needs to get documented (not only that there is a minimum, but why... e.g. the rate limiting alg falls apart when you try to do something too low).

conorbros:
Opened a PR for docs #573

benbromhead (approved):

Issue #563, opened 2022-03-11
Parse cassandra BATCH messages


rukai:
Batch messages contains queries just like QUERY messages but there is currently no way for shotover transforms to process them.

~~If possible, an implementation where CassandraOperation::Query and CassandraOperation::Batch are combined into a single variant would give really nice transform UX.
we would disambiguate between them by checking the length of the queries Vec.
if queries.len() == 1 then its a QUERY message otherwise its a BATCH message.
There is a bunch of complicated flags and stuff going on in the BATCH message protocol though so this might not be possible.~~

Didnt end up taking that approach.
Instead we currently have the half finished `QueryIterator` abstraction.
Just need to finish it off.

benbromhead:
Can we close this out now?

PR #570, opened 2022-03-15
Add CassandraOperation::queries method to abstract over QUERY and BATCH

rukai:
Progress towards https://github.com/shotover/shotover-proxy/issues/563
I ended up going for a different design then what I described in the issue.
QUERY and BATCH diverge too much to take that route.
So instead I propose we just add a common getter method so we can pretend batches and queries are the same thing in most cases but then fall back to directly accessing query/batch specific fields if we ever need it.

TODO in follow up PRs:
* CassandraOperation::Query should store a single Statement instead of a `Vec<Statement>` (we can use https://doc.rust-lang.org/std/slice/fn.from_mut.html to turn it into a `&mut [T]`)
* Return CassandraOperation::Batch queries in CassandraOperation::queries once we make use of https://docs.rs/cassandra-protocol/latest/cassandra_protocol/frame/frame_batch/struct.BodyReqBatch.html

We can land this PR as is if we are ok with the CassandraOperation::queries API introduced by the PR and can confirm that its use in transforms is sound.

When thinking about maintaining transform invariants when supporting batch requests remember that batch requests only support INSERT/UPDATE/DELETE queries so we dont have to worry about them returning values. (But they could still return an error)


conorbros:
Is BATCH parsing not already in cassandra-protocol?

rukai:
Oh, turns out it is already implemented: https://docs.rs/cassandra-protocol/latest/cassandra_protocol/frame/frame_batch/struct.BodyReqBatch.html

I still want to make use of it in a follow up PR though.

rukai:
Hmmm I might have to think about this more, the way the batch protocol mixes prepared and regular queries makes it impossible to expose `Statement`s as a slice without butchering the way we store batch queries.

rukai:
All good, I changed the slice to an iterator and we should be able to expose batch queries just fine with that.

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -286,6 +286,18 @@ pub enum CassandraOperation {
 }
 
 impl CassandraOperation {
+    /// Return all queries contained within CassandaOperation::Query and CassandraOperation::Batch
+    pub fn queries(&mut self) -> std::slice::IterMut<Statement> {
```
conorbros, line 290: Add an import

shotover-proxy/src/transforms/redis/cache.rs
```diff
@@ -72,38 +69,48 @@ impl SimpleRedisCache {
...
-        for message in &mut messages {
-            info!("Received reply from redis cache {:?}", message);
-            // TODO: Translate the redis reply into cassandra
-            *message = Message::from_frame(Frame::Cassandra(CassandraFrame {
+
+        // Replace cass_request messages with cassandra responses in place.
+        // We reuse the vec like this to save allocations.
+        let mut messages_redis_response_iter = messages_redis_response.iter_mut();
+        for cass_request in &mut messages_cass_request {
+            let _redis_response = messages_redis_response_iter.next();
```
conorbros, line 95: What's this doing?

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -286,6 +286,18 @@ pub enum CassandraOperation {
...
+        match self {
+            CassandraOperation::Query {
+                query: CQL::Parsed(query),
+                ..
+            } => query.iter_mut(),
+            // TODO: Return CassandraOperation::Batch queries once we add BATCH parsing to cassandra-protocol
```
conorbros, line 296: Are all the TODOs intended to be done in the follow up PRs you mention?

shotover-proxy/src/transforms/redis/cache.rs
rukai: Its demonstrating how to get the redis responses that correspond to this cassandra request.
Each cassandra request can have multiple redis responses due to batches getting split up into multiple redis messages.

There is nothing to actually use the redis responses as even before this PR this was left as a TODO: `// TODO: Translate the redis reply into cassandra`

shotover-proxy/src/frame/cassandra.rs
rukai: This is the only new TODO and yes I intend to do it in a follow up PR

conorbros: Oh true, the other one is just whitespace.

shotover-proxy/src/transforms/redis/cache.rs
conorbros: Could you add a comment explaining this, all the iterators are a bit hard to follow.

rukai: In order to understand what the iterators are doing you need to understand how the earlier vecs are constructed.
So I wrote some comments explaining my mental model of how the whole function works.
Does this help?

Also, thanks to writing out these comments I noticed and fixed a bug earlier in this function.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -964,6 +964,7 @@ mod functions {
             "CREATE TABLE test_function_keyspace.test_function_table (id int PRIMARY KEY, x int, y int);",
         );
         run_query(
+            // I think these are sending as query messages lolololol
```
rukai, line 967: oh right this.
I'll investigate.

shotover-proxy/src/transforms/redis/cache.rs
conorbros: Yep that clears it up, thanks!

conorbros (approved):

benbromhead (approved):
lgtm

PR #572, opened 2022-03-15
Add check for untracked/uncommitted files in CI

rukai:
This will prevent incidents like https://github.com/shotover/shotover-proxy/pull/571 from occurring again.
Will also ensure tests aren't leaving weird files around.


conorbros:
Already found one!

benbromhead (approved):

conorbros (approved):

PR #573, opened 2022-03-16
RequestThrottling docs

conorbros:


docs/src/transforms.md
```diff
@@ -367,3 +368,13 @@ The response from the down-chain transform is returned back up-chain but various
           filter: Read
       - Null
 ```
+
+
+### RequestThrottling
+
+This transform will backpressure requests to Shotover, ensuring that throughput does not exceed the `max_requests_per_second` value. It uses a [Generic Cell Rate Algorithm ](https://en.wikipedia.org/wiki/Generic_cell_rate_algorithm). `max_requests_per_second` has a minimum allowed value of 50 to ensure that drivers such as Cassandra are able to complete their startup procedure correctly.
```
rukai, line 375: We should also document how message count is calculated per message.

rukai (approved):
lgtm

benbromhead (approved):

PR #574, opened 2022-03-16
Batch query counting bug fix

conorbros:
As @rukai noticed, it is valid for a BATCH statement to have 0 statements which would cause our current implementation of counting queries to panic on trying to convert a `u16` of 0 to a `NonZeroU32`. The bug is exposed in the first commit containing tests, the second commit is a fix that uses `max(1)` to stop it from occurring.


shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1278,6 +1278,60 @@ mod protect {
...
+        let mut batch = Batch::new(BatchType::LOGGED);
+        for i in 0..2 {
+            let statement = format!("INSERT INTO batch_keyspace.batch_table (id, lastname, firstname) VALUES ({}, 'text', 'text')", i);
+            batch.add_statement(&stmt!(statement.as_str())).unwrap();
+        }
+        connection.execute_batch(&batch).wait().unwrap();
+        assert_query_result(
+            connection,
+            "SELECT lastname FROM batch_keyspace.batch_table WHERE id = 1;",
+            &[&[ResultValue::Varchar("text".into())]],
```
rukai, line 1298: Can we assert on every row and every field in this table?
I could imagine bugs where some of these get lost in shotovers parsing + reconstruction of batch messages.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1288,31 +1288,39 @@ fn test_batch_statements(connection: &Session) {
...
             batch.add_statement(&stmt!(statement.as_str())).unwrap();
         }
         connection.execute_batch(&batch).wait().unwrap();
         assert_query_result(
             connection,
-            "SELECT lastname FROM batch_keyspace.batch_table WHERE id = 1;",
-            &[&[ResultValue::Varchar("text".into())]],
+            "SELECT id, firstname, lastname FROM batch_keyspace.batch_table WHERE id = 1;",
+            &[&[
+                ResultValue::Int(1),
```
rukai, line 1299: And every row?

rukai:
Can we also extend this to test every case using both BATCH protocol and batch over QUERY protocol?
(we can put this in a follow up PR if you want)

conorbros:
Yeah, happy to do in a follow up PR. I'll keep this one as a fix for the bug only.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
rukai: still need this in case you missed it.

conorbros: Have I fixed everything you meant now?

rukai: I'm thinking to remove the WHERE and add an ORDER BY and then assert we get back both rows.

conorbros: We don't need to ORDER BY, the assert methods we have sort it for us.

rukai: oh excellent, lets remove the WHERE then.

The reason being: we arent trying to test select query handling, we are trying to test the batch successfully ran ALL inserts.

rukai (approved):
LGTM

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -34,7 +34,7 @@ fn get_batch_len(bytes: &[u8]) -> Result<NonZeroU32> {
     }
 
     let short_bytes = &bytes[1..3];
-    let short = u16::from_be_bytes(short_bytes.try_into()?);
+    let short = u16::from_be_bytes(short_bytes.try_into()?).max(1); // it is valid for a batch statement to have 0 statements, but for the purposes of shotover throttling we can count it as one query
```
benbromhead, line 37: I feel like I'm missing something, but I need to ask... should this be min? How do we ever get a batch len longer than 1 in this case?


I thought about commenting on this on the other PR. But here is my thoughts as this will have far ranging consquences to all transform authors for only the silliest of non-real world queries (batches used properly should never be 0 or 1 despite being valid in the protocol). 

We may need to distinguish how batches work in relation to query counts and define formally what a batch is to shotover. To this I would simply say we formally define that shotover will always count empty batches as a batch size of one. So transform authors need to be aware of this, if they care.

They can also get the real count from the AST if needed

rukai: Hmmm I havent checked if min or max is the correct one here but we should probably just replace it with `NonZeroU32::new(short).unwrap_or(nonzero!(1u32))`
The advantages are:
* its more obviously correct
* unwraps can have slightly worse performance due to the compiler failing to optimize out the panic. Not something thats usually worth worrying about but it adds another bullet point here :P

rukai: > batches used properly should never be 0 or 1 despite being valid in the protocol

I think its not too unreasonable for a programmatically generated batch to contain 0 or 1 messages.

> We may need to distinguish how batches work in relation to query counts and define formally what a batch is to shotover. To this I would simply say we formally define that shotover will always count empty batches as a batch size of one. So transform authors need to be aware of this, if they care.

I guess the outcome of this is `Message::message_count` needs a doc comment that explains the meaning of the returned value and how empty batches are defined,  and in a way that doesnt refer to throttling as there are other possible use cases.

conorbros: So we want to make `Message::get_message_count` available to transform authors? In that case it should be named a bit better, we might want to call it `Message::get_request_count` instead and define what a request is in Shotover instead. (i.e. every request except for batches counts as 1 request.

shotover-proxy/src/message/mod.rs
```diff
@@ -178,18 +178,20 @@ impl Message {
         }
     }
 
-    pub fn message_count(&self) -> Result<NonZeroU32> {
+    /// Get the count of "requests" in this message. In Shotover, all messages are only 1 request except for batch messages which can contain many requests. If a batch message has
+    /// no inner reqeusts, it's count is 1.
```
rukai, line 182: reqeusts -> requests
I think the wording "1 or more requests" gives more confidence to the possible values than "many requests"
a message can either be a response or a request but a response still contains 1 request? :thinking:

conorbros: Hmm, I'm thinking we frame it more in the context of rate limiting then. Since the terminology I tried to use here is too confusing. Perhaps "cell", like what the GCRA/library call it. 

https://docs.rs/governor/latest/governor/struct.RateLimiter.html#method.check

rukai: Replacing "request" with "cell" could work.
Maybe something like "submessage" would be more self explanatory though. e.g. `message.submessage_count()`

conorbros: To me that is a bit confusing, since BATCH messages are the only ones that actually have "submessages".

rukai: well at least with submessage its a word that doesnt already mean anything.
So we can define a meaning here and it means exactly what we say it means.
But message and request already have preexisting meanings in shotover.

I also dont see how that differs to cell, only BATCH messages can contain cells.

But I can see an argument for making this method designed around getting values useful for rate limiting and documenting it as such.
Update the docs with your cell approach and we'll see how it reads.

conorbros: Cell means "ticket" in the rate limiter. So most messages have one "ticket" that needs to get through the rate limiter, whereas BATCH messages have many. It's less about the actual content of the message and more about what the rate limiter considers them to be.

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -36,8 +36,8 @@ fn get_batch_len(bytes: &[u8]) -> Result<NonZeroU32> {
     }
 
     let short_bytes = &bytes[1..3];
-    let short = u16::from_be_bytes(short_bytes.try_into()?);
-    Ok(NonZeroU32::new(short.into()).unwrap())
+    let short = u16::from_be_bytes(short_bytes.try_into()?); // it is valid for a batch statement to have 0 statements, but for the purposes of shotover throttling we can count it as one query
```
rukai, line 39: Can we put this comment on a newline?
At the moment it looks like it applies line to line 39 but it actually applies to line 40

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -61,7 +61,7 @@ pub(crate) fn metadata(bytes: &[u8]) -> Result<CassandraMetadata> {
 }
 
 /// Count queries only from an unparsed Cassandra frame
-pub(crate) fn get_message_count(bytes: &[u8]) -> Result<NonZeroU32> {
+pub(crate) fn get_query_count(bytes: &[u8]) -> Result<NonZeroU32> {
```
rukai, line 64: get_query_count -> get_cell_count

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -93,7 +93,7 @@ impl CassandraFrame {
     }
 
     // Count the amount of queries in this `CassandraFrame`, this will either be the count of all queries in a BATCH statement or 1 for all other queries
-    pub(crate) fn get_message_count(&self) -> Result<NonZeroU32> {
+    pub(crate) fn get_query_count(&self) -> Result<NonZeroU32> {
```
rukai, line 96: get_query_count -> get_cell_count

shotover-proxy/src/frame/cassandra.rs
rukai: should we drop the get_?
We dont have it on the highest level function.

shotover-proxy/src/message/mod.rs
```diff
@@ -182,18 +182,21 @@ impl Message {
         }
     }
 
-    pub fn message_count(&self) -> Result<NonZeroU32> {
+    /// Get the count of "cells" in this message. In Shotover, all messages are only 1 cell except for batch messages which can contain 1 or more cells.
+    /// A cell in a batch message is equivalent to one of it's inner batched messages.
```
rukai, line 186: I find this sentence hard to understand.

We could write it more simply as:
```
Each inner message in a batch message is worth one cell.
```

Or maybe a bullet list of cell count logic would be more clear than a paragraph:
```
Cell count:
* Regular message - 1 cell
* Message containing submessages e.g. a batch request - 1 cell per submessage
* Message containing submessages with 0 submessages - 1 cell
```

rukai: The idea of the bullet points is to supercede all count explanations in the paragraph.
So we would get rid of "In Shotover, all messages are only 1 cell except for batch messages which can contain 1 or more cells. Batch messages have a cell count of 1 cell per inner message. "

Is that alright?

conorbros: Yes, I prefer your way of explaining things with the points. I'll switch it over.

rukai (approved):
LGTM

benbromhead (approved):
lgtm

PR #576, opened 2022-03-17
Parse cassandra batch messages

rukai:
Takes the same approach as CassandraOperation::Result, recreate a bunch of structures so that we can replace the string query with the query AST.
batch queries are now parsed by sql_parser (soon to be replaced with claudes parser)

### DIFF
--- a/shotover-proxy/src/frame/cassandra.rs
+++ b/shotover-proxy/src/frame/cassandra.rs
@@ -96,6 +96,7 @@ impl CassandraFrame {
     pub(crate) fn get_message_count(&self) -> Result<NonZeroU32> {
         Ok(match &self.operation {
             CassandraOperation::Batch(batch) => {
+                // it doesnt make sense to say a message is 0 messages, so when the batch has no queries we round up to 1
                 NonZeroU32::new(batch.queries.len() as u32).unwrap_or(nonzero!(1u32))
             }
             _ => nonzero!(1u32),
@@ -114,7 +115,7 @@ impl CassandraFrame {
                         params: body.query_params,
                     }
                 } else {
-                    unreachable!()
+                    unreachable!("We already know the operation is a query")
                 }
             }
             Opcode::Result => {
@@ -169,14 +170,14 @@ impl CassandraFrame {
                         ResResultBody::Void => CassandraOperation::Result(CassandraResult::Void),
                     }
                 } else {
-                    unreachable!()
+                    unreachable!("We already know the operation is a result")
                 }
             }
             Opcode::Error => {
                 if let ResponseBody::Error(body) = frame.response_body()? {
                     CassandraOperation::Error(body)
                 } else {
-                    unreachable!()
+                    unreachable!("We already know the operation is an error")
                 }
             }
             Opcode::Startup => CassandraOperation::Startup(frame.body),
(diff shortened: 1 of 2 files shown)


shotover-proxy/src/frame/cassandra.rs
```diff
@@ -432,3 +481,24 @@ pub enum CassandraResult {
...
+
+#[derive(PartialEq, Debug, Clone)]
+pub enum BatchStatementTy {
+    Statement(CQL),
+    PreparedId(CBytesShort),
+}
+
+#[derive(PartialEq, Debug, Clone)]
+pub struct BatchStatement {
+    ty: BatchStatementTy,
```
conorbros, line 493: Can we just say `type`?

rukai: nope, `type` is a keyword

rukai: ahhh I should do BatchStatementTy -> BatchStatementType though

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -93,7 +95,9 @@ impl CassandraFrame {
     // Count the amount of queries in this `CassandraFrame`, this will either be the count of all queries in a BATCH statement or 1 for all other queries
     pub(crate) fn get_message_count(&self) -> Result<NonZeroU32> {
         Ok(match &self.operation {
-            CassandraOperation::Batch(bytes) => get_batch_len(bytes)?,
+            CassandraOperation::Batch(batch) => {
+                NonZeroU32::new(batch.queries.len() as u32).unwrap_or(nonzero!(1u32))
```
conorbros, line 99: Similar to the bug-fix PR, can we add a comment explaining this? i.e. BATCH statements query length can be 0 and for our purposes counting as 1 is fine.

conorbros (approved):
Just some some things. Other than that, LGTM!

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -110,7 +115,7 @@ impl CassandraFrame {
                         params: body.query_params,
                     }
                 } else {
-                    unreachable!()
+                    unreachable!("We already know the operation is a query")
```
benbromhead, line 118: we can do this in a different PR, but I feel like all the nested `if let else` statements in each match arm could be brought up a level to the match it self:

```rust
let operation = match (frame.opcode, frame.request_body()?) {
            (Opcode::Query, RequestBody::Query(body)) => {}
...
}
```

Then put the unreachable in the fall through arm.

benbromhead, line 118: we can do this in a different PR, but I feel like all the nested `if let else` statements in each match arm could be brought up a level to the match it self:

```rust
let operation = match (frame.opcode, frame.request_body()?) {
            (Opcode::Query, RequestBody::Query(body)) => {}
...
}
```

Then put the unreachable in the fall through arm.

benbromhead (approved):

shotover-proxy/src/frame/cassandra.rs
rukai: Unfortunately the entire reason we have to do this is because cassandra-protocol splits the logic into separate request_body and response_body methods.
I tried to unify them before but got resistance from the maintainer.
Its something to look into some day but I considered it low priority.

PR #577, opened 2022-03-18
Split up message methods

conorbros:
Noticed this pattern of splitting up similar method in different `impl` on some other libraries.


rukai (approved):
fine by me

rukai:
@conorbros I reran it but you should look into making it more robust:
image

benbromhead (approved):

PR #346, opened 2021-11-03
Redis cluster with a cluster aware client docs

conorbros:
<s>Should land #347 first</s>

PR #578, opened 2022-03-18
Switch over to published Docker image on Cassandra example

conorbros:


rukai:
Now that we can run it in CI we should add a test for it.

conorbros:
We can't test it at the moment for the same reason described here: https://github.com/shotover/shotover-proxy/pull/346#discussion_r755561861

rukai:
Should we add support for that in this PR then?
Otherwise I dont know when we will pay off that debt.

conorbros:
Yep I think so.

test-helpers/src/docker_compose.rs
```diff
@@ -138,8 +163,7 @@ impl DockerCompose {
     /// # Panics
     /// * If `count` occurrences of `log_text` is not found in the log within 90 seconds.
     ///
-    fn wait_for_log(&self, log_text: &str, count: usize) {
-        let timeout_seconds = 110;
```
rukai, line 142: Can we just bump this to the safer 150 instead of allowing the choice?
If you really want to allow the choice per docker-compose.yaml then we should specify the value within the wait_for_containers_to_startup match, thats kind of its point after all.

conorbros: It will save us 40 seconds in some cases so I think having it as low as possible is good. 

I can move the logic to `wait_for_containers_to_startup`.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1443,6 +1443,26 @@ fn test_cassandra_peers_rewrite() {
...
+async fn test_cassandra_rewrite_peers_example() {
+    let _docker_compose =
+        DockerCompose::new("example-configs/cassandra-rewrite-peers/docker-compose.yml");
+
+    let shotover_manager = ShotoverManager::from_topology_file(
+        "example-configs/cassandra-rewrite-peers/topology.yaml",
+    );
+
+    let connection = shotover_manager.cassandra_connection("172.16.1.2", 9043);
+    table::test(&connection); // run some basic tests to confirm it works as normal
```
rukai, line 1457: We dont actually want to run any tests on it.
The image is old shotover code so who knows what features its missing.

Just starting up and shutting down the container should be enough to ensure that the configuration remains valid.

Also given that, I think this should go in its own file.
We definitely used to have a file for these sorts of "test config loads, but dont actually test functionality" cases, but I think we deleted that file due to all the tests being removed due to those examples being deleted.

shotover-proxy/example-configs/cassandra-rewrite-peers/config.yaml
```diff
@@ -0,0 +1,3 @@
+---
```
rukai, line 1: What is this for? Dont we instruct the user to use the generic config.yaml?

conorbros: Needs to be mounted. The reason we can't use the base one is because then it won't mount on the user's machine. This means as long as they put it in the same directory it will work.

rukai: ah, makes sense

shotover-proxy/tests/examples/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+use serial_test::serial;
+use test_helpers::docker_compose::DockerCompose;
+
+#[tokio::test(flavor = "multi_thread")]
+#[serial]
+async fn test_cassandra_rewrite_peers_example() {
+    let _docker_compose =
+        DockerCompose::new("example-configs/cassandra-rewrite-peers/docker-compose.yml");
+
+    let shotover_manager = ShotoverManager::from_topology_file(
```
rukai, line 12: why do we have a shotover manager?
Shouldnt shotover be launched by docker?

conorbros: To connect to the cassandra cluster, but I'll make the method available as a function.

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -1,10 +1,23 @@
-use cassandra_cpp::{stmt, Error, Session, Value, ValueType};
+use cassandra_cpp::{stmt, Cluster, Error, Session, Value, ValueType};
 use ordered_float::OrderedFloat;
 
 mod basic_driver_tests;
 
+#[allow(unused)]
+pub fn cassandra_connection(contact_points: &str, port: u16) -> Session {
```
rukai, line 7: Isnt this now completely duplicated with the existing method?

conorbros: Yes

rukai: yeah understandable to not blow up this PR, but at least make the other cassandra_connection call into this one?

conorbros: Ended up moving stuff around a bit to make this change, so I opened another PR to make the change a bit easier: #584 584

rukai: perfect

rukai, line 7: Make sure we use the one from https://github.com/shotover/shotover-proxy/pull/584 before merging

rukai (approved):
LGTM!

benbromhead:
Shouldn't we be using the local docker image built using the shotover executable for the current build?

conorbros:
I don't think so, that would mean the user would be downloading the Shotover source code and building it themselves. The purpose of this example is to demonstrate to user's how to setup Shotover for production use and we want to encourage them to use a released version of Shotover.

rukai:
I want to talk about this more on monday, but I think bens confusion here might be valid and possibly we should do one of:
* move these shotover docker examples into a seperate repo
* or rewrite the examples to build from local shotover.

conorbros:
What would be the advantage of those over the current setup?

rukai:
here is the downsides for each approach:

## example-configs contains examples that use ShotoverManager and examples that use release shotover image
* confusing for developers - Easy to accidentally write an integration test that is only run against the latest release.
* confusing for users browsing examples
   + its confusing that some examples in the folder need a shotover binary to be acquired separately while other examples do not.
   + its not clear that some examples can be used on the latest release and some examples are not yet compatible with the latest release.

## example-configs contains examples that use ShotoverManager and examples that use locally built shotover image
* not as useful for users as they will have to change the examples to run them

## example-configs only contains examples that use ShotoverManager, a separate folder or repo is used for examples with release shotover image
* users have to search multiple folders to find examples.

I'm currently thinking the best solution here would be to put docker-compose.yml's that use release shotover images in a separate `example-configs-published-docker/` folder?
By keeping a consistent naming scheme with `example-configs` it should help both developers and users differentiate between these two kinds of examples.

conorbros:
Thanks for that great write-up. To be honest I didn't think it was that confusing but I'm happy to go forward with that recommendation given it's pretty inconsequential.

benbromhead:
keep in mind we can build shotover as a docker image and tag it useing `-t shotover.io/shotover` which will resolve the above path and use the locally built version if you dont have the locally built version it will pull from the repo.

I think part of the confusion as well arises from the fact we use the examples as part of the tests (which is good cause it keeps examples up to date), but also forces us a bit into a corner with how we use them. 

Having said all of the above...I'm happy with splitting into the two different ones (published vs built locally)

conorbros:
BTW, the `base64` is already a dependency.

shotover-proxy/example-configs-docker/redis-cluster/docker-compose.yml
```diff
@@ -0,0 +1,100 @@
+version: '2'
```
rukai, line 1: What is example-configs-docker/redis-cluster for?

conorbros: For the "Unaware client" Redis example, instead of running the Shotover binary yourself, I added the Docker image to the `docker-compose.yml` to be more consistent with the other examples.

I copied the config files for `example-configs/redis-cluster` to the new folder, changed them and left the original ones in the `example-configs` dir because they are used by some tests.

conorbros: For the "Unaware client" Redis example, instead of running the Shotover binary yourself, I added the Docker image to the `docker-compose.yml` to be more consistent with the other examples.

I copied the config files for `example-configs/redis-cluster` to the new folder, changed them and left the original ones in the `example-configs` dir because they are used by some tests.

rukai: Thanks
I'll try and review this all in place.
For next time though, this PR has a lot of stuff going on now and some of it could have been in follow up PRs.

rukai (approved):
LGTM

benbromhead (approved):

docs/src/examples/cassandra-cluster-shotover-sidecar.md
```diff
@@ -12,10 +12,10 @@ In this example, we will be connecting to a Cassandra cluster that has the follo
 
 ### Rewriting the peer ports
 
-Shotover will be deployed as a sidecar to each node in the Cassandra cluster, listening on `9043`. Use the following [docker-compose.yml](https://raw.githubusercontent.com/conorbros/shotover-proxy/cassandra-docs/shotover-proxy/example-configs/cassandra-rewrite-peers/docker-compose.yml) to run the Cassandra cluster and Shotover sidecars. In this example we want to ensure that all our traffic to Cassandra goes through Shotover.
+Shotover will be deployed as a sidecar to each node in the Cassandra cluster, listening on `9043`. Use the following [docker-compose.yml](https://raw.githubusercontent.com/conorbros/shotover-proxy/cassandra-docs/shotover-proxy/example-configs-docker/cassandra-rewrite-peers/docker-compose.yml) to run the Cassandra cluster and Shotover sidecars. In this example we want to ensure that all our traffic to Cassandra goes through Shotover.
```
benbromhead, line 15: for some reason @conorbros your repo link snuck in here

PR #579, opened 2022-03-21
Bump up Cassandra request throttling tests threshold

conorbros:
Bumping up due to intermittent failures on the CI, as demonstrated on https://github.com/shotover/shotover-proxy/pull/577


rukai:
A bit of extra context: We recently bumped the expected count from 25 -> 50, so it makes sense that CI lag spikes would also double so we should double our buffer from 5 -> 10. Hence the resulting 50 + 10 = 60 max value.

rukai (approved):

benbromhead (approved):

PR #580, opened 2022-03-21
Fail to start shotover when configured port already taken

rukai:
Previously if you run two `cargo run` and then run another `cargo run` at the same time, you would get a success exit code and the error `failed to accept connection cause=Address already in use (os error 98) address=127.0.0.1:6379`

With this PR doing the same will give you a failure exit code and the error: `Address already in use (os error 98) address=127.0.0.1:6379`

The original implementation shutting down shotover only occurs when all sources have their port taken, what happens is shotover detects that there are no longer any sources running and performs its natural shutdown procedure.
With this change it will always fail to startup even if only one of many sources have their port taken.

I was motivated to make this change to make integration tests fail in a clearer way when there is a stray shotover process.
But I believe it will also help users to quickly catch mistakes such as running multiple shotover instances or running shotover on the same port as the currently running DB.

### DIFF
--- a/shotover-proxy/src/server.rs
+++ b/shotover-proxy/src/server.rs
@@ -100,7 +100,7 @@ pub struct TcpCodecListener<C: Codec> {
 
 impl<C: Codec + 'static> TcpCodecListener<C> {
     #![allow(clippy::too_many_arguments)]
-    pub fn new(
+    pub async fn new(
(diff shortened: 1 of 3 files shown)


conorbros (approved):
Very welcome change.

benbromhead (approved):

PR #583, opened 2022-03-23
Add DebugForceParse transform

rukai:
This will be useful for evaluating the performance of the cassandra parser.

To get realistic evaluation of the performance impact we would want to run cassandra-passthrough-parse-request/topology.yaml etc on the cassandra_bench example.
The benches here could be useful but at the very least they ensure that these yaml's dont bitrot and that DebugForceParse the transform doesnt explode.

### DIFF
--- a/shotover-proxy/tests/test-configs/cassandra-passthrough-parse-request/topology.yaml
+++ b/shotover-proxy/tests/test-configs/cassandra-passthrough-parse-request/topology.yaml
@@ -1,0 +1,16 @@
+---
+sources:
+  cassandra_prod:
+    Cassandra:
+      listen_addr: "127.0.0.1:9042"
+chain_config:
+  main_chain:
+    - DebugForceParse:
+        parse_requests: true
+        parse_responses: false
+    - CassandraSinkSingle:
+        remote_address: "127.0.0.1:9043"
+named_topics:
+  testtopic: 5
+source_to_chain_mapping:
+  cassandra_prod: main_chain
--- a/shotover-proxy/benches/cassandra_benches.rs
+++ b/shotover-proxy/benches/cassandra_benches.rs
@@ -87,7 +87,53 @@ fn cassandra(c: &mut Criterion) {
         });
         for query in &queries {
             group.bench_with_input(
-                format!("passthrough_{}", query.name),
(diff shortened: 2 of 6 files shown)


conorbros:
Are you intending to release this transform to users?

shotover-proxy/tests/test-configs/cassandra-passthrough-parse-request/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: "3.3"
+services:
+  cassandra-one:
+    image: bitnami/cassandra:3.11
```
conorbros, line 4: Can't we just reuse the `cassandra-passthrough` docker-compoose file

rukai:
I dont intend to make this public at least for now.

I'm pretty sure `#[cfg(test)]` cant be used in integration tests, right?
I dont have any problem with transforms that we dont want in our public API being put behind the alpha feature.

conorbros:
My understanding was the alpha transform feature was so user's could use soon-to-be released transforms early? Anyhow I'm not too fussed about it, just thought I'd ask.

rukai:
Heres some explanation where it was added for reference: https://github.com/shotover/shotover-proxy/pull/494#issue-1119036656
It had to be introduced to allow us to run integration tests on transforms not in the public API.
Users getting access to unreleased transforms via a flag was just a secondary benefit. (or even possibly a negative depending on your point of view)

shotover-proxy/tests/test-configs/cassandra-passthrough-parse-request/topology.yaml
```diff
@@ -0,0 +1,16 @@
+---
```
conorbros, line 1: Could we combine these `topology.yaml` files into the same folder. I don't think they're different enough to warrant separate ones.

rukai: I think the naming makes a lot more sense when we can give each its own folder and call the file topology.yaml
Otherwise you cant tell from the filename if its a toplogy file or docker-compose file :cry:

conorbros (approved):

benbromhead (approved):

PR #584, opened 2022-03-24
Move Cassandra test helpers to helpers module

conorbros:


shotover-proxy/tests/helpers/mod.rs
```diff
@@ -133,15 +136,7 @@ impl ShotoverManager {
 
     #[allow(unused)]
     pub fn cassandra_connection(&self, contact_points: &str, port: u16) -> Session {
```
rukai, line 138: we could `pub use cassandra::cassandra_connection` or even just change all the calls to use `helpers::cassandra::cassandra_connection(..)`, this PR doesnt have much else going on so the extra noise would be fine.

rukai (approved):
nice!

XA21X (approved):

PR #585, opened 2022-03-24
Cleanup and improve assertions of ShotoverProcess

rukai:
Changes:
* The initial cargo build was swapped to `run_command` as that will print output on failure.
* `.env("RUST_LOG", "debug,shotover_proxy=debug")` was removed as it did nothing (didnt bother investigating why) and it is incorrect (we should be logging at info level not debug level)
* `wait()` now returns WaitOutput so we dont have to guess what tuple value contains what.
* implemented AssertExitsWith which will trigger the requested assert when ShotoverProcess is dropped.

### DIFF
--- a/shotover-proxy/examples/cassandra_bench.rs
+++ b/shotover-proxy/examples/cassandra_bench.rs
@@ -1,6 +1,6 @@
 use clap::Parser;
 use test_helpers::docker_compose::DockerCompose;
-use test_helpers::shotover_process::ShotoverProcess;
+use test_helpers::shotover_process::{AssertExitsWith, ShotoverProcess};
 
 /// e.g.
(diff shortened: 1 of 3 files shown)


conorbros (approved):

benbromhead (approved):
lgtm

rukai:
I realized the API I made here was really confusing.

The AssertExitsWith logic isnt run if wait is called.
Its also not intuitive to setup the AssertExitsWith logic in the constructor.

So I instead I rewrote it such that the Drop impl always panics if some kind of wait was not called.

At the very least @conorbros should rereview the PR.
@benbromhead can also take a look if he wants

benbromhead:
@conorbros can relook at it thanks

conorbros (approved):

PR #587, opened 2022-03-25
Accept base64 string instead of individual bytes in protect transform config

### LINKED ISSUE #510: Read Protect key in config as an encoded string
Read the key as an encoded string (base64 or hex) and then convert into the bytes format that `sodiumoxide` expects. 

This makes it more convenient for the user as they won't have to provide the key byte by byte. 

_Originally posted by @conorbros in https://github.com/shotover/shotover-proxy/pull/494#discussion_r800287726_

conorbros:
Closes #510


shotover-proxy/example-configs/cassandra-protect-local/topology.yaml
```diff
@@ -8,39 +8,7 @@ chain_config:
...
-              - 0x2c
-              - 0x03
-              - 0xce
-              - 0x21
-              - 0x49
-              - 0x43
-              - 0x02
-              - 0x9b
-              - 0x59
+            kek: 0x1e0xdf0x0c0xd60x700xce0xff0xb70xda0xcb0xe70x1f0xb70xbd0x4c0xd90x7c0xbb0x8f0x7d0x040x9c0xb00x2c0x030xce0x210x490x430x020x9b0x59
```
rukai, line 11: Is this some kind of standard I dont know about?
Wouldnt this be a lot nicer if each byte was exactly two characters and without the 'x's?

conorbros: It's the same format as before

rukai: ok ... well no need to stick to it if we can come up with something better.

conorbros: I assumed that you get the key from AWS in that format.

rukai: > I assumed that you get the key from AWS in that format, I'll see if I can find out what format it is and then we can accept that.

The current format isnt based on anything, its literally just deserializing directly into a vec of bytes.
So I agree, investigating what format these bytes actually come in is a great way forward!

> That said it's not uncommon for hex to have the leading 0x
Leading `0x` is good!
I wouldnt want that removed.

conorbros: >The current format isnt based on anything

Ah that makes sense, I assumed someone had made a demo key in AWS to use. They might even come in a different format (such as Base64) so it would be good to know.

conorbros: After a bit of research I think if we can make it accept hex and base64, that's pretty much all bases covered.

rukai: I would write out a design that we can go over before you get too stuck into implementation problems. Could save you some time.

XA21X: > you get the key from AWS in that format

With KMS, you're typically never exposed to the key material itself, you simply ask AWS to encrypt/decrypt things with the key on your behalf. There is the option of uploading your own key material, but I've never seen the latter used. Re: format, AWS just lets you specify the path to the (encrypted) binary file. I was expecting the same for us - path to key file, rather than embedding the key in the config. Although our approach is more convenient. :)

> accept hex and base64, that's pretty much all bases covered

Yeah, that's a reasonable combination. Although base64 is standard, but the hex format arguably is not? What tool do we use to generate these keys, and how do we get it into the expected format? Should each byte be prefixed with `0x`, or is it enough to have only one at the start? Do we specify the format with another config, or is it auto-detected?

benbromhead:
Yup agreed something like base64 would be good in this case

conorbros:
Just base64 orr the ability to accept both base64 and hex?

rukai:
I havent looked closely at this but in general I would prefer to only support a single format, otherwise we have to maintain both formats forever.

benbromhead:
i think base64 is probably the way to go... please double check how you get keys out of AWS + whether openssl supports key export in that format (as the swiss army knife of working with crypto) and if that looks right lets go with that

XA21X:
> i think base64 is probably the way to go... please double check how you get keys out of AWS + whether openssl supports key export in that format (as the swiss army knife of working with crypto) and if that looks right lets go with that

AWS uses base64 in places where binary won't work: (needs to be printed, or in JSON)
https://docs.aws.amazon.com/kms/latest/APIReference/API_GenerateDataKeyPair.html
> When you use the HTTP API or the AWS CLI, the value is Base64-encoded. Otherwise, it is not Base64-encoded.

conorbros:
BTW, the `base64` crate is already a dependency.

rukai (approved):
alright, thanks for the investigation @XA21X 
LGTM

XA21X (approved):
LGTM

Issue #538, opened 2022-02-20
Remove checked in certs from the repo


conorbros:
_Originally posted by @benbromhead in https://github.com/shotover/shotover-proxy/issues/437#issuecomment-1043810289_

PR #588, opened 2022-03-29
Generate redis TLS certs at test time

rukai:
This makes progress towards https://github.com/shotover/shotover-proxy/issues/538
To be done in follow up PRs:
* rename redis-tls/tls_keys -> redis-tls/certs to match the naming scheme used by cassandra
* generate cassandra certs at test time
But they are low priority, I was only working on this due to being blocked on other work.


shotover-proxy/benches/redis_benches.rs
```diff
@@ -49,6 +50,9 @@ fn redis(c: &mut Criterion) {
     group.bench_with_input(
         "passthrough",
         || {
+            test_helpers::cert::generate_test_certs(Path::new(
+                "example-configs/redis-tls/tls_keys",
+            ));
```
conorbros, line 55: Why generate them in this bench? `redis-passthrough` doesn't need keys.

rukai: good catch

conorbros (approved):
Just that one comment, LGTM.

PR #589, opened 2022-03-29
store single cql statement instead of a vec

rukai:
Makes progress towards https://github.com/shotover/shotover-proxy/issues/563
Also makes progress towards using the new cql parser as that will return only a single statement.

A cassandra query can only contain a single statement, the only reason it was a vec here was because that was what the sqlparser crate returned.


conorbros:
Why is this progress towards #563?

rukai:
ah I guess its not technically progress but ... its an obvious refactor that I spotted while working on https://github.com/shotover/shotover-proxy/pull/570

conorbros (approved):
Sounds good to me

benbromhead (approved):
yay!

PR #595, opened 2022-04-04
CassandraOperation::queries returns batch queries

rukai:
closes https://github.com/shotover/shotover-proxy/issues/563

This PR allows transforms like redis cache and protect transform to work on batch messages.
This however only works for batch messages sent over the BATCH protocol, batch messages sent over QUERY are still broken.

You can verify the integration test tests the new functionality by disabling the new `CassandraOperation::Batch(batch)` handling and observing the test fail.

### DIFF
--- a/shotover-proxy/src/frame/cassandra.rs
+++ b/shotover-proxy/src/frame/cassandra.rs
@@ -308,6 +308,7 @@ pub enum CassandraOperation {
 }
 
 pub enum QueryIterator<'a> {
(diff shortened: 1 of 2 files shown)


conorbros:
Will this work for the peers rewrite transform?

rukai:
batches only apply to UPDATE/INSERT/DELETE and the peers rewrite transform only operates on SELECT, so peers rewrite never needs to touch batches anyway.

conorbros:
True, forgot about that.

conorbros (approved):

benbromhead (approved):

PR #596, opened 2022-04-04
Store Wrapper.transforms as an IterMut instead of a cloned Vec

rukai:
I noticed this when investigating redis pubsub support.

This change is more semantically accurate and avoids an extra allocation per request.


conorbros (approved):

benbromhead (approved):
Nice!

PR #599, opened 2022-04-05
Remove entry from deny.toml & cargo update

conorbros:
https://github.com/hsivonen/encoding_rs/issues/75 has been closed so we can remove `encoding_rs` from `deny.toml`


rukai:
If you set regex to 1.5.5 it should clear the "possible security issue" warning from deps.rs
This won't affect dependency resolution as we already use 1.5.5 as per the cargo.lock but deps.rs is silly and only checks the cargo.toml

rukai (approved):
nice!

benbromhead (approved):

PR #600, opened 2022-04-06
Reenable parallelization of packet parsing in packet tests

rukai:
Ended up down this rabbit hole because I went to remove the num_cpus dep as its now built into rust 1.59, however discovered that we didnt need it at all in the first place.

* Removes commented out code + dependencies used by commented out code
* Reduces test runtime by 10s by parallelizing the workload, theres still 10s of runtime left that could be optimized but I dont have high confidence in the approach of this test in the first place so i'll leave as is.
* Simplifies implementation, rayon is much simpler to use than raw Threadpools

### DIFF
--- a/shotover-proxy/tests/codec/util/packet_capture.rs
+++ b/shotover-proxy/tests/codec/util/packet_capture.rs
@@ -1,11 +1,9 @@
-use crate::codec::util::packet_parse::{PacketHeader, PacketParse, ParsedPacket};
-
+use crate::codec::util::packet_parse::{OwnedPcapPacket, PacketHeader, PacketParse, ParsedPacket};
 use anyhow::Result;
 use pcap::{Active, Capture, Device};
+use rayon::prelude::*;
 use std::net::IpAddr;
 use std::path::Path;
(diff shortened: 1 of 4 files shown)


shotover-proxy/tests/codec/util/packet_capture.rs
```diff
@@ -135,41 +127,23 @@ impl PacketCapture {
         file_name: &Path,
         filter: Option<String>,
     ) -> Vec<Result<ParsedPacket, String>> {
-        // TODO: Fix flakiness from out-of-order futures.
-        // let pool = ThreadPool::new(num_cpus::get() * 2);
-        let pool = ThreadPool::new(1);
```
conorbros, line 140: Is there anything more to the simplification than getting to use `par_iter()` instead of creating the threadpool and using `pool.execute`?

rukai: we also ditch the Arc<Mutex<T>>

conorbros (approved):

benbromhead (approved):

PR #602, opened 2022-04-06
Fix redis tls bench to actually use tls

rukai:
* The non TLS source in redis-tls is removed, I have no idea what its purpose was.
* The TLS source is given the default port 6379
* The redis tls integration test is changed to use 6379
* The redis tls bench is left as 6379 so it now correctly uses TLS

### DIFF
--- a/shotover-proxy/example-configs/redis-tls/topology.yaml
+++ b/shotover-proxy/example-configs/redis-tls/topology.yaml
@@ -1,11 +1,8 @@
 ---
 sources:
-  redis_prod:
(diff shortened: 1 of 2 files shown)


rukai:
oh I see whats going on, the benches are currently testing shotover <-> db TLS but not client <-> shotover TLS.
While I think the benches should be changed to include TLS both ways, it will require complex redis-rs changes so lets just leave it for now.

I originally thought this issue was blocking the pubsub work but I've now realized it isnt actually.

PR #603, opened 2022-04-07
Update to rust 1.60

rukai:
https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.58"
+channel = "1.60"
 components = [ "rustfmt", "clippy" ]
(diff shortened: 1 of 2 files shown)


conorbros (approved):

rukai:
I tested that bench locally and it does seem to be giving a ~5-10% regression locally.
While I dont think its worth blocking this PR, it might be worth investigating?
I'm leaning towards not investigating though, rust 1.60 includes an upgrade to LLVM 14 so we probably just got unlucky with the chaotic nature of LLVM optimization.

conorbros:
Curious it's only one benchmark regressing.

rukai:
Its quite likely lots of benchmarks are affected but this was the only one that got combined with enough CI performance variance to tip past the 30% reporting threshold

benbromhead:
Do you want to just confirm the other benchmarks. We can wait till the next rust ver if we are seeing regressions across the board.

rukai:
Running every bench once there is various +-5% results which is easily within expected noise.
So im not seeing any trend up or down.
I think this is fine to merge.

benbromhead (approved):

PR #604, opened 2022-04-08
Bench logging changes

conorbros:
This change makes the benches a bit more readable, at the moment they are full of logs which makes it hard to read the results.

- Changed the `"failed to fetch from cache"` `tracing::error` in `RedisCache` to a warning. It makes the logs pretty noisy.
- Added a new config file to use during benches.
- Added a new method to `ShotoverManager` for starting with a custom config file.

### DIFF
--- a/shotover-proxy/benches/cassandra_benches.rs
+++ b/shotover-proxy/benches/cassandra_benches.rs
@@ -275,10 +275,7 @@ pub struct BenchResources {
 impl BenchResources {
     fn new(shotover_topology: &str, compose_file: &str) -> Self {
         let compose = DockerCompose::new(compose_file);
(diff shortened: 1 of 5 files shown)


rukai:
> Changed the "failed to fetch from cache" tracing::error in RedisCache to a warning. It makes the logs pretty noisy.

Yeah sorry about that, I was intending to quickly make use of those errors to fix the problems. But then I wasnt able to work on RedisCache for a while.
I'm quite happy for that to become an `info!`

> Added a new config file to use during benches.

I vote to make this apply to both tests and benches.
I also think it would make sense to use warn error level.

benbromhead (approved):

PR #607, opened 2022-04-11
Only record metrics if using the observability interface

conorbros:
## Changes

As part of benchmarking and profiling I noticed that we spend a lot of time running metrics code. This can be seen in the flamegraph by searching for "metrics". You may to download the file and open it in your browser from your local files for the search function to work properly.

with_cassandra_peers_write
 
While there is other optimisations we could make (exploring how registering 
 and using the handles instead of macros affects performance) I think this optimisation is a good start. 

## Benchmarks

Below is the results of running the latte benchmark example with metrics enabled and disabled. The commands are:
- With observability: `cargo run --release --example cassandra_bench -- --topology-dir example-configs/cassandra-passthrough -r 1000`
- Without observability: `cargo run --release --example cassandra_bench -- --topology-dir example-configs/cassandra-passthrough -r 1000 --config-path tests/helpers/config_without_observability.yaml`

### Without 

image

### With

image


rukai:
The cost is on registering right?
Is there register calls that need to be moved out of hot code?

conorbros:
I'm intending to optimise metrics when enabled as well but this PR is just an easy performance win for cases where they definitely don't want metrics.

If there's a problem with the implementation, I can investigate how fast I can get it while enabled and if it ends up being fast enough we could just leave it on all the time.

conorbros:
Closing as stale, opened an issue to revisit in the future.

Issue #605, opened 2022-04-11
Redis - CLIENT SETNAME behaves incorrectly


johndelcastillo:
## Describe the bug
The CLIENT SETNAME command it sent to all nodes, and the response is concatenated to the client, which breaks some client libraries. 
Jedis at this stage is confirmed to not be working.

## To Reproduce
shotover-proxy.0afaa9ad7f664117babac37425a07b0d.cu.dev.instaclustr.com:6378> CLIENT SETNAME hello
1) OK
2) OK
3) OK

## Configuration
DR Chain is set, but I can't confirm if it matters

### Topology
3 node DC + 3 node dc in DR

```yaml
sources:
  redis_prod:
    Redis: {batch_size_hint: 4, listen_addr: '0.0.0.0:6378', connection_limit: 20000,
      hard_connection_limit: true}
chain_config:
  redis_chain:
  - Tee:
      behavior: Ignore
      buffer_size: 10000
      chain:
      - QueryTypeFilter: {filter: Read}
      - Coalesce:
          max_behavior:
            CountOrWait: [2000, 100]
      - Tee:
          buffer_size: 100
          timeout_micros: 10000
          chain:
          - QueryCounter: {name: DR_chain}
          - RedisSinkCluster:
              first_contact_points: ['44.232.82.142:6379']
      - 'Null'
  - QueryCounter: {name: Main_chain}
  - RedisClusterPortsRewrite: {new_port: 6378}
  - RedisSinkCluster:
      first_contact_points: ['3.216.108.90:6379']
source_to_chain_mapping: {redis_prod: redis_chain}
```

Please include any configuration for setting up services such as `docker-compose` that the `topology.yaml` depends on.

## Expected behaviour
3.216.108.90:6379> client setname goo
OK

## Systems and Version:
 - OS - Debian buster
 - Shotover Version - 0.1.2
 - Redis 6.0.9

rukai:
I've reproduced this in an integration test.
The issue is present in SET messages so its impressive this made it to production.

Issue #520, opened 2022-02-14
Mixing message types in a single transform


rukai:
In general transforms are supposed to handle any message type but there are some transforms that only make sense with one message type e.g. CassandraPeersRewrite and would indicate a user configuration error if they encountered another message type.
To improve UX we could add a system to our validation system that allows transforms to define what message types they are compatible with.
Then we can fail to start shotover on a bad configuration allowing users to catch config issues earlier and with better diagnostics.

The transform invariants currently don't specify whether a transform instance will always receive messages of the same type.
If we cant think of a use case for multiple message types going through the same transform then we should:
*    explicitly specify in the invariants that transforms will always receive messages of the same type and that transforms should never allow messages of mixed types to flow down the chain. (They can completely replace messages of type A with type B though)
*    implement the allowed message type configuration validation
    
Or we could go the other way and explicitly allow transforms to receive and send multiple message types.
Then we could have a sushi belt of messages and transforms could filter and split by message type.

Which choice we make has all sorts of design implications and we dont need an answer right now, so we should come back to this problem later, I just wanted to write out my thoughts on the problem.

rukai:
Resolved by https://github.com/shotover/shotover-proxy/pull/1601
A little more work could be done in this area to do with improving diagnostics and removing redundant is_terminating method but that can be tracked separately.

PR #608, opened 2022-04-11
Fix RedisSinkCluster returning an array of responses instead of a single response

rukai:
This adds an integration test case `test_client_name` that reproduces the issue explained in #605
This PR also adds asserts on the return value of every write only command.
Doing so discovered that this bug also affects the FLUSHDB command.
In fact it should affect every command listed in here https://github.com/shotover/shotover-proxy/blob/1a0f731ea01423d919f588b8c5bc7d5aefa1f575/shotover-proxy/src/transforms/redis/sink_cluster.rs#L552+#L555

Im not sure if these assert_int/assert_ok helpers are the best approach, maybe I should have used the connection.get/set helpers combined with assert_eq instead.
But oh well we can always swap to that later.

The PR includes a fix that fixes the issue but I have set the PR as draft because I still want to investigate if there is a more appropriate fix.

### DIFF
--- a/shotover-proxy/src/transforms/redis/sink_cluster.rs
+++ b/shotover-proxy/src/transforms/redis/sink_cluster.rs
@@ -130,6 +130,7 @@ impl RedisSinkCluster {
         };
 
         Ok(match channels.len() {
+            // Return an error as we cant send anything if there are no channels.
             0 => {
                 let (one_tx, one_rx) = immediate_responder();
                 match self.connection_error {
@@ -140,11 +141,15 @@ impl RedisSinkCluster {
                 };
                 Box::pin(one_rx)
             }
+            // Send to the single channel and return its response.
             1 => {
                 let channel = channels.get(0).unwrap();
                 let one_rx = self.choose_and_send(channel, message).await?;
                 Box::pin(one_rx.map_err(|_| anyhow!("no response from single channel")))
             }
+            // Send to all senders.
+            // If any of the responses were a failure then return that failure.
+            // Otherwise return the first successful result
             _ => {
                 let responses = FuturesUnordered::new();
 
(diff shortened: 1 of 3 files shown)


benbromhead:
For some context here, this was done intentionally as some of those commands would require the client to know about the result of each command and the lazy approach was to "leak" some of the clusterness to the clients. Some clients were ok with this behaviour.

Going forward and to complete the cluster hiding nature of this transform. The approach should probably be something like if one of the results is an error, return the error otherwise return the top result (e.g. the "OK").

For commands where rerunning a succesful command is not idempotent, then we probably need to at least return a customized error, or better work out how to support it (or not support it!).

E.g. rerunning FLUSHDB against a redis node that already ran it succesfully will be fine.

benbromhead:
You might find this interesting. RedisLabs has a redis proxy that behaves in a similar manner to some parts of shotover. The list of commands they support can be found here:

https://github.com/RedisLabs/redis-cluster-proxy/blob/unstable/COMMANDS.md

rukai:
Marking as no longer a draft, as I think this change is now correct.
However I am not considering this as closing https://github.com/shotover/shotover-proxy/issues/605 anymore, I believe we need to do a large rewrite of the command mapping done at `for_command_frame` first.

shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -140,48 +141,52 @@ impl RedisSinkCluster {
...
-                                    |mut message| match message.frame().unwrap() {
-                                        Frame::Redis(frame) => frame.take(),
-                                        _ => unreachable!(),
-                                    },
-                                )),
-                                Ok(Response {
-                                    response: Err(e), ..
-                                }) => acc.push(RedisFrame::Error(e.to_string().into())),
-                                Err(e) => acc.push(RedisFrame::Error(e.to_string().into())),
+                    let response = responses
```
conorbros, line 161: Can we add a comment for what this is doing?

shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -140,48 +141,52 @@ impl RedisSinkCluster {
...
+                            } else {
+                                match response {
+                                    Ok(Response {
+                                        response: Ok(mut messages),
+                                        ..
+                                    }) => Some(messages.pop().map_or(
+                                        RedisFrame::Null,
+                                        |mut message| match message.frame().unwrap() {
+                                            Frame::Redis(frame) => frame.take(),
+                                            _ => unreachable!(),
```
conorbros, line 174: Can we add a message explaining why this is unreachable?

conorbros (approved):
Just that one small thing, otherwise LGTM.

shotover-proxy/src/transforms/redis/sink_cluster.rs
rukai: lol, its not unreachable.
While its bad code, its existing bad code.

But I think we are better off leaving it as is until we start resolving https://github.com/shotover/shotover-proxy/issues/520

conorbros: Ah, that is confusing! Sounds good to me.

XA21X: When would we get a non-Redis frame as a response here? i.e. how exactly do we 'reach' here? 🤔

rukai: User hooks up a CassandraSource to a RedisSink

XA21X (approved):

shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -140,48 +141,52 @@ impl RedisSinkCluster {
...
-                                    |mut message| match message.frame().unwrap() {
-                                        Frame::Redis(frame) => frame.take(),
-                                        _ => unreachable!(),
-                                    },
-                                )),
-                                Ok(Response {
-                                    response: Err(e), ..
-                                }) => acc.push(RedisFrame::Error(e.to_string().into())),
-                                Err(e) => acc.push(RedisFrame::Error(e.to_string().into())),
+                        .fold(None, |acc, response| async move {
```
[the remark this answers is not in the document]
rukai: My thinking is that in the follow up work any commands that are expected to return unique values will be marked as unusable and shotover will return an error instead of attempting to handle them. e.g. https://redis.io/commands/time/
but this will be determined on a case-by-case basis, for some commands it might be worth pretending that they return the same value.

shotover-proxy/src/transforms/redis/sink_cluster.rs
rukai: Ah!
Thankyou for thinking about this properly
You are 100% correct.

I added a proper message to the unreachable explaining this.
It is actually unreachable because the codec will only ever return a Redis message.

PR #609, opened 2022-04-12
Don't re-register counter on cassandra sink clone

conorbros:
A simple metrics improvement. The previous implementation was re-registering the metric on each new connection instead of just cloning the existing handle. 

### This PR
image

### Main
image

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/sink_single.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_single.rs
@@ -48,11 +48,13 @@ pub struct CassandraSinkSingle {
 
 impl Clone for CassandraSinkSingle {
     fn clone(&self) -> Self {
-        CassandraSinkSingle::new(
-            self.address.clone(),
-            self.chain_name.clone(),
-            self.tls.clone(),
-        )
+        CassandraSinkSingle {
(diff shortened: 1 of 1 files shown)


rukai:
Is the bench results trying to demonstrate something?
It just looks like noise to me.

The actual change looks great though.
Internally Counter is just an `Option<Arc<T>>`, so cloning will be cheaper than creating a new one.

rukai (approved):

conorbros:
50th percentile results go from -74% to -67%.

benbromhead (approved):

rukai:
The only other transform needing this is RedisSinkSingle, you can add it to this PR or (considering that we already got 2 reviews) you could do it in a follow up.

conorbros:
I'll put it in this PR.

shotover-proxy/src/transforms/chain.rs
```diff
@@ -190,6 +190,7 @@ impl TransformChain {
     }
 
     pub fn new(transform_list: Vec<Transforms>, name: String) -> Self {
+        tracing::info!("registering here");
```
rukai, line 193: oops?

rukai (approved):

PR #610, opened 2022-04-12
Simplify eval routing

rukai:
* Removes the TODO: It seems the author got confused and thought redis was supposed to use `RedisFrame::Integer` within commands.
However the spec clearly says that commands only consist of `RedisFrame::BulkString`: https://redis.io/docs/reference/protocol-spec/#resp-protocol-description
* Removes the `RedisFrame::Integer` handling
* One less allocation (`&str` instead of `String`)


conorbros (approved):

XA21X (approved):
Good catch on the BulkString/Integer handling. I didn't realise what the comment meant.

PR #611, opened 2022-04-12
Fix invalid command routes

rukai:
"CLIENT LIST",  "CLIENT SETNAME" and "SCRIPT KILL" would never match because a command containing a " " will never match here as commands are already split by spaces.
* "SCRIPT KILL" is reimplemented with the intended mapping
* "CLIENT LIST" is removed, it will continue to receive RoutingInfo::AllNodes as intended
* "CLIENT SETNAME" is removed, it will continue to receive RoutingInfo::AllNodes even though this wasnt the intention.

I'll do another pass over these mappings in a follow up PR but I wanted to remove these nonsense cases first.

### DIFF
--- a/shotover-proxy/src/transforms/redis/sink_cluster.rs
+++ b/shotover-proxy/src/transforms/redis/sink_cluster.rs
@@ -549,12 +549,16 @@ impl RoutingInfo {
         };
 
         Ok(match command_name.as_slice() {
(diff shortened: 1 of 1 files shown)


conorbros (approved):

XA21X (approved):
Could do with a unit test that covers these special cases, but otherwise LGTM.

PR #612, opened 2022-04-13
Rewrite redis_int_tests to assert on write only responses

rukai:
Pulls the int test changes out of https://github.com/shotover/shotover-proxy/pull/608 so that we can land this ASAP to avoid merge conflicts.
Naturally I had to disable the assert in assert_ok as this PR doesnt include the bug fix, but that will be enabled when we land the fix.

### DIFF
--- a/shotover-proxy/tests/redis_int_tests/basic_driver_tests.rs
+++ b/shotover-proxy/tests/redis_int_tests/basic_driver_tests.rs
@@ -1,9 +1,10 @@
 use crate::helpers::ShotoverManager;
+use crate::redis_int_tests::assert::*;
 use rand::{thread_rng, Rng};
 use rand_distr::Alphanumeric;
 use redis::aio::Connection;
 use redis::cluster::ClusterConnection;
-use redis::{AsyncCommands, Cmd, Commands, ErrorKind, RedisError, Value};
+use redis::{AsyncCommands, Commands, ErrorKind, RedisError, Value};
 use serial_test::serial;
 use shotover_proxy::tls::TlsConfig;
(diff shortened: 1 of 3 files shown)


shotover-proxy/tests/redis_int_tests/basic_driver_tests.rs
```diff
@@ -21,18 +21,30 @@ const STRESS_TEST_MULTIPLIER: usize = 1;
 #[cfg(not(debug_assertions))]
 const STRESS_TEST_MULTIPLIER: usize = 100;
 
+async fn assert_nil(cmd: &mut Cmd, connection: &mut Connection) {
```
conorbros, line 24: Can these helpers live in the mod.rs file for this module?

rukai: sure, that sounds reasonable

rukai: Ended up putting it in its own submodule, that tends to be the preferred way in rust.

conorbros: LGTM. If there's other existing helpers in for redis that could live there I'm fine with including that in this PR if you want to move them. Otherwise lets move forward with this.

conorbros (approved):

shotover-proxy/tests/redis_int_tests/basic_driver_tests.rs
rukai: nah, theres nothing else in there I would consider a helper

benbromhead (approved):

PR #614, opened 2022-04-20
Add edition to rustfmt.toml

conorbros:
Fixes an issue with my IDE.


rukai (approved):
https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#edition
Looks like in some cases it falls back to this config if I cant find it in the Cargo.toml
And by default its 2015 so we need to set to 2021

XA21X (approved):

PR #616, opened 2022-04-22
Remove Cassandra certs from repo

### LINKED ISSUE #538: Remove checked in certs from the repo

conorbros:
closes #538


conorbros:
Alright, fixed the failing test. One thing to remember from when this PR is merged is that if a test relies on the current directory being something particular it needs to be marked with `#[serial]`.

test-helpers/src/cert.rs
```diff
@@ -34,3 +37,17 @@ pub fn generate_test_certs(path: &Path) {
     .unwrap();
     std::fs::write(path.join("redis.key"), cert.serialize_private_key_pem()).unwrap();
 }
+
+pub fn generate_cassandra_test_certs() {
+    let current_path = current_dir().unwrap();
+    set_current_dir("example-configs/cassandra-tls/certs").unwrap();
+
+    if let Err(ErrorKind::NotFound) = Command::new("./gen_certs.sh")
```
rukai, line 45: Shouldnt we fail on all errors?

test-helpers/src/cert.rs
```diff
@@ -34,3 +37,17 @@ pub fn generate_test_certs(path: &Path) {
     .unwrap();
     std::fs::write(path.join("redis.key"), cert.serialize_private_key_pem()).unwrap();
 }
+
+pub fn generate_cassandra_test_certs() {
+    let current_path = current_dir().unwrap();
+    set_current_dir("example-configs/cassandra-tls/certs").unwrap();
```
rukai, line 43: If we panic in the middle that will leave other tests with a different working directory right?
Can this be written such that it doesnt change the directory?
Maybe the script should be responsible for settings its own directory?

TBH I would rather we just rewrite the script entirely using rust, I would need some convincing that the existing `generate_test_certs` can not be adapted for cassandra.
That said this stuff can be a bit complicated so im willing to compromise here for the sake of moving things forward.

shotover-proxy/example-configs/cassandra-tls/certs/gen_certs.sh
```diff
@@ -1,17 +1,63 @@
...
-
 # generate keystore
 openssl pkcs12 -export -out keystore.p12 -inkey localhost.key -in localhost.crt -passout pass:password
 keytool -importkeystore -destkeystore keystore.jks -srcstoretype PKCS12 -srckeystore keystore.p12 -deststorepass "password" -srcstorepass "password"
 
 # generate truststore
 openssl pkcs12 -export -out truststore.p12 -inkey localhost.key -in localhost.crt -passout pass:password
 keytool -importkeystore -destkeystore truststore.jks -srcstoretype PKCS12 -srckeystore truststore.p12 -deststorepass "password" -srcstorepass "password"
+
+sudo chown -R 1001 keystore.p12
```
rukai, line 60: We should be able to just set the appropriate permission bits via chmod which wont need root.

rukai (approved):
Would love to see this using the cert generation rewritten in rust one day for maintainability and cross platform reasons but as is this PR is an improvement over the status quo.

LGTM

.gitignore
```diff
@@ -7,6 +7,12 @@ tests/soaktest-*
 *.gz
 /.vscode
 /shotover-proxy/example-configs/redis-tls/certs
+/shotover-proxy/example-configs/cassandra-tls/certs/*.jks
+/shotover-proxy/example-configs/cassandra-tls/certs/*.p12
+/shotover-proxy/example-configs/cassandra-tls/certs/*.key
+/shotover-proxy/example-configs/cassandra-tls/certs/*.srl
+/shotover-proxy/example-configs/cassandra-tls/certs/*.csr
+/shotover-proxy/example-configs/cassandra-tls/certs/*.crt
```
XA21X, line 15: Any reason to use specific file extensions instead of excluding the whole folder like for Redis?

Doesn't matter, just curious.

XA21X: Nvm, I missed the fact that the script is in the same folder.

XA21X (approved):

PR #618, opened 2022-04-26
Implement redis cluster sink ResponseJoin abstraction

rukai:
* A `ResponseJoin` enum is added to `RoutingInfo::AllMasters`.
   + The ResponseJoin defines how the redis responses should be joined into a single response for returning to the client.
   + Currently it only has `First` and `ArrayJoin` but I have also identified a need for the following to be added later on:
       - `ResponseJoin::IntegerSum` - needed for DBSIZE
       - `ResponseJoin::EarliestDate` - needed for LASTSAVE
       - There is likely more undiscovered cases
   + Other `RoutingInfo` variants dont need to specify a `ResponseJoin`, even `AllNodes` doesnt need it as `AllNodes` should only really be used by commands like SAVE/BGSAVE.
* The KEYS command is now properly handled by the cluster sink.
   + This isn't a high priority feature but it was the easiest way to demonstrate this abstraction will work.
* `get_channels` now returns a `RoutingInfo::Unsupported` instead of a None, this makes it more obvious what the None case means. In the future we can even further split up `RoutingInfo::Unsupported` to indicate different error cases.
* unreachable! message was shortened because rustfmt wasnt formatting it properly

The only functional change caused by this PR should be that the KEYS command now correctly returns every key across the entire cluster.

Still need another follow up PR after this to fix a bunch of incorrect mappings.

### DIFF
--- a/shotover-proxy/src/transforms/redis/sink_cluster.rs
+++ b/shotover-proxy/src/transforms/redis/sink_cluster.rs
@@ -589,20 +589,21 @@ impl RoutingInfo {
                 RoutingInfo::Unsupported
             }
             b"EVALSHA" | b"EVAL" => match args.get(2) {
-                Some(RedisFrame::BulkString(key_count)) => std::str::from_utf8(key_count)
-                    .unwrap_or("0")
-                    .parse::<u64>()
-                    .ok(),
-                _ => None,
-            }
-            .and_then(|key_count| {
-                if key_count == 0 {
-                    Some(RoutingInfo::Random)
-                } else {
-                    args.get(3).and_then(RoutingInfo::for_key)
+                Some(RedisFrame::BulkString(key_count)) => {
+                    if key_count.as_ref() == b"0" {
+                        RoutingInfo::Random
+                    } else {
+                        // For an EVAL to succeed every key must be present on the same node.
+                        // So we use the first key to find the correct destination node and if any of the
+                        // remaining keys are not on the same node as the first key then the destination
+                        // node will handle will correctly handle this for us.
+                        args.get(3)
+                            .and_then(RoutingInfo::for_key)
+                            .unwrap_or(RoutingInfo::Unsupported)
+                    }
                 }
-            })
-            .unwrap_or(RoutingInfo::Unsupported),
+                _ => RoutingInfo::Unsupported,
+            },
             b"XGROUP" | b"XINFO" => args
                 .get(2)
                 .and_then(RoutingInfo::for_key)
--- a/shotover-proxy/src/transforms/redis/sink_cluster.rs
+++ b/shotover-proxy/src/transforms/redis/sink_cluster.rs
@@ -122,7 +122,8 @@ impl RedisSinkCluster {
             message => bail!("syntax error: bad command: {message:?}"),
         };
 
-        let channels = match self.get_channels(command).await? {
+        let routing_info = RoutingInfo::for_command_frame(command)?;
+        let channels = match self.get_channels(routing_info) {
             ChannelResult::Channels(channels) => channels,
             ChannelResult::Other(Command::AUTH) => {
                 return self.on_auth(command).await;
@@ -156,7 +157,6 @@ impl RedisSinkCluster {
                 for channel in channels {
                     responses.push(self.choose_and_send(&channel, message.clone()).await?);
                 }
-
                 Box::pin(async move {
                     let response = responses
                         .fold(None, |acc, response| async move {
@@ -170,8 +170,16 @@ impl RedisSinkCluster {
                                     }) => Some(messages.pop().map_or(
                                         RedisFrame::Null,
                                         |mut message| match message.frame().unwrap() {
-                                            Frame::Redis(frame) => frame.take(),
-                                            _ => unreachable!("This is a direct response from a redis sink so it must be a redis message"),
+                                            Frame::Redis(frame) => {
+                                                let new_frame = frame.take();
+                                                match acc {
+                                                    Some(prev_frame) => routing_info
+                                                        .response_join()
+                                                        .join(prev_frame, new_frame),
+                                                    None => new_frame,
+                                                }
+                                            }
+                                            _ => unreachable!("direct response from a redis sink"),
                                         },
                                     )),
                                     Ok(Response {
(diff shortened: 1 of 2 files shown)


shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -547,23 +565,29 @@ enum ChannelResult {
...
+                }
+                _ => RoutingInfo::AllMasters(ResponseJoin::First),
             },
-            b"SCAN" | b"SHUTDOWN" | b"SLAVEOF" | b"REPLICAOF" | b"MOVE" | b"BITOP" => None,
+            b"SCAN" | b"SHUTDOWN" | b"SLAVEOF" | b"REPLICAOF" | b"MOVE" | b"BITOP" => {
+                RoutingInfo::Unsupported
+            }
             b"EVALSHA" | b"EVAL" => match args.get(2) {
                 Some(RedisFrame::BulkString(key_count)) => std::str::from_utf8(key_count)
                     .unwrap_or("0")
```
conorbros, line 593: Not your changes, the `key_count` variable below this should be a `u64`

shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -529,10 +537,20 @@ impl SlotMap {
 #[derive(Debug, Clone, Copy)]
 pub enum RoutingInfo {
     AllNodes,
-    AllMasters,
+    AllMasters(ResponseJoin),
     Random,
     Slot(u16),
     Other(Command),
+    Unsupported,
```
conorbros, line 544: Does `Unsupported` represent an error getting a `RoutingInfo` struct for the message and we can't route it?

rukai: It represents a command that the cluster sink does not support.
There are certain commands that there is no sensible way for shotover to handle here.

shotover-proxy/src/transforms/redis/sink_cluster.rs
rukai: Hmmmm I cant imagine ever needing that bit https://redis.io/commands/eval/

But maybe we should do a string check against "0" instead anyway?

rukai: I'll just make it a u64 and leave it at that :shrug: the current logic looks wrong (why do we only use one key?) and will probably need to be redone later on.

shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -547,27 +565,33 @@ enum ChannelResult {
...
-                Some(RedisFrame::BulkString(a)) if a.to_ascii_uppercase() == b"KILL" => None,
-                _ => Some(RoutingInfo::AllMasters),
+                Some(RedisFrame::BulkString(a)) if a.to_ascii_uppercase() == b"KILL" => {
+                    RoutingInfo::Unsupported
+                }
+                _ => RoutingInfo::AllMasters(ResponseJoin::First),
             },
-            b"SCAN" | b"SHUTDOWN" | b"SLAVEOF" | b"REPLICAOF" | b"MOVE" | b"BITOP" => None,
+            b"SCAN" | b"SHUTDOWN" | b"SLAVEOF" | b"REPLICAOF" | b"MOVE" | b"BITOP" => {
+                RoutingInfo::Unsupported
```
johndelcastillo, line 589: What is the practical implication here, if these commands are submitted to shotover, they are dropped completely?

rukai: No messages are sent to the sink redis instances and an error is returned to the client.

shotover-proxy/src/transforms/redis/sink_cluster.rs
benbromhead: from a routing perspective eval commands get routed on the first key passed into the lua script

let me find the doc ref that specifies this

rukai: If thats true then we can write a much simpler implementation:
```rust
            b"EVALSHA" | b"EVAL" => match args.get(2) {
                Some(RedisFrame::BulkString(key_count)) => {
                    if key_count.as_ref() == b"0" {
                        RoutingInfo::Random
                    } else {
                        args.get(3)
                            .and_then(RoutingInfo::for_key)
                            .unwrap_or(RoutingInfo::Unsupported)
                    }
                }
                _ => RoutingInfo::Unsupported,
            },
```

rukai: oh cool, redis doesnt accept 000 so we dont care
image

XA21X: > why do we only use one key?

Lua scripts on Redis is normally only allowed to access data that is in the same ~node~ slot, so I guess it was a shortcut to only consider the first key and defer the validation to Redis. Although after digging deeper, this is more complicated... If you declare all the keys upfront as you're supposed to, Redis will return CROSSSLOT errors if the keys don't all hash to the same slot. But if you access undeclared keys that are cross slots, Redis v6 will allow that. This behaviour has only just been disabled by default on Redis v7 (redis/redis#10615).

This doesn't change the fact that "Redis scripts block access to non-local keys", so for Shotover, it is probably a reasonable shortcut to keep? Is there any benefit with pre-validating if Redis will do it again? :/

https://github.com/redis/redis/blob/efcd1bf394668e418df1a93cd28cf9e8b0c09ce5/src/script.c#L399
https://github.com/redis/redis/blob/3cd8baf61610416aab45e0bcedcaab9beae80184/src/cluster.c#L6677

> should do a string check against "0" instead anyway?

Is this referring to the `unwrap_or("0")`? 🤔 The key_count should only be an ASCII number. If we can't parse it as UTF-8, it's a protocol error. I think we should just abort and return an error back to the client?

rukai: ooooh thanks @XA21X that explanation makes a lot of sense, I will go ahead and use the code snippet I posted earlier.

> Is this referring to the unwrap_or("0")? thinking The key_count should only be an ASCII number. If we can't parse it as UTF-8, it's a protocol error. I think we should just abort and return an error back to the client?

I just meant that I will do the check directly against the string bytes rather than parsing to an int.

conorbros (approved):

XA21X (approved):

PR #598, opened 2022-04-04
Use cql3 parser

Claude-at-Instaclustr:
This is a merge of the CQL3 parser into the shotover-proxy codebase.

- [ ] requires cql3-parser v 0.2.0 to be releases first

PR #622, opened 2022-04-27
Avoid triggering a message parse in the Null transform

rukai:
As mentioned in https://github.com/shotover/shotover-proxy/pull/598#issuecomment-1110593846

The fix works because `.frame()` triggers a full message parse but `.metadata()` just manually grabs a few bits from the header.


conorbros (approved):
Good pick up

rukai:
Well that result confirms the fix works!

Claude-at-Instaclustr (approved):
Nice catch.

Issue #606, opened 2022-04-11
Redis - CLUSTER SLOTS command doesn't return NOAUTH


johndelcastillo:
## Describe the bug
The CLUSTER SLOTS command does not return a NOAUTH response, it just hangs in the CLI.

## To Reproduce
3.216.108.90:6378> cluster slots

## Configuration
DR Chain is set, but I can't confirm if it matters

### Topology
3 node DC + 3 node dc in DR

```yaml
sources:
  redis_prod:
    Redis: {batch_size_hint: 4, listen_addr: '0.0.0.0:6378', connection_limit: 20000,
      hard_connection_limit: true}
chain_config:
  redis_chain:
  - Tee:
      behavior: Ignore
      buffer_size: 10000
      chain:
      - QueryTypeFilter: {filter: Read}
      - Coalesce:
          max_behavior:
            CountOrWait: [2000, 100]
      - Tee:
          buffer_size: 100
          timeout_micros: 10000
          chain:
          - QueryCounter: {name: DR_chain}
          - RedisSinkCluster:
              first_contact_points: ['44.232.82.142:6379']
      - 'Null'
  - QueryCounter: {name: Main_chain}
  - RedisClusterPortsRewrite: {new_port: 6378}
  - RedisSinkCluster:
      first_contact_points: ['3.216.108.90:6379']
source_to_chain_mapping: {redis_prod: redis_chain}
```

Please include any configuration for setting up services such as `docker-compose` that the `topology.yaml` depends on.

## Expected behaviour
3.216.108.90:6379> cluster slots
(error) NOAUTH Authentication required.
3.216.108.90:6379>

## Systems and Version:
 - OS - Debian buster
 - Shotover Version - 0.1.2
 - Redis 6.0.9

## Additional context
The Lettuce client doesn't work using the shotover port, its getting NOAUTH errors but I can't be sure its due to this issue.

Edit: I'm pretty confident it is due to this error being returned: 
Cannot retrieve cluster partitions from [redis://icredis:****************@3.216.108.90:6378]

PR #623, opened 2022-04-28
rework redis cluster command mapping

rukai:
closes https://github.com/shotover/shotover-proxy/issues/605

Justification for changes to command mappings:
* DBSIZE
  + Query every master and sum the result. Allows us to pretend to be a single redis instance.
* KEYS
  + Query every master and combine the results.
  + Running `KEYS *` in a large production environment could OoM shotover, but I assume in such an environment the instances would be configured to not allow running KEYS, so each redis instance would return an error and then we would return a single error to the client which would be fine.
* ACL
  + Not sure about this one but it seemed like the majority of the ACL commands should be run uniformly against every instance.
  + Possibly we will need to revisit this in the future.
* SAVE + BGSAVE + BGREWRITEAOF
  + If the user wants to force redis to save to disk it makes sense that they want this to happen to EVERY instance not just the masters
* LASTSAVE
  + The LASTSAVE command is needed to confirm that a previous BGSAVE command has succeed.
     - In order to maintain this use case we query every instance and return the oldest save time. This way the return value wont change until every instance has completed their BGSAVE.
* CLIENT GETNAME/SETNAME
  + We cant reasonably support these commands in shotover
     - Connections are pooled so we cant forward it to the redis instance
     - We cant just implement the functionality shotover side because the connection names are supposed to be viewable from CLIENT LIST
  + However we dont want to just fail them either as clients like jedis would break so:
     - We just pretend to accept CLIENT SETNAME returning an "OK" but without hitting any instances
     - We just pretend to handle CLIENT GETNAME always returning nil without hitting any instances
* CONFIG
   + I believe each instance will need unique configuration for some things so allowing a command to reconfigure every instance with the same config sounded wrong so I disabled it.
* SLOWLOG
   + Instance specific results that cant be combined so I set to Unsupported
* INFO
  + Each instance will return different information and no reasonable way to combine them so I just set Unsupported.
* TIME
  + I just set to Unsupported because of potential differences in time state and I didnt think the functionality was that important?
  + Maybe we should just pick a random instance though?
* ECHO + PING
  + Stateless commands that return a response, we just need a single instance to handle this for us so shotover can pretend to be a single instance. So we just pick a random instance.

It might be useful to refer to how redis-cluster-proxy handles these commands as well: https://github.com/RedisLabs/redis-cluster-proxy/blob/unstable/COMMANDS.md

### DIFF
--- a/shotover-proxy/src/transforms/redis/sink_cluster.rs
+++ b/shotover-proxy/src/transforms/redis/sink_cluster.rs
@@ -587,10 +587,22 @@ impl RoutingInfo {
         };
 
         Ok(match command_name.as_slice() {
+            // These commands write data but need to touch every shard so we send it to every master
             b"FLUSHALL" | b"FLUSHDB" => RoutingInfo::AllMasters(ResponseJoin::First),
+            // The total count of all keys can be obtained by summing the count of keys from every shard.
             b"DBSIZE" => RoutingInfo::AllMasters(ResponseJoin::IntegerSum),
+            // A complete list of every key can be obtained by joining the key lists provided by every shard.
+            //
+            // With this implementation, running `KEYS *` in a large production environment could OoM shotover.
+            // But I assume in such an environment the nodes would be configured to not allow running KEYS.
+            // So each redis node would return an error and then we would return a single error to the client which would be fine.
             b"KEYS" => RoutingInfo::AllMasters(ResponseJoin::ArrayJoin),
+            // The LASTSAVE command is needed to confirm that a previous BGSAVE command has succeed.
+            // In order to maintain this use case we query every node and return the oldest save time.
+            // This way the return value wont change until every node has completed their BGSAVE.
             b"LASTSAVE" => RoutingInfo::AllNodes(ResponseJoin::IntegerMin),
+            // When a command that forces writing to disk occurs we want it to occur on every node.
+            // Replica nodes receive updates from their master nodes and we want those to be written to disk too.
             b"BGSAVE" | b"SAVE" | b"BGREWRITEAOF" | b"ACL" => {
                 RoutingInfo::AllNodes(ResponseJoin::First)
             }
@@ -600,6 +612,12 @@ impl RoutingInfo {
                 }
                 _ => RoutingInfo::AllMasters(ResponseJoin::First),
             },
+            // * We cant reasonably support CLIENT SETNAME/GETNAME in shotover
+            //     - Connections are pooled so we cant forward it to the redis node
+            //     - We cant just implement the functionality shotover side because the connection names are supposed to be viewable from CLIENT LIST
+            // * However we dont want to just fail them either as clients like jedis would break so:
+            //     - We just pretend to accept CLIENT SETNAME returning an "OK" but without hitting any nodes
+            //     - We just pretend to handle CLIENT GETNAME always returning nil without hitting any nodes
             b"CLIENT" => match args.get(1) {
                 Some(RedisFrame::BulkString(sub_command)) => {
                     let sub_command = sub_command.to_ascii_uppercase();
@@ -611,6 +629,7 @@ impl RoutingInfo {
                 }
                 _ => RoutingInfo::Random,
             },
+            // These commands can not reasonably be supported by shotover, so we just return an error to the client when they are used
             b"SCAN" | b"SHUTDOWN" | b"SLAVEOF" | b"REPLICAOF" | b"MOVE" | b"BITOP" | b"CONFIG"
             | b"SLOWLOG" | b"INFO" | b"TIME" => RoutingInfo::Unsupported,
             b"EVALSHA" | b"EVAL" => match args.get(2) {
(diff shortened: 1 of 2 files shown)


shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -482,7 +492,10 @@ impl RedisSinkCluster {
     #[inline(always)]
     fn short_circuit(&self, one_tx: oneshot::Sender<Response>) {
         warn!("Could not route request - short circuiting");
-        if let Err(e) = self.send_error_response(one_tx, "ERR Could not route request") {
+        if let Err(e) = self.send_error_response(
+            one_tx,
+            "ERR Shotover RedisSinkCluster does not not support this command used in this way",
```
conorbros, line 497: Are we able to give more detailed error message?

conorbros, line 497: Are we able to give more detailed error message?

rukai: What sort of details do you want?

conorbros: What way the command is being used

rukai: ah, well we could try and store an error in the RoutingInfo::Unsupported but I think that should go in a follow up PR

conorbros: Happy for that to be attempted in a followup PR.

conorbros (approved):
Would you be able to add your justification of why certain commands are being handled a certain way as comments to the match statement. Especially for the ones where it's not clear why we do that, will save someone having to find this PR to make sense of it.

rukai:
oh good point!
I'll do that

shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -572,22 +587,51 @@ impl RoutingInfo {
...
+                        b"SETNAME" => RoutingInfo::ShortCircuitOk,
+                        b"GETNAME" => RoutingInfo::ShortCircuitNil,
+                        _ => RoutingInfo::Unsupported,
+                    }
+                }
+                _ => RoutingInfo::Random,
+            },
+            // These commands can not reasonably be supported by shotover, so we just return an error to the client when they are used
+            b"SCAN" | b"SHUTDOWN" | b"SLAVEOF" | b"REPLICAOF" | b"MOVE" | b"BITOP" | b"CONFIG"
+            | b"SLOWLOG" | b"INFO" | b"TIME" => RoutingInfo::Unsupported,
```
XA21X, line 634: Does redis-cli still work if we entirely forbid `CONFIG`? :/ I was planning to test it myself, but macOS + M1 has broken the (lack of) bridge networking workaround I was using, and I haven't had time to setup a Linux VM.

I thought of this because I remember it running CONFIG commands, but at a glance, I'm not sure whether it's required to succeed: https://github.com/redis/redis/blob/87131a5fa612c0d2b688fc65c722ebc95ae8f42a/src/redis-cli.c#L7043

XA21X: Should we have some basic client driver integration tests, to prevent regressions like #606?

(e.g. redis-cli + Lettuce + Jedis)

New ticket of course. :)

rukai: this works:
image

this does not:
image

Do we need to support --cluster though?
I dont think it makes sense when shotover is pretending to be a single non clustered redis instance.

rukai: > Should we have some basic client driver integration tests, to prevent regressions like https://github.com/shotover/shotover-proxy/issues/606?

Yeah this would be really valuable if we can find a way to nicely pull in these non rust dependencies.

XA21X: > Do we need to support --cluster though?
> I dont think it makes sense when shotover is pretending to be a single non clustered redis instance.

For our managed service, cluster hiding is a bonus feature that isn't actually needed at the moment. The primary use case is Shotover mirroring for Redis clusters, so it's always `redis-cli` with `-c` in our docs/examples.

> find a way to nicely pull in these non rust dependencies

One more self-contained docker compose? :)

rukai: > For our managed service, cluster hiding is a bonus feature that isn't actually needed at the moment. The primary use case is Shotover mirroring for Redis clusters, so it's always redis-cli with -c in our docs/examples.

Oh huh, I'll have to go over things again then.
When ben linked https://github.com/RedisLabs/redis-cluster-proxy/blob/unstable/COMMANDS.md I must have taken it too literally.

rukai: I have confirmed this does work on main branch, so I'll investigate whats going on to make sure we dont regress here
image

rukai: From my testing setting CONFIG to Unsupported does not prevent the running of --cluster commands.
But setting the actual cluster command to Unsupported will naturally prevent the running of that --cluster command.

So I think there is no issue here.
If there are any --cluster commands that we can support then they will be supported.

benbromhead (commented):
I'm happy with this approach on the surface level. I'm happy for this to get merged if @XA21X is happy with the approach.

XA21X (approved):

PR #624, opened 2022-05-02
Fix RedisClusterPortsRewrite noauth handling

rukai:
closes https://github.com/shotover/shotover-proxy/issues/606

The issue was that we were failing as soon as we received an unexpected message in this transform.
To handle NOAUTH and likely other error cases too we should just ignore messages in an unexpected format.

### DIFF
--- a/shotover-proxy/example-configs/redis-cluster-auth/topology.yaml
+++ b/shotover-proxy/example-configs/redis-cluster-auth/topology.yaml
@@ -5,6 +5,10 @@ sources:
       listen_addr: "127.0.0.1:6379"
 chain_config:
   redis_chain:
+    # This transform is only here to ensure that the transform correctly handles the case where
+    # redis returns an error due to being unauthenticated
+    - RedisClusterPortsRewrite:
+        new_port: 6379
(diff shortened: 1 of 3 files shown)


XA21X (approved):

shotover-proxy/example-configs/redis-cluster-auth/topology.yaml
```diff
@@ -5,6 +5,10 @@ sources:
       listen_addr: "127.0.0.1:6379"
 chain_config:
   redis_chain:
+    # This transform is only here to ensure that the transform correctly handles the case where
+    # redis returns an error due to being unauthenticated
```
conorbros, line 9: Can we move this to `test-configs`? It's no different from the other cluster ones and now contains testing specific stuff.

conorbros (approved):

shotover-proxy/example-configs/redis-cluster-auth/topology.yaml
rukai: ah true, i'll do in a follow up.

Issue #183, opened 2021-09-10
use anyhow .context for error handling


rukai:
There are many cases where we do something like `error.map_err(|e| anyhow!("more context {}", e))` instead of using `error.context("more context")`.
We should first double check that `.context()` actually provides better errors and then convert the entire codebase to use it.

PR #625, opened 2022-05-02
Use anyhow .context() instead of manually constructing error messages

rukai:
closes https://github.com/shotover/shotover-proxy/issues/183

The reason I jumped on this was because I hit an issue in https://github.com/shotover/shotover-proxy/pull/624 due to the incomplete use of context.
What happened was a contextified anyhow was printed with `error!("{}")` instead of `error!("{:?}")`.
This results in a log that lacks the innermost error.

So the core fix here is the changes to server.rs
However I wanted to make the rest of shotover make use of `.context` while I was doing this to make our errors uniform.

That said, I'm not 100% sure that using context is actually the right call.
The `{}`/`{:?}` difference is definitely a footgun.
So maybe we would be better off just entirely removing the use of `.context` instead?

Here is what logged error created via `.context()` looks like:
```
2022-05-02T01:22:59.373422Z ERROR request{id=2 source="RedisSource"}: shotover_proxy::server: chain failed to send and/or receive messages

Caused by:
    0: failed to rewrite CLUSTER NODES port
    1: RedisClusterPortsRewrite intercepted an incorrect message
```

### DIFF
--- a/shotover-proxy/src/codec/redis.rs
+++ b/shotover-proxy/src/codec/redis.rs
@@ -47,7 +47,7 @@ impl Decoder for RedisCodec {
 
     fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>> {
         loop {
-            match decode_mut(src).map_err(|e| anyhow!("Error decoding redis frame {}", e))? {
+            match decode_mut(src).map_err(|e| anyhow!(e).context("Error decoding redis frame)"))? {
                 Some((frame, _size, bytes)) => {
                     self.messages
                         .push(Message::from_bytes_and_frame(bytes, Frame::Redis(frame)));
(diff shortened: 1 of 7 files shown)


conorbros (approved):
I much prefer the results from `.context()` and I think we can be disciplined with the use of `{:?}`.

Claude-at-Instaclustr:
Unless I am doing something wrong, the amount of information in the error message is significantly reduced in the attached code.  Simply change the context_style variable on line 4 to see the difference

[test.zip](https://github.com/shotover/shotover-proxy/files/8637937/test.zip)

rukai:
image
Are you maybe missing the "Caused by" section?

Claude-at-Instaclustr:
weird, when I run it from the command line as you did I get the result you did.  However, inside the IDE I get the following:
Screenshot from 2022-05-06 07-40-01

rukai:
I think your IDE is hiding it behind the `<1 internal call>`
Can you try clicking the `+`?

Claude-at-Instaclustr:
DOH! That was it.

Claude-at-Instaclustr (approved):

PR #627, opened 2022-05-03
Only parse the opcode when checking for errors in Cassandra sink

conorbros:


shotover-proxy/src/message/mod.rs
```diff
@@ -130,6 +130,14 @@ impl Message {
         }
     }
 
+    pub(crate) fn get_raw_bytes(&self) -> &Bytes {
+        match self.inner.as_ref().unwrap() {
+            MessageInner::RawBytes { bytes, .. } => bytes,
+            MessageInner::Parsed { bytes, .. } => bytes,
+            MessageInner::Modified { .. } => panic!("can't get raw bytes of modified"),
```
rukai, line 137: I do not like the idea of a public method on Message that just panics if the message is in a certain state.

conorbros: Yes, I'm intending to change this. Apologies, I should've marked as a draft.

conorbros: Not sure what the best approach here is. I only want to return a reference to the bytes because this function is used in a very hot path so cloning the bytes so we can own it is not possible, but to return the `bytes` for a `MessageInner::Modified` means we have to encode it and then that value will be owned by the function.

Some options:
- Leaving the `panic` in here, keeping it as `pub (crate)` so it can't be used by transform developers and marking it clearly with a comment that this should not be used where it's possible the message might be in that state.
- Making `inner` `pub (crate)` so I can write this logic in the codec to ensure it's only used there.

rukai: I'm confused your comment mentions using it in a codec, but the only place its used is in a transform.

conorbros: Oops, updated my comment.

rukai: A possible option is making a reference version of  `fn into_encodable(self, expected_message_type: MessageType) -> Result<Encodable> {`

Something like: `as_encodable(self, expected_message_type) -> Result<EncodableRef>` 
where:
```rust
enum EncodableRef {
    Bytes(&Bytes),
    Frame(&Frame),
}
```

If we did that we should probably rename Encodable to something more generic though.

conorbros:
Small performance gain with this change. Bench results: 

image

shotover-proxy/src/message/mod.rs
```diff
@@ -173,6 +173,47 @@ impl Message {
         }
     }
 
+    pub fn as_encodable(&self, expected_message_type: MessageType) -> Result<EncodableRef> {
```
rukai, line 176: Now that I've taken a closer a look at how we use this is in CassandraSinkSingle this suggestion feels like overkill. Its possible the extra functionality here might be useful in the future but I would rather stick with a simpler API and refactor it later if needed. Theres also the problem of confusing the meaning of Encodable, as its now actually used for decoding!

So I'm thinking we would be better off with something like:
```rust
/// Placeholder comment: only use straight out of the oven
pub fn as_raw_bytes(self) -> Option<&Bytes> {
    match &self.inner {
        MessageInner::RawBytes { bytes, .. } => Some(bytes),
        _ => None,
    }
}
```

A reason I dont want to expose the inner field as public API is because we might want to add hidden caching to Message in the future and we cant do both of these.

My assumption here is that MessageInner::RawBytes is the only case we will need this for as its the only time we could possibly make an assumption about the state of a message.
In any other case we would be better off just using `.frame()`.

shotover-proxy/src/message/mod.rs
```diff
@@ -173,6 +173,16 @@ impl Message {
         }
     }
 
+    /// Only use for messages read straight from the socket
+    /// that are definitely in an unparsed state
+    /// (haven't passed through any transforms where they might have been parsed or modified)
+    pub fn as_raw_bytes(&self) -> Option<&Bytes> {
+        match &self.inner.as_ref().unwrap() {
```
rukai, line 180: nit: this can be just `match self.inner.as_ref().unwrap() {`

rukai (approved):
I am a bit concerned that punching a pole through our abstractions could hurt us later on but I cant see any other way to performantly provide this failed request counting feature.

conorbros:
I'll change it to `pub (crate)`?

rukai:
atm that is equivalent to `pub` anyway but sure lets do that as it better reflects intent.

rukai:
oh.
I wonder if we should have just added an opcode field to CassandraMetadata?

Claude-at-Instaclustr (approved):

PR #633, opened 2022-05-05
Read multiple Cassandra frames

conorbros:
The Cassandra protocol is designed to allow multiple inflight messages, Shotover currently doesn't take advantage of this: we read one frame off the socket, process it through the transform chain, send to the upstream and wait for the response. This change allows us to read multiple frames off the socket and process a batch instead.

Bench results with latte:

image

### DIFF
--- a/shotover-proxy/src/codec/cassandra.rs
+++ b/shotover-proxy/src/codec/cassandra.rs
@@ -1,8 +1,8 @@
 use crate::frame::cassandra::CassandraOperation;
 use crate::frame::{CassandraFrame, Frame, MessageType};
 use crate::message::{Encodable, Message, Messages};
-use anyhow::anyhow;
-use bytes::{BufMut, BytesMut};
+use anyhow::{anyhow, Error, Result};
+use bytes::{Buf, BufMut, BytesMut};
 use cassandra_protocol::compression::Compression;
 use cassandra_protocol::frame::frame_error::{AdditionalErrorInfo, ErrorBody};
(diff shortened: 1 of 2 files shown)


shotover-proxy/src/codec/cassandra.rs
```diff
@@ -52,60 +55,65 @@ impl CassandraCodec {
...
+                }
+                Err(ParseFrameError::NotEnoughBytes) => {
+                    if self.messages.is_empty() || src.remaining() != 0 {
+                        return Ok(None);
+                    } else {
+                        return Ok(Some(std::mem::take(&mut self.messages)));
+                    }
+                }
+                Err(ParseFrameError::UnsupportedVersion(version)) => {
+                    let _ = src.split_to(9); // we will have read the entire header so clear those bytes
```
rukai, line 89: I dont see the point in partially handling this here?
We have to force close the connection anyway as happens on the following line.

conorbros: We still have to clear the buffer so the loop can exit.

rukai: Clearing just the header isnt enough.
I guess you could clear the entire buffer but I think it would be more clear if you just returned from the function.

rukai (approved):
Incredible performance win here.
Good stuff.

Logic matches what we have for redis and I went over it again and it still makes sense to me.

Claude-at-Instaclustr (approved):

PR #635, opened 2022-05-05
Benches for CassandraPeersRewrite

conorbros:


rukai (approved):
A Null transform would cause problems if we were benching the rewrite case, but seeing as we're only benching the passthrough case... seems fine to me.

conorbros:
Yep, that's the intention. I'm intending to work on the performance for passthrough cases of this transform.

Claude-at-Instaclustr (approved):
LGTM

Issue #630, opened 2022-05-05
Fix RedisSinkCluster random node handling


rukai:
1. RoutingInfo::Random does not route to a random node but to the first node in the list of master nodes
2. Many commands fallback to the default routing implementation of using the 2nd command arg as the key for a slot
    + This results in commands like "CLUSTER REPLICAS" using REPLICAS as a key which is absurd, we should instead use RoutingInfo::Random for this command. Actually thinking about it more, maybe we are better off setting "CLUSTER" commands as RoutingInfo::Unsupported for cluster hiding.

rukai:
The second point is not fixed yet.
CLUSTER REPLICAS is still routed with "REPLICAS" treated as a key.

PR #636, opened 2022-05-06
RoutingInfo::Random is actually random

rukai:
Progress towards: https://github.com/shotover/shotover-proxy/issues/630

The advantages of actually being random are:
* Better load balancing
* Reveals bugs where the client relies on a command not being randomly routed but the bug is hidden due to the order usually being the same
* More accurately reflects the name used in shotover

If performance is a concern then we could just as easily implement this with an incrementing u32.
(I wanted to add a benchmark but cant until https://github.com/shotover/shotover-proxy/pull/623 lands)

If we really truly want it to not be random then we should instead rename RoutingInfo::Random to RoutingInfo::FirstMaster or similar.


shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -19,7 +19,7 @@ use derivative::Derivative;
 use futures::stream::FuturesUnordered;
 use futures::{Future, StreamExt, TryFutureExt};
 use metrics::{counter, register_counter};
-use rand::prelude::SmallRng;
+use rand::prelude::*;
```
conorbros, line 22: Why change this to the `*`?

rukai: the idea of a prelude is it just provides the main types so that you can glob import them all.
So using a prelude without a glob import just looks wrong to me.
But I agree, lets just not use the prelude at all.

Changed.

conorbros: Yeah I was wondering if we were using more imports from there now. I'd always prefer to not use the `*` if possible.

conorbros (approved):

benbromhead (approved):
From a perf perspective most of the stuff that gets routed via random isnt going to be on the hot path of most applications.

PR #637, opened 2022-05-06
use check_frame instead of from_buffer

conorbros:
Use a new method that only checks the frame length and version of incoming Cassandra frames before fully parsing them. 

Latte benchmarks against main:
image

<s>Waiting on https://github.com/krojew/cdrs-tokio/pull/93 to be merged.</s>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -291,7 +291,7 @@ checksum = "8234d29d30873ab5a41e3557b8515d3ecbaefb1ea5be579425b3b0074b6d0e40"
 [[package]]
 name = "cassandra-protocol"
 version = "1.1.1"
-source = "git+https://github.com/krojew/cdrs-tokio#62ce78e5592fc8ea00a13fc4c30df5343693b7ce"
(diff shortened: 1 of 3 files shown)


Claude-at-Instaclustr:
Are we going to release the cdrs-tokio package ourselves?

conorbros:
No, my changes were just upstreamed.

rukai:
What about the from_buffer calls in metadata and cell_count?

cell_count needs to access the body so not much we can do there.

But maybe for metadata we need a method that only reads the header?
Should that be combined with the check_frame_size method?
Do we care about the performance of metadata in the first place?

conorbros:
Yeah, we can definitely speed that up too. Although it's in a much less hot path than this so I would prefer to get this merged first and investigate later.

rukai (approved):
Given that you've already landed `check_frame_size` in cassandra-protocol, I agree completely lets just land this.

If we were still running off a cassandra-protocol fork I would have pushed for more experimentation in the cassandra-protocol API before trying to upstream it to reduce upstream churn.

This is fine though.
Very nice win.

XA21X (approved):

PR #638, opened 2022-05-09
Redis benchmarks test both set and get commands

rukai:
On my machine `cargo bench redis/` takes: 2m 6s
on the main branch it takes: 1m 33s
So a bit of a regression in run time but I think the extra cases will be useful.

The results are actually stable enough to see a difference between set and get commands which is nice.
In follow up PRs I would like to add more cases (but probably only run them against `example-configs/redis-cluster`)

I copied the abstraction used by the cassandra benches to efficiently run benches with different queries against the same docker setup.


conorbros (approved):

benbromhead (approved):
lgtm!

PR #640, opened 2022-05-11
CassandraPeersRewrite bugfix

conorbros:
If you send a query to the `system.peers_v2` table to an older version of Cassandra (`cqlsh` does this for example) it will cause a panic. 

- Added a test case exposing the bug
- Removed the panic.


rukai:
If we support cassandra 3 should we run all of our test cases against it?

rukai (approved):
I still think this is a good idea regardless of what versions we want to support.
As noted in https://github.com/shotover/shotover-proxy/pull/598 there are various reasons the server could return an error to a query that should be infallible.
So panicking is clearly not an option.

Possibly we want to log as an error.
But this could easily become unrelated log spam when there is a persistent issue with the server causing the query to fail.

conorbros:
> If we support cassandra 3 should we run all of our test cases against it?

I'll investigate this. I think it would depend on the CQL version that each Cassandra version uses, given all of the other test don't rely on any version specific Cassandra internals like this one does.

benbromhead (approved):

Issue #631, opened 2022-05-05
Support clusters without hiding


rukai:
Currently we have:
* RedisSinkCluster which is designed for redis cluster hiding
* RedisSinkSingle which is designed to handle non-clustered and clustered redis nodes.

Another use case that we dont yet have a suitable solution for is an equivalent of `redis-cli -c` where MOVE/ASK is handled by shotover but commands are otherwise passed directly to a single specific node.

I think the best way to do this would be separate sinks:
* RedisSinkClusterHandling or maybe RedisSinkClusterSingle
* RedisSinkClusterHiding
* RedisSinkSingle

Alternatively we could add a `node_destination: Option<String>` field to the RedisSinkClusterConfig to enable non-hiding mode.

PR #642, opened 2022-05-12
RedisSinkCluster Hiding/Handling modes

rukai:
Closes https://github.com/shotover/shotover-proxy/issues/631

* I opted to expose this to the user via an enum in the destination field, the changes to the topology files should demonstrate how this is used.
   + A result of this approach is that you can no longer specify a list of default nodes beyond the initial node.
      + I'm not sure if having the handling destination node as the only source of other nodes makes shotover more or less robust. On one hand shotover is pretending to be that node so if it cant contact it then its better to fall over to avoid hiding the issue. On the other hand working in more cases is always better right?
      + A clear benefit is it makes configuration easier as you only need to provide the one true destination and then other peers are sourced from that.
   + It seems reasonable to me, but I dont feel strongly about this approach and would be interested to hear feedback or other ideas.
   + I feel like a pretty good reason to not do this approach would be if cluster hiding mode was a much more common use case than cluster handling.
      + in that case we would be better off having cluster hiding as the default and then having a `handling_destination: Option<String>` field that enables handling mode
* the test_script case is absent from the handling integration test because script no longer runs against every instance of the cluster, which means that EVALSHA (which is routed by key) is likely to be run against a redis instance without the script.
* The `Flusher` abstraction needed to be added to the integration tests because the handling test cases were failing due to FLUSHDB only running on the destination instance instead of every instance of the cluster.
   + I honestly really dislike this solution but I cant think of a better one.


shotover-proxy/tests/redis_int_tests/basic_driver_tests.rs
```diff
@@ -1476,3 +1572,46 @@ async fn run_all(connection: &mut Connection) {
     test_ping_echo(connection).await;
     test_time(connection).await;
 }
+
+struct Flusher {
```
conorbros, line 1576: Another option for this could be integrating it into `ShotoverManager` and then passing that around instead of the Redis connection. That way we could remove some duplicate functionality.

Would have to rework lots of `ShotoverManager` though, so I'll let you decide if you want to do that. Otherwise this is fine.

rukai: I'm trying to imagine what such an API would look like.

```rust
let shotover_manager = ShotoverManager::from_topology_file("example-configs/redis-multi/topology.yaml");
// Store connections internally within shotover_manager: 6379 is shotovers redis port and 1111, 1112, 1113 are the ports of the actual redis instances.
shotover_manager.redis_connections_async(6379, &[1111, 1112, 1113]).await; 
run_test_case(&shotover_manager);
    
fn run_test_case(shotover_manager: &ShotoverManager) {
   shotover_manager.flush();
   shotover_manager.shotover_redis_connection.set("foo", "bar");
   for connection in shotover_manager.direct_redis_connections {
      assert_eq!(connection.get("foo"), "bar");
   }
}
```

I do quite like the idea of this but I'm worried of the extra statefulness within ShotoverManager that this introduces.
I think the simplified and more flexible API would be worth it though?

Something this doesnt account for is secondary shotover connections which can be useful for some testcases.
Maybe we can just externally manage that connection for those cases?

conorbros: >secondary shotover connections

That could work in a similar way to the `shotover_redis_connection` field you showed in your example above. They could be created lazily, reusing the port we originally specified to connect to Shotover.
```rust
let shotover_manager = ShotoverManager::from_topology_file("example-configs/redis-multi/topology.yaml");
// Store connections internally within shotover_manager: 6379 is shotovers redis port and 1111, 1112, 1113 are the ports of the actual redis instances.
shotover_manager.redis_connections_async(6379, &[1111, 1112, 1113]).await; 
run_test_case(&shotover_manager);
    
fn run_test_case(shotover_manager: &ShotoverManager) {
    shotover_manager.flush();
    shotover_manager.shotover_redis_connection(0).unwrap().set("foo", "bar"); // get our first shotover connection with the index 0
    shotover_manager.shotover_redis_connection(1).unwrap().set("foo", "bar"); // get our second shotover connection with the index 1
   for connection in shotover_manager.direct_redis_connections {
      assert_eq!(connection.get("foo"), "bar");
   }
}
```

shotover-proxy/example-configs/redis-cluster-dr/topology.yaml
```diff
@@ -19,10 +19,12 @@ chain_config:
...
           - RedisSinkCluster:
-              first_contact_points: [ "127.0.0.1:2120", "127.0.0.1:2121", "127.0.0.1:2122", "127.0.0.1:2123", "127.0.0.1:2124", "127.0.0.1:2125" ]
+              destination:
+                Hiding: [ "127.0.0.1:2120", "127.0.0.1:2121", "127.0.0.1:2122", "127.0.0.1:2123", "127.0.0.1:2124", "127.0.0.1:2125" ]
     - QueryCounter:
         name: "Main chain"
     - RedisSinkCluster:
-        first_contact_points: [ "127.0.0.1:2220", "127.0.0.1:2221", "127.0.0.1:2222", "127.0.0.1:2223", "127.0.0.1:2224", "127.0.0.1:2225" ]
+        destination:
+          Hiding: [ "127.0.0.1:2220", "127.0.0.1:2221", "127.0.0.1:2222", "127.0.0.1:2223", "127.0.0.1:2224", "127.0.0.1:2225" ]
```
benbromhead, line 28: Hiding here as a keyword feels confusing (are we hiding these nodes or are we including these in the hidden cluster).

shotover-proxy/example-configs/redis-cluster-handling/topology.yaml
```diff
@@ -0,0 +1,15 @@
...
+  redis_prod:
+    # define how shotover listens for incoming connections from our client application (`redis-benchmark`).
+    Redis:
+      listen_addr: "127.0.0.1:6379"
+chain_config:
+  redis_chain:
+    # configure Shotover to connect to the Redis cluster via the shotover instances passthrough node
+    - RedisSinkCluster:
+        destination:
+          Handling: "127.0.0.1:2220"
```
benbromhead, line 12: As above... I think destination (or something like it) needs to be the contact point keyword and Handling / Hiding modes defined above. Or even an explicity `mode: Handling` property.

I know this makes the rust side less ergonomic, but right now the user configuration feels confusing with the way it's set up.

shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -65,23 +65,32 @@ impl RedisSinkClusterConfig {
     }
 }
 
+#[derive(Deserialize, Debug, Clone)]
+pub enum Destination {
```
benbromhead, line 69: call this mode

shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -65,23 +65,32 @@ impl RedisSinkClusterConfig {
...
+    /// Contains a single instance to connect to
+    Handling(String),
```
benbromhead, line 71: make the string a named field called destination or something

benbromhead (commented):
Having only one contact/discovery points is problematic and likely to lead to the occasional discovery related outage (C* recommends bootstrapping nodes have 2 contact seeds and that's just on bootstrap which is a once off process) , I would err on the side of 2.

Could you also update the docs describing the Redis transform to be in line with this ability to support two different modes.

rukai:
I changed the public facing API to use a single direct_connection field.
I think that introducing an enum here isnt worth the cost because enums can be confusing to configure in yaml.
And I think the hiding/handling distinction I invented is not clear by itself (the user would have to check the documentation anyway)
So we may as well just use an Option that changes functionality when set and document this functionality in the config example.

But I guess you could also argue that having a mandatory Hiding/Handling enum ensures that users always know that such a configuration is active rather than hiding it behind a default value?

I'll look into what we want to do with the possible ShotoverManager refactor.

shotover-proxy/tests/redis_int_tests/basic_driver_tests.rs
rukai: After thinking about it I would like to land the Flusher as is and raise an issue to investigate combining with ShotoverManager.

My reasoning is:
It will touch a lot of things and so is out of scope of this PR and would therefore need to be a prereq or postreq. Making it a postreq has the advantage of allowing us to directly compare the ShotoverManager and Flusher approaches.
Additionally if the ShotoverManager approach doesnt pan out we can always leave it as the Flusher approach.

conorbros: I agree with that. Let's not overengineer it, if we end up needing the abstraction we'll do it.

rukai:
I was reading this article on the weekend, maybe we should use oncecell to store the direct_connection https://matklad.github.io/2022/06/11/caches-in-rust.html

shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -108,6 +115,25 @@ impl RedisSinkCluster {
         Ok(sink_cluster)
     }
 
+    async fn direct_connection(&mut self) -> Result<&UnboundedSender<Request>> {
```
rukai, line 118: Im a bit conflicted on if caching the direct connection in self.direct_connection is the right call here.

reasons to cache:
Its more correct as we can maintain connection state for the direct connection for the lifetime of the chain.
Bonus: faster lookup.

reasons not to cache:
maybe we shouldnt be relying on connection state in the first place as that state wont exist when querying a SLOT off of this connection.
bonus: the connection pool will auto-detect and reopen closed connections for us.

At the moment I'm kind of thinking it doesnt matter a whole lot either way right now and we should properly address this when we look into connection state issues like RESP3 support.

docs/src/transforms.md
```diff
@@ -275,10 +275,18 @@ This transform is a full featured Redis driver that will connect to a Redis clus
 - RedisSinkCluster:
     # A list of IP address and ports of the upstream redis nodes/services.
     first_contact_points: ["127.0.0.1:2220", "127.0.0.1:2221", "127.0.0.1:2222", "127.0.0.1:2223", "127.0.0.1:2224", "127.0.0.1:2225"]
+
+    # By default RedisSinkCluster will attempt to emulate a single non-clustered redis node by completely hiding the fact that redis is a cluster.
+    # However, when this field is provided, this cluster hiding is disabled.
+    # Instead other nodes in the cluster will only be accessed when performing a command that accesses a slot.
+    # All other commands will be passed directly to the direct_connection node.
+    direct_connection: "127.0.0.1:2220"
```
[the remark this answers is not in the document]
conorbros: I agree, I think these examples should be the most sensible options.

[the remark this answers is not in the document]
rukai: alright, done

shotover-proxy/example-configs/redis-cluster-hiding/docker-compose.yml
```diff
@@ -0,0 +1,90 @@
+version: '2'
```
conorbros, line 1: I know you didn't write this whole file, but any idea why this is '2' while the others are '3.3'

rukai: yeah its a bit all over the place.
I think its out of scope this PR but you could raise an issue to unify onto a single version if you wanted.

shotover-proxy/src/transforms/redis/sink_cluster.rs
benbromhead: I'm ok to revisit this properly when we look into resp3

conorbros (approved):

PR #644, opened 2022-05-13
Change timeout in TcpCodecListener

conorbros:
Change the timeout in `TcpCodecListener` to more closely replicate the behaviour of the source that's representing. 

- Redis by [default](https://redis.io/docs/reference/clients/#client-timeouts) has no timeout but the user can configure one.
- Cassandra by [default](https://github.com/apache/cassandra/blob/e0247d6833f59901e6849e168f6e2363351c6f4c/conf/cassandra.yaml#L801) has no timeout either but the user can configure one.

### DIFF
--- a/shotover-proxy/src/server.rs
+++ b/shotover-proxy/src/server.rs
@@ -96,6 +96,8 @@ pub struct TcpCodecListener<C: Codec> {
     message_count: u64,
 
     available_connections_gauge: Gauge,
+
+    timeout: Option<u64>,
 }
 
 impl<C: Codec + 'static> TcpCodecListener<C> {
@@ -109,6 +111,7 @@ impl<C: Codec + 'static> TcpCodecListener<C> {
         limit_connections: Arc<Semaphore>,
         trigger_shutdown_rx: watch::Receiver<bool>,
         tls: Option<TlsAcceptor>,
+        timeout: Option<u64>,
     ) -> Result<Self> {
         let available_connections_gauge =
             register_gauge!("shotover_available_connections", "source" => source_name.clone());
@@ -128,6 +131,7 @@ impl<C: Codec + 'static> TcpCodecListener<C> {
             tls,
             message_count: 0,
             available_connections_gauge,
+            timeout,
         })
     }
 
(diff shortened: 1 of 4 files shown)


docs/src/source-types.md
```diff
@@ -20,6 +20,9 @@ Cassandra:
   # * when true: the connection is dropped.
   # * when false: the connection will wait until a connection can be made within the limit.
   hard_connection_limit: false
+ 
+  #  Timeout in seconds after which to terminate an idle connection. This field is optional, if not provided, idle connections will never be terminated.
+  # timeout: 60
```
rukai, line 25: Hmmm, all the other optional fields we have provided uncommented as it makes it more obvious that it is the field and not just a comment. see .e.g. tls field on RedisSinkSingle

However I think in this situation you have made the right call as we want people to use the default value of no timeout.

docs/src/source-types.md
```diff
@@ -20,6 +20,9 @@ Cassandra:
   # * when true: the connection is dropped.
   # * when false: the connection will wait until a connection can be made within the limit.
   hard_connection_limit: false
+ 
+  #  Timeout in seconds after which to terminate an idle connection. This field is optional, if not provided, idle connections will never be terminated.
```
rukai, line 24: Should we use a single space after the `#`?

conorbros: Yep, that wasn't intentional.

shotover-proxy/src/server.rs
```diff
@@ -437,13 +447,13 @@ impl<C: Codec + 'static> Handler<C> {
...
-                            if idle_time_seconds < 35 {
-                                trace!("Connection Idle for more than {} seconds {}",
-                                    idle_time_seconds, self.conn_details);
-                            } else {
-                                debug!("Dropping. Connection Idle for more than {} seconds {}",
-                                    idle_time_seconds, self.conn_details);
-                                return Ok(());
+                            if let Some(timeout) =  self.timeout {
+                                if idle_time_seconds < timeout {
+                                    trace!("Connection Idle for more than {} seconds {}", timeout, self.conn_details);
```
rukai, line 452: seems wrong to have a louder trace level for not hitting the timeout than actually hitting the timeout?

rukai: Oh it was always like that...

rukai: maybe just drop it to `debug!` level while we are here?

rukai (approved):
lgtm

benbromhead (approved):
lgtm

PR #647, opened 2022-05-18
Add RequestThrottling docs to the Cassandra example

conorbros:


docs/src/examples/cassandra-cluster-shotover-sidecar.md
```diff
@@ -115,3 +115,32 @@ and use the same query again. You should see the following results returned, not
 ```
 
 If everything has worked, you will be able to use Cassandra, with your connection going through Shotover!
+
+
+#### Adding Rate Limiting
+
+The next section of this tutorial will cover adding rate limiting to your Cassandra cluster with Shotover. We will add the `RequestThrottling` transform to our `topology.yaml` as shown below. This transform should go at the front of the chain to prevent any unnessecary operations from occurring if a query is going to be rate limited.
```
rukai, line 122: unnessecary -> unnecessary

rukai:
When following the instructions on `cassandra-cluster-shotover-sidecar.html` I forget to create a topology.yaml file and `docker-compose up -d` still seemed to succeed.
Is there a way we can make it fail more explicitly,

In fact even when setup up like this:
image
I still cant connect to shotover (but direct to cassandra works fine)
image

conorbros:
I can investigate more but it will require changes out of the scope of this PR.

rukai:
I'm pretty sure the instructions are broken because they never mention a `./config` folder but the docker-compose.yml expects there to be one.

conorbros:
It mounts the PWD into a config folder in the docker container.

rukai:
right, but the instructions dont mention that at all?

rukai:
ah, alright

rukai:
image
The shotovers just end up in a restarting loop

rukai (approved):
Not sure if this should go in its own page or not, we can always refactor it later though

benbromhead (approved):
thanks!

PR #648, opened 2022-05-18
Temp CI fix

conorbros:
Until the cert is updated we will need this for the CI to work. I will re enable cert checking once it's updated.


benbromhead (approved):

rukai (approved):
ah, lol, thanks for investigating
fix makes sense to me

PR #649, opened 2022-05-18
Remove unused example

conorbros:
Not sure how this slipped through...


rukai (approved):
nice cleanup

rukai:
Oh, I believe I was using this for benchmarking the rate-limiting with examples/cassandra_bench.rs and it accidentally got included in unrelated PR.

Even though it is useful for benchmarking, I think it should still be removed considering it will bitrot without a related integration test.

conorbros:
Ah, makes sense. I actually assumed it was one of mine lol

rukai:
After thinking more, maybe we should just add an integration test to avoid bitrot.
But also in place of that, deleting it is fine.

benbromhead (approved):

PR #651, opened 2022-05-19
Rename connection_error -> reason_for_no_nodes

rukai:
This is a more accurate name for the variable as its only purpose is to provide a better error message if possible, it doesnt actually affect whether an error will occur or not.

This variable might end up deleted in another planned refactor but at least by landing this first it will make the next PR easier to review.

### DIFF
--- a/shotover-proxy/src/transforms/redis/sink_cluster.rs
+++ b/shotover-proxy/src/transforms/redis/sink_cluster.rs
@@ -76,7 +76,7 @@ pub struct RedisSinkCluster {
     rng: SmallRng,
     connection_count: usize,
     connection_pool: ConnectionPool<RedisCodec, RedisAuthenticator, UsernamePasswordToken>,
-    connection_error: Option<&'static str>,
+    reason_for_no_nodes: Option<&'static str>,
     rebuild_connections: bool,
     first_contact_points: Vec<String>,
     token: Option<UsernamePasswordToken>,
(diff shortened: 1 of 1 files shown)


conorbros (approved):

shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -141,7 +141,7 @@ impl RedisSinkCluster {
             // Return an error as we cant send anything if there are no channels.
             0 => {
                 let (one_tx, one_rx) = immediate_responder();
-                match self.connection_error {
+                match self.reason_for_no_nodes {
```
benbromhead, line 144: unreachable_cluster_error ?

Not to bikeshed on names, but i feel like the name error should be in here?

rukai: The reason I dont want error in the name is because it makes it sound like None means there is an error and Some("") means there is an error.
Which isnt true.
The variable does not decide if there is an error, it is just used to enhance the error message if available.

But ... maybe this isnt how other people see this at all.
So im happy for `unreachable_cluster_error` if you think that makes more sense.

benbromhead: Ahh yup gotcha

benbromhead (approved):

PR #658, opened 2022-05-25
RedisSinkCluster ChannelResult and short_circuit refactor

rukai:
There are two parts to this refactor:
1. ChannelResult is removed - instead of returning a ChannelResult in get_channels that describes how we should use the channel we just directly use the channel.
    * its easier to follow what each case is doing in dispatch_message_hiding - just check the inner function call rather than having to backtrack
    * its more flexible for uniquely handling a case without having to add a new hyperspecific ChannelResult variant. e.g. we can now directly send a single message which was impossible before.
    * the dispatch_message_handling approach used in https://github.com/shotover/shotover-proxy/pull/642/files is more feasible
        * we can just directly use the direct_connection channel without having to feed it through ChannelResult somehow.
    * returning proper error messages to the redis client that describe what went wrong should now be easier.
2. the message sending functions no longer take a one_tx and instead just construct one internally and return a `Result<ResponseFuture>`
    * The logic for immediate_responder and send_frame_response now live in short_circuit
    * The benefit here should be obvious, the API for sending a response back to the client is a lot simpler and doesn't require constructing oneshots all over the place.
    * This will result in slightly more allocations due to the Box in ResponseFuture - but I think the improved API is worth it.
         + I think there might be ways to remove the Box that we can investigate later.
    * If you are having trouble reviewing this PR let me know and I can pull this part into its own prereq PR.
    
There should be no functional changes caused by this PR, with the exception of some improved or changed error messages.
Please report any possible functional changes you might find.

### DIFF
--- a/shotover-proxy/src/transforms/redis/sink_cluster.rs
+++ b/shotover-proxy/src/transforms/redis/sink_cluster.rs
@@ -8,16 +8,15 @@ use crate::transforms::redis::RedisError;
 use crate::transforms::redis::TransformError;
 use crate::transforms::util::cluster_connection_pool::{Authenticator, ConnectionPool};
 use crate::transforms::util::{Request, Response};
-use crate::transforms::ResponseFuture;
-use crate::transforms::CONTEXT_CHAIN_NAME;
-use crate::transforms::{Transform, Transforms, Wrapper};
+use crate::transforms::{ResponseFuture, Transform, Transforms, Wrapper, CONTEXT_CHAIN_NAME};
 use anyhow::{anyhow, bail, ensure, Context, Result};
 use async_trait::async_trait;
 use bytes::Bytes;
 use bytes_utils::string::Str;
 use derivative::Derivative;
 use futures::stream::FuturesUnordered;
-use futures::{Future, StreamExt, TryFutureExt};
+use futures::{StreamExt, TryFutureExt};
+use itertools::Itertools;
 use metrics::{counter, register_counter};
 use rand::rngs::SmallRng;
 use rand::seq::IteratorRandom;
@@ -27,8 +26,7 @@ use serde::Deserialize;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use tokio::sync::mpsc::UnboundedSender;
 use tokio::sync::oneshot;
-use tokio::time::timeout;
-use tokio::time::Duration;
+use tokio::time::{timeout, Duration};
 use tracing::{debug, error, info, trace, warn};
 
 const SLOT_SIZE: usize = 16384;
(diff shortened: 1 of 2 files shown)


shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -859,27 +856,21 @@ async fn receive_frame_response(receiver: oneshot::Receiver<Response>) -> Result
...
-    impl Future<Output = Result<Response>>,
-) {
+fn short_circuit(frame: RedisFrame) -> Result<ResponseFuture> {
     let (one_tx, one_rx) = oneshot::channel::<Response>();
-    (one_tx, async {
+
+    one_tx
+        .send(Response {
+            original: Message::from_frame(Frame::None),
+            response: Ok(vec![Message::from_frame(Frame::Redis(frame))]),
```
conorbros, line 865: Do we ever put more than one message in this vec? I couldn't find any case of it.

shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -378,66 +381,85 @@ impl RedisSinkCluster {
...
-                    vec![lookup.clone()]
-                } else {
-                    vec![]
-                },
-            ),
+            RoutingInfo::Slot(slot) => self.send_message_to_slot(slot, message).await,
             RoutingInfo::AllNodes(_) => {
-                ChannelResult::Channels(self.slots.nodes.iter().cloned().collect())
+                self.send_message_to_channels(
+                    &self.slots.nodes.iter().cloned().collect_vec(),
```
conorbros, line 419: Is it possible to not clone here?

shotover-proxy/src/transforms/redis/sink_cluster.rs
rukai: Very nice find!
I raised a separate PR https://github.com/shotover/shotover-proxy/pull/660/files

shotover-proxy/src/transforms/redis/sink_cluster.rs
rukai: It could be possible with a bit more fiddling and would be a nice win but this was as far as I got before leaving it as is.
image

conorbros (approved):

benbromhead:
I love this change. It really cleans things up and simplifies our approach.

benbromhead (approved):

benbromhead:
FYI I've approved but given the complexity it might be worth having another set of eyes over this

rukai:
Alright, if @XA21X gets a chance to look at this today then great, otherwise I'll just merge tomorrow.

shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -124,47 +122,56 @@ impl RedisSinkCluster {
...
-                    Some(message) => {
-                        self.send_error_response(one_tx, message).ok();
-                    }
-                    None => self.short_circuit(one_tx),
-                };
-                Box::pin(one_rx)
-            }
+            0 => self.send_error_response(
+                self.reason_for_no_nodes
+                    .unwrap_or("ERR Shotover RedisSinkCluster does not have any nodes"),
```
XA21X, line 157: I wouldn't expect to ever see this, but "cluster does not have any nodes" would be confusing (impossible?). 😂

XA21X (approved):

shotover-proxy/src/transforms/redis/sink_cluster.rs
rukai: Well it is `RedisSinkCluster does not have any nodes` which should have a different meaning.
Does it make more sense like that?

Or do you have a recommendation for a better error? (The current one isnt great but not sure how to improve it)

rukai: Maybe `RedisSinkCluster does not know of any nodes` makes it more clear?

rukai: I went ahead and changed it to `RedisSinkCluster does not know of any nodes`.
I'll queue the merge as is.

Issue #656, opened 2022-05-24
Cannot connect to shotover on redis cluster using lettuce.io client


carrie-instaclustr:
## Describe the bug
Cannot use the lettuce.io client to connect to shotover installed on a Redis cluster.

## To Reproduce
1. Have tried the below code block with on clusters with and without authentication, and with and without SSL. Have modified code depending on cluster configuration (taking out SSL for example).
```java
       final RedisURI redisUri = RedisURI.Builder.redis("<ip>") 
                .withPort(6378)
                .withAuthentication("<username>", "<password>")
                .withSsl(true)
                .withVerifyPeer(false)
                .build();
```
2.  Connecting directly to the redis port works, however trying to go via the shotover port results in 
```
Exception in thread "main" io.lettuce.core.RedisConnectionException: Unable to establish a connection to Redis Cluster
	at io.lettuce.core.cluster.RedisClusterClient.lambda$assertInitialPartitions$26(RedisClusterClient.java:922)
	at io.lettuce.core.cluster.RedisClusterClient.get(RedisClusterClient.java:941)
	at io.lettuce.core.cluster.RedisClusterClient.assertInitialPartitions(RedisClusterClient.java:921)
	at io.lettuce.core.cluster.RedisClusterClient.connect(RedisClusterClient.java:398)
	at io.lettuce.core.cluster.RedisClusterClient.connect(RedisClusterClient.java:375)
	at com.mycompany.app.App.main(App.java:63)
Caused by: io.lettuce.core.cluster.topology.DefaultClusterTopologyRefresh$CannotRetrieveClusterPartitions: Cannot retrieve cluster partitions from [redis://<user>:****************@<ip>:6378]

Details:
	[redis://<username>:****************@<ip>:6378]: Unable to connect to [<ip>:6378]: NOAUTH Authentication required (cached)
```
3. Connecting to the same cluster works using the redis-cli and the jedis client.
  
### Topology

3 node DC + 3 node dc in DR

// Your topology.yaml file
```yaml
sources:
  redis_prod:
    Redis: {batch_size_hint: 4, listen_addr: '0.0.0.0:6378', connection_limit: 20000,
      hard_connection_limit: true}
chain_config:
  redis_chain:
  - Tee:
      behavior: Ignore
      buffer_size: 10000
      chain:
      - QueryTypeFilter: {filter: Read}
      - Coalesce:
          max_behavior:
            CountOrWait: [2000, 100]
      - Tee:
          buffer_size: 100
[...]

PR #659, opened 2022-05-26
RedisSinkCluster handle attempts to use RESP3

rukai:
Closes https://github.com/shotover/shotover-proxy/issues/656

The way to enable RESP3 is to send a "HELLO 3" message - https://redis.io/commands/hello/
The lettuce client attempts that by default but fallsback to resp2 if the error message contains "unknown" - an attempt to match against the `-ERR unknown command 'hello'` error message.

I believe the best way for shotover's RedisSinkCluster transform to handle RESP3 is to just tell clients that we dont support it.
Handling RESP3 would be fairly tricky due to the extra state that needs to be taken into account in our connection pool.

RedisSinkSingle on the other hand could much more easily support RESP3 but doesnt yet.
Currently RedisSinkSingle will also break if it receives a "HELLO 3" but I would rather add proper RESP3 support there and dont want to have to force message parsing just to force rejection of "HELLO 3".
So that problem is left for later.

Relevant lettuce logic can be seen here:
* https://github.com/lettuce-io/lettuce-core/blob/5d3310d79967a580d172af7f6d9718ec9ed0325f/src/main/java/io/lettuce/core/RedisHandshake.java#L262
* https://github.com/lettuce-io/lettuce-core/blob/5d3310d79967a580d172af7f6d9718ec9ed0325f/src/main/java/io/lettuce/core/RedisHandshake.java#L102

### DIFF
--- a/shotover-proxy/src/transforms/redis/sink_cluster.rs
+++ b/shotover-proxy/src/transforms/redis/sink_cluster.rs
@@ -495,7 +495,7 @@ impl RedisSinkCluster {
         warn!("Could not route request - short circuiting");
         if let Err(e) = self.send_error_response(
             one_tx,
-            "ERR Shotover RedisSinkCluster does not not support this command used in this way",
+            "ERR unknown command - Shotover RedisSinkCluster does not not support this command",
         ) {
             trace!("short circuiting - couldn't send error - {:?}", e);
         }
@@ -664,6 +664,7 @@ impl RoutingInfo {
             // We just need a single redis node to handle this for us so shotover can pretend to be a single node.
             // So we just pick a node at random.
             b"ECHO" | b"PING" => RoutingInfo::Random,
+            b"HELLO" => RoutingInfo::Unsupported,
             _ => match args.get(1) {
                 Some(key) => RoutingInfo::for_key(key).unwrap_or(RoutingInfo::Unsupported),
                 None => RoutingInfo::Random,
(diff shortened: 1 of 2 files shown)


XA21X (approved):
LGTM. It's a bit surprising that Lettuce checks a specific word in the error.

shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -495,7 +495,7 @@ impl RedisSinkCluster {
         warn!("Could not route request - short circuiting");
         if let Err(e) = self.send_error_response(
             one_tx,
-            "ERR Shotover RedisSinkCluster does not not support this command used in this way",
+            "ERR unknown command - Shotover RedisSinkCluster does not not support this command",
```
conorbros, line 498: Should this be `unsupported command` instead?

shotover-proxy/tests/redis_int_tests/basic_driver_tests.rs
```diff
@@ -150,7 +150,20 @@ async fn test_time_cluster(connection: &mut Connection) {
             .unwrap_err()
             .detail()
             .unwrap(),
-        "Shotover RedisSinkCluster does not not support this command used in this way".to_string(),
+        "unknown command - Shotover RedisSinkCluster does not not support this command".to_string(),
```
conorbros, line 153: Same here.

conorbros (approved):
<s>Just the one question,</s> LGTM.

shotover-proxy/src/transforms/redis/sink_cluster.rs
XA21X: [`unknown` is actually necessary.](https://github.com/lettuce-io/lettuce-core/blob/main/src/main/java/io/lettuce/core/RedisHandshake.java#L262)

conorbros: Ah, thanks for that.

PR #660, opened 2022-05-26
store response as single Message instead of Vec<Message>

rukai:
Only one message was ever stored and allowing for multiple messages is incorrect when the `original` field only has one Message.


conorbros (approved):

benbromhead:
Doesn't this break redis pipelining?

rukai:
I dont think it could "break" anything as nothing ever tries to store more than one message.
Maybe what your describing is a yet unimplemented use case for it?

benbromhead (approved):

PR #661, opened 2022-05-27
CQL parser rewrite

rukai:
A fork of https://github.com/shotover/shotover-proxy/pull/598 to keep it up to date with main and respond to review feedback


benbromhead:
Thanks for taking this on.

rukai:
Alright, I am happy for reviews to start again on this PR:
* All TODOs left in the code are planned to be checked in as is, but please do bring up any TODOs that you think need to be addressed immediately.
* The plan for the cache rewrite is to leave it in the PR until the last moment so that it can be used to review the usage of the new cassandra parser AST.
   + However there is no need to review the logic of the cache rewrite, the cache transform will be entirely removed from this PR before checking in. A separate PR will be made to reintroduce the cache transform
       - This will allow reviewers to approve the cache and parser changes separately allowing for better quality reviews.

shotover-proxy/tests/helpers/mod.rs
```diff
@@ -49,6 +49,29 @@ impl ShotoverManager {
         }
     }
 
+    #[allow(dead_code)] // to make clippy happy
+    pub fn from_topology_file_without_observability(topology_path: &str) -> ShotoverManager {
```
conorbros, line 53: Can we pull this out?

shotover-proxy/tests/cassandra_int_tests/cache/assert.rs
```diff
@@ -0,0 +1,82 @@
...
+        if let DebugValue::Counter(counter) = value {
+            if key.key().name() == "cache_miss" && *counter > result {
+                result = *counter;
+            }
+        }
+    }
+    result
+}
+
+fn assert_increment(
```
conorbros, line 20: Are we able to pull out the changes to the cache tests?

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -74,10 +75,14 @@ fn test_source_tls_and_single_tls() {
 #[test]
 #[serial]
 fn test_cassandra_redis_cache() {
+    let recorder = DebuggingRecorder::new();
+    let snapshotter = recorder.snapshotter();
```
conorbros, line 79: And these changes?

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -320,7 +284,10 @@ impl CassandraFrame {
...
 #[derive(PartialEq, Debug, Clone)]
 pub enum CassandraOperation {
-    Query { query: CQL, params: QueryParams },
+    Query {
+        query: Box<CassandraStatement>,
+        params: QueryParams,
+    },
     Result(CassandraResult),
     Error(ErrorBody),
     // operations for protocol negotiation, should be ignored by transforms
```
benbromhead, line 293: Should be ignored by *most* transforms

shotover-proxy/tests/cassandra_int_tests/mod.rs
rukai: Yep, I'll pull all of these out at the end once we are happy with the rest of the PR as per https://github.com/shotover/shotover-proxy/pull/661#issue-1250234838
Or would I better off completely pulling out the cache changes into a dependent PR right now and just letting people refer to the other PR to see how the new parser API works in transforms?

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
```diff
@@ -24,50 +27,76 @@ impl CassandraPeersRewriteConfig {
...
 }
 
-fn is_system_peers(message: &mut Message) -> bool {
-    if let Some(Frame::Cassandra(_)) = message.frame() {
-        if let Some(namespace) = message.namespace() {
-            if namespace.len() > 1 {
-                return namespace[0] == "system" && namespace[1] == "peers_v2";
+/// determine if the message contains a SELECT from `system.peers_v2` that includes the `native_port` column
+/// return a list of column names (or their alias) for each `native_port`.
+fn extract_native_port_column(peer_table: &FQName, message: &mut Message) -> Vec<Identifier> {
```
benbromhead, line 73: This will need some perf revalidation, @conorbros had to spend some time getting the fast path (e.g. exit early if it's not a query we are interested in) right previously

shotover-proxy/src/transforms/protect/crypto.rs
```diff
@@ -38,21 +38,24 @@ pub async fn encrypt(
         kek_id: sym_key.key_id,
     };
 
-    Ok(SQLValue::SingleQuotedString(serde_json::to_string(
-        &protected,
-    )?))
+    // TODO: investigate using storing as blob instead of text and then use textAsBlob here
+    // We should functionally verify which has better performance before sticking with one.
```
benbromhead, line 42: +1

shotover-proxy/src/frame/cassandra.rs
rukai: Resolved in https://github.com/shotover/shotover-proxy/pull/679

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
rukai: I dont think this transform ever had a fast path.
A proper fast path would need to skip query parsing and nothing of that sort was ever implemented.
Maybe you are thinking of one of the many other cassandra optimizations conor did e.g. https://github.com/shotover/shotover-proxy/pull/637

A fast path would be useful but we should investigate as a follow up to this PR.

shotover-proxy/src/transforms/protect/crypto.rs
rukai: Are you requesting that I resolve this TODO before we land the PR?

shotover-proxy/example-configs/cassandra-redis-cache/topology.yaml
```diff
@@ -9,10 +9,10 @@ chain_config:
         caching_schema:
           test_cache_keyspace_batch_insert.test_table:
             partition_key: [id]
-            range_key: [id]
+            range_key: []
           test_cache_keyspace_simple.test_table:
             partition_key: [id]
-            range_key: [id]
+            range_key: []
```
conorbros, line 15: What does this change do?

conorbros, line 15: What does this change do?

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -483,40 +457,134 @@ impl CassandraOperation {
...
+        CassandraType::Bigint(n)
+    } else if let Ok(n) = f64::from_str(value) {
+        CassandraType::Double(n)
+    } else if let Ok(uuid) = Uuid::parse_str(value) {
+        CassandraType::Uuid(uuid)
+    } else if let Ok(ipaddr) = IpAddr::from_str(value) {
+        CassandraType::Inet(ipaddr)
+    } else {
+        CassandraType::Null
+    }
```
conorbros, line 555: Could we write this as: 
```rust
    if value.starts_with('\'') || value.starts_with("$$") {
        CassandraType::Varchar(Operand::unescape(value))
    } else if value.starts_with("0X") || value.starts_with("0x") {
        hex::decode(&value[2..])
            .map(|x| CassandraType::Blob(Blob::from(x)))
            .unwrap_or(CassandraType::Null)
    }
   
    if let Ok(n) = i64::from_str(value) {
        CassandraType::Bigint(n)
    } else if let Ok(n) = f64::from_str(value) {
        CassandraType::Double(n)
    } else if let Ok(uuid) = Uuid::parse_str(value) {
        CassandraType::Uuid(uuid)
    } else if let Ok(ipaddr) = IpAddr::from_str(value) {
        CassandraType::Inet(ipaddr)
    } else {
        CassandraType::Null
    }
```

I think it more clearly illustrates that we must check the first 2 cases first and the order of the others doesn't matter.

rukai: I dont see why the order of the others doesnt matter.
I'm pretty sure swapping f64::from_str and i64::from_str would break things.

conorbros: True, my mistake.

shotover-proxy/example-configs/cassandra-redis-cache/topology.yaml
rukai: I dont know what it used to do but I believe with the current changes partition_key and range_key just alters whether a column goes in the hash's key or the hash's field.
e.g.
`hset "table:'foo'" "* WHERE " $CACHE`
instead of 
`hset "table'" "* WHERE id=foo " $CACHE`
I dont actually know why we want this though.


However I'm pretty sure that having partition_key and range_key both containing `id` doesnt make sense, so better to just remove from range_key.

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
conorbros: +1 re follow up PR

conorbros (approved):
LGTM. I ran the latte benchmarks against main and it appears there is no significant perf differences. Nice work @rukai will be good to get this done.

shotover-proxy/src/transforms/protect/crypto.rs
benbromhead: no lets raise as an issue and resolve as we are fixing up the crypto stuff later

shotover-proxy/src/transforms/redis/cache.rs
```diff
@@ -1,457 +1,555 @@
...
-use sqlparser::ast::{Assignment, BinaryOperator, Expr, Ident, Query, SetExpr, Statement, Value};
-use std::borrow::Borrow;
-use std::collections::HashMap;
-
-const TRUE: [u8; 1] = [0x1];
-const FALSE: [u8; 1] = [0x0];
+use std::collections::{BTreeMap, HashMap};
+use tracing_log::log::{error, warn};
+
+/// Data is stored in Redis as a Hash (hset/hget) and constructed from the cassandra SELECT statement
```
benbromhead, line 21: I'm not reviewing the cache changes for this PR, but I understand they are included as part of the CQL rewrite

shotover-proxy/src/transforms/protect/crypto.rs
rukai: Sounds good, I added it to https://github.com/shotover/shotover-proxy/issues/676

shotover-proxy/tests/cassandra_int_tests/mod.rs
benbromhead: I've just ignored the cache stuff for this review, but it has been helpful to see the changes illustrated.

Up to you whether you pull out the cache changes now or just before merge

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
benbromhead: cool I'm happy with that

benbromhead (approved):
LGTM

Thanks for picking this up and great work, I'll let you manage the cache side of the PR as you see fit.

PR #662, opened 2022-05-27
Pushed messages

conorbros:
Opening for some stages early feedback. Still working out some bugs.


conorbros:
Hmm, can't get that test to fail locally.

shotover-proxy/examples/cassandra_bench.rs
```diff
@@ -3,14 +3,17 @@ use test_helpers::docker_compose::DockerCompose;
...
+/// cargo run --release --example cassandra_bench -- --docker-compose example-configs/cassandra-passthrough/docker-compose-4.0.yml --topology example-configs/cassandra-passthrough/topology.yaml -r 1000
 /// or
-/// cargo run --release --example cassandra_bench -- --config-dir tests/test-configs/cassandra-passthrough-parse-request -r 1000
+/// cargo run --release --example cassandra_bench -- --docker-compose tests/test-configs/cassandra-passthrough-parse-request/docker-compose.yml --topology tests/test-configs/cassandra-passthrough-parse-request/topology.yaml -r 1000
 #[derive(Parser, Clone)]
 #[clap(name = "cassandra_bench", arg_required_else_help = true)]
 pub struct Args {
     #[clap(short, long)]
-    pub config_dir: String,
+    pub docker_compose: String,
```
rukai, line 13: Do these changes relate to the rest of the PR? Can they be pulled into a separate PR?

shotover-proxy/src/transforms/cassandra/sink_single.rs
```diff
@@ -44,6 +44,7 @@ pub struct CassandraSinkSingle {
     chain_name: String,
     failed_requests: Counter,
     tls: Option<TlsConnector>,
+    pushed_messages_tx: Option<tokio::sync::mpsc::UnboundedSender<Messages>>,
```
rukai, line 47: `tokio::sync::mpsc::UnboundedSender` can come from a use.

shotover-proxy/src/transforms/mod.rs
```diff
@@ -521,6 +584,27 @@ pub trait Transform: Send {
...
+    /// This transform method is not the same request/response model as the other transform method.
+    /// This method processes one pushed message before sending it in reverse on the chain back to the source.
+    ///
+    /// You can modify the messages in the wrapper struct to achieve your own designs. Your transform can
+    /// also modify the response from `message_wrapper.call_next_transform_rev` if it needs to. As long as the message
+    /// carries on through the chain, it will function correctly.
+    ///
+    /// ## Invariants
+    /// * _Non-terminating_ - Your `transform_rev` method should not be terminating as the messages should get passed back to the source, where they will terminate.
+    /// * _Message count_ - you can modify pushed messages but you should not remove or add them.
```
rukai, line 600: What about message ordering?

rukai: Also... Is there actually a requirement for message count to remain the same here?
Its not like the protocol has to match requests up with responses here

shotover-proxy/src/transforms/cassandra/sink_single.rs
rukai, line 47: Does this get used anywhere?

shotover-proxy/examples/cassandra_bench.rs
conorbros: I needed to change it because of the different versions of the docker-compose added to `cassandra-passthrough`. I'll pull those changes out though.

shotover-proxy/src/transforms/cassandra/sink_single.rs
conorbros: In the constructor of `CassandraConnection`.

shotover-proxy/src/transforms/cassandra/sink_single.rs
```diff
@@ -167,4 +170,11 @@ impl Transform for CassandraSinkSingle {
     fn is_terminating(&self) -> bool {
         true
     }
+
+    fn add_pushed_messages_tx(
+        &mut self,
+        pushed_messages_tx: tokio::sync::mpsc::UnboundedSender<Messages>,
```
rukai, line 176: and here

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
```diff
@@ -1459,6 +1469,111 @@ fn test_source_tls_and_single_tls() {
...
+        .build()
+        .await
+        .unwrap();
+
+    let session = TcpSessionBuilder::new(RoundRobinLoadBalancingStrategy::new(), config).build();
+    let mut event_recv = session.create_event_receiver();
+
+    let mut tries = 0;
+    loop {
+        if tries > 3 {
```
rukai, line 1500: Why do we have a retries system?
Shouldnt the timeout on event_recv.recv() be enough?

shotover-proxy/tests/test-configs/cassandra-peers-rewrite/docker-compose-3.11-cassandra.yaml
```diff
@@ -44,3 +46,14 @@ services:
     environment: *environment
     depends_on:
       - cassandra-one
+
+  cassandra-three:
```
rukai, line 50: Why have we introduced a third cassandra instance here?

conorbros: So we can kill it to test the events.

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
conorbros: I need to confirm this with someone that has knowledge of Cassandra internals, but I tested it with just a straight connection to Cassandra with the driver. It's possible for Cassandra to lose events and not deliver them to clients that are subscribed, so I've allowed 3 tries.

shotover-proxy/tests/test-configs/cassandra-peers-rewrite/docker-compose-3.11-cassandra.yaml
rukai: Oh I see, makes sense :+1:

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
rukai: I see, that research would be much appreciated.
I would want to see some kind of documentation or demonstration of this behavior before allowing this retry.
We dont want our tests to allow flakiness to be introduced into shotover

shotover-proxy/src/server.rs
```diff
@@ -436,6 +445,8 @@ impl<C: Codec + 'static> Handler<C> {
         while !self.shutdown.is_shutdown() {
             // While reading a request frame, also listen for the shutdown signal
             debug!("Waiting for message");
+            let mut reverse_chain = false;
```
rukai, line 448: We should be able to rewrite without this extra boolean state:
image

rukai: oh wait, my suggestion moves the chain processing within the timeout!
I dont think thats a good idea.

conorbros: Should this be resolved then?

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -331,3 +340,102 @@ async fn test_cassandra_request_throttling() {
...
+
+    let n: u8 = rng.gen();
+    let create_ks = format!("CREATE KEYSPACE IF NOT EXISTS test_ks_{} WITH REPLICATION = {{ 'class' : 'SimpleStrategy', 'replication_factor' : 1 }};", n);
+
+    session
+        .query(create_ks)
+        .await
+        .expect("Keyspace creation error");
+
+    match timeout(Duration::from_secs(10), event_recv.recv()).await {
```
rukai, line 380: unwrap this

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -331,3 +340,102 @@ async fn test_cassandra_request_throttling() {
...
+        Ok(recvd) => {
+            if let Ok(event) = recvd {
```
rukai, line 382: unwrap this

shotover-proxy/tests/cassandra_int_tests/basic_driver_tests.rs
conorbros: The issue was that the Cassandra driver was not finished registering for the event before the keyspace was created and the event fired off.

The issue ideally should be fixed within the driver itself. It should not return a `Session` object that is technically not ready to use yet. I can add a sleep to the test for the time being.

shotover-proxy/Cargo.toml
```diff
@@ -63,7 +63,7 @@ halfbrown = "0.1.11"
 
 # Transform dependencies
 redis-protocol = { version = "4.0.1", features = ["decode-mut"] }
-cassandra-protocol = { git = "https://github.com/krojew/cdrs-tokio" }
+cassandra-protocol = { git = "https://github.com/krojew/cdrs-tokio", rev = "dd564cbb6e348952c8b8a9c5c87ff4f98c28077e" }
```
rukai, line 66: Why are we pinning here? was there a regression in a later commit?

If you just want to pin here so that a `cargo update` wont hit breaking changes, then thats fine but just wanted to understand the reasoning.

rukai: A quick test reveals that yes the latest commit does have breaking changes in it.

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -342,3 +349,46 @@ async fn test_cassandra_request_throttling() {
...
+
+    sleep(Duration::from_secs(3)).await; // let the driver finish connecting to the cluster and registering for the events
+
+    let create_ks = "CREATE KEYSPACE IF NOT EXISTS test_events_ks WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 };";
+
+    session.query(create_ks).await.unwrap();
+
+    match timeout(Duration::from_secs(10), event_recv.recv()).await {
+        Ok(recvd) => {
+            if let Ok(event) = recvd {
```
rukai, line 386: I think this should be unwrapped to avoid the test silently failing.

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -342,3 +349,46 @@ async fn test_cassandra_request_throttling() {
...
+    let mut event_recv = session.create_event_receiver();
+
+    sleep(Duration::from_secs(3)).await; // let the driver finish connecting to the cluster and registering for the events
+
+    let create_ks = "CREATE KEYSPACE IF NOT EXISTS test_events_ks WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 };";
+
+    session.query(create_ks).await.unwrap();
+
+    match timeout(Duration::from_secs(10), event_recv.recv()).await {
+        Ok(recvd) => {
```
rukai, line 385: This could also be written more simply as an unwrap.

shotover-proxy/src/transforms/mod.rs
benbromhead: I think we can allow transform authors to add or remove messages? Given it's not expecting request/reponse pairs?

shotover-proxy/src/transforms/cassandra/sink_single.rs
conorbros: Can this be resolved?

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
```diff
@@ -66,6 +68,22 @@ impl Transform for CassandraPeersRewrite {
...
+
+    async fn transform_rev<'a>(&'a mut self, mut message_wrapper: Wrapper<'a>) -> ChainResponse {
+        for message in &mut message_wrapper.messages {
+            if let Some(Frame::Cassandra(frame)) = message.frame() {
+                if let Event(ServerEvent::StatusChange(StatusChange { addr, .. })) =
+                    &mut frame.operation
+                {
+                    addr.addr.set_port(self.port);
+                }
+                message.invalidate_cache();
```
rukai, line 80: The cache invalidation should go in the innermost if let

shotover-proxy/src/transforms/mod.rs
```diff
@@ -521,6 +582,64 @@ pub trait Transform: Send {
...
+    ///         info!("{} Event content: {:?}", self.counter, message_wrapper.messages);
+    ///         let response = message_wrapper.call_next_transform_rev().await;
+    ///         response
+    ///     }
+    /// }
+    /// ```
+    ///
+    /// In this example `counter` will contain the count of the number of messages seen for this connection.
+    /// Wrapping it in an [`Arc<Mutex<_>>`](std::sync::Mutex) would make it a global count of all messages seen by this transform.
+    async fn transform_rev<'a>(&'a mut self, message_wrapper: Wrapper<'a>) -> ChainResponse {
```
rukai, line 638: I wonder if we should call this function `transform_event`?
While it certainly does go in reverse, rev doesnt explain the type of messages that it receives.
Like it might sound like all message replies come through this transform.
Whereas `transform_event` still implies that the pipeline goes in reverse because events go from the server to the client.
Maybe even `transform_server_event` would make it more obvious that its an event coming from the server?

`event` refers to the type of messages that cassandra sends down this pipeline.
The equivalent for redis would be `publish` messages, that doesnt quite explain whats going on as well, so I think `event` is the better general name.

conorbros: Perhaps `pushed`? I think that describes both well.

rukai: Does pushed imply a direction?
I think a client could be considered pushing to a server?
`event` implies coming from the server which is useful property.
But maybe I only think that because of my knowledge of cassandra...

conorbros: "Pushed" as in, the messages are not a response to a request, but rather pushed from the server to the client. Think "push notifications" on a mobile phone.

rukai: Right true, I agree it is natural to consider it from the clients perspective so pushed sounds like it is being pushed from the server.
pushed sounds reasonable to me :+1:

shotover-proxy/src/server.rs
```diff
@@ -211,8 +211,13 @@ impl<C: Codec + 'static> TcpCodecListener<C> {
...
             socket.set_nodelay(true)?;
 
+            let (pushed_messages_tx, pushed_messages_rx) =
+                tokio::sync::mpsc::unbounded_channel::<Messages>();
+
             let mut handler = Handler {
-                chain: self.chain.clone(),
+                chain: self
+                    .chain
+                    .clone_with_pushed_messages_tx(pushed_messages_tx.clone()),
```
rukai, line 220: Should this clone_with_pushed_messages_tx logic be moved into the constructor?
Seems hacky to have this important initialization to occur in some extra method called later on.

shotover-proxy/src/transforms/chain.rs
```diff
@@ -291,6 +291,38 @@ impl TransformChain {
...
+        self.chain_total.increment(1);
+        if result.is_err() {
+            self.chain_failures.increment(1);
+        }
+
+        histogram!("shotover_chain_latency", start.elapsed(),  "chain" => self.name.clone());
+        result
+    }
+
+    pub fn clone_with_pushed_messages_tx(
```
rukai, line 309: Can we write like this:
```rust
    pub fn clone_with_pushed_messages_tx(
        &self,
        pushed_messages_tx: tokio::sync::mpsc::UnboundedSender<Messages>,
    ) -> Self {
        let mut result = self.clone();

        for transform in &mut result.chain {
            transform.add_pushed_messages_tx(pushed_messages_tx.clone());
        }

        result
}
```

shotover-proxy/src/transforms/chain.rs
```diff
@@ -291,6 +290,34 @@ impl TransformChain {
         histogram!("shotover_chain_latency", start.elapsed(),  "chain" => self.name.clone(), "client_details" => client_details);
         result
     }
+
+    pub async fn process_request_rev(&mut self, mut wrapper: Wrapper<'_>) -> ChainResponse {
+        let start = Instant::now();
+        wrapper.reset_reverse(&mut self.chain);
```
rukai, line 296: This call actually reverses the order of self.chain.
That sounds wrong, wont that affect the order of non pushed messages?

rukai: I think this would be the best fix:
image

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -342,3 +349,41 @@ async fn test_cassandra_request_throttling() {
...
+
+    let create_ks = "CREATE KEYSPACE IF NOT EXISTS test_events_ks WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 };";
+
+    session.query(create_ks).await.unwrap();
+
+    let event = timeout(Duration::from_secs(10), event_recv.recv())
+        .await
+        .unwrap()
+        .unwrap();
+    assert!(matches!(event, ServerEvent::SchemaChange { .. }));
```
rukai, line 388: Can we write like this to ensure that the server event comes from the creation of the keyspace we just did:
```rust
    assert_eq!(
        event,
        ServerEvent::SchemaChange(SchemaChange {
            change_type: SchemaChangeType::Created,
            target: SchemaChangeTarget::Keyspace,
            options: SchemaChangeOptions::Keyspace("test_events_ks".to_string())
        })
    );
 ```

File: shotover-proxy/src/transforms/chain.rs (hunk 17 above)
Comment (conorbros) [reply shotover-proxy/src/transforms/chain.rs]: Yeah, you are correct I believe. That code was written when the chain was being cloned before calling that method so the code was leftover.

File: shotover-proxy/src/transforms/cassandra/connection.rs [hunk 19]
```diff
@@ -108,14 +116,23 @@ async fn rx_process<C: CodecReadHalf, T: AsyncRead>(
                 match maybe_req {
                     Ok(req) => {
                         for m in req {
+                            if let Some(ref pushed_messages_tx) = pushed_messages_tx {
```
rukai, line 119: We can use an `else if` instead of a continue here, makes the logic easier to follow.
`unwrap`ing the `as_raw_bytes` ensures we dont silently fail here if a refactor causes the message to be parsed.
Also pay attention to the formatting here, rust fmt isnt running because we are in a `select!` macro, if we wanted we could avoid that by pulling the logic into a function.
```rust
                            if let Ok(Opcode::Event) = cassandra::raw_frame::get_opcode(m.as_raw_bytes().unwrap()) {
                                if let Some(ref pushed_messages_tx) = pushed_messages_tx {
                                    pushed_messages_tx.send(vec![m]).unwrap();
                                }
                            }
                            else if let Some(stream_id) = m.stream_id() {
                                match return_channel_map.remove(&stream_id) {
                                    None => {
                                        return_message_map.insert(stream_id, m);
                                    },
                                    Some((return_tx, original)) => {
                                        return_tx.send(Response {original, response: Ok(m) })
                                            .map_err(|_| anyhow!("couldn't send message"))?;
                                    }
                                }
                            }
```

Looking at this code i'm pretty sure we previously handled events by sending them through the regular pipeline which would have caused transform invariants to break, so I dont think the cassandra_cpp client has been using events under the hood as we previously theorized.

shotover-proxy/src/server.rs
```diff
@@ -470,18 +485,28 @@ impl<C: Codec + 'static> Handler<C> {
             debug!("Received raw message {:?}", messages);
             debug!("client details: {:?}", &self.client_details);
 
-            match self
```
rukai, line 473: We can simplify this by pulling out wrapper creation:
```rust
            let wrapper = Wrapper::new_with_client_details(
                messages,
                self.client_details.clone(),
                self.chain.name.clone(),
            );
            let chain_result = if reverse_chain {
                self.chain.process_request_rev(wrapper).await
            } else {
                self.chain
                    .process_request(wrapper, self.client_details.clone())
                    .await
            };
```

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
```diff
@@ -66,6 +68,22 @@ impl Transform for CassandraPeersRewrite {
...
+                if let Event(ServerEvent::StatusChange(StatusChange { addr, .. })) =
+                    &mut frame.operation
+                {
+                    addr.addr.set_port(self.port);
+                    message.invalidate_cache();
+                }
+            }
+        }
+
+        let response = message_wrapper.call_next_transform_pushed().await?;
```
rukai, line 84: What if we removed the need for this call, directly returned results in transform_pushed and had the chain manage feeding the result of one transform into the next?
My reasoning:
1. simpler API for users that better represents the flow of data, its truly one way now.
2. branching transforms with side chains wont be able to handle cassandra events anyway, so no benefit to having that flexibility here.

My only fear here is that reason 2 is incorrect and we might end up needing to handle at least some of the branching transforms. Maybe redis pubsub could support branching transforms?

So... in conclusion, unless you can see that 2 is definitely true, I'm happy to just land as is for now, we can always refactor later.

shotover-proxy/src/transforms/mod.rs
```diff
@@ -369,6 +404,27 @@ impl<'a> Wrapper<'a> {
...
+        };
+
+        let transform_name = transform.get_name();
+        let chain_name = self.chain_name.clone();
+
+        let start = Instant::now();
+        let result = CONTEXT_CHAIN_NAME
+            .scope(chain_name, transform.transform_pushed(self))
+            .await;
+        counter!("shotover_transform_total", 1, "transform" => transform_name);
```
rukai, line 420: These metrics should probably be seperate to regular transform metrics `shotover_transform_pushed_total` or something.
My reasoning is that the values would be completely different and would mess up the regular data. e.g. latency would be wayyyy smaller.

shotover-proxy/src/transforms/mod.rs
```diff
@@ -521,6 +577,64 @@ pub trait Transform: Send {
...
+    ///
+    /// impl Printer {
+    ///     pub fn new() -> Printer {
+    ///         Printer { counter: 0 }
+    ///     }
+    /// }
+    ///
+    /// #[async_trait]
+    /// impl Transform for Printer {
+    ///     async fn transform<'a>(&'a mut self, message_wrapper: Wrapper<'a>) -> ChainResponse {
```
rukai, line 622: This is documenting `transform` in the `transform_pushed` docs.

Maybe there should just be a single code snippet example on the trait itself?

Although maybe we dont need any of these code snippets at all, we could always just refer to an actual transform implementation? When we actually provide custom transforms to users I imagine the API will look quite different and we can create appropriate examples at that time.

shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
conorbros: I think we should keep it close to the original transform model for now. After we implement more we will have a better idea of what to do.

rukai: yep, that sounds good

shotover-proxy/src/transforms/mod.rs
conorbros: I agree with that. I'll remove it for now, no need to keep them as no one is reading the docs.rs at the moment.

rukai (approved):
LGTM, this approach makes sense to me.

As noted in some of the comments theres some API questions that can be further worked out once we have more uses of pushed messages.

benbromhead (approved):
LGTM

PR #664, opened 2022-05-30
Refactor Protect transform

rukai:
Refactor pulled out of https://github.com/shotover/shotover-proxy/pull/661

This should have no functional changes.
Technically the serialization format for protected values has changed making it a breaking change, (as can be seen in the changed integration test) but its impossible to use this transform in production yet anyway.

Changes included in this refactor:
* private_cached_fetch key_id an `&str` now to save an allocation as we never needed a String anyway (key_id only ever gets used in a `format!()`)
* KeyMaterial is moved into the key_management module - its pretty closely related to KeyManager, so makes sense to keep it together.
* All the encryption/decryption logic is moved into its own module and is reduced into a simple encrypt/decrypt pair of functions.
   + the logic is now much simpler, so we can follow the entire encryption or decryption process in just 15 lines instead of 50 lines spread across different functions
   + the logic is now stateless making it less bug prone. Previously we had to rely on asserts to know that encryption/decryption was actually occuring, now it just happens.

### DIFF
--- a/shotover-proxy/src/transforms/protect/aws_kms.rs
+++ b/shotover-proxy/src/transforms/protect/aws_kms.rs
@@ -1,4 +1,4 @@
-use crate::transforms::protect::KeyMaterial;
+use crate::transforms::protect::key_management::KeyMaterial;
 use anyhow::anyhow;
 use anyhow::Result;
 use bytes::Bytes;
--- a/shotover-proxy/src/transforms/protect/crypto.rs
+++ b/shotover-proxy/src/transforms/protect/crypto.rs
@@ -48,19 +48,19 @@ pub async fn decrypt(
     key_management: &KeyManager,
     key_id: &str,
 ) -> Result<MessageValue> {
-    let varchar = match value {
-        MessageValue::Varchar(varchar) => varchar,
+    let bytes = match value {
+        MessageValue::Bytes(bytes) => bytes,
         _ => bail!("expected varchar to decrypt but was {:?}", value),
     };
-    let protected: Protected = serde_json::from_str(varchar)?;
+    let protected: Protected = serde_json::from_slice(bytes)?;
 
     let sym_key = key_management
         .cached_get_key(key_id, Some(protected.enc_dek), Some(protected.kek_id))
         .await?;
 
     let decrypted_bytes = secretbox::open(&protected.cipher, &protected.nonce, &sym_key.plaintext)
-        .map_err(|_| anyhow!("couldn't open box"))?;
+        .map_err(|_| anyhow!("couldn't decrypt value"))?;
 
     //TODO make error handing better here - failure here indicates an authenticity failure
-    bincode::deserialize(&decrypted_bytes).map_err(|_| anyhow!("couldn't open box"))
+    bincode::deserialize(&decrypted_bytes).map_err(|_| anyhow!("couldn't decrypt value"))
 }
--- a/shotover-proxy/src/transforms/protect/crypto.rs
+++ b/shotover-proxy/src/transforms/protect/crypto.rs
@@ -1,0 +1,66 @@
+use crate::message::MessageValue;
+use crate::transforms::protect::key_management::KeyManager;
+use anyhow::{anyhow, bail, Result};
+use serde::{Deserialize, Serialize};
+use sodiumoxide::crypto::secretbox;
+use sodiumoxide::crypto::secretbox::Nonce;
+use sqlparser::ast::Value as SQLValue;
+
+// A protected value meets the following properties:
+// https://doc.libsodium.org/secret-key_cryptography/secretbox
+// This all relies on crypto_secretbox_easy which takes care of
+// all padding, copying and timing issues associated with crypto
+#[derive(Serialize, Deserialize)]
+struct Protected {
+    cipher: Vec<u8>,
+    nonce: Nonce,
+    enc_dek: Vec<u8>,
+    kek_id: String,
+}
+
+pub async fn encrypt(
+    value: &SQLValue,
+    key_management: &KeyManager,
+    key_id: &str,
+) -> Result<SQLValue> {
+    let value = MessageValue::from(&*value);
+
+    let sym_key = key_management.cached_get_key(key_id, None, None).await?;
+
+    let ser = bincode::serialize(&value)?;
+    let nonce = secretbox::gen_nonce();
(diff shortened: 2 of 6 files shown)


conorbros (approved):

shotover-proxy/src/transforms/protect/encrypt.rs
```diff
@@ -0,0 +1,66 @@
+use crate::message::MessageValue;
```
benbromhead, line 1: Can we change the name of the file to `cryptography.rs` or something as it does both encryption and decryption operations

shotover-proxy/src/transforms/protect/encrypt.rs
```diff
@@ -0,0 +1,66 @@
...
+    let cipher = secretbox::seal(&ser, &nonce, &sym_key.plaintext);
+
+    let protected = Protected {
+        cipher,
+        nonce,
+        enc_dek: sym_key.ciphertext_blob.to_vec(),
+        kek_id: sym_key.key_id,
+    };
+
+    Ok(SQLValue::SingleQuotedString(serde_json::to_string(
```
benbromhead, line 41: We used to store the Protected value in a blob / bytes type. Admittedly the prior state of protect was broken on the serialization side of things. Whats the rationale for storing as a string vs bytes?

shotover-proxy/src/transforms/protect/encrypt.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    let sym_key = key_management
+        .cached_get_key(key_id, Some(protected.enc_dek), Some(protected.kek_id))
+        .await?;
+
+    let decrypted_bytes = secretbox::open(&protected.cipher, &protected.nonce, &sym_key.plaintext)
+        .map_err(|_| anyhow!("couldn't open box"))?;
+
+    //TODO make error handing better here - failure here indicates an authenticity failure
+    bincode::deserialize(&decrypted_bytes).map_err(|_| anyhow!("couldn't open box"))
```
benbromhead, line 65: Given this will bubble up to a chain error back to the client. We should make these error messages a bit more informative. E.g. 

"Couldn't decrypt value"

We probably shouldn't give much more away from that at the client error as it can inform an attacker what part of the decrypt / deserialise stage they have gotten to. But we could look at trace level logging which will stay on the shotover side of things and provide some useful errors if we need to dig into it if you feel that it's worth having diagnostics here (which it might not be).

shotover-proxy/src/transforms/protect/encrypt.rs
```diff
@@ -0,0 +1,66 @@
...
+pub async fn encrypt(
+    value: &SQLValue,
+    key_management: &KeyManager,
+    key_id: &str,
+) -> Result<SQLValue> {
+    let value = MessageValue::from(&*value);
+
+    let sym_key = key_management.cached_get_key(key_id, None, None).await?;
+
+    let ser = bincode::serialize(&value)?;
```
benbromhead, line 30: as with decrypt should we customise this error message back to the client?

benbromhead (commented):
Much better structural change and more understandable. 

sodiumoxide our crypto lib is also deprecated. Given this is a breaking change on the serialization side of things. Let's make the jump to an actively developed rust crypto lib. RustCrypto and the nacl wrapper seems to be a possible path to take -> https://github.com/RustCrypto/nacl-compat

shotover-proxy/src/transforms/protect/encrypt.rs
rukai: I dont think we ever stored as a blob as there was no way to do that with the old sqlparser implementation.
That will become possible once we move to cql3-parser though.

shotover-proxy/src/transforms/protect/encrypt.rs
rukai: good point, done

shotover-proxy/src/transforms/protect/encrypt.rs
rukai: oh, wait, you are correct actually!

What we did was use MessageValue::Bytes in protect.
However shotover was internally converting that into a SQLValue::SingleQuotedString

What I will do is revert the change here to use a MessageValue::Bytes and then let the cql3-parser PR resolve the bytes/string issues.

rukai: actually, actually.
This specific line is fine.

This is equivalent to the old logic where MessageValue::Bytes was internally converted to a SQLValue::SingleQuotedString.
However now we are just directly using the SingleQuotedString.

However there is a line in `fn decrypt` that was expecting a MessageValue::Varchar instead of a MessageValue::Bytes, so I changed that to MessageValue::Bytes.
decrypt is broken anyway, but at least this is closer to something that works.

As before though, the cql3-parser PR actually fixes this brokenness and has a full integration test proving that encryption/decryption round trips correctly.

shotover-proxy/src/transforms/protect/encrypt.rs
rukai: Yep, thats a better error than the existing code had, ill use that.
I dont think any of these errors are routed to the client though.
I'm pretty sure shotover just immediately closes the connection and logs a tracing::error

shotover-proxy/src/transforms/protect/encrypt.rs
rukai: nah, I think we want to return as much information as possible when handling a a serialization error.
An error like this should never ever occur so if it does we will need as much information as possible to diagnose it.

The value is already encrypted at this point so no risk of leaking data by directly returning whatever the serialization error contains.

rukai:
> sodiumoxide our crypto lib is also deprecated. Given this is a breaking change on the serialization side of things. Let's make the jump to an actively developed rust crypto lib. RustCrypto and the nacl wrapper seems to be a possible path to take -> https://github.com/RustCrypto/nacl-compat

I'm going to leave that for future work, the true purpose of this PR is to reduce the size of the cql3-parser PR and I dont want to distract too much from that.

shotover-proxy/src/transforms/protect/encrypt.rs
benbromhead: ok cool

shotover-proxy/src/transforms/protect/encrypt.rs
benbromhead: ok cool, I'm happy for more detailed error messages then

PR #668, opened 2022-06-07
Cassandra events

conorbros:
pulled from #662


rukai (approved):
The unreachable is fine because `Opcode::Event` and `ResponseBody::Event` are mutually inclusive.

We will also need the parsed ServerEvent to implement the CassandraPeersRewrite transform regardless of the event/pubsub solution we arrive at.

LGTM

benbromhead (approved):

PR #669, opened 2022-06-08
remove depends_on from Cassandra docker clusters

conorbros:
@rukai noticed we don't need this and it gives us a speedup on test runtime.


rukai (approved):
Nice!

XA21X (approved):

PR #671, opened 2022-06-08
Update deps

rukai:
`metrics::clear_recorder();` was removed from metrics (technically just marked unsafe and hidden from docs) due to being dangerous to use.
So I changed our metrics init code to instead ignore errors due to metrics already being initialized.


conorbros (approved):

shotover-proxy/src/runner.rs
```diff
@@ -77,7 +77,11 @@ impl Runner {
     pub fn with_observability_interface(self) -> Result<Self> {
         let recorder = PrometheusBuilder::new().build_recorder();
         let handle = recorder.handle();
-        metrics::set_boxed_recorder(Box::new(recorder))?;
+
+        // Result::Err can only occur when set_boxed_recorder has already been called.
+        // So we just ignore the error with .ok() as integration tests will cause shotover
+        // to be initialized multiple times thus recalling this function.
+        metrics::set_boxed_recorder(Box::new(recorder)).ok();
```
rukai, line 84: Ah...
Probably this doesnt actually work because some of the state here gets dropped and deinitializes the recorder when shotover shutsdown

conorbros: Oops, I thought the test suite had passed when I approved.

rukai: I pushed a fix.
The only reason im considering this fix is because the issue existed before, the only difference then was the clear_recorder function wasnt marked as unsafe. refer to https://github.com/metrics-rs/metrics/pull/302
So this is actually an improvement as the unsafe call will now show up in an audit whereas before the issue would have gone unnoticed.

And I really have no idea how to go about fixing it properly.
The proper fix would probably be actually spawning shotover in a separate process, that would fix this issue and a very similar issue in tracing init.
But that has considerable downsides and implementation time.

rukai:
@conorbros would be good to rereview this considering there was a drastic change since your previous approval.

conorbros (approved):

shotover-proxy/tests/helpers/mod.rs
```diff
@@ -157,7 +157,10 @@ impl Drop for ShotoverManager {
     fn drop(&mut self) {
         // Must clear the recorder before skipping a shutdown on panic; if one test panics and the recorder is not cleared,
         // the following tests will panic because they will try to set another recorder
-        metrics::clear_recorder();
+        // TODO: This is unsafe due to a possible race condition and we are doing nothing to prevent that race condition.
+        // It is left as is because there are no reasonable solutions to this problem at the moment.
+        // A possible way to trigger this issue would be creating a ShotoverManager instance in one thread while dropping a ShotoverManager instance in another thread.
+        unsafe { metrics::clear_recorder() };
```
conorbros, line 163: Moving forward what can we do to remove this `unsafe`?

rukai: Ah good point, ill raise an issue for it.

conorbros: Thanks!

benbromhead (approved):

PR #673, opened 2022-06-09
Move cassandra_int_tests modules into separate files

rukai:
About time we split this up. (And it was causing some insane nesting in something I was working on)

Moves everything out of the basic_driver_tests.rs file.
The actual tests go into cassandra_int_tests/mod.rs
The helper modules go into their own files.

No code has been changed only moved to different files.


shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -1 +1,460 @@
...
+mod udt;
+
+use crate::helpers::cassandra::{assert_query_result, execute_query, run_query, ResultValue};
+use crate::helpers::ShotoverManager;
+use cassandra_cpp::{stmt, Batch, BatchType, Error, ErrorKind, Session};
+use futures::future::{join_all, try_join_all};
+use serial_test::serial;
+use test_helpers::docker_compose::DockerCompose;
+
+fn test_batch_statements(connection: &Session) {
```
conorbros, line 19: Might as well move this too so this file is just tests

rukai: ah good point

conorbros (approved):

benbromhead (approved):
lgtm!

PR #677, opened 2022-06-16
remove sodium oxide

conorbros:
Replaced with https://github.com/RustCrypto/AEADs

### DIFF
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -68,11 +68,12 @@ crc16 = "0.4.0"
 ordered-float = { version = "3.0.0", features = ["serde"] }
 
 #Crypto
-sodiumoxide = "0.2.5"
 rusoto_kms = "0.48.0"
 rusoto_signature = "0.48.0"
 csv = "1.1.6"
 strum_macros = "0.24"
+chacha20poly1305 = "0.9.0"
+rand_core = "0.6"
 
 [dev-dependencies]
 rayon = "1.5.1"
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -72,7 +72,7 @@ rusoto_kms = "0.48.0"
 rusoto_signature = "0.48.0"
 csv = "1.1.6"
 strum_macros = "0.24"
-chacha20poly1305 = "0.9.0"
+chacha20poly1305 = { version = "0.9.0", features = ["std"] }
 generic-array = { version = "0.14", features = ["serde"] }
 
 [dev-dependencies]
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -73,6 +73,7 @@ rusoto_signature = "0.48.0"
 csv = "1.1.6"
 strum_macros = "0.24"
 chacha20poly1305 = "0.9.0"
+generic-array = { version = "0.14", features = ["serde"] }
 
 [dev-dependencies]
 rayon = "1.5.1"
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -73,7 +73,6 @@ rusoto_signature = "0.48.0"
 csv = "1.1.6"
 strum_macros = "0.24"
 chacha20poly1305 = "0.9.0"
-rand_core = "0.6"
 
 [dev-dependencies]
 rayon = "1.5.1"
--- a/shotover-proxy/src/transforms/protect/crypto.rs
+++ b/shotover-proxy/src/transforms/protect/crypto.rs
@@ -14,7 +14,7 @@ use sqlparser::ast::Value as SQLValue;
 #[derive(Serialize, Deserialize)]
 struct Protected {
     cipher: Vec<u8>,
-    nonce: Vec<u8>,
+    nonce: Nonce,
     enc_dek: Vec<u8>,
     kek_id: String,
 }
@@ -37,7 +37,7 @@ pub async fn encrypt(
 
     let protected = Protected {
         cipher: ciphertext,
-        nonce: nonce.to_vec(),
+        nonce,
         enc_dek: sym_key.ciphertext_blob.to_vec(),
         kek_id: sym_key.key_id,
     };
--- a/shotover-proxy/src/transforms/protect/crypto.rs
+++ b/shotover-proxy/src/transforms/protect/crypto.rs
@@ -1,9 +1,10 @@
 use crate::message::MessageValue;
 use crate::transforms::protect::key_management::KeyManager;
 use anyhow::{anyhow, bail, Result};
+use chacha20poly1305::aead::{Aead, NewAead};
+use chacha20poly1305::{ChaCha20Poly1305, Key, Nonce};
+use rand::RngCore;
 use serde::{Deserialize, Serialize};
-use sodiumoxide::crypto::secretbox;
-use sodiumoxide::crypto::secretbox::Nonce;
 use sqlparser::ast::Value as SQLValue;
 
 // A protected value meets the following properties:
@@ -13,7 +14,7 @@ use sqlparser::ast::Value as SQLValue;
 #[derive(Serialize, Deserialize)]
 struct Protected {
     cipher: Vec<u8>,
-    nonce: Nonce,
+    nonce: Vec<u8>,
     enc_dek: Vec<u8>,
     kek_id: String,
 }
@@ -28,12 +29,15 @@ pub async fn encrypt(
     let sym_key = key_management.cached_get_key(key_id, None, None).await?;
 
     let ser = bincode::serialize(&value)?;
-    let nonce = secretbox::gen_nonce();
-    let cipher = secretbox::seal(&ser, &nonce, &sym_key.plaintext);
+    let nonce = gen_nonce();
+    let cipher = ChaCha20Poly1305::new(Key::from_slice(&sym_key.plaintext));
+    let ciphertext = cipher
+        .encrypt(&nonce, &*ser)
+        .map_err(|_| anyhow!("couldn't encrypt value"))?;
 
     let protected = Protected {
-        cipher,
-        nonce,
+        cipher: ciphertext,
+        nonce: nonce.to_vec(),
         enc_dek: sym_key.ciphertext_blob.to_vec(),
         kek_id: sym_key.key_id,
     };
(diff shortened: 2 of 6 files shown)


rukai:
as per the sodiumoxide docs: https://docs.rs/sodiumoxide/latest/sodiumoxide/crypto/secretbox/index.html
the secretbox module uses xsalsa20poly1305 

As per wikipedia: https://en.wikipedia.org/wiki/ChaCha20-Poly1305 and this SO post: https://crypto.stackexchange.com/questions/27007/why-should-one-prefer-salsa20-over-chacha
chacha20 is an improved version of salsa20

So the use of the chacha20poly1305 is equivalent to the existing sodiumoxide implementation

shotover-proxy/Cargo.toml
```diff
@@ -68,11 +68,12 @@ crc16 = "0.4.0"
 ordered-float = { version = "3.0.0", features = ["serde"] }
 
 #Crypto
-sodiumoxide = "0.2.5"
 rusoto_kms = "0.48.0"
 rusoto_signature = "0.48.0"
 csv = "1.1.6"
 strum_macros = "0.24"
+chacha20poly1305 = "0.9.0"
+rand_core = "0.6"
```
rukai, line 76: this seems unused

shotover-proxy/src/transforms/protect/crypto.rs
```diff
@@ -58,9 +62,29 @@ pub async fn decrypt(
...
 
     //TODO make error handing better here - failure here indicates an authenticity failure
     bincode::deserialize(&decrypted_bytes).map_err(|_| anyhow!("couldn't decrypt value"))
 }
+
+pub fn gen_key() -> Key {
+    let mut key_bytes = [0; 32];
+    let mut rng = rand::thread_rng();
+    rng.fill_bytes(&mut key_bytes);
+    let key = Key::from_slice(&key_bytes);
```
rukai, line 80: no need to use an intermediate variable here

shotover-proxy/src/transforms/protect/crypto.rs
```diff
@@ -1,9 +1,10 @@
...
 use anyhow::{anyhow, bail, Result};
+use chacha20poly1305::aead::{Aead, NewAead};
+use chacha20poly1305::{ChaCha20Poly1305, Key, Nonce};
+use rand::RngCore;
 use serde::{Deserialize, Serialize};
-use sodiumoxide::crypto::secretbox;
-use sodiumoxide::crypto::secretbox::Nonce;
 use sqlparser::ast::Value as SQLValue;
 
 // A protected value meets the following properties:
```
rukai, line 10: This comment still talks about libsodium.
Also does our new library take care of "all padding, copying and timing issues associated with crypto" as well?

shotover-proxy/src/transforms/protect/crypto.rs
```diff
@@ -13,7 +14,7 @@ use sqlparser::ast::Value as SQLValue;
 #[derive(Serialize, Deserialize)]
 struct Protected {
     cipher: Vec<u8>,
-    nonce: Nonce,
+    nonce: Vec<u8>,
```
rukai, line 17: ughh, really unfortunate the chacha crate was written before const generics, not sure how feasibile it is to just go and change something like that in a crypto crate under a lot of scrutiny.

That said we can still make Nonce and friends serializable by adding the following to our cargo.toml
`generic-array = { version = "0.14", features = ["serde"] }`
generic-array is what the chacha crate is using instead of const generics.

shotover-proxy/Cargo.toml
conorbros: Oops, leftover from an earlier implementation.

shotover-proxy/src/transforms/protect/crypto.rs
conorbros: I'm not sure whether padding refers to [this type of padding](https://datatracker.ietf.org/doc/html/rfc7539#:~:text=padding2%20%2D%2D%20the%20padding) or this [type](https://en.wikipedia.org/wiki/Padding_(cryptography)). Also unsure what copying refers to. I would assume that timing refers to the nonce which is handled.

@benbromhead ?

shotover-proxy/src/transforms/protect/crypto.rs
conorbros: I'll raise an issue on the repo for it but that's an alright workaround.

shotover-proxy/src/transforms/protect/crypto.rs
conorbros: Though it appears padding is handled with chacha20: https://loup-vaillant.fr/tutorials/chacha20-design

shotover-proxy/src/transforms/protect/key_management.rs
```diff
@@ -118,5 +117,5 @@ async fn private_cached_fetch(
 pub struct KeyMaterial {
     pub ciphertext_blob: Bytes,
     pub key_id: String,
-    pub plaintext: Key,
+    pub plaintext: Vec<u8>,
```
rukai, line 120: We can just directly use Key now:
image

shotover-proxy/src/transforms/protect/crypto.rs
conorbros: Alright, there's an issue on the repo for it which I've subscribed to. If it gets any updates I'll fix it in our codebase.

shotover-proxy/src/transforms/protect/crypto.rs
benbromhead: Padding is handled by chacha20.
The AEAD spec rfc 7539 claims the design is supposed to lend itself to a constant time implementation such that timing attacks are minimised. At this point I can only assume and rely on audits such as https://github.com/RustCrypto/AEADs/issues/426 that the implementation is kind of sound. 

At this point in time, my priority is to move to a maintained / popular crypto library. 

For what its worth secretbox in sodium oxide is just a re-export of xsalsa20poly1305 and the libsodium implementation is the actual AEAD implementation https://github.com/jedisct1/libsodium/blob/master/src/libsodium/crypto_secretbox/xsalsa20poly1305/secretbox_xsalsa20poly1305.c 

I'm happy that with this change we are getting the same thing we _thought_ we were getting previously. Ideally when we need to use this in anger, we will get an independent audit of our use (reading the RFC, how we generate our nonce appears to be the main area we can trip up badly).

shotover-proxy/src/transforms/protect/crypto.rs
```diff
@@ -58,9 +62,27 @@ pub async fn decrypt(
...
+
+pub fn gen_key() -> Key {
+    let mut key_bytes = [0; 32];
+    let mut rng = rand::thread_rng();
+    rng.fill_bytes(&mut key_bytes);
+    *Key::from_slice(&key_bytes)
+}
+
+pub fn gen_nonce() -> Nonce {
+    let mut rng = rand::thread_rng();
```
benbromhead, line 84: We need to use a cryptographically secure random byte source here. Previously we called gen_nonce which would end up calling out to libsodiums random bytes function. Does RustCryptos ecosystem give us a better alternative

benbromhead (changes requested):
Need to fix nonce source

shotover-proxy/src/transforms/protect/crypto.rs
rukai: huh.
rand 0.6.5 claims that thread_rng is cryptographically secure: https://docs.rs/rand/0.6.5/rand/rngs/struct.ThreadRng.html
However the latest rand 0.8.5 does not make any claims in regards to being cryptographically secure: https://docs.rs/rand/0.8.5/rand/rngs/struct.ThreadRng.html

benbromhead: I would feel safer if we called a rustcrypto rand source, even if it was just a re-export of rand::thread_rng

safety in numbers and all that :)

conorbros: I can use`rand_core` because it's re-exported but not `rand`.

benbromhead (approved):

shotover-proxy/src/transforms/protect/crypto.rs
```diff
@@ -58,9 +59,27 @@ pub async fn decrypt(
         .cached_get_key(key_id, Some(protected.enc_dek), Some(protected.kek_id))
         .await?;
 
-    let decrypted_bytes = secretbox::open(&protected.cipher, &protected.nonce, &sym_key.plaintext)
+    let nonce = Nonce::from_slice(&protected.nonce);
+    let cipher = ChaCha20Poly1305::new(Key::from_slice(&sym_key.plaintext));
```
rukai, line 63: same here

shotover-proxy/src/transforms/protect/key_management.rs
```diff
@@ -69,9 +69,8 @@ impl KeyManagerConfig {
             })),
             KeyManagerConfig::Local { kek, kek_id } => {
                 let decoded_base64 = base64::decode(&kek)?;
-                let kek =
-                    Key::from_slice(&decoded_base64).ok_or_else(|| anyhow!("Not a valid key"))?;
-                Ok(KeyManager::Local(LocalKeyManagement { kek, kek_id }))
+                let kek = Key::from_slice(&decoded_base64);
```
rukai, line 72: The kek comes from the topology.yaml so we need some kind of validation on the size of value decoded from base64 otherwise we will hit the panic inside Key::from_slice that used to be a Result::Err.

rukai: Oh I just realized that this logic would be better off going in the transform config validation anyway.
So we should probably leave as a panic and then raise an issue to add the relevant validation.

shotover-proxy/src/transforms/protect/local_kek.rs
```diff
@@ -18,12 +19,22 @@ pub struct DEKStructure {
 }
 
 impl LocalKeyManagement {
-    pub async fn get_key(&self, dek: Option<Vec<u8>>) -> Result<KeyMaterial> {
+    pub async fn get_key(
+        &self,
+        dek: Option<Vec<u8>>,
+        _kek_alt: Option<String>,
```
rukai, line 25: Lets not introduce an unused parameter

rukai (approved):
LGTM

PR #680, opened 2022-06-22
Update deps, lots of cassandra-protocol breaking changes

rukai:
Box changes are needed for clippy stack usage lints

Sticking `Version::V4` everywhere doesnt bother me because we were already doing that before.
We can address that properly when we add support for protocol v5

The new Duration type will need to be refined in the future to handle multiple protocols
But for now it lets us roundtrip cassandra Durations which is all we really care about.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -64,21 +64,21 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.57"
+version = "1.0.58"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "08f9b8508dccb7687a1d6c4ce66b2b0ecef467c94667de27d8d7fe1f8d2a9cdc"
+checksum = "bb07d2053ccdbe10e2af2995a2f116c1330396493dc1269f6a91d0ae82e19704"
+
+[[package]]
+name = "arc-swap"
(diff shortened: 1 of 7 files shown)


conorbros:
Cassandra protocol changes from v4 -> v5

https://github.com/apache/cassandra/blob/36d0c4e9ae195518df72d17fef97b420f451ded3/doc/native_protocol_v5.spec#L1437

rukai:
> Introduces outer framing format wrapping the "frames" of v4 and earlier, which are
    now referred to as "envelopes" (Sections 2.1, 2.2 and 2.3)
    
0.0 that doesnt sound fun.
Not a problem for this PR though.

conorbros:
Are we able to look at adding tests for a Cassandra cluster running the V5 protocol?

rukai:
I dont see why that would go in this PR.
We will need extensive changes to get V5 working.

conorbros:
I thought it might as well since we are upgrading to a cassandra-protocol version that supports it, but sure it can go later.

rukai:
My main concerns are unhardcoding all the `Protocol::V4`'s and whatever this new envelope/frame system is.

conorbros (approved):

benbromhead (approved):

Issue #666, opened 2022-05-31
test_cluster_replication is not robust to timing


rukai:
While working on https://github.com/shotover/shotover-proxy/pull/662 a very confusing test failure was encountered due to temporary regressions in performance.
I investigated the failure and figured out how to reproduce it.

By adding a sleep here:
![image](https://user-images.githubusercontent.com/5120858/171096649-13c1e915-0722-44ba-ac98-2713e66e1e94.png)

We get the following test failure:
```
---- redis_int_tests::basic_driver_tests::test_cluster_dr_redis stdout ----
thread 'redis_int_tests::basic_driver_tests::test_cluster_dr_redis' panicked at 'assertion failed: replication_connection.get::<&str, i32>(\"foo\").is_err()', shotover-proxy/tests/redis_int_tests/basic_driver_tests.rs:1090:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:143:14
   2: core::panicking::panic
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:48:5
   3: lib::redis_int_tests::basic_driver_tests::test_cluster_replication::{{closure}}
             at ./tests/redis_int_tests/basic_driver_tests.rs:1090:9
   4: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/future/mod.rs:91:19
   5: lib::redis_int_tests::basic_driver_tests::test_cluster_dr_redis::{{closure}}::{{closure}}
             at ./tests/redis_int_tests/basic_driver_tests.rs:1354:75
   6: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/future/mod.rs:91:19
   7: tokio::park::thread::CachedParkThread::block_on::{{closure}}
             at /home/rukai/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.18.2/src/park/thread.rs:263:54
   8: tokio::coop::with_budget::{{closure}}
[...]

PR #681, opened 2022-06-22
Fix intermittent failure in test_cluster_dr_redis

rukai:
Closes https://github.com/shotover/shotover-proxy/issues/666

The problem was that flush_when_millis_since_last_flush was too low. (10s)
So if for whatever reason the test case takes longer than 10s, then the flush occurs due to timeout before it occurs due to the message count.
As a result the foo key is set before 2000 iterations have completed and the assertion fails.

The fix is to set the timeout to 100000s (about 27 hours)

The changes to basic_driver_tests.rs is just to provide a more useful assert message.

You can verify this fixes the issue by following the instructions to reproduce from the issue.
First run on the main branch and then on this branch


benbromhead (approved):
lgtm

conorbros (approved):

PR #682, opened 2022-06-22
change peers rewrite port to u16

conorbros:
pulled from #662


rukai (approved):
I feel like I'd once heard that ports can be over u16 in some cases, but after searching online I must have completely imagined that

LGTM

conorbros:
Same here. But the std library says it's fine so 🤷

https://doc.rust-lang.org/std/net/enum.SocketAddr.html

XA21X (approved):

PR #685, opened 2022-06-27
store encrypted as blob

conorbros:
No noticeable change in perf

### DIFF
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -14,7 +14,7 @@ alpha-transforms = []
 
 [dependencies]
 pretty-hex = "0.3.0"
-hex = { version = "0.4.3", features = ["serde"] }
+hex = "0.4.3"
 tokio = { version = "1.14.0", features = ["full", "macros"] }
 tokio-util = { version = "0.7.0", features = ["full"] }
 tokio-stream = "0.1.2"
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -14,7 +14,7 @@ alpha-transforms = []
 
 [dependencies]
 pretty-hex = "0.3.0"
-hex = "0.4.3"
+hex = { version = "0.4.3", features = ["serde"] }
 tokio = { version = "1.14.0", features = ["full", "macros"] }
 tokio-util = { version = "0.7.0", features = ["full"] }
 tokio-stream = "0.1.2"
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -91,6 +91,7 @@ hex-literal = "0.3.3"
 nix = "0.24.0"
 reqwest = "0.11.6"
 metrics-util = "0.13.0"
+cdrs-tokio = { git = "https://github.com/krojew/cdrs-tokio" }
 
 [[bench]]
 name = "redis_benches"
--- a/shotover-proxy/src/transforms/protect/crypto.rs
+++ b/shotover-proxy/src/transforms/protect/crypto.rs
@@ -42,8 +42,8 @@ pub async fn encrypt(
     // TODO: investigate using storing as blob instead of text and then use textAsBlob here
     // We should functionally verify which has better performance before sticking with one.
     Ok(Operand::Const(format!(
-        "'{}'",
-        serde_json::to_string(&protected)?
+        "textAsBlob('{}')",
+        hex::encode(bincode::serialize(&protected)?)
     )))
 }
 
@@ -53,10 +53,10 @@ pub async fn decrypt(
     key_id: &str,
 ) -> Result<MessageValue> {
     let varchar = match value {
-        MessageValue::Varchar(varchar) => varchar,
+        MessageValue::Bytes(varchar) => varchar,
         _ => bail!("expected varchar to decrypt but was {:?}", value),
     };
-    let protected: Protected = serde_json::from_str(varchar)?;
+    let protected: Protected = bincode::deserialize(&hex::decode(varchar)?)?;
 
     let sym_key = key_management
         .cached_get_key(key_id, Some(protected.enc_dek), Some(protected.kek_id))
--- a/shotover-proxy/src/transforms/protect/crypto.rs
+++ b/shotover-proxy/src/transforms/protect/crypto.rs
@@ -39,10 +39,8 @@ pub async fn encrypt(
         kek_id: sym_key.key_id,
     };
 
-    // TODO: investigate using storing as blob instead of text and then use textAsBlob here
-    // We should functionally verify which has better performance before sticking with one.
     Ok(Operand::Const(format!(
-        "textAsBlob('{}')",
+        "0x{}",
         hex::encode(bincode::serialize(&protected)?)
     )))
 }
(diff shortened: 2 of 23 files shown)


shotover-proxy/src/transforms/protect/crypto.rs
```diff
@@ -42,7 +42,7 @@ pub async fn encrypt(
     // TODO: investigate using storing as blob instead of text and then use textAsBlob here
```
rukai, line 42: todo is still here

shotover-proxy/src/transforms/protect/crypto.rs
```diff
@@ -42,7 +42,7 @@ pub async fn encrypt(
...
     // We should functionally verify which has better performance before sticking with one.
     Ok(Operand::Const(format!(
-        "'{}'",
+        "textAsBlob('{}')",
         serde_json::to_string(&protected)?
```
rukai, line 46: To actually make use of storing as a blob we need to encode with bincode or similar instead of json.

conorbros: Isn't this storing a JSON string as blob within Cassandra?

rukai: correct.
Storing json as a blob or a text should take up the same amount of space.

conorbros: Wouldn't it going from bincode -> string -> bincode then?

rukai: Oh I see where it went wrong.
We shouldnt be using textAsBlob.
Instead we can insert directly to the blob column by writing our binary data as a hex string.

https://stackoverflow.com/questions/22141620/is-it-possible-in-cassandra-cql3-insert-statement-binary-data

conorbros: the struct wouldn't be valid hex.

rukai: the struct?
The struct gets encoded into a Vec<u8> by bincode.
Then we turn the Vec<u8> into a hex encoded String

shotover-proxy/tests/cassandra_int_tests/protect.rs
```diff
@@ -1,11 +1,21 @@
 use crate::helpers::cassandra::{assert_query_result, execute_query, run_query, ResultValue};
 use cassandra_cpp::{stmt, Batch, BatchType, Session};
+use chacha20poly1305::Nonce;
+use serde::Deserialize;
+
+#[derive(Deserialize)]
+pub struct Protected {
```
rukai, line 7: Why have we duplicated this type?

shotover-proxy/Cargo.toml
```diff
@@ -14,7 +14,7 @@ alpha-transforms = []
 
 [dependencies]
 pretty-hex = "0.3.0"
-hex = "0.4.3"
+hex = { version = "0.4.3", features = ["serde"] }
```
rukai, line 17: I dont think we make use of this serde feature anywhere?
Seems to enable this module that we dont use: https://docs.rs/hex/0.4.3/hex/serde/index.html
That module sure is interesting though, does it let us avoid allocating the Vec<u8> in between?

shotover-proxy/tests/cassandra_int_tests/protect.rs
conorbros: To avoid making it public

rukai: Oh yeah, I see that is a bit unfortunate, oh well.

shotover-proxy/Cargo.toml
conorbros: True, I must've left it in there trying something else and assumed I still needed it.

shotover-proxy/src/transforms/protect/crypto.rs
```diff
@@ -39,11 +39,9 @@ pub async fn encrypt(
         kek_id: sym_key.key_id,
     };
 
-    // TODO: investigate using storing as blob instead of text and then use textAsBlob here
-    // We should functionally verify which has better performance before sticking with one.
     Ok(Operand::Const(format!(
-        "'{}'",
-        serde_json::to_string(&protected)?
+        "textAsBlob('{}')",
```
rukai, line 43: We need to remove the textAsBlob otherwise the data would end up stored as a hex string instead of as the raw bytes.

rukai: The quotes need to go as well. (refer to https://stackoverflow.com/questions/22141620/is-it-possible-in-cassandra-cql3-insert-statement-binary-data )
I would think the int test would fail as is?

rukai:
## Positives
On main we get benchmark results around:
```
transform/cassandra_protect_protected
                        time:   [6.0411 us 6.0443 us 6.0474 us]
                        change: [-1.6616% -1.4996% -1.3351%] (p = 0.00 < 0.05)
                        Change within noise threshold.
```
On average this branch appears to be around 0.3 us faster than main.

To verify storage space improvements I added logging to the size of the bincode output on this branch and the size of the json output on main.

* main - 689 bytes
* this branch - 261 bytes

Taking these raw numbers as is assumes that cassandra does the sensible thing with storage of the bytes.
But this definitely sounds liked a very nice win!

## Negatives
Bincode has no backwards compatibility between major versions of bincode and when adding/removing/changing struct fields.

cbor might give us similar space wins while maintaining backwards compatibility due to cbor efficiently storing the encrypted `Vec<u8>`.
But the speed win might be lost.

shotover-proxy/src/transforms/protect/crypto.rs
rukai: Oh, it also needs a `0x` prefix as `hex::encode` doesnt include that.

shotover-proxy/src/transforms/protect/crypto.rs
```diff
@@ -53,10 +51,10 @@ pub async fn decrypt(
     key_id: &str,
 ) -> Result<MessageValue> {
     let varchar = match value {
-        MessageValue::Varchar(varchar) => varchar,
+        MessageValue::Bytes(varchar) => varchar,
         _ => bail!("expected varchar to decrypt but was {:?}", value),
     };
-    let protected: Protected = serde_json::from_str(varchar)?;
+    let protected: Protected = bincode::deserialize(&hex::decode(varchar)?)?;
```
rukai, line 57: the hex::decode needs to be removed here.
Cassandra results include the bytes directly as Bytes.

shotover-proxy/src/transforms/protect/crypto.rs
```diff
@@ -53,10 +51,10 @@ pub async fn decrypt(
     key_id: &str,
 ) -> Result<MessageValue> {
     let varchar = match value {
-        MessageValue::Varchar(varchar) => varchar,
+        MessageValue::Bytes(varchar) => varchar,
```
rukai, line 54: varchar -> bytes (or similar)

shotover-proxy/tests/cassandra_int_tests/protect.rs
```diff
@@ -86,8 +96,8 @@ APPLY BATCH;"
         assert_eq!(row[1], ResultValue::Varchar("cluster".into()));
 
         // protected values are stored encrypted
-        if let ResultValue::Varchar(value) = &row[2] {
-            assert!(value.starts_with("{\"cipher"), "but was {:?}", value);
+        if let ResultValue::Blob(value) = &row[2] {
+            let _: Protected = bincode::deserialize(&hex::decode(value).unwrap()).unwrap();
```
rukai, line 100: and remove hex::decode from here too

rukai (approved):
I tried replacing bincode with cbor to test performance, it should have worked but for some reason just didnt.
I gave up on investigating though because I realized that we have 2 layers of serialization here, one of which is already bincode, so the advantages of cbor would be lost unless we changed both layers to cbor.

So lets just merge this PR so that both layers are bincode and then later on we can design a system to handle backwards compatibility with old serialized data. (possibly using cbor or maybe something more schema-like)

XA21X (approved):

PR #687, opened 2022-06-29
Add benches for protect transform

rukai:
Adds benches to directly bench the protect transform without going through the full shotover<->cassandra stack
You can confirm that the `cassandra_protect_protected` hits the encryption path by observing that it performs a lot worse than `cassandra_protect_unprotected`

```
transform/cassandra_protect_unprotected
                        time:   [1.7484 us 1.7582 us 1.7666 us]
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high severe
transform/cassandra_protect_protected
                        time:   [6.0483 us 6.0769 us 6.1084 us]
```

These benches will help with evaluating the impact of https://github.com/shotover/shotover-proxy/pull/685


conorbros (approved):
Thanks!

benbromhead (approved):

PR #688, opened 2022-06-30
Small chain.rs cleanup

rukai:
* use AtomicU64 instead of a mutex, its technically faster but that doesnt matter as its just test code, I only changed it because its simpler
   + The Ordering value doesnt matter here at all, as they are all suitable for an atomic counter.
* reworded various tracing log messages


conorbros (approved):

XA21X (approved):
`std::sync::atomic::Ordering::Relaxed` is a bit verbose, but LGTM.

rukai:
Yeah its one of those, "import once, use once" scenarios. Either way isnt great. I'll go ahead and merge.

PR #689, opened 2022-06-30
Remove unneeded Result from source return type

rukai:
* ~~runner.rs changes cause panics in sources to unwind all the way up such that shotover terminates.~~
   + ~~This is desirable for integration tests - any panic in shotover should cause the int test to fail.~~
   + ~~I also believe this is desirable for end users, a panic is an unrecoverable error and so should be handled as such.~~
   + ~~Previously it looked like we were getting that behavior in tests but that was just because:~~
      - ~~the test itself was failing due to assertions causing a panic + full unwind~~
      - ~~while tokio was catching the source panic at the task boundary, displaying the error and then never unwinding the rest of the way.~~
* `JoinHandle<Result<()>>` -> `JoinHandle<()>` is changed because there is no expected runtime error for a source. They should be handling all of their errors internally.

### DIFF
--- a/shotover-proxy/src/runner.rs
+++ b/shotover-proxy/src/runner.rs
@@ -232,11 +232,7 @@ pub async fn run(
 
     match topology.run_chains(trigger_shutdown_rx).await {
         Ok(sources) => {
-            for result in
-                futures::future::join_all(sources.into_iter().map(|x| x.into_join_handle())).await
-            {
-                result.unwrap();
-            }
+            futures::future::join_all(sources.into_iter().map(|x| x.into_join_handle())).await;
             info!("Shotover was shutdown cleanly.");
             Ok(())
         }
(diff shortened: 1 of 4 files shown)


conorbros (approved):

rukai:
Having thought about this more, I'm not sure I still agree with propagating the panic here.
I'll get back to this.

rukai:
I ended up removing the main change here.
The panics only propagated after all the shotover source tasks had finished so it didnt make shotover any less robust.
But I no longer think panicking here brings any value, we already get the panic message on stdout and we now have https://github.com/shotover/shotover-proxy/pull/691 to catch panics in integration tests.
There might be value in including the panic message in an `error!` so these issues get reported properly but I couldnt find a reasonable way to get the panic message for use in an `error!`.

So now the only change remaining in this PR is the small cleanup to the return type.

XA21X (approved):

Issue #498, opened 2022-02-01
Tee should propagate panics that occur in the sub chain


rukai:
If a panic occurs in the main chain shotover is brought down.
If a panic occurs in the tee sub chain shotover is not brought down.
This means that panics in the sub chain can go uncaught by tests.

To fix this we should propagate panics that occur in the sub chain to the main process.

Example of panic on main branch that has slipped through because of this:
![image](https://user-images.githubusercontent.com/5120858/151906638-d7ff02c2-e073-4b19-9a53-a234d0299042.png)

`cargo test test_cluster_dr_redis -- --nocapture`

Issue #499, opened 2022-02-01
Possible design flaw - Filter, Protect etc. do not follow transform invariant


rukai:
The transform message invariants allow for less messages to be returned than were sent, as long as every message eventually receives a reply.
The reason for this is to allow for the implementation of coalesce which holds onto messages until a certain condition is met.

However multiple transforms expect the number of messages received to be the same as the number that went out.
If every transform had to manually buffer messages to avoid this problem it would result in a lot of complexity in many transforms.
What kind of abstractions can we provide to keep the transform logic simple here?

Another related issue is protocols that support out of order messages like cassandra could optimize their sinks by having the sink return messages as soon as they are ready as long as they are returned in order. But that would explode when combined with these problematic transforms.

rukai:
I just realized that the sinks have the same behaviour as Coalesce.
Or at the very least it looks like RedisSinkSingle will sometimes not return the full set of responses.

rukai:
The mesageID rewrite is far enough along now that the issues with filter and coalesce are resolved.

PR #691, opened 2022-07-04
Detect panics in integration tests

rukai:
closes https://github.com/shotover/shotover-proxy/issues/498 - although the design ended up different to what I initially described in that issue
I went through many design and implementation attempts before finally landing on this one.
Some other things I tried that didnt work out:
* always joining + unwrapping every task. Gave up on this because holding the JoinHandle around to do that is very tricky for long running tasks.
* writing a #[cfg(debug_assertions)] wrapper of tokio::spawn that catches panics in the task before it reaches tokio. I never got this working and the design it also relied on the developer to always call the wrapper.

But I'm now quite happy with the design I ended up with.
By using set_hook we can ensure that any panic at all will fail an integration test without any affect on the functionality of real shotover and without introducing any difference in functionality between tests and real shotover.
This leaves us free to choose to join or not join tasks purely based on if it makes sense from the perspective of achieving the desired logic.
This design also means shotover is more robust as a single panic wont bubble all the way up and cause shotover to shutdown.

Ideally I would fix the Filter transform issue that the panic was exposing but the solution will likely be complicated and require a lot of design work.
So instead I just downgraded it from a panic to an `error!` so we can land this important improvement to our test framework.
The real problem will continue to be tracked by https://github.com/shotover/shotover-proxy/issues/499

### DIFF
--- a/shotover-proxy/tests/helpers/mod.rs
+++ b/shotover-proxy/tests/helpers/mod.rs
@@ -6,6 +6,7 @@ use shotover_proxy::runner::{ConfigOpts, Runner};
 use shotover_proxy::tls::{TlsConfig, TlsConnector};
 use std::fs::read_to_string;
 use std::pin::Pin;
+use std::sync::mpsc;
 use std::time::Duration;
 use tokio::runtime::{Handle as RuntimeHandle, Runtime};
 use tokio::sync::watch;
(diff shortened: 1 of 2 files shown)


conorbros (approved):

shotover-proxy/tests/helpers/mod.rs
```diff
@@ -21,44 +22,49 @@ pub struct ShotoverManager {
...
     }
 
-    #[allow(dead_code)] // to make clippy happy
+    #[allow(dead_code)]
     pub fn from_topology_file_without_observability(topology_path: &str) -> ShotoverManager {
+        ShotoverManager::from_topology_file_inner(topology_path, false)
+    }
+
+    fn from_topology_file_inner(topology_path: &str, observability: bool) -> ShotoverManager {
+        // There is a race condition between the call to take_hook and set_hook but not much we can do.
```
XA21X, line 41: Could be useful to link to https://github.com/rust-lang/rust/issues/92649

shotover-proxy/tests/helpers/mod.rs
```diff
@@ -21,44 +22,49 @@ pub struct ShotoverManager {
...
+            Runner::new(opts)
+                .unwrap_or_else(|x| panic!("{x} occurred processing {topology_path:?}"))
+                .with_observability_interface()
+                .unwrap_or_else(|x| panic!("{x} occurred processing {topology_path:?}"))
+                .run_spawn()
+        } else {
+            Runner::new(opts)
+                .unwrap_or_else(|x| panic!("{x} occurred processing {topology_path:?}"))
+                .run_spawn()
+        };
```
XA21X, line 67: Looked a bit odd to have the same message repeated three times, but the alternative I could come up with is a bit messy, and your existing code has the benefit of distinguishing which line the error was from. 👍 

```
        let spawn = if observability {
            Runner::new(opts)
            .and_then(|r| r.with_observability_interface())
        } else {
            Runner::new(opts)
        }.map(|r| r.run_spawn())
        .unwrap_or_else(|x| panic!("{x} occurred processing {topology_path:?}"));
```

XA21X (approved):

PR #692, opened 2022-07-05
Remove byteorder dep

rukai:
The one use of this library can be better expressed using methods from std anyway.


conorbros (approved):

XA21X (approved):
It's been part of the std since Rust 1.32.0, maybe it was missed. Huh.

Issue #174, opened 2021-09-07
investigate every instance of converting a redis bulk string to a rust string


rukai:
investigate every instance of converting a redis bulk string to a rust string to make sure we arent treating fallible conversions as infallible.
e.g. in redis_codec.rs and timestamp_tagging.rs

Ben said:
> Based on the redis protocol, we should never actually get a single bulk string as a client query/request. All redis requests are arrays of bulk strings, with the command portions being parseable as strings and the userdata potentially not being so. So this branch should never actually be reached by a correctly implemented RESP2 client!

PR #693, opened 2022-07-05
redis bulkstring handling fixes

rukai:
closes https://github.com/shotover/shotover-proxy/issues/174

Looks like all the actually worrying cases had already been fixed by other work.

That leaves us with the following cases that are worth fixing but are extremely rare:

* length checks - check the length of any array/vec before addressing it or use `.get` to ensure panics cannot be caused maliciously
* String::from_utf8_lossy -> str::from_utf8 - in this case we are processing a server reply, so we should assume the server will provide utf8 data but we should also not panic when its not in case something malicious is going on.

### DIFF
--- a/shotover-proxy/src/frame/cassandra.rs
+++ b/shotover-proxy/src/frame/cassandra.rs
@@ -37,7 +37,7 @@ use uuid::Uuid;
 /// Functions for operations on an unparsed Cassandra frame
 pub mod raw_frame {
     use super::{CassandraMetadata, RawCassandraFrame};
-    use anyhow::{anyhow, Result};
+    use anyhow::{anyhow, bail, Result};
     use cassandra_protocol::{compression::Compression, frame::Opcode};
     use nonzero_ext::nonzero;
     use std::convert::TryInto;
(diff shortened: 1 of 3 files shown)


shotover-proxy/src/frame/cassandra.rs
```diff
@@ -84,6 +83,9 @@ pub mod raw_frame {
     }
 
     pub(crate) fn get_opcode(bytes: &[u8]) -> Result<Opcode> {
+        if bytes.len() < 9 {
+            bail!("Cassandra message too short, needs at least 9 bytes for header");
```
conorbros, line 87: Cassandra frame*

conorbros (approved):

shotover-proxy/src/transforms/redis/sink_cluster.rs
```diff
@@ -785,7 +785,8 @@ fn build_slot_to_server(
     ensure!(frames.len() >= 2, "expected at least two fields");
 
     let ip = if let RedisFrame::BulkString(ref ip) = frames[0] {
-        String::from_utf8_lossy(ip.as_ref()).to_string()
+        std::str::from_utf8(ip.as_ref())
+            .map_err(|e| anyhow!("Failed to parse IP address as utf8 {e}"))?
```
XA21X, line 789: Would context be better? `std::str::from_utf8(ip.as_ref()).context("Failed to parse IP as utf8")?`

XA21X (approved):

shotover-proxy/src/transforms/redis/sink_cluster.rs
rukai: Oh! thats very cool, I hadnt seen the `.context` method available on `Result` before.
Previously I had only seen the `.context` on the `Error`.

Issue #294, opened 2021-10-22
Remove allocations from redis command equality checks


rukai:
I noticed this API which we could use for allocationless redis command name checks.
https://doc.rust-lang.org/std/primitive.slice.html#method.eq_ignore_ascii_case

And possibly this one as well if we have ownership of the Vec<u8> https://doc.rust-lang.org/std/primitive.slice.html#method.make_ascii_uppercase

PR #694, opened 2022-07-05
Remove allocation in simple redis command checks

rukai:
closes https://github.com/shotover/shotover-proxy/issues/294

All the cases that match on multiple commands are left alone as I have no idea if it would help there and it would make it a lot harder to read.


conorbros:
I'm for removing the fast-path because it's not actually doing anything performance wise at the moment. If we do regress on the regular path we can address that in the best way at the time (maybe reintroducing the fast-path or a different way).

conorbros (approved):

shotover-proxy/src/transforms/redis/cluster_ports_rewrite.rs
```diff
@@ -180,51 +171,34 @@ fn rewrite_port_node(frame: &mut Frame, new_port: u16) -> Result<()> {
...
-                _ => None,
-            })
-            .take_while(Option::is_some)
-            .map(Option::unwrap)
-            .collect::<Vec<Vec<u8>>>()
+    if let Frame::Redis(RedisFrame::Array(array)) = frame {
+        match array.as_slice() {
+            [RedisFrame::BulkString(one), RedisFrame::BulkString(two), ..] => {
+                one.eq_ignore_ascii_case(b"CLUSTER")
+                    && (two.eq_ignore_ascii_case(b"NODES") || two.eq_ignore_ascii_case(b"REPLICAS"))
```
XA21X, line 181: This is much nicer to read. :)

XA21X (approved):

PR #697, opened 2022-07-07
update deps

conorbros:
Because we can't merge #696 due to the CLA bot. Also updated our other outdated deps.


XA21X (approved):
I've added `dependabot[bot]` to the exemption list so we should be good in the future.
https://github.com/cla-assistant/cla-assistant#can-i-allow-bot-user-contributions

rukai (approved):

PR #699, opened 2022-07-08
Minor improvements

rukai:
The previous warning `warn!("failed to decode message: {}", error)` resulted in the very confusing output: 

```
WARN request{id=8 source="RedisSource"}: shotover_proxy::server: failed to decode message: Connection reset by peer (os error 104)
```

But now the message acknowledges that the error could also be from receiving.

The previous error `error!("Stream ended with error {:?}", err);` was changed to match the format of the receive/decode message


conorbros (approved):

benbromhead (approved):
very sensible

Issue #702, opened 2022-07-14
Invalid redis commands can break shotover's transform invariants


rukai:
Pulled out of https://github.com/shotover/shotover-proxy/pull/645 which uncovered the issue:

By dumping the requests/responses in RedisSinkSingle and running `cargo test cassandra_int_tests::test_cassandra_redis_cache` we get:

```
ERROR request{id=5 source="CassandraSource"}: shotover_proxy::transforms::redis::sink_single: request: [Message { inner: Some(Modified { frame: Redis(BulkString(b"FLUSHDB")) }), return_to_sender: false, meta_timestamp: None }]
ERROR request{id=5 source="CassandraSource"}: shotover_proxy::transforms::redis::sink_single: result:  [Message { inner: Some(Parsed { bytes: b"-ERR unknown command `$7`, with args beginning with: \r\n", frame: Redis(Error("ERR unknown command `$7`, with args beginning with: ")) }), return_to_sender: false, meta_timestamp: None }]
ERROR request{id=5 source="CassandraSource"}: shotover_proxy::transforms::redis::sink_single: request: [Message { inner: Some(Modified { frame: Redis(Array([BulkString(b"DEL"), BulkString(b"test_cache_keyspace_simple.test_table:1")])) }), return_to_sender: false, meta_timestamp: None }]
ERROR request{id=5 source="CassandraSource"}: shotover_proxy::transforms::redis::sink_single: result:  [Message { inner: Some(Parsed { bytes: b"+OK\r\n", frame: Redis(SimpleString(b"OK")) }), return_to_sender: false, meta_timestamp: None }]
ERROR request{id=5 source="CassandraSource"}: shotover_proxy::transforms::redis::sink_single: request: [Message { inner: Some(Modified { frame: Redis(Array([BulkString(b"DEL"), BulkString(b"test_cache_keyspace_simple.test_table:2")])) }), return_to_sender: false, meta_timestamp: None }]
ERROR request{id=5 source="CassandraSource"}: shotover_proxy::transforms::redis::sink_single: result:  [Message { inner: Some(Parsed { bytes: b":0\r\n", frame: Redis(Integer(0)) }), return_to_sender: false, meta_timestamp: None }]
[...]

PR #704, opened 2022-07-18
Reject non-array redis commands

rukai:
closes https://github.com/shotover/shotover-proxy/issues/702

In order to catch invalid commands I inserted a validation check both just after the command is decoded and just before it is encoded.

This seems like a reasonable way to do it:
* we catch invalid commands from the client asap
* we catch invalid commands generated by transforms
* If a transform really wants to for performance reasons it can bypass the check by directly generating bytes

Previously RedisSinkSingle was dropping errors on send, so I changed that to bubble up the error to make this PR work.
The effect of this is if we encounter a connection/protocol error when sending to the redis instance, shotover will now:
* terminate the connection when RedisSinkSingle is used directly in a chain 
* log a cache failure and keep running when RedisSinkSingle is in the cache chain of SimpleRedisCache

You can observe the new validation logic functioning correctly by running `cargo test cassandra_int_tests::test_cassandra_redis_cache`
and observing the output log:
`Cache error: Redis command must be an array but was BulkString(b"FLUSHDB")`

### DIFF
--- a/shotover-proxy/src/codec/redis.rs
+++ b/shotover-proxy/src/codec/redis.rs
@@ -8,8 +8,15 @@ use redis_protocol::resp2::prelude::encode_bytes;
 use tokio_util::codec::{Decoder, Encoder};
 
 #[derive(Debug, Clone)]
+pub enum RedisDirection {
+    Source,
+    Sink,
+}
+
+#[derive(Debug, Clone)]
 pub struct RedisCodec {
     messages: Messages,
+    direction: RedisDirection,
 }
 
 #[inline]
@@ -29,15 +36,12 @@ pub fn redis_query_type(frame: &RedisFrame) -> QueryType {
     QueryType::Write
 }
 
-impl Default for RedisCodec {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
 impl RedisCodec {
-    pub fn new() -> RedisCodec {
-        RedisCodec { messages: vec![] }
+    pub fn new(direction: RedisDirection) -> RedisCodec {
+        RedisCodec {
+            messages: vec![],
+            direction,
+        }
     }
 }
 
@@ -53,8 +57,15 @@ impl Decoder for RedisCodec {
                         "incoming redis message:\n{}",
                         pretty_hex::pretty_hex(&bytes)
                     );
-                    self.messages
-                        .push(Message::from_bytes_and_frame(bytes, Frame::Redis(frame)));
+                    let mut message = Message::from_bytes_and_frame(bytes, Frame::Redis(frame));
+                    if let RedisDirection::Source = self.direction {
+                        if let Some(frame) = message.frame() {
+                            validate_command(frame)?;
+                        } else {
+                            return Err(anyhow!("redis message could not be parsed as a message"));
+                        }
+                    }
+                    self.messages.push(message);
                 }
                 None => {
                     if self.messages.is_empty() || src.remaining() != 0 {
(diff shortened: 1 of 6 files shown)


shotover-proxy/src/codec/redis.rs
```diff
@@ -29,15 +36,12 @@ pub fn redis_query_type(frame: &RedisFrame) -> QueryType {
...
-impl Default for RedisCodec {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
 impl RedisCodec {
-    pub fn new() -> RedisCodec {
-        RedisCodec { messages: vec![] }
+    pub fn new(direction: RedisDirection) -> RedisCodec {
```
conorbros, line 40: nit: *Self

shotover-proxy/src/codec/redis.rs
```diff
@@ -53,8 +57,15 @@ impl Decoder for RedisCodec {
...
                         pretty_hex::pretty_hex(&bytes)
                     );
-                    self.messages
-                        .push(Message::from_bytes_and_frame(bytes, Frame::Redis(frame)));
+                    let mut message = Message::from_bytes_and_frame(bytes, Frame::Redis(frame));
+                    if let RedisDirection::Source = self.direction {
+                        if let Some(frame) = message.frame() {
+                            validate_command(frame)?;
+                        } else {
+                            return Err(anyhow!("redis message could not be parsed as a message"));
```
conorbros, line 65: redis *frame could not be parsed?

shotover-proxy/src/codec/redis.rs
```diff
@@ -68,6 +79,26 @@ impl Decoder for RedisCodec {
...
+                    match value {
+                        RedisFrame::BulkString(_) => {} // valid case
+                        _ => return Err(anyhow!("Redis command must be an array of bulk strings but one of the values was not a bulk string {frame:?}")),
+                    }
+                }
+                Ok(())
+            }
+            frame => Err(anyhow!("Redis command must be an array but was {frame:?}")),
+        },
+        _ => unreachable!("Message from a redis codec must be a redis message"),
```
conorbros, line 98: will always*

shotover-proxy/src/codec/redis.rs
```diff
@@ -7,9 +7,16 @@ use redis_protocol::resp2::prelude::decode_mut;
 use redis_protocol::resp2::prelude::encode_bytes;
 use tokio_util::codec::{Decoder, Encoder};
 
+#[derive(Debug, Clone)]
+pub enum RedisDirection {
```
conorbros, line 11: So my understanding is we use this for perf reasons? 

instead of first checking whether something is a Redis command and then checking it's validity.

conorbros: My thinking is that we could get rid of this enum and just check everything on the codec methods. Assuming the sink won't be sending us commands that is.

shotover-proxy/src/codec/redis.rs
```diff
@@ -68,6 +79,26 @@ impl Decoder for RedisCodec {
     }
 }
 
+/// It is critical that shotover only send valid redis commands.
+/// An invalid command could result in the redis instance returning multiple responses breaking shotovers transform invariants.
+fn validate_command(frame: &Frame) -> Result<()> {
```
conorbros, line 84: This could just take a Redis frame, and we can unwrap on `into_redis` in the function that calls it.

conorbros:
Some minor concerns with implementing this in the codec. To me this seems like it's leaking Shotover implementation details (validation and protecting our transform invariant) into the codec, which should have the single responsibility of encoding/decoding  messages. Is there a reason we couldn't do these checks in the source and sink?

However I won't block on this as the PR clearly works.

shotover-proxy/src/codec/redis.rs
rukai: We can't always check because the server responses are not valid commands

conorbros: Ah my bad, I was thinking in Cassandra land where we have op codes.

shotover-proxy/src/codec/redis.rs
rukai: we cant use into_redis because into_redis consumes the Frame

conorbros (approved):

rukai:
>Some minor concerns with implementing this in the codec. To me this seems like it's leaking Shotover implementation details (validation and protecting our transform invariant) into the codec, which should have the single responsibility of encoding/decoding messages. Is there a reason we couldn't do these checks in the source and sink?


To me this seems reasonable because this validation is not just required by shotover but by any redis implementation.
Remember that it is redis itself that does not properly handle non-array requests.
The advantage of putting it in the codec over a transform is it means that all redis sink transforms benefit from it without each of them needing to reimplement the logic.

rukai:
As per discussion on slack I should first:
* investigate upstreaming a fix so that redis returns a single request instead of the checks added by this PR
* investigate redis telnet mode compatibility

rukai:
Closing due to low priority right now, but the issue is still valid and tracked at https://github.com/shotover/shotover-proxy/issues/702

PR #706, opened 2022-07-18
Add redis/passthrough_new_connection bench

rukai:
Tried writing this bench and hit a shotover failure that appears to be caused by opening lots of connections really fast.
Opening a PR to put visibility on the problem but I'll keep investigating whats going on.

interestingly, it always seems to fail on the 28234th request

### DIFF
--- a/shotover-proxy/benches/redis_benches.rs
+++ b/shotover-proxy/benches/redis_benches.rs
@@ -1,6 +1,7 @@
 use criterion::{criterion_group, criterion_main, Criterion};
-use redis::Cmd;
+use redis::{Client, Cmd};
 use std::path::Path;
+use std::time::Duration;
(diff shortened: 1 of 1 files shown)


rukai:
I confirmed this is an issue in redis or linux itself.
Some kind of internal resource appears to be running out.
I used two different redis clients https://crates.io/crates/redis and https://crates.io/crates/simple_redis directly against a redis instance and they both gave the same behavior of connection resets after the ~28230th connection.

This bench isnt very useful (The idea of benching how long a connection takes to make is useful, but our bench infrastructure isnt good enough to create a useful implementation of that idea) and I dont think its worth reporting this issue upstream, they'll probably just tell us to not make and destroy so many connections.
So I'm going to close this PR.

PR #711, opened 2022-07-26
Fix transform TLS docs

rukai:
* CassandraSinkSingle was missing
* RedisSinkCluster had the comments mixed up
* RedisSinkSingle had the fields in an inconsistent order with RedisSinkCluster

### DIFF
--- a/docs/src/transforms.md
+++ b/docs/src/transforms.md
@@ -59,11 +59,11 @@ This transform will take a query, serialise it into a CQL4 compatible format and
     # Removing this field will disable TLS.
     #tls:
     #  # Path to the certificate authority file, typically named with a .crt extension.
-    #  certificate_authority_path: "example-configs/cassandra-tls/certs/localhost_CA.crt"
+    #  certificate_authority_path: "tls/localhost_CA.crt"
     #  # Path to the certificate file, typically named with a .crt extension.
-    #  certificate_path: "example-configs/cassandra-tls/certs/localhost.crt"
+    #  certificate_path: "tls/localhost.crt"
     #  # Path to the private key file, typically named with a .key extension.
-    #  private_key_path: "example-configs/cassandra-tls/certs/localhost.key"
+    #  private_key_path: "tls/localhost.key"
 ```
 
 Note: this will just pass the query to the remote node. No cluster discovery or routing occurs with this transform.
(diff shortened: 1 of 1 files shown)


docs/src/transforms.md
```diff
@@ -54,6 +54,16 @@ This transform will take a query, serialise it into a CQL4 compatible format and
 - CassandraSinkSingle:
     # The IP address and port of the upstream cassandra node/service.
     remote_address: "127.0.0.1:9042"
+
+    # When this field is provided TLS is used when connecting to the remote address.
+    # Removing this field will disable TLS.
+    tls:
```
conorbros, line 60: I'd have this commented out so it can be copied and used immediately.

conorbros: Same for the other examples with TLS configs.

conorbros (approved):
LGTM, just that one small change.

docs/src/transforms.md
rukai: Commenting it out does make it harder to read the tls example.
But you make a good point and I think commenting it out is the right move.

docs/src/transforms.md
```diff
@@ -353,13 +362,13 @@ This transform will take a query, serialise it into a RESP2 compatible format an
...
-    tls:
-      # Path to the certificate file, typically named with a .crt extension.
-      certificate_path: "tls/redis.crt"
-      # Path to the private key file, typically named with a .key extension.
-      private_key_path: "tls/redis.key"
-      # Path to the certificate authority file typically named ca.crt.
-      certificate_authority_path: "tls/ca.crt"
+    #tls:
+    #  # Path to the certificate authority file, typically named ca.crt.
+    #  certificate_authority_path: "tls/ca.crt"
```
XA21X, line 367: Should the paths be different between examples? i.e. do all the paths make sense when uncommented?

XA21X (approved):

docs/src/transforms.md
rukai: oh yeah those path prefixes are a bit of mess, I'll tidy that up a bit.

rukai: @XA21X @conorbros Please check the latest commit and let me know what you think of that change.
The file names are left as is because they reflect the naming used by the example key generator for redis and cassandra respectively but the prefix is normalized because there is no need to care about the shotovers internal example/test key paths.

XA21X: LGTM, thanks for the cleanup.

conorbros: LGTM

PR #715, opened 2022-07-28
Replace pushed_messages_tx if let with exhaustive match

rukai:
This way we enforce at compile time that the transforms add_pushed_messages implementation is hit.


conorbros (approved):
Good catch, this would've caused a headache 😂

XA21X (approved):

PR #717, opened 2022-07-28
Add CassandraSinkCluster

rukai:
Introduces the CassandraSinkCluster behind the alpha-transforms feature so that we can start collaboratively working on it.
Is currently identical to the implementation of CassandraSinkSingle except for the config which reflects the final config we want to have for CassandraSinkCluster.
Currently has the same integration test cases as CassandraSinkSingle but test cases will be added and removed as needed in future PRs.

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster.rs
@@ -1,0 +1,165 @@
+use super::connection::CassandraConnection;
+use crate::codec::cassandra::CassandraCodec;
+use crate::concurrency::FuturesOrdered;
+use crate::error::ChainResponse;
+use crate::frame::cassandra;
+use crate::message::Messages;
+use crate::tls::TlsConfig;
+use crate::tls::TlsConnector;
+use crate::transforms::util::Response;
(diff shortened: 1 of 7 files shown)


shotover-proxy/src/transforms/cassandra/sink_cluster.rs
```diff
@@ -0,0 +1,165 @@
...
+pub struct CassandraSinkCluster {
+    contact_points: Vec<String>,
+    outbound: Option<CassandraConnection>,
+    chain_name: String,
+    failed_requests: Counter,
+    tls: Option<TlsConnector>,
+    pushed_messages_tx: Option<mpsc::UnboundedSender<Messages>>,
+}
+
+impl Clone for CassandraSinkCluster {
```
conorbros, line 49: Why do we need this?

conorbros (approved):

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
rukai: We need a custom clone for the same reason as for CassandraSinkSingle.
we need to not clone `self.outbound` so that the new transform creates its own outbound connection.

benbromhead (approved):

PR #718, opened 2022-08-01
small cleanup in CassandraSinkSingle

rukai:
I'm pretty sure the TODO was because `Response { result, .. }` used to contain multiple `Message`s but that has since been fixed separately so we can remove the TODO.

The timeout info should really be an `error!` it indicates a failure in communication with the destination node.
The second info can just be removed as it doesnt provide anything we need that isnt already in the first log.


conorbros (approved):

rukai:
Those two regressions were from separate runs, and I dont believe they indicate anything as there are no functional changes here.
I checked and the `error!` log is not triggered in this benchmark.

benbromhead (approved):

Issue #491, opened 2022-01-31
Fix hang in redis_int_tests::basic_driver_tests::test_source_tls_and_single_tls


rukai:
The test `redis_int_tests::basic_driver_tests::test_source_tls_and_single_tls` has a small chance of hanging.
It occurs in something like 1-5% of --release test runs and I have only been able to reproduce it in --release.

A good way to reproduce the issue is: 
```sh
set -e
set -u
while :;
do
    cargo test --release redis_int_tests::basic_driver_tests::test_source_tls_and_single_tls;
done
```
This will keep rerunning the test until it hangs.

By breaking the logic in `RedisCodec::decode` a hang can be caused that only occurs in `test_source_tls_and_single_tls`, this may or may not be related to this issue.

If this issue is related to our implementation of `RedisCodec::decode`, I wonder if it has something to do with returning multiple messages in a single item causing issues with assumptions made in the implementation of https://github.com/sfackler/tokio-openssl

PR #645, opened 2022-05-17
Redis pubsub support

rukai:
Closes https://github.com/shotover/shotover-proxy/issues/491 (We now ensure we read a response for every request we send out, which fixes the hang)
Implements the `transform_subscribe` idea described in https://github.com/shotover/shotover-proxy/issues/541

This only implements support for RedisSinkSingle.
I will investigate RedisSinkCluster later on, not sure what will be involved for that one.

PR #719, opened 2022-08-02
Fix timeout in RedisSinkSingle

rukai:
closes https://github.com/shotover/shotover-proxy/issues/491 (We now ensure we read a response for every request we send out, which fixes the hang)

Recreates these fixes that were incidentally fixed by https://github.com/shotover/shotover-proxy/pull/645/files :
* Fix timing issue where only a subset of the responses are returned. (fixed by looping until result.len() < requests_len)
* Fix errors going unreported and leaving the connection in an invalid state, we now terminate the connection and log an error when an error is encountered `return Err(anyhow!("encountered error in redis stream: {err:?}"))`

I've pulled these fixes out because they are important and the pubsub PR is very complicated and will take a while to review and land.


benbromhead (approved):

rukai:
Superceded by https://github.com/shotover/shotover-proxy/pull/645

PR #721, opened 2022-08-03
Implement system.peers_v2 rewriting in CassandraSinkCluster

conorbros:
Opening a draft for some early feedback. 

Just removing the peers table entirely will assign ourselves the entire token range. In the future we might need to actually fill this table with the Shotover nodes. I've allowed the use of both the new port field and this option for that reason, you will want to assign token ranges and specify the port of the Shotover nodes.

I could rename this to cluster hiding to more closely match the Redis one?
 
TODO:
- [x] unit tests
- [ ] update documentation

### DIFF
--- a/shotover-proxy/benches/chain_benches.rs
+++ b/shotover-proxy/benches/chain_benches.rs
@@ -214,7 +214,7 @@ fn criterion_benchmark(c: &mut Criterion) {
     {
         let chain = TransformChain::new(
             vec![
-                Transforms::CassandraPeersRewrite(CassandraPeersRewrite::new(9042)),
+                Transforms::CassandraPeersRewrite(CassandraPeersRewrite::new(9042, false)),
                 Transforms::Null(Null::default()),
             ],
             "bench".into(),
--- a/shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
+++ b/shotover-proxy/src/transforms/cassandra/peers_rewrite.rs
@@ -332,7 +332,7 @@ mod test {
 
     #[test]
     fn test_rewrite_port_match() {
-        let col_spec = vec![ColSpec {
+        let col_spec = &[ColSpec {
             table_spec: None,
             name: "native_port".into(),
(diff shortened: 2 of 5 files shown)


rukai:
> Just removing the peers table entirely will assign ourselves the entire token range. 

I recently learnt that the peer you are connected to will not appear in the peers table, so this does sound possible to me.
Do you have a source or line of code in a driver to justify this claim though?

rukai:
The PR is titled: "Implement system.peers_v2 rewriting in CassandraSinkCluster" 
But the changes are actually to the CassandraPeersRewrite transform.

If we do want to put in CassandraPeersRewrite we should retitle the PR.
But I think that putting it in CassandraSinkCluster might be better.
If we cant think of a scenario where the flexibility of having the functionality in a separate transform would enable more use cases then we are better off having CassandraSinkCluster just work by default.

conorbros:
>But I think that putting it in CassandraSinkCluster might be better.

If we are doing that then maybe we should just remove this transform entirely? Add the port rewriting functionality to the sinks instead.

>Do you have a source or line of code in a driver to justify this claim though?
There will be no tokens, so the driver is unable to do any token aware routing. Every request goes to Shotover, thus it has the entire token range. Poor wording on my part.

rukai:
My thinking is:
* the existing transform still serves a purpose - rewriting the ip or port for paired shotover/cassandra instances.
* the new system.peers_v2 and system.local rewriting functionality that is needed by CassandraSinkCluster should be part of CassandraSinkCluster because CassandraSinkCluster will be unusable without it. i.e. CassandraSinkCluster is pretending to be a single Cassandra instance and if we dont perform rewriting of those tables then we are leaking that there is actually a cluster on the other side.

benbromhead:
I think I had previously mentioned we should try and seperate out the peers rewriting into a seperate transfrom altogether.

Seeing the comment above, this might be a premature optimization and if it's easier / quicker to do peers rewriting in the sink then let's do that.

PR #722, opened 2022-08-03
cassandra-cluster-tls int test

conorbros:


shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -70,27 +104,17 @@ fn test_cluster() {
...
+    let _compose = DockerCompose::new("example-configs/cassandra-cluster-tls/docker-compose.yml");
 
     let shotover_manager =
-        ShotoverManager::from_topology_file("example-configs/cassandra-tls/topology.yaml");
+        ShotoverManager::from_topology_file("example-configs/cassandra-cluster-tls/topology.yaml");
 
     let ca_cert = "example-configs/cassandra-tls/certs/localhost_CA.crt";
 
-    {
-        // Run a quick test straight to Cassandra to check our assumptions that Shotover and Cassandra TLS are behaving exactly the same
```
rukai, line 83: Could we include this direct check for the cluster test as well?
It would help prevent the case where the test cluster is not setup for TLS properly and shotover is not implementing TLS properly.

rukai (approved):
LGTM!
Thanks for doing this, it will be good to have the extra int test to develop against.

benbromhead (approved):
Thanks

PR #723, opened 2022-08-03
Implement basic cassandra cluster routing

rukai:
Messages are never routed to the initial contact point as it is not considered a peer.
I consider this a bug but i'm not addressing it in this PR to keep the scope down 

I had previously mentioned implementing round robin for this PR but considering the final goal is "best of two choices", just doing a single random selection seemed like the best step towards that while still being a sound decision for an MVP.

I pulled the receive logic out of the two sink transforms and into connection.rs

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/connection.rs
+++ b/shotover-proxy/src/transforms/cassandra/connection.rs
@@ -14,7 +14,7 @@ use futures::StreamExt;
 use halfbrown::HashMap;
 use std::time::Duration;
 use tokio::io::{split, AsyncRead, AsyncWrite, ReadHalf, WriteHalf};
-use tokio::net::TcpStream;
+use tokio::net::{TcpStream, ToSocketAddrs};
 use tokio::sync::{mpsc, oneshot};
 use tokio::time::timeout;
 use tokio_stream::wrappers::UnboundedReceiverStream;
@@ -36,8 +36,8 @@ pub struct CassandraConnection {
 }
 
 impl CassandraConnection {
-    pub async fn new<C: Codec + 'static>(
-        host: String,
+    pub async fn new<C: Codec + 'static, A: ToSocketAddrs>(
+        host: A,
         codec: C,
         mut tls: Option<TlsConnector>,
         pushed_messages_tx: Option<mpsc::UnboundedSender<Messages>>,
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster.rs
@@ -2,22 +2,24 @@ use super::connection::CassandraConnection;
 use crate::codec::cassandra::CassandraCodec;
 use crate::concurrency::FuturesOrdered;
 use crate::error::ChainResponse;
+use crate::frame::cassandra::parse_statement_single;
 use crate::frame::{CassandraFrame, CassandraOperation, CassandraResult, Frame};
 use crate::message::{Message, MessageValue, Messages};
 use crate::tls::{TlsConnector, TlsConnectorConfig};
 use crate::transforms::{Transform, Transforms, Wrapper};
 use anyhow::Result;
 use async_trait::async_trait;
-use cql3_parser::cassandra_statement::CassandraStatement;
+use cassandra_protocol::consistency::Consistency;
+use cassandra_protocol::frame::Version;
+use cassandra_protocol::query::QueryParams;
 use cql3_parser::common::{FQName, Identifier};
-use cql3_parser::select::SelectElement;
 use metrics::{register_counter, Counter};
 use rand::prelude::*;
 use serde::Deserialize;
-use std::collections::HashMap;
 use std::net::IpAddr;
 use std::sync::Arc;
 use std::time::Duration;
+use tokio::net::ToSocketAddrs;
 use tokio::sync::{mpsc, oneshot, RwLock};
 
 #[derive(Deserialize, Debug, Clone)]
@@ -43,9 +45,10 @@ impl CassandraSinkClusterConfig {
 
 pub struct CassandraSinkCluster {
     contact_points: Vec<String>,
-    contact_point_handshake: Vec<Message>,
-    contact_point_handshake_complete: bool,
     contact_point_connection: Option<CassandraConnection>,
+    init_handshake: Vec<Message>,
+    init_handshake_address: Option<String>,
+    init_handshake_complete: bool,
     chain_name: String,
     failed_requests: Counter,
     tls: Option<TlsConnector>,
@@ -53,18 +56,24 @@ pub struct CassandraSinkCluster {
     read_timeout: Option<Duration>,
     peer_table: FQName,
     data_center: String,
-    nodes: Vec<CassandraNode>,
-    nodes_shared: Arc<RwLock<Vec<CassandraNode>>>,
+    /// A local clone of topology_task_nodes
+    /// Internally stores connections to the nodes
+    local_nodes: Vec<CassandraNode>,
+    /// Only written to by the topology task
+    /// Transform instances should never write to this.
+    topology_task_nodes: Arc<RwLock<Vec<CassandraNode>>>,
     rng: SmallRng,
+    task_handshake_tx: mpsc::Sender<TaskHandshake>,
 }
 
 impl Clone for CassandraSinkCluster {
     fn clone(&self) -> Self {
         CassandraSinkCluster {
             contact_points: self.contact_points.clone(),
-            contact_point_handshake: vec![],
-            contact_point_handshake_complete: false,
             contact_point_connection: None,
+            init_handshake: vec![],
+            init_handshake_address: None,
+            init_handshake_complete: false,
             chain_name: self.chain_name.clone(),
             tls: self.tls.clone(),
             failed_requests: self.failed_requests.clone(),
@@ -72,9 +81,10 @@ impl Clone for CassandraSinkCluster {
             read_timeout: self.read_timeout,
             peer_table: self.peer_table.clone(),
             data_center: self.data_center.clone(),
-            nodes: vec![],
-            nodes_shared: self.nodes_shared.clone(),
+            local_nodes: vec![],
+            topology_task_nodes: self.topology_task_nodes.clone(),
             rng: SmallRng::from_rng(rand::thread_rng()).unwrap(),
+            task_handshake_tx: self.task_handshake_tx.clone(),
         }
     }
 }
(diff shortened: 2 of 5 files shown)


shotover-proxy/src/transforms/cassandra/sink_cluster.rs
```diff
@@ -32,29 +33,44 @@ impl CassandraSinkClusterConfig {
...
+    contact_point_handshake: Vec<Message>,
+    contact_point_handshake_complete: bool,
+    contact_point_connection: Option<CassandraConnection>,
     chain_name: String,
     failed_requests: Counter,
     tls: Option<TlsConnector>,
     pushed_messages_tx: Option<mpsc::UnboundedSender<Messages>>,
+    peer_table: FQName,
+    data_center: String,
+    nodes: Vec<CassandraNode>,
```
conorbros, line 53: I think we should put this stuff (cluster metadata) in a struct.

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
```diff
@@ -93,60 +123,185 @@ impl CassandraSinkCluster {
...
+                    std::mem::drop(write_lock);
+                    std::mem::drop(expensive_drop);
+                }
+            }
+        }
+
+        Ok(responses)
+    }
+
+    fn get_system_peers_index(&self, messages: &mut [Message]) -> Option<QueryOccured> {
```
conorbros, line 195: Can we combine this with the similar function in the peers rewrite transform?

shotover-proxy/src/transforms/cassandra/connection.rs
```diff
@@ -158,3 +157,46 @@ async fn rx_process<C: CodecReadHalf, T: AsyncRead>(
...
 }
+
+pub async fn receive(
+    failed_requests: &metrics::Counter,
+    mut results: FuturesOrdered<oneshot::Receiver<Response>>,
+) -> Result<Messages> {
+    let expected_size = results.len();
+    let mut responses = Vec::with_capacity(expected_size);
+    loop {
+        match timeout(Duration::from_secs(5), results.next()).await {
```
benbromhead, line 168: This timeout should probably be configurable

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
```diff
@@ -93,60 +123,185 @@ impl CassandraSinkCluster {
...
-                            responses.push(original);
+                    }
+                }
+            }
+        }
+        system_peers_index
+    }
+}
+
+fn get_nodes_from_system_peers(
```
benbromhead, line 237: this function (or higher up in the call stack) would be a good candidate for an integration test to pick up changes to the system.peers table.

benbromhead (commented):
Would be good to land a basic integration test with this PR (if we don't already have one), where we discover the cluster and test that we are getting all the bits of info we need from system.peers that we need.

It will be a good canary for breaking peers table changes in the future as well.

benbromhead:
> Would be good to land a basic integration test with this PR (if we don't already have one), where we discover the cluster and test that we are getting all the bits of info we need from system.peers that we need.

If this is coverred in the addition of the cassandra_cpp / cdrs test cases, feel free to ignore.

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
rukai: which stuff is that?
peer_table and data_center are immutable while nodes is mutable so I dont think they belong together?

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
rukai: I deleted the function instead ;)

shotover-proxy/src/transforms/cassandra/connection.rs
rukai: Why does cassandra even have a timeout here?
Redis doesnt seem to have one?

Should we just remove the timeout from cassandra to match redis?
Or should we add configurable timeouts to redis as well?

I pulled this logic into a prereq PR, so the decision we make can go into there: https://github.com/shotover/shotover-proxy/pull/727

benbromhead: hmm we could either let the client handle the timeout. E.g. if the client ends up timing out... all this will get dropped anyway right? If the client hangs on waiting for a connection, so do we?

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
conorbros: I anticipate us moving stuff out of this file into a struct holding the cluster stuff and doing things like handling different routing strategies and topology rebuilding in there rather than the main transform file.

But probably premature optimisation so lets go ahead with this for now.

shotover-proxy/src/transforms/cassandra/connection.rs
rukai: A well behaving client should close the connection when it times out.
Shotover would definitely terminate the transform if its waiting for the client to send a message and the client closes the connection.
But I think thats not true if the transform is stuck waiting on something, in this case for a response from the server...

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
```diff
@@ -31,29 +34,48 @@ impl CassandraSinkClusterConfig {
...
+    init_handshake_address: Option<String>,
+    init_handshake_complete: bool,
     chain_name: String,
     failed_requests: Counter,
     tls: Option<TlsConnector>,
     pushed_messages_tx: Option<mpsc::UnboundedSender<Messages>>,
+    peer_table: FQName,
+    data_center: String,
+    nodes: Vec<CassandraNode>,
+    nodes_shared: Arc<RwLock<Vec<CassandraNode>>>,
```
conorbros, line 56: Are you still planning to remove the node sharing stuff and just send our own system.peers query?

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
```diff
@@ -62,92 +84,244 @@ impl CassandraSinkCluster {
...
             }
         }
 
+        let mut responses_future = FuturesOrdered::new();
+        for message in messages {
+            let (return_chan_tx, return_chan_rx) = oneshot::channel();
+            if self.nodes.is_empty() || !self.init_handshake_complete {
+                self.contact_point_connection.as_mut().unwrap()
+            } else {
+                let random_node = self.nodes.choose_mut(&mut self.rng).unwrap();
```
conorbros, line 177: What are your plans for abstracting the query planning/routing strategies?

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
rukai: We do send our own system.peers query!
It gets shared between all the transforms through nodes_shared ... I should go rename nodes_shared now though.

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
rukai: I didnt know there were any current plans to allow multiple planning/routing strategies?

conorbros: Oh true, I just assumed we would. e.g.

https://docs.rs/cdrs-tokio/latest/cdrs_tokio/load_balancing/index.html

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
```diff
@@ -31,29 +34,52 @@ impl CassandraSinkClusterConfig {
...
     tls: Option<TlsConnector>,
     pushed_messages_tx: Option<mpsc::UnboundedSender<Messages>>,
+    peer_table: FQName,
+    data_center: String,
+    /// A local clone of topology_task_nodes
+    /// Internally stores connections to the nodes
+    local_nodes: Vec<CassandraNode>,
+    /// Only written to by the topology task
+    /// Transform instances should never write to this.
+    topology_task_nodes: Arc<RwLock<Vec<CassandraNode>>>,
```
conorbros, line 60: Might be a candidate for https://docs.rs/arc-swap/latest/arc_swap/.

Then we can avoid the lock/drop thing.

rukai: oh! I was researching these other types of locks earlier.
I decided it wasnt worth blocking progress over and I could come back to it later.
I did get a bit confused trying to understand how arc_swap works earlier, but I'll take a another look when I get the chance.

shotover-proxy/src/transforms/cassandra/connection.rs
rukai: Having thought about it.
If we want to handle the server suddenly never talking to us again, then we need to use a timeout like we already have for cassandra.
Redis currently doesnt do that, so I'll raise an issue to get that implemented.
For cassandra I'll make the timeout configurable and do that in the other PR.

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
conorbros: Sure, let's not block progress on it then, especially since the solution is just an extra line. I'll open an issue to investigate properly later.

conorbros (approved):
Alrighty LGTM. Nicely done

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
rukai: The two straightforward ways I can think of doing an integration test here are:
* call `create_topology_task` pass it a hardcoded handshake and then wait until it writes some nodes to the RwLock, and run some assertions on them
   + doesnt test the full stack
   + least invasive approach
   + lets us make assertions directly on the system.peers table that the task returns.
* Add some some metrics counters to CassandraSinkSingle that record each hit to each outgoing node. Then assert that each outgoing node is hit at least once.
    + Could fail if we got really unlucky, it is unlikely though as our integration tests perform a lot of queries.
    + Can only make a very weak assertion against our routing. e.g. impossible to test that we have implemented token aware routing well.

Which of these sounds better?

Something else thats probably too complicated to invest in:
If we wanted a really robust test we could mock out the cassandra nodes and then use data from the mocks to assert that our token aware routing is implemented properly.

rukai: Ended up just going with the first option.

benbromhead (approved):

PR #724, opened 2022-08-04
Update deps

rukai:
Looks like chacha20 crate used `NewAead` to introduce a change to `Aead` without breaking changes and is now actually performing the breaking change by replacing `Aead` with `NewAead`

I did not update tokio in the cargo.lock as it hits a false positive clippy lint.
We can update tokio once https://github.com/rust-lang/rust-clippy/pull/9246 hits stable, probably rust 1.64

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,12 +19,12 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aead"
-version = "0.4.3"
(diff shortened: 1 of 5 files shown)


rukai:
blocked on: https://github.com/dtolnay/unicode-ident/issues/10#issuecomment-1204783373

rukai:
Discussed with @justinmclean and the unicode license is compatible due to being permissive, we just need to add it to `cargo deny`s allow list.

This PR can proceed

benbromhead (approved):

conorbros (approved):

PR #731, opened 2022-08-08
Int test function propogation

rukai:
I originally tried implementing a check for this via cdrs-tokio or cassandra_cpp but found there was no way to hit a specific node via those drivers.
Instead I ended up finding the desired functionality already implemented in the scylla rust driver.
I have no problems adding it to our dependencies as it would be nice to eventually add it to the list of drivers that our int tests abstract over.

I had to remove test::functions test cases from the TLS cases as scylla doesnt seem to support TLS? But thats not really a problem as the TLS tests are just duplicating the regular test logic in the hope of catching some absurd edge case, but such an edge case would be truly absurd.

I have been able to reproduce the failure locally without this change.
I have not been able to reproduce the failure with this change.


benbromhead:
> But thats not really a problem as the TLS tests are just duplicating the regular test logic in the hope of catching some absurd edge case, but such an edge case would be truly absurd.

I feel like we are tempting fate with this sentence 🤣 

LGTM

benbromhead (approved):

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -24,25 +25,27 @@ mod native_types;
...
 #[serial]
-fn test_passthrough() {
+async fn test_passthrough() {
     let _compose = DockerCompose::new("example-configs/cassandra-passthrough/docker-compose.yml");
 
     let shotover_manager =
         ShotoverManager::from_topology_file("example-configs/cassandra-passthrough/topology.yaml");
 
     let connection = shotover_manager.cassandra_connection("127.0.0.1", 9042);
+    let schema_awaiter = SchemaAwaiter::new("127.0.0.1:9043").await;
```
conorbros, line 41: Is this one meant to connect to a different port? The rest connect directly to Shotover it seems.

rukai: hmmm they are all meant to connect directly to cassandra

rukai: I dont see any that connect directly to shotover

conorbros: Oops, I looked at the cluster one and didn't notice the IP was different.

conorbros (approved):

PR #732, opened 2022-08-09
refactor redis_int_tests

rukai:
Moves the actual test cases from `redis_int_tests::basic_driver_tests` -> `redis_int_tests`
No changes to test logic occured.

This way it is easier to remember unique test names: `cargo test redis_int_tests::test_passthrough` instead of `cargo test redis_int_tests::basic_driver_tests::test_passthrough` 
It also matches the names already used by cassandra: `cargo test cassandra_int_tests::test_passthrough`

Organization wise it also makes sense, we have split up an 1851 line long file into two smaller files at a nice boundary.


conorbros (approved):
Thank you!!

benbromhead (approved):

PR #733, opened 2022-08-09
update serde_yaml to 0.9 (by adding `!` everywhere)

rukai:
Just publishing this to demonstrate the new serialization format of serde_yaml.
We will have to decide if we want to actually upgrade to this new format, find an alternative crate or just stick to the old version of serde_yaml.

The changes to invalid_subchains.yaml allows `cargo test test_validate_chain_multiple_subchains` to pass.


rukai:
Superceded by https://github.com/shotover/shotover-proxy/pull/1015

PR #734, opened 2022-08-09
Fix another intermittent cluster failure

rukai:
Fixes https://github.com/shotover/shotover-proxy/runs/7739411831?check_suite_focus=true

This time the problem is in table::test
... and another showed up in udt::test

The problem is because we arent waiting for the cluster to reach schema agreement before running certain commands, so we can fix it by using the SchemaAwaiter like we do in functions::test

Rather than remove another test case from the TLS tests I reintroduced the previously removed functions::test case by adding a new_noop constructor

### DIFF
--- a/shotover-proxy/tests/cassandra_int_tests/schema_awaiter.rs
+++ b/shotover-proxy/tests/cassandra_int_tests/schema_awaiter.rs
@@ -5,22 +5,30 @@ use scylla::{Session, SessionBuilder};
 // But for cases like adding a new function we hit issues where the function is not yet propagated to all nodes.
 // So we make use of the scylla drivers await_schema_agreement logic to wait until all nodes are on the same schema.
 pub struct SchemaAwaiter {
-    session: Session,
+    session: Option<Session>,
 }
 
 impl SchemaAwaiter {
     pub async fn new(node: &str) -> Self {
         SchemaAwaiter {
-            session: SessionBuilder::new()
-                .known_node(node)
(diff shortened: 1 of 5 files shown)


benbromhead (approved):

shotover-proxy/tests/cassandra_int_tests/schema_awaiter.rs
```diff
@@ -5,22 +5,30 @@ use scylla::{Session, SessionBuilder};
...
+                    .build()
+                    .await
+                    .unwrap(),
+            ),
         }
     }
 
+    pub fn new_noop() -> Self {
+        SchemaAwaiter { session: None }
+    }
```
XA21X, line 27: I'm just curious, why have a no-op awaiter instead of not having an awaiter (i.e. Option<SchemaAwaiter>)?
Is it so we don't end up with confusing `None`s as positional arguments? I'm not familiar with Rust idioms.

XA21X (approved):

shotover-proxy/tests/cassandra_int_tests/schema_awaiter.rs
rukai: It makes implementing test cases easier, the test case can just call await_schema_agreement without caring whether the particular test needs an await or not.

Usually putting the Option outside of the type would make more sense as its more composable.
I consider what I've done here to be an unusual case.

PR #735, opened 2022-08-10
CassandraSinkCluster query system.local for local node info

rukai:
The change to the integration test demonstrates the change: we are now representing the entire cluster in the nodes list, previously we only included the 2 peers of the node we are querying but now we include the node itself as well.

Additionally I improved the error handling by:
* system_peers_into_node now returns errors for cases that are unrecoverable instead of silently ignoring the case.
   + A nice improvement this will bring is `CassandraOperation::Error`s are now logged - a possible case that could cause that is an auth error which is a realistic case.
* Retry querying the same peer 3 times and then move onto another handshake provided by another clients connection.

The old QueryParams values were blindly copy pasted in and on closer inspection they are not very good. e.g. page_size and timestamp should not be set.
I used the QueryParams::default because all of the default values happen to be exactly what we want.

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster.rs
@@ -264,7 +264,7 @@ pub fn create_topology_task(
             let mut attempts = 0;
             while let Err(err) = topology_task_process(&tls, &nodes, &handshake, &data_center).await
             {
-                tracing::error!("{err:?}");
+                tracing::error!("topology task failed, retrying, error was: {err:?}");
                 attempts += 1;
                 if attempts > 3 {
                     // 3 attempts have failed, lets try a new handshake
@@ -298,7 +298,9 @@ async fn topology_task_process(
             tracing_id: None,
             warnings: vec![],
             operation: CassandraOperation::Query {
-                query: Box::new(parse_statement_single("SELECT * FROM system.peers")),
+                query: Box::new(parse_statement_single(
+                    "SELECT peer, rack, data_center, tokens FROM system.peers",
+                )),
                 params: Box::new(QueryParams::default()),
             },
         })),
@@ -313,7 +315,9 @@ async fn topology_task_process(
             tracing_id: None,
             warnings: vec![],
             operation: CassandraOperation::Query {
-                query: Box::new(parse_statement_single("SELECT * FROM system.local")),
+                query: Box::new(parse_statement_single(
+                    "SELECT listen_address, rack, data_center, tokens FROM system.local",
+                )),
                 params: Box::new(QueryParams::default()),
             },
         })),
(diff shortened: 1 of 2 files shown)


shotover-proxy/src/transforms/cassandra/sink_cluster.rs
```diff
@@ -229,37 +235,50 @@ pub fn create_topology_task(
...
     data_center: &str,
 ) -> Result<()> {
-    let outbound = new_connection(handshake.address, &handshake.handshake, tls, &None).await?;
+    let outbound = new_connection(&handshake.address, &handshake.handshake, tls, &None).await?;
+
+    let (peers_tx, peers_rx) = oneshot::channel();
+    outbound.send(
+        Message::from_frame(Frame::Cassandra(CassandraFrame {
+            version: Version::V4,
+            stream_id: 1000,
```
conorbros, line 247: Why 1000? I thought setting to 0 was more common when you don't care about the stream id.

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
```diff
@@ -229,37 +235,50 @@ pub fn create_topology_task(
...
+        peers_tx,
+    )?;
 
-    let (return_chan_tx, return_chan_rx) = oneshot::channel();
+    let (local_tx, local_rx) = oneshot::channel();
     outbound.send(
         Message::from_frame(Frame::Cassandra(CassandraFrame {
             version: Version::V4,
-            stream_id: 0,
+            stream_id: 1001,
```
conorbros, line 264: Oh I see, we are using 1000 and 1001 for these 2 messages.

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
rukai: These are cleaned up frames that the driver output and thats where the 1000 value came from.
We could probably do 0 and 1 and to make it look less strange.

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
```diff
@@ -271,69 +290,70 @@ async fn topology_task_process(
...
-                        });
+                })
+                .map(|row| {
+                    if row.len() != 4 {
+                        return Err(anyhow!("expected 4 columns but was {}", row.len()));
                     }
-                }
-            }
+
+                    let tokens = if let Some(MessageValue::List(list)) = row.pop() {
```
conorbros, line 316: This is actually a bug in cassandra-protocol, the tokens should be a set. Happy to move ahead with as is, but once https://github.com/krojew/cdrs-tokio/pull/103 is merged and we update cassandra-protocol this will need to be fixed.

rukai: oooh good catch, yeah lets merge and fix it once the cdrs-tokio fix lands

conorbros (approved):

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
rukai: Hmmm before we land this I need to look into how robust our stream id handling is against repeating stream ids.

rukai: Oh its fine, in that case the value at stream_id in return_channel_map and return_message_map just gets replaced with the new message.

I've now replaced the stream_id's with 0 and 1.
I also tweaked the async logic a bit so that we can start processing whichever message comes back first.

benbromhead (approved):

PR #738, opened 2022-08-10
Fix intermittent failure (final I swear)

rukai:
I had originally thought that failures were a rare edge case that we should handle individually.
However its become clear that approach wont work.

To fix this issue once and for all I have replaced the illadvised SchemaAwaiter abstraction with logic in our CassandraConnection::execute method that will check for any schema altering commands and call await_schema_agreement() to block until the schema altering command is truly complete.

The rest of the PR should be just asyncification noise


rukai:
Still getting failures on this?
I wonder if a better fix would be to add a retry mechanism to CassandraConnection::execute?

benbromhead (approved):

conorbros (approved):

PR #739, opened 2022-08-10
CassandraSinkCluster implement USE routing

rukai:
When we receive a USE statement we need to:
* route it to all open connections
* route it to all future connections
* respond to the client with a single response

We could drastically simplify the USE statement error handling by just terminating all connections when we get a use statement.
I assume its worth keeping the connections alive but its possible that we are better off just terminating them.

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster.rs
@@ -244,7 +244,7 @@ impl CassandraSinkCluster {
                 .await
                 .map(|x| x.map_err(|e| anyhow!(e)));
             // If any errors occurred close the connection as we can no
-            // longer make any gaurantees about the current state of the connection
+            // longer make any guarantees about the current state of the connection
             if !is_use_statement_successful(response) {
                 self.local_nodes[node_index].outbound = None;
             }
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster.rs
@@ -6,13 +6,16 @@ use crate::frame::cassandra::parse_statement_single;
 use crate::frame::{CassandraFrame, CassandraOperation, CassandraResult, Frame};
 use crate::message::{Message, MessageValue, Messages};
 use crate::tls::{TlsConnector, TlsConnectorConfig};
+use crate::transforms::util::Response;
 use crate::transforms::{Transform, Transforms, Wrapper};
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use async_trait::async_trait;
 use cassandra_protocol::consistency::Consistency;
 use cassandra_protocol::frame::Version;
 use cassandra_protocol::query::QueryParams;
+use cql3_parser::cassandra_statement::CassandraStatement;
 use cql3_parser::common::{FQName, Identifier};
+use futures::StreamExt;
 use metrics::{register_counter, Counter};
 use rand::prelude::*;
 use serde::Deserialize;
@@ -49,6 +52,7 @@ pub struct CassandraSinkCluster {
     init_handshake: Vec<Message>,
     init_handshake_address: Option<SocketAddr>,
     init_handshake_complete: bool,
+    init_handshake_use_received: bool,
     chain_name: String,
     failed_requests: Counter,
     tls: Option<TlsConnector>,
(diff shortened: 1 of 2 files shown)


shotover-proxy/src/transforms/cassandra/sink_cluster.rs
```diff
@@ -336,6 +379,35 @@ fn get_nodes_from_system_peers(
...
+        if let CassandraOperation::Query { query, .. } = &mut frame.operation {
+            if let CassandraStatement::Use(_) = query.as_mut() {
+                return true;
+            }
+        }
+    }
+    false
+}
+
+fn is_use_statement_succesful(response: Option<Result<Response>>) -> bool {
```
conorbros, line 394: *successful

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
```diff
@@ -200,6 +231,18 @@ impl CassandraSinkCluster {
             super::connection::receive(self.read_timeout, &self.failed_requests, responses_future)
                 .await?;
 
+        for node_index in use_future_index_to_node_index {
+            let response = responses_future_use
+                .next()
+                .await
+                .map(|x| x.map_err(|e| anyhow!(e)));
+            // If any errors occurred close the connection as we can no
+            // longer make any gaurantees about the current state of the connection
```
conorbros, line 240: *guarantees

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
```diff
@@ -336,6 +379,35 @@ fn get_nodes_from_system_peers(
...
+}
+
+fn is_use_statement_successful(response: Option<Result<Response>>) -> bool {
+    if let Some(Ok(Response {
+        response: Ok(mut response),
+        ..
+    })) = response
+    {
+        if let Some(Frame::Cassandra(CassandraFrame {
+            operation: CassandraOperation::Result(CassandraResult::SetKeyspace(_)),
```
conorbros, line 401: Should we be checking that the keyspace set was the one we expect?

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
```diff
@@ -356,7 +428,7 @@ pub struct CassandraNode {
     pub address: IpAddr,
     pub _rack: String,
     pub _tokens: Vec<String>,
-    outbound: Option<CassandraConnection>,
+    pub outbound: Option<CassandraConnection>,
```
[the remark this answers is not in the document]
rukai: Because if I use `get_connection` then that will implicitly create a new connection which would perform really badly in the new case where I access outbound directly.

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
rukai: Are there any known cases where that could occur?
I feel its safe to assume that cassandra does its documented to do.

If the assumption could lead to a shotover panic then I would be more inclined to double check cassandra is behaving properly.

benbromhead (approved):

benbromhead:
> We could drastically simplify the USE statement error handling by just terminating all connections when we get a use statement.
> I assume its worth keeping the connections alive but its possible that we are better off just terminating them.

I would normally agree and users shouldn't be chaning keyspaces constantly over the course of the lifetime of a session. But I could also imagine someone useing USE between transactions rather than fully qualifying their keyspaces

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
conorbros: I was imagining some sort of race condition that with multiple consecutive `USE` statements but I think you're right and we should be right with this.

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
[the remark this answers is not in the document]
conorbros: Doesn't `get_connection` check if it's none before creating a new one?

conorbros (approved):

shotover-proxy/src/transforms/cassandra/sink_cluster.rs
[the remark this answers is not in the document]
conorbros: Approved because I'm certainly missing something here.

PR #741, opened 2022-08-11
cdrs-tokio test suite integration

conorbros:


conorbros:
Not ready for review yet, just trying to get the CI to run so I can check it's working.

benbromhead:
This seems stuck, are we pursuing this (or a todo later)?

shotover-proxy/Cargo.toml
```diff
@@ -11,6 +11,7 @@ license = "Apache-2.0"
 [features]
 # Include WIP alpha transforms in the public API
 alpha-transforms = []
+cpp-driver-tests = []
```
rukai, line 14: The `cpp` is a bit confusing, maybe one of:
* `cassandra-cpp-driver-tests` be specific about which driver we are enabling
* `external-dependency-integration-tests` be as generic as possible

shotover-proxy/src/message/mod.rs
```diff
@@ -300,7 +300,7 @@ impl Message {
             Metadata::Cassandra(metadata) => {
                 let body = CassandraOperation::Error(ErrorBody {
                     error_code: 0x1001,
-                    message: "".into(),
+                    message: "Server overloaded".into(),
```
rukai, line 303: Is this related to this PR?

shotover-proxy/Cargo.toml
```diff
@@ -92,6 +93,7 @@ reqwest = "0.11.6"
 metrics-util = "0.14.0"
 cdrs-tokio = { git = "https://github.com/krojew/cdrs-tokio" }
 scylla = { version = "0.4.7", features = ["ssl"] }
+rstest = "0.15.0"
```
rukai, line 96: is this unused?

shotover-proxy/tests/cassandra_int_tests/functions.rs
```diff
@@ -1,40 +1,29 @@
 use crate::helpers::cassandra::{assert_query_result, run_query, CassandraConnection, ResultValue};
 
 async fn drop_function(session: &CassandraConnection) {
-    assert_query_result(
-        session,
-        "SELECT test_function_keyspace.my_function(x, y) FROM test_function_keyspace.test_function_table WHERE id=1;",
-        &[&[ResultValue::Int(4)]]
-    ).await;
+    assert_query_result(session, "SELECT test_function_keyspace.my_function(x, y) FROM test_function_keyspace.test_function_table WHERE id=1;", &[&[ResultValue::Int(4)]]).await;
```
rukai, line 4: I know rustfmt isnt running on these files but can we keep manually formatting something like what rustfmt would do?
Its a lot easier to read with the function args split each on their own line instead of the entire function call on one line.

conorbros:
This PR enables us to run `cdrs-tokio` for our Cassandra integration tests. The cpp driver is hidden behind a feature that will also run if the feature is enabled. `rstest` is used to define what tests we want to run each driver for.

There is still some problems running cdrs-tokio on all the tests:
- I will need to get a fix upstreamed to cdrs-tokio to re-enable the collections tests (https://github.com/krojew/cdrs-tokio/pull/103).
- Many of the tests don't work with cdrs-tokio, still need to investigate the reasons for all of these. They have been marked with TODOs

I need to re-enable the collections test before this can be merged.

After that, need to decide whether to:
 - Only merge once cdrs-tokio runs on all the tests. 
-  Merge as is and follow up on problems in later PRs. 

I'm leaning towards merging once cdrs-tokio passes on all the tests.

Getting all the setup code for adding new drivers in could be useful though, for example the scylla driver could also be added.

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -1,61 +1,71 @@
...
 mod batch_statements;
 mod cache;
+#[cfg(feature = "cpp-driver-tests")]
 #[cfg(feature = "alpha-transforms")]
 mod cluster;
+#[cfg(feature = "cpp-driver-tests")]
 #[cfg(feature = "alpha-transforms")]
 mod cluster_multi_rack;
-mod collections;
+//mod collections;
```
rukai, line 28: Just sticking a comment here to remind myself this is unaddressed

shotover-proxy/Cargo.toml
rukai: Oh sorry, searching `rstest` didnt turn up any results, I must have failed to expand the hidden file it was in.
:+1:

rukai:
> I'm leaning towards merging once cdrs-tokio passes on all the tests.

As long as we have integration test cases running on at least the cassandra_cpp driver and we are happy with the multi-driver abstraction then I am happy to merge.

shotover-proxy/Cargo.toml
conorbros: Went with `cassandra-cpp-driver-tests` until we need to be generic.

shotover-proxy/src/message/mod.rs
conorbros: Yes. Now that we are testing the error messages generically I added this to make that possible. Plus it makes sense to include this message so might as well.

rukai:
The test run time has halved from 40m -> 20m, I think a bunch of tests are not longer running.

conorbros:
Fixed the issue with it missing some tests, review can continue now.

shotover-proxy/tests/cassandra_int_tests/functions.rs
```diff
@@ -22,14 +24,15 @@ async fn create_function(session: &CassandraConnection) {
...
+              "CREATE KEYSPACE test_function_keyspace WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 };"
     ).await;
+
     run_query(
-        session,
-        "CREATE TABLE test_function_keyspace.test_function_table (id int PRIMARY KEY, x int, y int);",
-    ).await;
+            session,
+            "CREATE TABLE test_function_keyspace.test_function_table (id int PRIMARY KEY, x int, y int);",
+        ).await;
```
rukai, line 34: I'm not sure how this happened but the changes here are just messing up the formatting.

shotover-proxy/tests/helpers/cassandra.rs
```diff
@@ -1,36 +1,201 @@
...
+                let user = "cassandra";
+                let password = "cassandra";
+                let auth = StaticPasswordAuthenticatorProvider::new(&user, &password);
+
+                let node_addresses = contact_points
+                    .split(',')
+                    .map(|contact_point| NodeAddress::from(format!("{contact_point}:{port}")))
+                    .collect::<Vec<NodeAddress>>();
+
+                let config = futures::executor::block_on(
```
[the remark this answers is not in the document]
conorbros: Just thought it was a bit more convenient since `cassandra-cpp` is sync.

shotover-proxy/tests/cassandra_int_tests/collections.rs
```diff
@@ -237,16 +239,31 @@ mod list {
 
         // test selecting list of sets
         for (i, native_col_type) in NATIVE_COL_TYPES.iter().enumerate() {
+            let expected: Vec<Vec<ResultValue>> = match driver {
```
rukai, line 242: If we stick a single one of these at the top:
```rust
            let new_set = match driver {
                CassandraDriver::CdrsTokio => ResultValue::Set,
                CassandraDriver::Datastax => ResultValue::List,
            };
```

Then we can implement each of these with something like:
```rust
            assert_query_result(
                session,
                format!(
                    "SELECT my_set FROM test_collections_keyspace.test_set_table_{}_0;",
                    i
                )
                .as_str(),
                &[&[ResultValue::List(vec![new_set(vec![
                    get_type_example_result_value(*native_col_type),
                ])])]],
            )
            .await;
```

shotover-proxy/tests/helpers/cassandra.rs
```diff
@@ -157,21 +392,81 @@ impl CassandraConnection {
...
+                ResResultBody::Void => vec![],
+            },
+            _ => todo!(),
         }
     }
 }
 
 #[derive(Debug, Clone, PartialOrd, Eq, Ord)]
 pub enum ResultValue {
+    #[cfg(feature = "cassandra-cpp-driver-tests")]
```
rukai, line 469: why is this feature flagged out?

conorbros: I definitely had some reason 😂, but not needed anymore. I'll remove that.

conorbros: Because only the cassandra-cpp driver uses it. I'm inclined to remove it but another driver we add might use it.

rukai: I took a look at the protocol spec.
It is a bit of a mess but what I'm reading from it is that varchar and text share the same id: `0x000D` so theres no way to differentiate them at the protocol level.
So lets completely remove the text variant and just map anything that comes up as text into Varchar. 

IMO text is a nicer name and we should just use that instead of varchar everywhere but thats for a future cleanup PR.

shotover-proxy/tests/helpers/cassandra.rs
```diff
@@ -298,18 +609,61 @@ impl ResultValue {
...
+            CassandraType::Set(set) => {
+                let mut elements = Vec::new();
+                for element in set {
+                    elements.push(ResultValue::new_from_cdrs(element, version));
+                }
+                ResultValue::Set(elements)
+            }
+            CassandraType::Udt(_) => todo!(),
+            CassandraType::Tuple(_) => todo!(),
+            CassandraType::Null => todo!(),
```
rukai, line 665: TODO's left

conorbros: Those were already there in the cpp impl. Could we not block this PR on those?

shotover-proxy/tests/helpers/cassandra.rs
```diff
@@ -274,42 +568,98 @@ impl ResultValue {
...
                         .unwrap()
-                        .map(|(k, v)| (ResultValue::new(k), ResultValue::new(v)))
+                        .map(|(k, v)| (ResultValue::new_from_cpp(k), ResultValue::new_from_cpp(v)))
                         .collect(),
                 ),
                 ValueType::UNKNOWN => todo!(),
                 ValueType::CUSTOM => todo!(),
                 ValueType::UDT => todo!(),
                 ValueType::TUPLE => todo!(),
+                ValueType::TEXT => unimplemented!(),
```
rukai, line 604: Can we include in the unimplemented!() a message explaining that text is represented by the same id as varchar at the protocol level and therefore will never be instantiated by the datastax cpp driver

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -541,67 +589,51 @@ async fn test_cassandra_request_throttling() {
...
         for i in 0..60 {
-            let statement = format!("INSERT INTO test_keyspace.my_table (id, lastname, firstname) VALUES ({}, 'text', 'text')", i);
-            batch.add_statement(&stmt!(statement.as_str())).unwrap();
+            queries.push(("INSERT INTO test_keyspace.my_table (id, lastname, firstname) VALUES (?, 'text', 'text')".into(), i));
         }
-        let result = connection.execute_batch_expect_err(&batch);
-        assert!(matches!(
-            result,
-            Error(
-                ErrorKind::CassErrorResult(cassandra_cpp::CassErrorCode::SERVER_OVERLOADED, ..),
```
rukai, line 565: I dont want to regress on important testing here.
The error code is the actual important bit here that will alter how the client processes the error so we need to at least be testing that.
Maybe execute_batch_expect_err should return a:
```rust
struct CassandraError {
    code: i32,
    message: String 
}
```
That we map cdrs-tokio and casandra_cpp types into.

The `code` field could be given a custom enum instead of an i32 but not necessary.

shotover-proxy/tests/cassandra_int_tests/batch_statements.rs
```diff
@@ -9,12 +8,15 @@ pub async fn test(connection: &CassandraConnection) {
     }
 
     {
-        let mut batch = Batch::new(BatchType::LOGGED);
+        let mut queries: Vec<(String, i32)> = vec![];
         for i in 0..2 {
-            let statement = format!("INSERT INTO batch_keyspace.batch_table (id, lastname, firstname) VALUES ({}, 'text1', 'text2')", i);
-            batch.add_statement(&stmt!(statement.as_str())).unwrap();
+            queries.push((
+                "INSERT INTO batch_keyspace.batch_table (id, lastname, firstname) VALUES (?, 'text1', 'text2')".into(),
```
rukai, line 14: I dont understand why we are changing execute_batch to perform binding?
Arent we better off just using plain old queries?

conorbros: Hmm I had assumed you couldn't use it that way with `cdrs-tokio` but I might be wrong. Will have a look into that.

shotover-proxy/tests/helpers/cassandra.rs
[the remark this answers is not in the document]
conorbros: Ended up just switching to async.

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -457,39 +508,42 @@ async fn test_cassandra_peers_rewrite_cassandra3() {
...
-    assert!(matches!(
+    assert_eq!(
         result,
-        Error(
-            ErrorKind::CassErrorResult(cassandra_cpp::CassErrorCode::SERVER_INVALID_QUERY, ..),
-            _
-        )
-    ));
+        CassandraError {
+            code: 0x2200.into(),
```
rukai, line 522: wouldnt it make sense to just use the enum variant name considering we went to the trouble of setting it up?

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -542,67 +596,57 @@ async fn test_cassandra_request_throttling() {
...
+        let result = connection.execute_batch_expect_err(queries);
+        assert_eq!(
             result,
-            Error(
-                ErrorKind::CassErrorResult(cassandra_cpp::CassErrorCode::SERVER_OVERLOADED, ..),
-                ..
-            )
-        ));
+            CassandraError {
+                code: 0x1001.into(),
```
rukai, line 618: wouldnt it make sense to just use the enum variant name considering we went to the trouble of setting it up?

shotover-proxy/tests/cassandra_int_tests/prepared_statements.rs
```diff
@@ -20,66 +18,42 @@ async fn delete(session: &CassandraConnection) {
...
 }
 
-async fn insert(session: &CassandraConnection) {
-    let prepared = session
-        .prepare("INSERT INTO test_prepare_statements.table_1 (id, x, name) VALUES (?, ?, ?);");
+fn insert(session: &CassandraConnection) {
+    let prepared = session.prepare("INSERT INTO test_prepare_statements.table_1 (id) VALUES (?);");
 
-    let mut statement = prepared.bind();
-    statement.bind_int32(0, 1).unwrap();
```
rukai, line 28: Our tests here are becoming worse with this change, we no longer test binding a string variable.
But the original tests are pretty lackluster in the first place so im not super worried.

That said, what is your plan for extending execute_prepared into an API that will allow more coverage? Might be worth leaving a TODO so we have some idea of the intention down the line.

If we decide its not worth abstracting the APIs here maybe we could just match on the connection and handle each type seperately?
Otherwise something like this might be the best way forward?
```rust
    pub fn execute_prepared(
        &self,
        prepared_query: &PreparedQuery,
        value: Vec<ResultValue>
    ) -> Vec<Vec<ResultValue>> {
```

shotover-proxy/tests/cassandra_int_tests/protect.rs
```diff
@@ -12,30 +11,29 @@ pub struct Protected {
...
 pub async fn test(shotover_session: &CassandraConnection, direct_session: &CassandraConnection) {
-    run_query(shotover_session, "CREATE KEYSPACE test_protect_keyspace WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 };").await;
     run_query(
         shotover_session,
-        "CREATE TABLE test_protect_keyspace.test_table (pk varchar PRIMARY KEY, cluster varchar, col1 blob, col2 int, col3 boolean);",
+        "CREATE KEYSPACE test_protect_keyspace WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 };"
+    ).await;
+
+    run_query(shotover_session,
+              "CREATE TABLE test_protect_keyspace.test_table (pk varchar PRIMARY KEY, cluster varchar, col1 blob, col2 int, col3 boolean);"
```
rukai, line 20: double indent

shotover-proxy/tests/helpers/cassandra.rs
```diff
@@ -1,36 +1,234 @@
...
+                cluster.set_ssl(&mut ssl);
+
+                CassandraConnection::Datastax {
+                    session: cluster.connect().unwrap(),
+                    schema_awaiter: None,
+                }
+            }
+            // TODO actually implement TLS for cdrs-tokio
+            CassandraDriver::CdrsTokio => {
+                let user = "cassandra";
```
rukai, line 212: maybe just make this branch into a todo!() instead of a bunch of code that doesnt work.

shotover-proxy/tests/helpers/cassandra.rs
```diff
@@ -42,7 +240,19 @@ impl CassandraConnection {
             context.build()
         });
         match self {
-            CassandraConnection::Datastax { schema_awaiter, .. } => {
+            #[cfg(feature = "cassandra-cpp-driver-tests")]
+            Self::Datastax { schema_awaiter, .. } => {
+                *schema_awaiter = Some(
+                    SessionBuilderScylla::new()
```
rukai, line 246: why dont we first get a mutable reference to schema_awaiter so that we dont have to duplicate this logic?

shotover-proxy/tests/helpers/cassandra.rs
```diff
@@ -56,99 +266,163 @@ impl CassandraConnection {
         }
     }
 
+    async fn await_schema_agreement(&self) {
+        match self {
+            #[cfg(feature = "cassandra-cpp-driver-tests")]
+            Self::Datastax { schema_awaiter, .. } => {
+                if let Some(schema_awaiter) = schema_awaiter {
+                    schema_awaiter.await_schema_agreement().await.unwrap();
```
rukai, line 274: deduplicate this logic

shotover-proxy/tests/helpers/cassandra.rs
rukai: null and tuple now have implementations for cpp and are very easy to implement here.
I very much agree with Udt though, dont implement that here.

shotover-proxy/tests/cassandra_int_tests/prepared_statements.rs
conorbros: I could have it similar to how I had the batch queries previously. 

You could pass the query and the variables you wanted to bind and handle for each driver type separately. Would be annoying to handle separate types so we could just test binding ints, which should be enough testing for our purposes.

shotover-proxy/tests/helpers/cassandra.rs
conorbros: Sure, I'll implement null and tuple then.

shotover-proxy/tests/cassandra_int_tests/prepared_statements.rs
rukai: I dont think just testing ints is enough, I think we would want to properly exercise the protocol level handling of different types here.

Gonna go ahead and resolve this, as theres nothing to really block on here, its just future work.

shotover-proxy/tests/helpers/cassandra.rs
conorbros: Where's the impl of tuple and null for cpp?

rukai: image

rukai: wow I cant believe rust didnt warn us about the duplicated ValueType::TUPLE :no_mouth: 
image

conorbros: Oh lol. Somehow both were hidden

rukai (approved):
I think we might end up removing the rstest approach so that we can reuse Cassandra clusters between tests that have identical Cassandra setups.

The rstest approach is fine for now though, let's land this and iterate further in main.

benbromhead (approved):

PR #742, opened 2022-08-11
Update to rust 1.63

rukai:
I opted to disable the new derive_partial_eq_without_eq lint, because blanket enforcing full Eq seems like a bad idea when the type might not meet the semantics of Eq


conorbros:
Does #738 fix that failure?

rukai:
yep, that is the kind of intermittent failure that #738 fixes

conorbros (approved):

benbromhead (approved):

PR #744, opened 2022-08-12
CassandraSinkCluster rewrite system.local

rukai:
Misc changes:
* The `v4` uuid feature is needed to enable random uuid generation
* The changes to `ResultValue::new` are needed to allow handling null values in cases that I hit where it wasnt. There are still other null value cases that its not handling properly but I dont know where cassandra will and wont return null values so I'm just leaving the rest to be dealt with when we hit them.
* I introduced an `Any` variant to `ResultValue`, the idea here is to allow `assert_query_results` in cases where there are values that are non deterministic. assert_query_results provides an intuitive way to assert on both the length and the contents of the results so I think `Any` is a lot better approach then just manually doing something like `assert_eq!(results[3], ResultValue::Varchar("blah"))`
* I had to stub out the system.peers table as the cassandra_cpp driver kept routing system.local queries directly to cassandra nodes instead of going through shotover causing my assertions on the system.local table results to fail.

system.local fields:
* release_version - I take the lowest value from any node as this field is most likely to be used for the existence of features so if some nodes dont have a feature then we should consider shotover as not having that feature
* tokens - combine all tokens from nodes in our dc-rack into a single token list, token generation is non-deterministic so we cant hardcode this.
* schema_version - if any peer nodes or the local node (routed to init_handshake_connection) disagree on the schema version a random uuid is returned, otherwise the agreed schema version is returned. DDL queries are also routed to init_handshake_connection which means that a schema disagreement will always be conveyed immediately after a DDL query and it will last until the local node has received gossip that schema agreement has occurred.
* listen_address - extract shotovers address from the client<->shotover socket, this gives us a sane default and the user may still rewrite this address via the CassandraPeersRewrite transform if needed.
* broadcast_address - same as listen_address
* host_id - configured host id. cassandra nodes will generate their own host id and store it without configuration. I didnt take that approach however because we want shotover to be completely stateless. And I think asking the user to manually set a host_id is better than introducing statefulness. Another option is to just randomly generate the host_id each time shotover starts up, but I think its better to be persistent in case drivers rely on that.
* rack - configured rack
* data_center - configured data center
* everything else is just left as is from the init_handshake_connection's response

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
@@ -16,6 +16,7 @@ use cql3_parser::common::{FQName, Identifier};
 use cql3_parser::select::SelectElement;
 use futures::stream::FuturesOrdered;
 use futures::StreamExt;
+use itertools::Itertools;
 use metrics::{register_counter, Counter};
 use node::CassandraNode;
 use rand::prelude::*;
@@ -178,22 +179,7 @@ impl CassandraSinkCluster {
             .collect();
 
         for table_to_rewrite in tables_to_rewrite.iter().rev() {
-            let mut stream_id = 0;
-            let mut restart = true;
-            while restart {
-                restart = false;
-                for message in &mut messages {
-                    if let Some(Frame::Cassandra(frame)) = message.frame() {
-                        if stream_id == frame.stream_id {
-                            match stream_id.checked_add(1) {
-                                Some(new_stream_id) => stream_id = new_stream_id,
-                                None => return Err(anyhow!("Ran out of stream_ids")),
-                            }
-                            restart = true;
-                        }
-                    }
-                }
-            }
+            let stream_id = get_unused_stream_id(&messages)?;
 
             if let RewriteTableTy::Local = table_to_rewrite.ty {
                 messages.insert(
@@ -772,6 +758,21 @@ fn is_use_statement_successful(response: Option<Result<Response>>) -> bool {
         }
     }
     false
+}
+
+fn get_unused_stream_id(messages: &Messages) -> Result<i16> {
+    // start at an unusual number to hopefully avoid looping many times when we receive stream ids that look like [0, 1, 2, ..]
+    // We can quite happily give up 358 stream ids as that still allows for shotover message batches containing 2 ** 16 - 358 = 65178 messages
+    for i in 358..i16::MAX {
+        if !messages
+            .iter()
+            .filter_map(|message| message.stream_id())
+            .contains(&i)
+        {
+            return Ok(i);
+        }
+    }
+    Err(anyhow!("Ran out of stream ids"))
 }
 
 struct SystemPeer {
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
@@ -2,16 +2,13 @@ use crate::codec::cassandra::CassandraCodec;
 use crate::error::ChainResponse;
 use crate::frame::cassandra::parse_statement_single;
 use crate::frame::{CassandraFrame, CassandraOperation, CassandraResult, Frame};
-use crate::message::{IntSize, Message, MessageValue, Messages};
+use crate::message::{Message, MessageValue, Messages};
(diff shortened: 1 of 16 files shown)


shotover-proxy/tests/cassandra_int_tests/udt.rs
```diff
@@ -25,10 +25,14 @@ async fn test_drop_udt(session: &CassandraConnection) {
...
     .await;
     run_query(session, "DROP TYPE test_udt_keyspace.test_type_drop_me;").await;
-    session.execute_expect_err_contains(
-        "CREATE TABLE test_udt_keyspace.test_delete_table (id int PRIMARY KEY, foo test_type_drop_me);",
-        "Unknown type test_udt_keyspace.test_type_drop_me",
-    );
+    // TODO: This exposes a bug in at least cassandra 3
+    // The error is supposed to be a 0x2200 syntax error but rarely cassandra will return a 0x0000 server error instead.
+    // The cpp driver interprets 0x0000 as different to 0x2200 and considers the node as down and will no longer talk to it.
+    // We should eventually reenable this test when we upgrade to cassandra 4 (the bug may also need to be reported and fixed in cassandra 4)
```
conorbros, line 31: Fixed in Cassandra 3*

rukai: The bug could still exist in cassandra 4, I've updated the comment to account for that.

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -121,7 +122,7 @@ async fn test_cluster() {
         native_types::test(&connection2).await;
     }
 
-    cluster::test().await;
+    cluster::test_topology_task().await;
```
conorbros, line 125: Why does this test not run `cluster::test()`?

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -161,6 +162,7 @@ async fn test_source_tls_and_cluster_tls() {
     collections::test(&connection).await;
     prepared_statements::test(&connection).await;
     batch_statements::test(&connection).await;
+    cluster::test(&connection).await;
```
conorbros, line 165: Can we not run the `test_topology_task` test as well here?

shotover-proxy/src/transforms/mod.rs
```diff
@@ -378,6 +379,7 @@ pub struct Wrapper<'a> {
     pub messages: Messages,
     transforms: IterMut<'a, Transforms>,
     pub client_details: String,
+    pub local_addr: SocketAddr,
```
conorbros, line 382: What does this field do?

rukai: I added a comment to document it.

shotover-proxy/tests/cassandra_int_tests/mod.rs
rukai: It took a sizeable change so I did it in a separate PR https://github.com/shotover/shotover-proxy/pull/754

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -23,13 +25,20 @@ use std::net::SocketAddr;
...
 use std::time::Duration;
 use tokio::sync::{mpsc, oneshot, RwLock};
+use uuid::Uuid;
+use version_compare::Cmp;
 
 mod node;
 
 #[derive(Deserialize, Debug, Clone)]
 pub struct CassandraSinkClusterConfig {
+    /// contact points must be within the specified data_center and rack.
```
conorbros, line 35: Would it be possible to emit a warning if this occurs?

rukai: there is a warning! :D

rukai: `racing::warn!("A contact point node is not in the configured rack, this node will receive traffic from outside of its rack");`

conorbros: Ah I see, missed that. Thanks

benbromhead:
Nice!

benbromhead (approved):

conorbros (approved):

PR #745, opened 2022-08-15
Fix intermittent redis failure

rukai:
Fixes this intermittent failure I saw on CI and could reproduce locally:
image

Its strange because `CLUSTER NODES` will succeed and then `CLUSTER REPLICAS` will return an empty list.
I imagine the cause is some state specific to `CLUSTER REPLICAS` that redis needs to populate during initialization before `CLUSTER REPLICAS` can return sensible values.

I have run this script for a while:
```bash
set -e
set -u
while :;
do
    cargo test --all-features redis_int_tests::test_cluster_handling_redis;
done
```
And the only failure I could get is another presumably unrelated redis failure: `An error was signalled by the server: The specified node is not a master`


conorbros:
image

Did you ever get this one?

rukai:
Never, seen that one before, are you getting that locally atm?

conorbros:
Yeah getting that on this branch.

benbromhead (approved):

conorbros:
Thanks to your new assert messages: 

image

rukai:
Does the failure occur every run or is it intermittent?
And does it occur on main branch as well?

conorbros:
Intermittent, and no I was not able to reproduce on main.

conorbros (approved):

PR #755, opened 2022-08-24
ConnectionFactory

conorbros:
A change pulled from the `NodePool` PR. No functional changes, just encapsulates the logic for creating new connections.

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster/node.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster/node.rs
@@ -32,12 +32,23 @@ impl CassandraNode {
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct ConnectionFactory {
     init_handshake: Vec<Message>,
     use_message: Option<Message>,
     tls: Option<TlsConnector>,
     pushed_messages_tx: Option<mpsc::UnboundedSender<Messages>>,
+}
+
+impl Clone for ConnectionFactory {
+    fn clone(&self) -> Self {
+        Self {
+            init_handshake: self.init_handshake.clone(),
+            use_message: None,
+            tls: self.tls.clone(),
+            pushed_messages_tx: None,
+        }
+    }
 }
 
 impl ConnectionFactory {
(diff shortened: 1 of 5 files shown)


shotover-proxy/src/transforms/cassandra/sink_cluster/node.rs
```diff
@@ -18,39 +18,86 @@ pub struct CassandraNode {
...
+    pub fn new(tls: Option<TlsConnector>) -> Self {
+        Self {
+            init_handshake: vec![],
+            use_message: None,
+            tls,
+            pushed_messages_tx: None,
+        }
+    }
+
+    pub fn clone_config(&self) -> Self {
```
rukai, line 53: its a bit concerning to me having multiple clone implementations and trying to remember which is which.
Maybe we could call this `new_with_same_tls` or `new_with_same_config` so it doesnt feel like we are cloning?

rukai: Oh we could completely avoid needing the second clone implementation by moving tls outside of this struct and back int o CassandraSinkCluster.
To me that separation makes sense because tls is immutable config while the rest of this struct is mutable state.

Up to you.

rukai (commented):
Heres some initial feedback

shotover-proxy/src/transforms/cassandra/sink_cluster/node.rs
conorbros: We need to be able to clone it to pass to the topology task. In that sense it's immutable, once it's given to the task it is no longer mutated.

Maybe I should add more docs but to me it was clear that this clone is for when you only want the config for the new connection (just TLS for now), and the default clone is for when you want everything within the struct.

rukai (approved):
I'm not 100% confident this is the best way to organize this state.
My main concern is the multiple clone implementations that clone various internal state, you have to understand the internals of these functions to know which one is the right one to call.
Edit: Oh the new name is clearer, I missed that before.

But the new abstraction does also have some advantages so... LGTM :+1:

benbromhead (approved):

PR #756, opened 2022-08-24
CassandraSinkCluster system.peers rewrite

rukai:
Introduced a new integration test `test_cluster_multi_rack` it tests all the same cases as `test_cluster` but each cassandra node is on a different rack and each cassandra node has its own shotover instance in front of it.
The existing `test_cluster` was renamed to `test_cluster_single_rack` and now runs two shotover test cases against the cassandra cluster.
One to test an empty peers list and one to test a consistent peers list.

The derived fields of system.peers should be non controversial as its the same as system.local.

## Config design thoughts
This PR configures the configurable fields with the following configuration:
```yaml
    - CassandraSinkCluster:
        first_contact_points: ["172.16.1.2:9042", "172.16.1.3:9042"]
        data_center: "dc1"
        rack: "rack1"
        host_id: "2dd022d6-2937-4754-89d6-02d2933a8f7a"
        shotover_peers:
          - address: "127.0.0.2:9042"
            data_center: "dc1"
            rack: "rack2"
            host_id: "3c3c4e2d-ba74-4f76-b52e-fb5bcee6a9f4"
          - address: "127.0.0.3:9042"
            data_center: "dc2"
            rack: "rack1"
            host_id: "fa74d7ec-1223-472b-97de-04a32ccdb70b"
```

I am strongly considering changing this such that each config specifies a complete list of nodes and then specifies which node is the local node:
```yaml
    - CassandraSinkCluster:
        first_contact_points: ["172.16.1.2:9042", "172.16.1.3:9042"]
        local_host_id: "2dd022d6-2937-4754-89d6-02d2933a8f7a"
        nodes:
          - address: "127.0.0.1:9042"
            data_center: "dc1"
            rack: "rack1"
            host_id: "2dd022d6-2937-4754-89d6-02d2933a8f7a"
          - address: "127.0.0.2:9042"
            data_center: "dc1"
            rack: "rack2"
            host_id: "3c3c4e2d-ba74-4f76-b52e-fb5bcee6a9f4"
          - address: "127.0.0.3:9042"
            data_center: "dc2"
            rack: "rack1"
            host_id: "fa74d7ec-1223-472b-97de-04a32ccdb70b"
```
This should make config generation/authoring drastically simpler for multi shotover instance setups but at the cost of some complexity for single shotover instance setups.

However I would like to land this PR with the first implementation as its a lot simpler to implement, then we can move to the second implementation (if we agree) post MVP.

### DIFF
--- a/shotover-proxy/example-configs/cassandra-cluster-multi-rack/topology_rack1.yaml
+++ b/shotover-proxy/example-configs/cassandra-cluster-multi-rack/topology_rack1.yaml
@@ -7,10 +7,12 @@ chain_config:
   main_chain:
     - CassandraSinkCluster:
         first_contact_points: ["172.16.1.2:9042", "172.16.1.3:9042"]
-        data_center: "dc1"
-        rack: "rack1"
-        host_id: "2dd022d6-2937-4754-89d6-02d2933a8f7a"
-        shotover_peers:
+        local_shotover_host_id: "2dd022d6-2937-4754-89d6-02d2933a8f7a"
+        shotover_nodes:
+          - address: "127.0.0.1:9042"
+            data_center: "dc1"
+            rack: "rack1"
+            host_id: "2dd022d6-2937-4754-89d6-02d2933a8f7a"
           - address: "127.0.0.2:9042"
             data_center: "dc1"
             rack: "rack2"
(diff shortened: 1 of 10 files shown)


benbromhead (approved):

rukai:
I ended up rewriting to use the second config approach because:
* I realized it was actually pretty simple implement
* I wanted to use this approach eventually so may as well just use it now
* I realized we lacked any kind of local ip rewriting and this was an elegant way to include it

It is added as a separate commit so its easy to back out and revert to the previous implementation.

conorbros (approved):

PR #757, opened 2022-08-25
Add test_cluster_multi_rack

rukai:
Pulled out of https://github.com/shotover/shotover-proxy/pull/756
Should reduce the scope of 756 a bit and will also let me test and fix a preexisting issue.

Introduced a new integration test `test_cluster_multi_rack` it tests all the same cases as `test_cluster` but each cassandra node is on a different rack and each cassandra node has its own shotover instance in front of it.

The fix to the issue is not complete but a full fix will require deciding between:
1. changing the design of the transform such that providing a node outside of the datacenter/rack in the contact points will cause shotover to return errors to every incoming message as upholding our invariants in that case is too tricky.
2. rewrite system.local rewriting to send an additional system.local

This PR makes enough progress to 2 to fix the test failure but does not commit to sending an additional system.local message yet.

Lets land this PR, its strictly an improvement to both implementation and integration tests, and we can decide between 1 or 2 later on.

### DIFF
--- a/shotover-proxy/example-configs/cassandra-cluster-multi-rack/docker-compose.yml
+++ b/shotover-proxy/example-configs/cassandra-cluster-multi-rack/docker-compose.yml
@@ -1,0 +1,73 @@
+version: "3.3"
+networks:
+  cluster_subnet:
+    name: cluster_subnet
+    driver: bridge
+    ipam:
+      driver: default
+      config:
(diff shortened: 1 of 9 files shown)


benbromhead (approved):
as discused on slack, g2g

conorbros (approved):

PR #761, opened 2022-08-26
Only show filter warning once

conorbros:
This makes the test logs very noisy, we shouldn't show it for every message. Just once per `Wrapper` would be enough.


rukai:
which test was hitting this problem again?

rukai:
Test the affect of this PR run: `cargo test redis_int_tests::test_cluster_dr_redis`

As currently implemented it doesnt have much of an affect because shown_error isnt persisted across multiple message batches.

rukai (approved):
LGTM

benbromhead (approved):
LGTM

PR #763, opened 2022-08-26
Drop docker logs to trace level

conorbros:
pulled from #741 

I use the debug level often and these logs get pretty noisy when it's on services that takes a while to start (such as the Cassandra cluster ones).


rukai (approved):
LGTM!

benbromhead (approved):

PR #765, opened 2022-08-29
Log cassandra connection errors

rukai:
tx_process and rx_process are run as tasks and we dont await on the results so the errors are just silently ignored. 
:no_mouth: 

Possibly we should be giving a more useful error for end users, "tx_process failed" feels very "implementation detail"y, but definitely an improvement over nothing.


rukai:
Marked as draft as it has false positives, so will need to investigate a bit.

rukai:
Nope, those are not false positives.
They are problems with our implementation that need to be fixed.
Lets land this improved logging and then fix the issues that the logging is surfacing.

benbromhead (approved):

conorbros (approved):

PR #766, opened 2022-08-29
CassandraConnection: Replace generic codec with direct usage of CassandraCodec

rukai:
There is no advantage to using generics here as CassandraConnection will only ever be used with a CassandraCodec so we should directly use CassandraCodec to improve code readability.


conorbros (approved):
lol, pretty sure this was me. Good catch.

benbromhead (approved):

PR #772, opened 2022-08-30
Cleanup cassandra protocol negotiation

rukai:
Improves error reporting and makes the cassandra protocol implementation easier to understand.
Downgrades the warnings caused by cassandra protocol negotiation to `info!` level.
It certainly makes the integration tests quieter, but maybe the slight performance loss of protocol negotiation is worth a warning?

Theres some magic trait type stuff going on here:
In these lines:
```rust
pub trait CodecReadHalf: Decoder<Item = Messages, Error = CodecReadError> + Clone + Send {}
impl<T: Decoder<Item = Messages, Error = CodecReadError> + Clone + Send> CodecReadHalf for T {}
```
We can set any type to be the Error as long as it implements `From<std::io::Error>`.
Internally tokio then uses that implementation to generate errors in our type when it hits an `io::Error` while receiving from the TcpStream.
So while we never create a `CodecReadError::Io` directly, tokio will return those sometimes when we call `FrameRead::next`.

### DIFF
--- a/shotover-proxy/src/codec/cassandra.rs
+++ b/shotover-proxy/src/codec/cassandra.rs
@@ -1,25 +1,18 @@
 use crate::frame::cassandra::CassandraOperation;
 use crate::frame::{CassandraFrame, Frame, MessageType};
 use crate::message::{Encodable, Message, Messages};
-use anyhow::{anyhow, Error, Result};
+use crate::server::CodecReadError;
+use anyhow::{anyhow, Result};
 use bytes::{Buf, BufMut, BytesMut};
 use cassandra_protocol::compression::Compression;
 use cassandra_protocol::frame::message_error::{AdditionalErrorInfo, ErrorBody};
 use cassandra_protocol::frame::{CheckEnvelopeSizeError, Envelope as RawCassandraFrame, Version};
 use tokio_util::codec::{Decoder, Encoder};
-use tracing::{debug, info};
+use tracing::info;
 
 #[derive(Debug, Clone)]
 pub struct CassandraCodec {
     compressor: Compression,
-    /// if force_close is Some then the connection will be closed the next time the
-    /// system attempts to read data from it.  This is used in protocol errors where we
-    /// need to return a message to the client so we can not immediately close the connection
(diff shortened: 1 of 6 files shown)


shotover-proxy/src/codec/cassandra.rs
```diff
@@ -113,14 +76,42 @@ impl Decoder for CassandraCodec {
...
+        "Negotiating protocol version: rejecting version {} (configure the client to use a supported version by default to improve connection time)",
+        version
+    );
+
+    CodecReadError::RespondAndThenCloseConnection(vec![Message::from_frame(Frame::Cassandra(
+        CassandraFrame {
+            version: Version::V4,
+            stream_id: 0,
+            operation: CassandraOperation::Error(ErrorBody {
+                error_code: 0xA, // https://github.com/apache/cassandra/blob/adf2f4c83a2766ef8ebd20b35b49df50957bdf5e/doc/native_protocol_v4.spec#L1053
```
benbromhead, line 105: Is this the right error code to throw for an unsupported version? The spec just says this if for a protocol error, an unsupported version may technically not be a protocol error itself?

What happens when we do a pcap on a bad version negotation with a direct connection to cassandra (or what does the cassandra code do)

rukai: A ProtocolException is thrown for invalid versions: https://github.com/apache/cassandra/blob/cd576a0d140c94007fd0181044df453c0681252f/src/java/org/apache/cassandra/transport/ProtocolVersion.java#L113

A ProtocolException corresponds to PROTOCOL_ERROR https://github.com/apache/cassandra/blob/cd576a0d140c94007fd0181044df453c0681252f/src/java/org/apache/cassandra/transport/ProtocolException.java#L43

Which is the magic 0xA int: https://github.com/apache/cassandra/blob/cd576a0d140c94007fd0181044df453c0681252f/src/java/org/apache/cassandra/exceptions/ExceptionCode.java#L35

benbromhead (approved):

conorbros (approved):

conorbros:
Let's merge this, I assume Ben just forgot to hit the resolve button.

PR #774, opened 2022-08-31
add timeout to new tcp connections

### LINKED ISSUE #769: Ensure all TcpStream::connect are wrapped in a timeout
`TcpStream::connect` will block indefinitely in some cases (I assume when the destination just completely ignores us or doesnt exist)
We should add a timeout to all `TcpStream::connect`s so that if a destination node goes down shotover doesnt get stuck and will instead return an error or retry.

conorbros:
Closes #769

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/connection.rs
+++ b/shotover-proxy/src/transforms/cassandra/connection.rs
@@ -40,7 +40,9 @@ impl CassandraConnection {
         mut tls: Option<TlsConnector>,
         pushed_messages_tx: Option<mpsc::UnboundedSender<Messages>>,
     ) -> Result<Self> {
-        let tcp_stream: TcpStream = TcpStream::connect(&host).await?;
+        let tcp_stream = timeout(Duration::from_secs(3), TcpStream::connect(&host))
+            .await?
+            .map_err(|e| anyhow::Error::new(e).context("Failed to connect to upstream"))?;
 
         let (out_tx, out_rx) = mpsc::unbounded_channel::<Request>();
         let (return_tx, return_rx) = mpsc::unbounded_channel::<Request>();
--- a/shotover-proxy/src/transforms/cassandra/connection.rs
+++ b/shotover-proxy/src/transforms/cassandra/connection.rs
@@ -34,7 +34,7 @@ pub struct CassandraConnection {
 }
 
 impl CassandraConnection {
-    pub async fn new<A: ToSocketAddrs>(
+    pub async fn new<A: ToSocketAddrs + std::fmt::Debug>(
         host: A,
         codec: CassandraCodec,
         mut tls: Option<TlsConnector>,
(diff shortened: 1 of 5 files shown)


shotover-proxy/src/transforms/cassandra/connection.rs
```diff
@@ -40,7 +40,9 @@ impl CassandraConnection {
         mut tls: Option<TlsConnector>,
         pushed_messages_tx: Option<mpsc::UnboundedSender<Messages>>,
     ) -> Result<Self> {
-        let tcp_stream: TcpStream = TcpStream::connect(&host).await?;
+        let tcp_stream = timeout(Duration::from_secs(3), TcpStream::connect(&host))
+            .await?
+            .map_err(|e| anyhow::Error::new(e).context("Failed to connect to upstream"))?;
```
rukai, line 45: I think we should include the `host` value in the error message, that should make it a lot easier to track down connection issues to specific nodes.

shotover-proxy/tests/redis_int_tests/basic_driver_tests.rs
```diff
@@ -1271,33 +1273,48 @@ pub async fn test_trigger_transform_failure_driver(connection: &mut Connection)
...
 /// CAREFUL: This lacks any kind of check that shotover is ready,
 /// so make sure shotover_manager.redis_connection is run on 6379 before calling this.
-pub fn test_trigger_transform_failure_raw() {
+pub async fn test_trigger_transform_failure_raw() {
     // Send invalid redis command
     // To correctly handle this shotover should close the connection
-    let mut connection = std::net::TcpStream::connect("127.0.0.1:6379").unwrap();
-    connection.write_all(b"*1\r\n$4\r\nping\r\n").unwrap();
-    connection
-        .set_read_timeout(Some(Duration::from_secs(10)))
```
rukai, line 1280: We've lost the read timeout which is a little concerning.
The test would hang if shotover never replied.
That sounds like quite an easy bug to accidentally introduce to shotover during development and a hung test is a pain to debug.

So probably the easiest fix is to just use https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.connect_timeout to perform a connection timeout on non async TcpStream.
Alternatively we could figure out the alternative for set_read_timeout on `tokio::net::TcpStream`

conorbros: I added the timeout below.

Happy to just switch back to the std TcpStream though.

rukai: oh duh, you are doing a tokio timeout on the connection.read, that works fine. :+1:

rukai (approved):
LGTM
I think we are fine using a hardcoded timeout for a connection because we always want at least some timeout here and its not like a query where we might have to wait for the server to do some complex operation.
3s is possibly a bit short but its what RedisSinkCluster is doing :shrug:

benbromhead (approved):

PR #775, opened 2022-08-31
Avoid creating AST for queries in CassandraSinkCluster

conorbros:
Creating the AST is pretty expensive. Since we are defining the queries ourselves and know they are not batch statements we can skip that check.

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
@@ -179,7 +179,15 @@ impl CassandraSinkCluster {
     }
 }
 
-fn create_query(messages: &Messages, query: String, version: Version) -> Result<Message> {
+//Creating the AST is pretty expensive. Since we are defining the queries
+// ourselves and know they are not batch statements we can skip that check.
+// This function checks the messages list to ensure that the stream id used is not
+// already being used
+fn create_query_with_generated_stream_id(
+    messages: &Messages,
+    query: String,
+    version: Version,
+) -> Result<Message> {
     let stream_id = get_unused_stream_id(messages)?;
     Ok(Message::from_frame(Frame::Cassandra(CassandraFrame {
         version,
@@ -191,6 +199,21 @@ fn create_query(messages: &Messages, query: String, version: Version) -> Result<
             params: Box::new(QueryParams::default()),
         },
     })))
+}
+
+//Creating the AST is pretty expensive. Since we are defining the queries
+// ourselves and know they are not batch statements we can skip that check.
+fn create_query(query: String, version: Version, stream_id: i16) -> Message {
+    Message::from_frame(Frame::Cassandra(CassandraFrame {
+        version,
+        stream_id,
+        tracing_id: None,
+        warnings: vec![],
+        operation: CassandraOperation::Query {
+            query: Box::new(CassandraStatement::Unknown(query)),
+            params: Box::new(QueryParams::default()),
+        },
+    }))
 }
 
 impl CassandraSinkCluster {
@@ -205,13 +228,21 @@ impl CassandraSinkCluster {
             let query = "SELECT rack, data_center, schema_version, tokens, release_version FROM system.peers";
             messages.insert(
                 table_to_rewrite.index + 1,
-                create_query(&messages, query.into(), table_to_rewrite.version)?,
+                create_query_with_generated_stream_id(
+                    &messages,
+                    query.into(),
+                    table_to_rewrite.version,
+                )?,
             );
             if let RewriteTableTy::Peers = table_to_rewrite.ty {
                 let query = "SELECT rack, data_center, schema_version, tokens, release_version FROM system.local";
                 messages.insert(
                     table_to_rewrite.index + 2,
-                    create_query(&messages, query.into(), table_to_rewrite.version)?,
+                    create_query_with_generated_stream_id(
+                        &messages,
+                        query.into(),
+                        table_to_rewrite.version,
+                    )?,
                 );
             }
         }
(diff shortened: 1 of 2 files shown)


rukai (approved):
Using `CassandraStatement::Unknown` in a non sink transform would cause problems because the query would be inaccessible to transforms further down the chain.

However using it in a sink transform is a clever performance hack, nice!

benbromhead (approved):
Nit: Might want to put a comment with "why" we do this, as it's not intuitive on the first read.

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -187,6 +195,21 @@ fn create_query(messages: &Messages, query: String, version: Version) -> Result<
...
+//Creating the AST is pretty expensive. Since we are defining the queries
+// ourselves and know they are not batch statements we can skip that check.
+fn create_query(query: String, version: Version, stream_id: i16) -> Message {
+    Message::from_frame(Frame::Cassandra(CassandraFrame {
+        version,
+        stream_id,
+        tracing_id: None,
+        warnings: vec![],
+        operation: CassandraOperation::Query {
+            query: Box::new(CassandraStatement::Unknown(query)),
```
rukai, line 207: To avoid someone accidentally copy/pasting this code into a non sink transform can we add a comment like this just before this line:

```rust
// Using CassandraStatement::Unknown in a non sink transform would cause problems because the query would be inaccessible to transforms further down the chain.
// However using it in a sink transform is a performance hack that is safe because no further transforms need access to the AST.
```

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -174,20 +173,47 @@ impl CassandraSinkCluster {
     }
 }
 
-fn create_query(messages: &Messages, query: &str, version: Version) -> Result<Message> {
+//Creating the AST is pretty expensive. Since we are defining the queries
+// ourselves and know they are not batch statements we can skip that check.
+// This function checks the messages list to ensure that the stream id used is not
+// already being used
+fn create_query_with_generated_stream_id(
```
rukai, line 180: How about naming `create_query_with_unique_stream_id` otherwise its not clear what kind of generated value is used.

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -174,20 +173,47 @@ impl CassandraSinkCluster {
...
+    messages: &Messages,
+    query: String,
+    version: Version,
+) -> Result<Message> {
     let stream_id = get_unused_stream_id(messages)?;
     Ok(Message::from_frame(Frame::Cassandra(CassandraFrame {
```
rukai, line 186: Can we call create_query here instead of duplicating the logic?

rukai: ahh, maybe we should just do something like `create_query(query, version, get_unused_stream_id(messages))` instead of having an entirely separate function for that.

conorbros: Good call.

conorbros:
Hmm how did that bug get there

rukai:
The error is that there are 256 elements in the tokens list instead of 128.
I'm not sure why this PR is consistently hitting it.

conorbros:
Ah true, thanks for the assist. I've pushed a temp fix to make the CI pass.

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -206,13 +208,21 @@ impl CassandraSinkCluster {
             let query = "SELECT rack, data_center, schema_version, tokens, release_version FROM system.peers";
             messages.insert(
                 table_to_rewrite.index + 1,
-                create_query(&messages, query, table_to_rewrite.version)?,
+                create_query(
+                    query.into(),
+                    table_to_rewrite.version,
```
rukai, line 213: found the bug, your missing `table_to_rewrite.index +1` here

rukai:
I think the issue is that queries arent being rewritten anymore?

rukai:
I think whats happening is:
we are creating a single connection to shotover and reusing it for the entire int test `connection1`.
That connection is returning bad results due to the control node not being part of the configured dc/rack. (the bug that https://github.com/shotover/shotover-proxy/pull/777 fixes)
So I suggest we review and merge https://github.com/shotover/shotover-proxy/pull/777 first and then rebase this and see if the issue still exists.

rukai:
I wrote out this test case to ensure that out both cases result in identical queries.
```rust
    #[test]
    pub fn test_problem_queries() {
        assert(
            "SELECT rack, data_center, schema_version, tokens, release_version FROM system.peers",
        );
        assert(
            "SELECT rack, data_center, schema_version, tokens, release_version FROM system.local",
        );
        assert("SELECT peer, rack, data_center, tokens FROM system.peers");
        assert("SELECT broadcast_address, rack, data_center, tokens FROM system.local");
    }

    fn assert(query: &str) {
        let unknown = CassandraStatement::Unknown(query.into()).to_string();
        let parsed = parse_statement_single(query).to_string();
        assert_eq!(unknown, parsed);
    }
```

The only remaining possibility is a race condition.

rukai:
I've figured out the issue.
`is_system_local_or_peers` relies on the message being correctly parsed.
Rather than continuing to hack around this, i'm leaning towards just abandoning this approach.
We've already hit one hard to track down case because of it, I'd rather not hit more.

If by profiling we find that parsing here is responsible for increasing our connection startup time then an alternate approach to this  I would like to take is directly calling Message::from_bytes on some bytes.
That we both completely skip the encoding stage + we get sane results when `.frame()` is called.

conorbros:
Thanks for the investigation. I'll close this and we can revisit if it turns out to be a perf issue. Perhaps caching the queries once we've created the AST and parsed them correctly.

PR #779, opened 2022-09-01
CassandraSinkCluster: return error when system.peers WHERE

rukai:
Returning an error for WHERE on system.peers makes complete sense to me as we currently make no attempt to handle the WHERE and I havent seen drivers use a WHERE before.
system.local is tricky though, I have seen drivers do redundant WHERE's that always evaluate to true for reasons unknown, so we need a more careful approach there.

Now that we have a third case of `RewriteTableTy` I opted for a little bit of duplication in the rewrite handling to make each specific case easier to reason about.

I imagine this error infrastructure will be useful for returning errors in more cases later on, e.g. returning error on tracing tables.

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
@@ -211,17 +211,29 @@ impl CassandraSinkCluster {
             .collect();
 
         for table_to_rewrite in tables_to_rewrite.iter().rev() {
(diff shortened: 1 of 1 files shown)


benbromhead:
What do the java, python, c driver all do with regards to queries with predicates on local and peers? We want these (and other) drivers to just work, so we might have to do the wrong thing to be right in this situation. E.g. just return the table,  ignore the where predicates and hope the driver gets what it needs?

rukai:
Superseded by https://github.com/shotover/shotover-proxy/pull/909

PR #780, opened 2022-09-01
test_cluster_single_rack: duplicate into v3 and v4 variants

rukai:
`cluster_sink_rack_single` is duplicated into `cluster_sink_rack_single_v3` and `cluster_sink_rack_single_v4`
To support this we split off the version specific parts from `cassandra_int_tests/cluster.rs` into `cassandra_int_tests/cluster_single_rack_v3.rs` and `cassandra_int_tests/cluster_single_rack_v4.rs`

Every other int test can be swapped over to v4 in a follow up PR


conorbros:
Happy to merge as is, but I think we should consider looking into seeing if there's a way to use the rstest library for this.

Alot of duplicated docker files

conorbros (approved):

benbromhead (approved):

PR #784, opened 2022-09-02
Fix cassandra protocol null handling

rukai:
Previously our implementation of the cassandra protocol was succesfully decoding null values but when it tried to reencode the null value it would output the bytes `00 00 00 04 FF FF FF FF` instead of just `FF FF FF FF`.
This resulted in e.g. null ip addresses showing up as 255.255.255.255 and integers as -1.

Unfortunately solving this requires completely rewriting `cassandra-protocol`s encoding implementation because of some poor fundamental decisions made there.
The problem is that the recursive generation (for e.g. List and Map types) relies on the `Bytes` various `impl From<T> for Bytes`.
However its impossible to represent a Null value on a `Bytes`, instead the `Bytes` needs to be wrapped in a CBytes in order to represent Null
This makes it impossible to handle both recursive types (list/map/set etc) and Null values using cassandra-protocol.

In order to get this working immediately for us I've just implemented the changes directly in shotover, its only a little more code shotover side but the custom implementation is:
* faster due to drastically reducing allocations
* easier to understand, all the logic is right in front of us rather than behind multiple layers of abstraction (IDE's have very poor support for navigating into `into()` implementations)
* we can look to upstream it in the future, but getting it working within shotover is a good first step, it at least illustrates what a good solution should look like when we go to upstream it.


conorbros:
Is it worth making a PR to cdrs-tokio or do you think just keeping it in Shotover is the best way to go? 

I'd be happy to merge as is, but wondering how about future plans.

rukai:
I absolutely want to investigate if we can upstream to cdrs-tokio as soon as we get the chance to, might be a while before we get that chance though.

conorbros (approved):

benbromhead (approved):
LGTM - I know it probably feels like a workaround at the moment given the underlying limitations, but happy to land this and work out pushing this upstream as mentioned in the comments above.

PR #785, opened 2022-09-02
Upgrade remaining cassandra int tests to cassandra 4.0.6

rukai:
Update all cassandra 3 integration tests to cassandra 4 (unless we are testing our handling of something cassandra 3 specific)

### DIFF
--- a/docs/src/examples/cassandra-cluster-shotover-sidecar.md
+++ b/docs/src/examples/cassandra-cluster-shotover-sidecar.md
@@ -114,7 +114,6 @@ and use the same query again. You should see the following results returned, not
 
 If everything has worked, you will be able to use Cassandra, with your connection going through Shotover!
 
-
 #### Adding Rate Limiting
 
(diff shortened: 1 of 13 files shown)


conorbros:
Why 4.0.6?

rukai:
oh huh, your right, 4.0.6 is old...
does bitnami not have images for recent releases? https://hub.docker.com/r/bitnami/cassandra/tags

conorbros:
4.0.6 is the latest. Just wondering why we should specify that?

rukai:
Pinning to a specific version sounds like the best idea to me.
Everything else in this project has versions pinned e.g. we have the Cargo.lock checked in protecting us in the case that some crate fails to follow semver or introduces a new bug or fixes a bug we relied on.
I dont see why that should be different for our int test db images.

If we find issues with 4.0.6 or we just want to make sure shotover continues to function on the latest release, we can always upgrade to a later version manually.

conorbros (approved):

benbromhead (approved):

PR #786, opened 2022-09-05
Dynamic port 

### LINKED ISSUE #773: CassandraSinkCluster dynamic port handling
Currently we hardcode the port 9042 everywhere, but we should handle the port dynamically from the config and system.peers_v2 results.

conorbros:
closes #773


conorbros:
Ready for some reviews, still need to add some tests and documentation about using non-standard ports.

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -103,68 +70,281 @@ async fn topology_task_process(
...
+                            let port = if let Some(broadcast_port_index) = broadcast_port_index {
+                                if let Some(MessageValue::Integer(value, _)) =
+                                    row.get(broadcast_port_index)
+                                {
+                                    *value
+                                } else {
+                                    return Err(anyhow!("port not an integer"));
+                                }
+                            } else {
+                                9042
```
rukai, line 188: can we add a comment just above the 9042, like: "cassandra 3.x does not have the broadcast_port field so we just assume the default port.

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -103,68 +70,281 @@ async fn topology_task_process(
...
+                            } else {
+                                false
+                            }
+                        })
+                        .map(|row| {
+                            let tokens =
+                                if let Some(MessageValue::List(list)) = row.get(tokens_index) {
+                                    list.iter()
+                                        .map::<Result<String>, _>(|x| match x {
+                                            MessageValue::Varchar(a) => Ok(a.clone()),
```
rukai, line 156: You can avoid cloning all of these strings like this:
image

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -103,68 +70,281 @@ async fn topology_task_process(
...
+                                value
+                            } else {
+                                return Err(anyhow!("port is not an integer"));
+                            }
+                        } else {
+                            9042
+                        };
+
+                        Ok(CassandraNode {
+                            address: SocketAddr::new(ip, port.try_into()?),
```
rukai, line 331: useless `.try_into()?`?

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -103,68 +70,281 @@ async fn topology_task_process(
...
+                        };
+
+                        let port = if let Some(message_value) = row.pop() {
+                            if let MessageValue::Integer(value, _) = message_value {
+                                value
+                            } else {
+                                return Err(anyhow!("port is not an integer"));
+                            }
+                        } else {
+                            9042
```
rukai, line 327: same here but comment about system.peers_v2 vs system.peers

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -103,68 +70,281 @@ async fn topology_task_process(
...
+                })),
+                tx,
+            )?;
+            response = rx.await?.response?;
+        }
+
+        into_nodes(response, data_center)
+    }
+
+    fn is_unconfigured_peers_v2_table_response(message: &mut Message) -> bool {
```
rukai, line 263: How about: `is_peers_v2_does_not_exist_error`.
That way you dont need to know that "unconfigured" is what cassandra says when a table does not exist.

rukai (commented):
General approach looks good, heres some minor feedback

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
conorbros: image

Am I missing something?

rukai: No, I was, my bad.

:+1:

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
rukai: comment doesnt quite make sense directly copied.
How about something like: "this method supports both system.peers and system.peers_v2, system.peers does not have a field for the port so we fallback to the default port."

conorbros: Ah yep, MB.

benbromhead (approved):

shotover-proxy/tests/helpers/config.yaml
```diff
@@ -1,3 +1,3 @@
 ---
-main_log_level: "warn,shotover_proxy=warn"
+main_log_level: "info,shotover_proxy=info"
```
rukai, line 2: ^

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -638,6 +640,8 @@ impl CassandraSinkCluster {
                         listen_address_alias = alias;
                     } else if column.name == Identifier::Unquoted("host_id".to_string()) {
                         host_id_alias = alias;
+                    } else if column.name == Identifier::Unquoted("rpc_port".to_string()) {
```
rukai, line 643: I dont believe such a field exists:
image

conorbros: Oops got mixed up with `system.local` clearly.

conorbros: Oh wait I was correct, the function that change is in is `rewrite_table_local`.

rukai: Oh true, I got the tables mixed up.
rpc_address/rpc_port is actually the fields describing where the cql protocol connection goes. :+1:

rukai (approved):
very nice!

PR #788, opened 2022-09-06
Remove unused MessageValue variants

rukai:
I have vague guesses at what these were for but they are unused so lets get rid of them.


conorbros (approved):

benbromhead (approved):
We can always add them back in, but the idea was to represent other data types (esp documents... say for mongodb).

PR #789, opened 2022-09-06
Add cassandra codec benches

rukai:
The message used in the `encode_cassandra_system.local_result` benchmark is taken directly from a real system.local response.
After implementing this I realized it was way too much work manually converting the message to the rusty frame types.
So I think we would probably be better off with just taking a raw binary message, and then programmatically decoding it into a frame before the bench runs.

However I've got this now and its fine was just more effort than was needed, so if others are happy to, we can land this and then rewrite to the raw binary approach if we want to add more cases later on.


conorbros (approved):
Do we want benches for decoding aswell?

If we are going down this path we might want benches for the `frame` methods too. 

Anyway that's just thinking aloud for future plans, gets go ahead with this.

rukai:
> Do we want benches for decoding aswell?

Yeah

> If we are going down this path we might want benches for the frame methods too.

Good point.

I think the complete state of these benchmarks would look like:
for each message we are testing:

* bench encode from raw bytes
* bench encode from frame
* bench decode without frame call
* bench decode with frame() call (or maybe just bench the frame() call itself)

Maybe we can automate all of these starting from just a single raw bytes message.

benbromhead (approved):

PR #795, opened 2022-09-08
Add DebugLogToFile transform

rukai:
A usecase I have in mind is extracting messages for the purposes of benchmarking parsing/encoding.
I think it could also be useful for debugging.

Each incoming connection to shotover results in a new numbered directory created in `message-log`.
Within those directories is a `request` and `response` directory in which messages are logged to disk under a `messageN.bin` name.
The easiest way to see what the log looks like is to comment out `std::fs::remove_dir_all("message-log").unwrap();` in `log_to_file` integration test and run it.

The comment:
```rust
    // TODO: Considering we already have the expected message type here maybe we should perform any required conversions and return a Result<Bytes> here.
    // I've left it as is to keep the PR simpler and there could be a need for codecs to control this process that I havent investigated.
```
has been deleted because the design it is proposing is no good. We need to return the `Encodable` so that the codec can encode the frame directly into the destination buffer.

The `MessageType` check had to be pulled out of `into_encodable` because `LogToFile` needs to operate on all message types unlike the codecs which always need a specific type of message.

### DIFF
--- a/shotover-proxy/tests/test-configs/log-to-file/docker-compose.yaml
+++ b/shotover-proxy/tests/test-configs/log-to-file/docker-compose.yaml
@@ -1,0 +1,6 @@
+version: "3.3"
+services:
+  redis-one:
+    image: library/redis:5.0.9
+    ports:
+      - "1111:6379"
--- a/shotover-proxy/tests/test-configs/log-to-file/topology.yaml
+++ b/shotover-proxy/tests/test-configs/log-to-file/topology.yaml
@@ -1,0 +1,13 @@
+---
+sources:
+  redis_prod:
+    Redis:
+      listen_addr: "127.0.0.1:6379"
+chain_config:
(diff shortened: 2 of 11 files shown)


shotover-proxy/src/transforms/debug/log_to_file.rs
```diff
@@ -0,0 +1,97 @@
...
+        }
+        Ok(response)
+    }
+}
+
+fn log_message(message: &Message, path: &Path) -> Result<()> {
+    info!("Logged message to {:?}", path);
+    match message.clone().into_encodable() {
+        Encodable::Bytes(bytes) => {
+            std::fs::write(path, bytes).context("failed to write message to disk")?;
```
conorbros, line 90: Should use tokio, this is blocking.

benbromhead (approved):

PR #796, opened 2022-09-08
Split cassandra_int_tests into its own CI job

rukai:
Caching isnt used on this PR due to changes to caching keys.
So despite the runtime looking about the same as before we should get at least a 10 minute improvement once checked in.
So this looks like a win to me.

Unfortunately this approach is a bit hacky and we'll have to manually split into more jobs again in the future, but I dont think we have the time to engineer a proper solution.


conorbros (approved):

rukai:
This is a potentially better (future?) solution https://nexte.st/book/partitioning.html

XA21X (approved):

PR #797, opened 2022-09-08
Cut CI bench compile time by a few minutes

rukai:
Every bench file in shotover-proxy/benches is individually compiled into its own binary.
This is really expensive because the compiler needs to reapply all the crazy optimizations for each binary.

Locally on my machine it takes ~1m 40s for each of these to compile.
Because I have plenty of cores this is all done in parallel and the fact that each binary takes so long to compile causes no slow down for me.
With this change it still takes the same ~1m 40s to compile all of the binaries.

However in CI we only have 2 cores, so having 4 benchmarks to compile doubles the time to compile from 2m to 4m.
So I expect this PR to cut off about ~2 minutes from our benchmark build time.

This also fixes the codec_benches not running in CI due to not being manually included in bench_against_master.sh.
Unfortunately this means we will need to force merge this PR as it relies on the main branch containing the `benches` bench.

### DIFF
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -94,17 +94,5 @@ cdrs-tokio = { git = "https://github.com/krojew/cdrs-tokio" }
 scylla = { version = "0.5.0", features = ["ssl"] }
 
 [[bench]]
-name = "redis_benches"
-harness = false
-
-[[bench]]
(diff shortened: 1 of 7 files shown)


conorbros (approved):
Any increase to the CI is very welcome at this point 😂

benbromhead (approved):

PR #798, opened 2022-09-09
Cassandra_int_test refactor connection creation

rukai:
prereq to https://github.com/shotover/shotover-proxy/pull/741

When we originally wrote the ShotoverManager abstraction the idea was to have it create connections so that we could one day make it guarantee that you were connecting directly to shotover. (by reading the topology.yaml or something)
But we never implemented that and then soon realized that we often wanted to create connections directly against the database bypassing shotover, and such an abstraction would have prevented that.
So considering we will never get that benefit this PR removes the ShotoverManager bit in the middle and just directly creates a CassandraConnection via `new` or `new_tls`.

This also fixes an intermittent failure due to some benchmark init logic manually creating their own connections without CassandraConnection and therefore missing out on CassandraConnection's retry logic.

### DIFF
--- a/shotover-proxy/benches/benches/cassandra.rs
+++ b/shotover-proxy/benches/benches/cassandra.rs
@@ -1,6 +1,6 @@
 use crate::helpers::cassandra::CassandraConnection;
 use crate::helpers::ShotoverManager;
-use cassandra_cpp::{stmt, Cluster, Session, Statement};
+use cassandra_cpp::{stmt, Session, Statement};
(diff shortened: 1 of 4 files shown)


conorbros (approved):
LGTM, especially makes sense given the changes in #741

https://github.com/shotover/shotover-proxy/blob/3513ab1c0d30e9c03e5bb711225f7384f8cd1758/shotover-proxy/tests/helpers/mod.rs

XA21X (approved):

Issue #776, opened 2022-08-31
CassandraSinkCluster misc routing improvements


rukai:
* system_schema quries should go through the control connection
* tracing quries should be rejected

yuzhouchen-instaclustr:
I'd like to point out a potential issue with system_schema queries:

If queries to the system_schema keyspace are all routed to a random node or a dedicated node in the rack, then the following might happen: if there exist schema disagreement in a rack then the schema version returned from shotover will be a random UUID, but if the node shotover goes to to get response for system_schema is not in schema disagreement there will be an inconsistency there, i.e. even if schema_version is outdated, queries to system_schema still return an up-to-date view of the actual schema.

Not sure if this is actually an issue

PR #799, opened 2022-09-09
CassandraSinkCluster fix routing for system keyspaces

rukai:
part of https://github.com/shotover/shotover-proxy/issues/776
Fixes an issue discovered during testing.
Queries against system keyspace tables need to be routed to the same node to remain consistent.

I took the list of system keyspaces from https://docs.datastax.com/en/dse/5.1/cql/cql/cql_using/useQuerySystem.html

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
@@ -176,7 +176,7 @@ impl CassandraSinkCluster {
             system_keyspaces: [
                 Identifier::parse("system"),
                 Identifier::parse("system_schema"),
-                Identifier::parse("repair_history"),
+                Identifier::parse("system_distributed"),
             ],
             local_shotover_node,
             local_nodes: vec![],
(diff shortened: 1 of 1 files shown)


conorbros (approved):

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -171,6 +173,11 @@ impl CassandraSinkCluster {
             local_table: FQName::new("system", "local"),
             peers_table: FQName::new("system", "peers"),
             peers_v2_table: FQName::new("system", "peers_v2"),
+            system_keyspaces: [
+                Identifier::parse("system"),
+                Identifier::parse("system_schema"),
+                Identifier::parse("repair_history"),
```
XA21X, line 179: Did you mean to specify the `system_distributed` keyspace? `repair_history` is the table name.

rukai: good catch!

XA21X (approved):

PR #801, opened 2022-09-13
store tokens as murmur3 instead of string

conorbros:
Part of token aware routing


rukai (approved):
LGTM
Mumur3Token is internally an int which will make our clones cheaper as well :tada:

XA21X (approved):

PR #802, opened 2022-09-13
Parse Cassandra execute statements

conorbros:
Unfortunately this had caused a clippy lint, not sure what the best course of action is:

- Leave the lint turned off
- Put the big variants in a box (alot of code changes because we will have to dereference all the Cassandra frames after parsing instead of just pattern matching)


rukai:
I think the answer is `Execute(Box<BodyReqExecuteOwned>)` right?
Thats what we already do for CassandraOperation::Query and CassandraOperation::Result.

conorbros:
Oh true! If only clippy was smart enough to tell me that. Thanks.

XA21X (approved):

rukai (approved):
LGTM

PR #803, opened 2022-09-14
Use statement integration tests

rukai:
The integration tests in this PR demonstrate some unfortunate realities about use statements: both batches and prepared statements take the `use`d keyspace into account.

This will affect how we go about handling keyspaces when identifying a keyspace/table pair.


rukai:
`cargo test --all-features test_cluster_multi_rack` fails locally and appears to reveal a bug in shotover

rukai:
To avoid blocking the PR I commented out the bit that fails, we can enable it once we fix the underlying shotover issue, which will require in depth investigation.

conorbros (approved):
LGTM. Did you open an issue for the problem?

rukai:
done https://github.com/shotover/shotover-proxy/issues/811

benbromhead (approved):

PR #804, opened 2022-09-14
cassandra_int_tests add case that forces cassandra message encoding

rukai:
I wanted a way to better test the encoding code path of changes like: https://github.com/shotover/shotover-proxy/pull/802
So I added this integration test which will exercise the encoding code path against all of our cassandra test cases.

### DIFF
--- a/shotover-proxy/src/transforms/debug/force_parse.rs
+++ b/shotover-proxy/src/transforms/debug/force_parse.rs
@@ -10,17 +10,40 @@ use anyhow::Result;
 use async_trait::async_trait;
 use serde::Deserialize;
 
+/// Messages that pass through this transform will be parsed.
+/// Must be individually enabled at the request or response level.
 #[derive(Deserialize, Debug, Clone)]
 pub struct DebugForceParseConfig {
-    parse_requests: Option<bool>,
-    parse_responses: Option<bool>,
+    parse_requests: bool,
+    parse_responses: bool,
 }
 
 impl DebugForceParseConfig {
     pub async fn get_transform(&self) -> Result<Transforms> {
         Ok(Transforms::DebugForceParse(DebugForceParse {
-            parse_requests: self.parse_requests.unwrap_or(true),
-            parse_responses: self.parse_responses.unwrap_or(true),
+            parse_requests: self.parse_requests,
+            parse_responses: self.parse_responses,
+            encode_requests: false,
+            encode_responses: false,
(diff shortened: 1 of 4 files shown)


shotover-proxy/src/transforms/debug/force_parse.rs
```diff
@@ -10,17 +10,40 @@ use anyhow::Result;
...
+    parse_requests: bool,
+    parse_responses: bool,
 }
 
 impl DebugForceParseConfig {
     pub async fn get_transform(&self) -> Result<Transforms> {
         Ok(Transforms::DebugForceParse(DebugForceParse {
-            parse_requests: self.parse_requests.unwrap_or(true),
-            parse_responses: self.parse_responses.unwrap_or(true),
+            parse_requests: self.parse_requests,
```
conorbros, line 24: Is there a reason for the configuration options?

I'm thinking to remove all this config, have a transform called DebugForceCodec and just test it all. 

I can't think of a situation where we would want to this separately so might as well keep it simple.

rukai: Only to match DebugForceParse, which does make use of the config for benchmarking purposes.
I have considered removing the config from DebugForceParse though and I'm happy to do that if you think it makes sense.

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -62,6 +62,30 @@ async fn test_passthrough(#[case] driver: CassandraDriver) {
     batch_statements::test(&connection).await;
 }
 
+#[cfg(feature = "cassandra-cpp-driver-tests")]
+#[rstest]
+#[cfg_attr(feature = "cassandra-cpp-driver-tests", case(Datastax))]
+#[tokio::test(flavor = "multi_thread")]
+#[serial]
+async fn test_passthrough_encode(#[case] driver: CassandraDriver) {
```
conorbros, line 70: Can't we just put another transform chain on the passthrough one and run the tests against that. Save starting another docker compose.

rukai: Currently the test suite does not clean up after itself so that wouldnt work.

conorbros (approved):

benbromhead (approved):

rukai:
Oh, looks like this caught an issue in between being written and now.

PR #806, opened 2022-09-14
CassandraSinkCluster keyspace based routing - handle use statements

rukai:
This PR handles use statements in every case except one:
* operates on both batch operations and query operations. 
* messages are modified such that if a keyspace is not specified the `use` keyspace is used, this ensures a single source of truth for transforms.
* messages are modified at the codec level ensuring all layers of shotover observe the same keyspace.

The one case we arent handling is prepared statements.
use statements also affect prepared statements, but currently shotover does not support parsing prepared statements so I have just left that functionality as a TODO.

There is a performance concern: calling metadata is currently more expensive than it needs to be due to internally copying the entire message body and then dropping it.
But that can be addressed later on in a follow up PR.

Making this change caused the `.as_raw_bytes()` hack to start blowing up due to the message already being parsed, so I also changed all instances of it to just use `.metadata()` instead.

### DIFF
--- a/shotover-proxy/src/codec/cassandra.rs
+++ b/shotover-proxy/src/codec/cassandra.rs
@@ -1,12 +1,16 @@
-use crate::frame::cassandra::CassandraOperation;
+use crate::frame::cassandra::{CassandraMetadata, CassandraOperation};
 use crate::frame::{CassandraFrame, Frame, MessageType};
-use crate::message::{Encodable, Message, Messages};
+use crate::message::{Encodable, Message, Messages, Metadata};
 use crate::server::CodecReadError;
 use anyhow::{anyhow, Result};
 use bytes::{Buf, BufMut, BytesMut};
 use cassandra_protocol::compression::Compression;
 use cassandra_protocol::frame::message_error::{AdditionalErrorInfo, ErrorBody};
-use cassandra_protocol::frame::{CheckEnvelopeSizeError, Envelope as RawCassandraFrame, Version};
+use cassandra_protocol::frame::{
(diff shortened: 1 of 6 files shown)


shotover-proxy/src/codec/cassandra.rs
```diff
@@ -89,6 +110,70 @@ impl Decoder for CassandraCodec {
...
+            if let CassandraStatement::Use(keyspace) = query.as_ref() {
+                return Some(keyspace.clone());
+            }
+        }
+    }
+    None
+}
+
+fn set_default_keyspace(message: &mut Message, keyspace: &Identifier) {
+    // TODO: rewrite Operation::Prepared in the same way
```
conorbros, line 125: Not sure what this means. What needs to be done for `Operation::Prepared`?

rukai: The cassandra server will apply the current `use`d keyspace to Operation::Prepared messages that it receives.
To make that visible to transforms we should apply the same modifications that we apply to queries and batches to prepared statements as well

conorbros: Yes I remember reading something interesting about prepared queries and keyspaces but I can't find it now. IIRC:

- it applies the used keyspace to `PREPARE` statements
- The keyspace set when the `PREPARE` statement was sent takes precedence over the current used keyspace (if it has changed)

rukai: yep, that matches my understanding too

benbromhead (approved):

conorbros (approved):

PR #807, opened 2022-09-15
CassandraSinkCluster: filter out Topology and Status events

rukai:
Implementation is pretty straightforward.
Just filter out the Topology and Status events to hide underlying cluster topology details from the clients.
Schema changes can still pass through as that is not part of the topology.

### DIFF
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -94,6 +94,7 @@ metrics-util = "0.14.0"
 cdrs-tokio = { git = "https://github.com/krojew/cdrs-tokio" }
 scylla = { version = "0.5.0", features = ["ssl"] }
 rstest = "0.15.0"
+docker-api = "0.10.1"
 
 [[bench]]
 name = "benches"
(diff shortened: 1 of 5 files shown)


conorbros:
Happy to merge without given it's simple, but this could be tested similar to the `test_events_keyspace` test.

rukai:
oh right, I'll look into that

rukai:
I had a go and couldnt get it working, I suspect its due to a bug in cdrs-tokio, lets move on: https://github.com/shotover/shotover-proxy/pull/808

rukai:
Ended up figuring it out after all
I've added the tests to this PR.

I have observed that sometimes the test doesnt fail without the changes in this PR.
But at least it will still (intermittently) catch the failure.

conorbros (approved):

shotover-proxy/Cargo.toml
```diff
@@ -94,6 +94,7 @@ metrics-util = "0.14.0"
 cdrs-tokio = { git = "https://github.com/krojew/cdrs-tokio" }
 scylla = { version = "0.5.0", features = ["ssl"] }
 rstest = "0.15.0"
+docker-api = "0.10.1"
```
benbromhead, line 97: are we ment to include this dep? I can see it in any of the imports

benbromhead (approved):

PR #812, opened 2022-09-16
CassandraSinkCluster: topology task now processes events

rukai:
an is_up field is introduced so that nodes that are up/down can be noted as such without completely deleting the node, this will be useful for recovering from scenarios where all nodes go down. (even though there is no such logic for that yet)

Control node selection and regular message routing is changed to use is_up

## Work left for future PRs

Currently the topology changes wont take affect until a new connection is made, a future PR will introduce logic to update existing connections. 

Another case that is currently not handled yet is when the topology task's control connection goes down.
Currently that is undetectable because we rely on the control connections events.
I believe the solution to that will be transforms reporting timeouts to the topology task (done in a follow up PR)
Another solution could be having a backup control connection.

Because this PR is missing those features I cant integration test it yet without intermittent failures.
So lets try and land this now, and add integration tests when we complete the functionality.

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
@@ -232,8 +232,11 @@ impl CassandraSinkCluster {
                 let nodes_shared = self.topology_task_nodes.read().await;
                 self.local_nodes = nodes_shared.clone();
             }
+            for node in &self.local_nodes {
+                tracing::error!("local nodes: {} {}", node.address, node.is_up);
+            }
 
-            let random_point = if self.local_nodes.is_empty() {
+            let random_point = if self.local_nodes.iter().all(|x| !x.is_up) {
                 tokio::net::lookup_host(self.contact_points.choose(&mut self.rng).unwrap())
                     .await?
                     .next()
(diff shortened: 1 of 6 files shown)


rukai:
I'm going to go ahead and mark this as ready for review, ideally I would also include an integration test.
But that would require blocking on https://github.com/shotover/shotover-proxy/pull/807 as we need to build upon the integration test included in that PR.
Lets add integration tests when we complete this functionality by extending it to update preexisting connections.

rukai:
The prereq is in, so I added the integration test, unfortunately its failing, so I'll make this a draft and investigate.

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -232,8 +232,11 @@ impl CassandraSinkCluster {
                 let nodes_shared = self.topology_task_nodes.read().await;
                 self.local_nodes = nodes_shared.clone();
             }
+            for node in &self.local_nodes {
+                tracing::error!("local nodes: {} {}", node.address, node.is_up);
```
conorbros, line 236: Wouldn't we only want to print this if it's down?

conorbros (approved):

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
rukai: Ooops those were from debugging, I'll remove

PR #814, opened 2022-09-19
Bench CI speedups

rukai:
# CI Change
Fixes an issue where bench_run does not use a cache when a PR is first submitted.

PRs do not share caches, instead each PR initially pulls from the cache of the main branch for the first run.
The bench_run workflow does not run on main, so to make use of a cache before this workflow has completed once on a PR, we need to manually recreate the key used by ubuntu-18.04 release builds.

This has the risk of the cache being overwritten by the run_bench workflow, but worst case the cache gets some extra bench logs included, Swatinem/rust-cache will not cache final binary output so we dont have to worry about losing that.
But I added a TODO to make use of a feature that will completely eliminate any risk here.

# Criterion change
Criterion's `html_reports` feature is generating ~15mb of html files across all of our benchmarks.
Neither me or Conor have ever used it so I suggest to remove it to save the space and cpu cycles, if anyone ever wants it they can go and manually enable the feature.

### DIFF
--- a/.github/workflows/bench_run.yml
+++ b/.github/workflows/bench_run.yml
@@ -21,8 +21,14 @@ jobs:
     - uses: actions/checkout@v3
     - uses: Swatinem/rust-cache@v2
       with:
-        # Unique key is used to avoid collisions with the Testing workflow
-        key: 'ubuntu-18.04-benchmark'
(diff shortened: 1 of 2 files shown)


conorbros (approved):

benbromhead (approved):
Yeah this can totally be opt in

PR #815, opened 2022-09-19
Update GA actions

rukai:
Motivation: I wanted to make use of a feature in rust-cache@v2 but figured I may as well keep other actions up to date, so we can easily try more recent features without having to update.

Updated checkout and rust-cache to latest versions.
Removed a stray `actions-rs/toolchain@v1`, not needed anymore as github actions images now have rustup installed by default.

The cache key format seems to have changed from v1 to v2, so I'll rerun CI to confirm its caching properly.

### DIFF
--- a/.github/workflows/bench_run.yml
+++ b/.github/workflows/bench_run.yml
@@ -18,8 +18,8 @@ jobs:
     runs-on: ubuntu-18.04
     name: Benchmarks
     steps:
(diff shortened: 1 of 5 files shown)


rukai:
Can confirm caching still works, lets land this.

conorbros (approved):

XA21X (approved):

PR #818, opened 2022-09-20
add host_id field to CassandraNode

conorbros:
I need the nodes to have a unique id for token aware routing and using the socket address is probably going to end up badly so I've added this field.


rukai (approved):
Sounds reasonable but would you mind going giving a description of how shotover will use the host_id?
Cassandra node up/down events use the SocketAddr as an id, so it is sufficient in some areas of cassandra.

conorbros:
Most drivers I've seen use the UUID for keys in a map of Nodes, haven't seen any using the socket address for that purpose.

benbromhead (approved):

PR #819, opened 2022-09-21
Update deps

rukai:
The cdrs-tokio changes are due to https://github.com/krojew/cdrs-tokio/pull/110

### DIFF
--- a/shotover-proxy/src/codec/cassandra.rs
+++ b/shotover-proxy/src/codec/cassandra.rs
@@ -5,7 +5,7 @@ use crate::server::CodecReadError;
 use anyhow::{anyhow, Result};
 use bytes::{Buf, BufMut, BytesMut};
 use cassandra_protocol::compression::Compression;
-use cassandra_protocol::frame::message_error::{AdditionalErrorInfo, ErrorBody};
+use cassandra_protocol::frame::message_error::{ErrorBody, ErrorType};
 use cassandra_protocol::frame::{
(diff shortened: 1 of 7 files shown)


shotover-proxy/src/codec/cassandra.rs
```diff
@@ -187,9 +187,8 @@ fn reject_protocol_version(version: u8) -> CodecReadError {
             version: Version::V4,
             stream_id: 0,
             operation: CassandraOperation::Error(ErrorBody {
-                error_code: 0xA, // https://github.com/apache/cassandra/blob/adf2f4c83a2766ef8ebd20b35b49df50957bdf5e/doc/native_protocol_v4.spec#L1053
                 message: "Invalid or unsupported protocol version".into(),
-                additional_info: AdditionalErrorInfo::Server,
+                ty: ErrorType::Protocol,
```
XA21X, line 191: Slightly surprised about the `ty` naming. I know `type` is reserved, but `error_type`? Too late lol.

XA21X (approved):

shotover-proxy/src/codec/cassandra.rs
rukai: `ty` is pretty standard due to `type` being reserved, but you make a good point, the redundant `error_` probably would have been worth it to avoid abbreviating `type`

conorbros (approved):

PR #821, opened 2022-09-21
Faster DockerCompose drop

rukai:
On my local machine this cuts 10s off the time to complete the test.
There is no need to be polite to these docker containers as we never intend to use them again and we dont care about cleanly closing connections to clients once the test is over.
We still need the call to `docker-compose down` as that will cleanup any left over networks for us.

I'm not sure what `docker-compose rm` was for, from looking at the documentation it seems to do the same thing as our call to `docker compose down`.
I traced it back to: https://github.com/shotover/shotover-proxy/commit/efb50e696313478844d8a1a68f3b73dd2e305f55#diff-6d1e3ec7ab6aaf7a94f9c1e9a423db6de6e4a1bec380dae17cc6e635f1db8854
Probably just a mistake.

### DIFF
--- a/test-helpers/src/docker_compose.rs
+++ b/test-helpers/src/docker_compose.rs
@@ -186,8 +186,8 @@ impl DockerCompose {
     fn clean_up(file_path: &str) -> Result<()> {
         trace!("bringing down docker compose {}", file_path);
 
+        run_command("docker-compose", &["-f", file_path, "kill"])?;
         run_command("docker-compose", &["-f", file_path, "down", "-v"])?;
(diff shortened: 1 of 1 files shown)


conorbros (approved):

XA21X:
The commit you traced it back to, `docker-compose down` was invoked without `-v` because that option was only supported since early 2021: https://github.com/docker/compose/commit/f07a4a72b92987db02fc47cc0b0815acc603afa6. The `docker-compose rm -v` _was_ needed to remove the named volumes.

XA21X (approved):

PR #824, opened 2022-09-23
Token aware routing

conorbros:
# Token Aware Routing

### `NodePool`

- Added this struct to contain the logic for holding our node list and selecting which nodes to use for queries
- We will need to add more fields for metadata (keyspace info) and I'll put those in there

### `TokenMap`

- Map of tokens->host_id

### `routing_key`

- Functions for calculating the routing key and generating Murmur3 token
- Tests for calculating the routing key and asserting that `NodePool` and `TokenMap` return the correct nodes

### Client sends a `PREPARE` statement to Shotover, we forward this to all nodes.

Drivers typically send a `PREPARE` statement to one node. If it tries to use an `EXECUTE` statement on a node that doesn't have the query prepared it will re-prepare on that node before sending the `EXECUTE` statement again. The problem with this on Shotover is that we don't know which node returned the unprepared error, so we don't know which one we should be sending the re-prepared statement to. By preparing on all nodes, we can avoid this. If a node is added and it returns an unprepared error, the driver will re-prepare it again and Shotover will forward that to all nodes.

### Cache the metadata from the `PREPARED` result to a `PREPARE` statement

We need to store this information to calculate the routing key to subsequent `EXECUTE` statements. Uses the prepared metadata id field as a key.

### Perform token aware routing on `EXECUTE` statements

Using the prepared metadata, calculate the routing key and then the Murmur3 token. If a replica doesn't exist just perform random routing. 

## Followup 
- Address the TODOs
- Some easy performance wins can be made, mostly around avoiding calling the `frame` method on every message and instead checking the opcode bytes directly first.
- We should be performing round robin load balancing instead of random.
- Create new connections concurrently for prepared statements
- Share prepared metadata across connections


shotover-proxy/src/transforms/cassandra/sink_cluster/routing/mod.rs
```diff
@@ -0,0 +1,2663 @@
...
+            .collect();
+
+        println!("{:?}", nodes);
+
+        let expected_node = "172.16.1.5:9044".parse().unwrap();
+        assert_eq!(nodes.len(), 1);
+        assert_eq!(nodes[0], expected_node);
+    }
+
+    fn prepare_nodes() -> Vec<CassandraNode> {
```
rukai, line 313: For the purposes of reviewing it would help if this was in a separate file so I can minimize the file, otherwise its hard to jump around.

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
```diff
@@ -0,0 +1,95 @@
...
+use rand::prelude::*;
+use std::collections::HashMap;
+
+#[derive(Debug)]
+pub struct NodePool {
+    prepared_metadata: HashMap<Vec<u8>, PreparedMetadata>,
+    token_map: TokenMap,
+    pub nodes: Vec<CassandraNode>,
+
+    rng: SmallRng,
```
rukai, line 17: we have two sources of rng now.
we should either pass the rng into NodePool methods when we need it here or use node_pool.rng when we need it in the transform.

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
```diff
@@ -0,0 +1,95 @@
...
+        self.nodes.is_empty()
+    }
+
+    pub fn set_nodes(&mut self, nodes: Vec<CassandraNode>) {
+        self.nodes = nodes;
+        self.token_map = TokenMap::new(self.nodes.as_slice());
+    }
+
+    pub fn add_prepared_result(&mut self, id: Vec<u8>, metadata: PreparedMetadata) {
+        tracing::warn!("Add prepared result metadata {id:?}");
```
rukai, line 40: I imagine these got left in by accident?

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
```diff
@@ -0,0 +1,95 @@
...
+        self.prepared_metadata.insert(id, metadata);
+    }
+
+    // TODO handle the unwrap
+    fn get_prepared_metadata(&self, id: &Vec<u8>) -> &PreparedMetadata {
```
rukai, line 45: This is only used internally, it would be easier to just directly call self.prepared_metadata.get(id).unwrap()

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub fn new(nodes: Vec<CassandraNode>) -> Self {
+        Self {
+            token_map: TokenMap::new(nodes.as_slice()),
+            nodes,
+            prepared_metadata: HashMap::new(),
+            rng: SmallRng::from_rng(rand::thread_rng()).unwrap(),
+        }
+    }
+
+    pub fn is_empty(&self) -> bool {
```
rukai, line 30: nodes is public, why not just directly call `self.nodes.is_empty()` ?

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -753,6 +810,46 @@ enum RewriteTableTy {
     Peers,
 }
 
+fn is_prepared_message(message: &mut Message) -> Option<(Vec<u8>, PreparedMetadata)> {
```
rukai, line 813: how about `is_result_prepared_message`.
`is_prepare_message` and `is_prepared_message` are confusingly similar.

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -313,6 +362,13 @@ impl CassandraSinkCluster {
             super::connection::receive(self.read_timeout, &self.failed_requests, responses_future)
                 .await?;
 
+        let _prepare_responses = super::connection::receive(
```
rukai, line 365: We should return an error if cassandra returns something unexpected here.

conorbros: Yeah this part is a bit tricky, we return the response to PREPARE from the handshake connection. But if an error occurs on one of the nodes, should we use that response or the handshake one?

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -753,6 +810,46 @@ enum RewriteTableTy {
...
 }
 
+fn is_prepared_message(message: &mut Message) -> Option<(Vec<u8>, PreparedMetadata)> {
+    if let Some(Frame::Cassandra(CassandraFrame {
+        operation: CassandraOperation::Result(CassandraResult::Prepared(prepared)),
+        ..
+    })) = message.frame()
+    {
+        return Some((
+            prepared.id.clone().into_bytes().unwrap(),
```
rukai, line 820: Can't unwrap here as a misbehaving server could send something that gets parsed into a None.
`.ok_or_else(|| anyhow!("result prepared contained empty id")?` should be enough.

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -753,6 +810,46 @@ enum RewriteTableTy {
...
+            prepared.metadata.clone(),
+        ));
+    }
+
+    None
+}
+
+fn is_execute_message(message: &mut Message) -> Option<(&BodyReqExecuteOwned, &Version)> {
```
rukai, line 828: I think `is_` prefix makes sense for `bool`s because it describes what the bool represents, but in cases where we are returning message contents a `get_` prefix would be better.

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
rukai: image

Looking at it closer returning an error is too extreme.

When we send use statements to every open connection, we close any connections that return strange results as we can no longer uphold our invariants on those connections.
We should probably do that here as well.

In both of these cases we should also log the error, seems I forgot to do so with the use statements.

conorbros: We don't need to close the connection because the query will remain unprepared, so it will just attempt to reprepare it when the node is hit with an execute.

I'd be happy with either approach TBH, closing or not closing would both work.

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    /// Get a token routed replica node for the supplied execute message (if exists)
+    pub fn replica_node(
+        &mut self,
+        execute: &BodyReqExecuteOwned,
+        version: &Version,
+    ) -> Option<&mut CassandraNode> {
+        let metadata = self
+            .prepared_metadata
+            .get(&execute.id.clone().into_bytes().unwrap())
```
rukai, line 64: We can avoid the clone here by just directly storing the CBytesShort.
image

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
conorbros, line 365: TODO: Assert that all nodes return the same id and metadata in response to PREPARE

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -265,46 +273,87 @@ impl CassandraSinkCluster {
...
+                self.init_handshake_connection
+                    .as_mut()
+                    .unwrap()
+                    .send(message, return_chan_tx)?;
+            } else if is_prepare_message(&mut message) {
+                // Send the PREPARE statement to all connections
+                for node in self.pool.nodes.iter_mut() {
+                    let (return_chan_tx, return_chan_rx) = oneshot::channel();
+
+                    node.get_connection(&self.connection_factory)
```
rukai, line 320: This will open a connection against every node in the cluster.
I'm not sure if this approach or doing what we do with use statement handling is better where we only use a connection if it already exists.
The current approach is better when the prepared statement is unpopulated and the execute will hit most nodes eventually.
Not opening every connection is better in cases where the prepared statement was already fully populated by a previous session.

We can just leave it as is for now, and think about this more when we need to improve performance

benbromhead: I think this is done to avoid the following behavior:
- Send prepared statement
- Not on the routed node
- return error to client
- client responds with prepare message
- shotover routes it to a different node

So as the size of the cluster increases, the number of errors shotover returns per unrecognized statement will increase by an average of (N - Nodes_already_prepped)/2 where N == number of nodes in the cluster.

So the tradeoff is, do we have an upfront hit in latency when we try to do this for every node in the cluster (with this approach it's also done sequentially rather than concurrently) or do bounce back and forth for every prepared query.

benbromhead: But yes, lets leave this in and figure out a better way to do it later (at a minimum this needs to send out the messages to the nodes in the cluster concurrently)

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
```diff
@@ -0,0 +1,88 @@
...
+            *version,
+        )
+        .unwrap();
+
+        // TODO this should use the keyspace info to properly select the replica count
+        let mut replica_host_ids = self
+            .token_map
+            .nodes_for_token_capped(Murmur3Token::generate(&routing_key), 1);
+
+        if let Some(host_id) = replica_host_ids.next() {
```
benbromhead, line 79: Not needed for this PR - but lets put in an issue to do a random choice from the replicas returned. 
We could do fancier routing (lowest latency of two random choices etc), but that's a much later thing to think about

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
rukai: yep that make sense to me, lets keep this approach.
Your absolutely right about the connections being opened sequentially instead of concurrently?
@conorbros Do you want to make that concurrent here, or just add a TODO and get to it in a follow up?

benbromhead (commented):
Looking good so far mate, def need to think a bit more about how we make sure queries are prepped though

shotover-proxy/src/transforms/cassandra/sink_cluster/token_map.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// Returns local nodes starting at given token and going in the direction of replicas.
+    pub fn nodes_for_token_capped(
+        &self,
+        token: Murmur3Token,
+        replica_count: usize,
+    ) -> impl Iterator<Item = Uuid> + '_ {
+        self.token_ring
+            .range(token..)
+            .chain(self.token_ring.iter())
+            .take(replica_count)
```
rukai, line 30: No need to take here, the caller can .take() if they need to.

rukai: wait... possibly I misunderstand what this is for.

shotover-proxy/src/transforms/cassandra/sink_cluster/token_map.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub fn new(nodes: &[CassandraNode]) -> Self {
+        TokenMap {
+            token_ring: nodes
+                .iter()
+                .flat_map(|node| node.tokens.iter().map(move |token| (*token, node.host_id)))
+                .collect(),
+        }
+    }
+
+    /// Returns local nodes starting at given token and going in the direction of replicas.
```
rukai, line 21: I would like to stop using the phrase "local nodes" (local_nodes has been renamed in https://github.com/shotover/shotover-proxy/pull/831)

conorbros: Yeah I did think it was confusing, espc because Cassandra has it's own concept of local nodes.

shotover-proxy/src/transforms/cassandra/sink_cluster/token_map.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub fn nodes_for_token_capped(
```
rukai, line 22: This gives a low level description of the functionality.
Instead we would be better off with a high level description like: `iter_replica_nodes`

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
conorbros: I'll add it to the follow up list, should be fine for now since it will only be slow for the first prepared statement, in subsequent ones the connections will already be there.

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
conorbros: Why random choice? Other drivers just use the first one as far as I know.

I'll open an issue for latency/topology aware routing.

shotover-proxy/src/transforms/cassandra/sink_cluster/token_map.rs
```diff
@@ -0,0 +1,109 @@
...
+        );
+    }
+
+    fn verify_tokens(node_host_ids: &[Uuid], token: Murmur3Token) {
+        let token_map = TokenMap::new(prepare_nodes().as_slice());
+        let nodes = token_map
+            .iter_replica_nodes(token, node_host_ids.len())
+            .collect_vec();
+
+        assert_eq!(nodes.len(), node_host_ids.len());
```
rukai, line 104: I believe we can simplify these assertions to just `assert_eq!(nodes, node_host_ids);`.
It should give better error messages too!

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
rukai: random choice will give better load balancing.
Its pretty easy to implement, so may as well just do it in this PR:
image

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
rukai: Assuming assert, means return or log an error instead of panicking then :+1:

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -753,6 +811,45 @@ enum RewriteTableTy {
     Peers,
 }
 
+fn get_prepared_result_message(
+    message: &mut Message,
+) -> Result<Option<(CBytesShort, PreparedMetadata)>> {
```
rukai, line 816: This never returns an `Err` so we should remove the Result from the return value.

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub fn set_nodes(&mut self, nodes: Vec<CassandraNode>) {
+        self.nodes = nodes;
+        self.token_map = TokenMap::new(self.nodes.as_slice());
+    }
+
+    pub fn add_prepared_result(&mut self, id: CBytesShort, metadata: PreparedMetadata) {
+        self.prepared_metadata.insert(id, metadata);
+    }
+
+    pub fn random_node(&mut self, rng: &mut SmallRng) -> &mut CassandraNode {
```
rukai, line 37: I dont think any of the usages of this function justify sending to a completely random node instead of to a random node within the rack we are proxying to.
Lets remove this function and call `get_random_node_in_dc_rack` instead.

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
```diff
@@ -0,0 +1,86 @@
...
+            .unwrap()
+    }
+
+    /// Get a token routed replica node for the supplied execute message (if exists)
+    pub fn replica_node(
+        &mut self,
+        execute: &BodyReqExecuteOwned,
+        version: &Version,
+    ) -> Option<&mut CassandraNode> {
+        let metadata = self.prepared_metadata.get(&execute.id).unwrap();
```
rukai, line 63: These unwraps should return errors instead right?

shotover-proxy/src/transforms/cassandra/sink_cluster/routing_key.rs
```diff
@@ -0,0 +1,2498 @@
...
+        assert_eq!(routing_key, vec![0, 0, 0, 4, 0, 0, 1, 164]);
+
+        let node = router.replica_node(execute_body, &Version::V4).unwrap();
+
+        assert_eq!(node.address, "172.16.1.5:9044".parse().unwrap());
+    }
+
+    fn prepare_nodes() -> Vec<CassandraNode> {
+        vec![
+            CassandraNode::new(
```
rukai, line 150: Can we either reduce the size of the test data or move it into a separate file?

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
```diff
@@ -0,0 +1,86 @@
...
+
+        let routing_key = calculate(
```
rukai, line 65: can we have a better name than calculate?

shotover-proxy/src/transforms/cassandra/sink_cluster/routing_key.rs
```diff
@@ -0,0 +1,2498 @@
...
+) -> Option<Vec<u8>> {
+    let values = match query_values {
+        QueryValues::SimpleValues(values) => values,
+        _ => panic!("handle named"),
+    };
+
+    serialize_routing_key_with_indexes(values, pk_indexes, version)
+}
+
+fn serialize_routing_key_with_indexes(
```
rukai, line 22: This is all copied from cdrs tokio, whats the plan for upstreaming?

conorbros: It could, though it's only a couple functions and not part of the "cassandra protocol" so doesn't make sense to move it there. Definitely driver functionality.

rukai: The way I see it is, cassandra-protocol is a collection of tools for constructing your own driver, so including things like this makes sense.
What we dont want to include is things that actually make driver level decisions, anything touching things like: tcp connections, error handling, building topologies.

We also dont want to unceremoniously dump things into cassandra-protocol and then find afterwards find the the API needs reworking.
So I am quite happy to leave it as is and cleanup later when we have a better understanding of our needs.
But lets at least add a link to the relevant cdrs-tokio lines this is from so we dont forget its origin.

conorbros: My understanding is it's an implementation of this document: 
https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v5.spec

Routing is a driver functionality.

rukai: I think we need to consider what is the purpose of cassandra-protocol.
Its to let users implement their own driver level functionality, not literally just to implement the contents of a document labelled protocol.spec

For example we did already move the Murmur3Token there.
This feels the same to me.

rukai: (And again, all im truly asking for here right now is a link back to the source so we remember where it came from and can deduplicate in the future)

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
conorbros: Ah yes poor choice of words given Rust's `assert` macro

shotover-proxy/src/transforms/cassandra/sink_cluster/routing_key.rs
conorbros: Ah good point. Anyway I'll add that source link

shotover-proxy/src/transforms/cassandra/sink_cluster/routing_key.rs
rukai: Yeah ok, looking at how we actually use this test data, I really dont think its justified being so long, ~50 lines should be sufficient for the kind of test we are actually using it for.

rukai: (I'm assuming you wrote this yourself and its not copied from cdrs-tokio right?)

conorbros: I just used a real token list from Cassandra

rukai: Its ok for a unittest to use synthetic data, we have integration tests to cover more realistic cases.

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
conorbros: Should we error or just return node for the replica?

rukai: If the client or server is misbehaving then we should return an error.
If its an expected case where we cant find a replica_node then I guess we should return None?

conorbros: Hmm, I could see both occurring. Imagine we have multiple Shotover nodes, the query is prepared on one, but sent to another. The prepared metadata won't be cached.

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
conorbros: I believe this has been addressed, we are now logging errors for when the response to a prepare statement is not equal on every node.

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -370,12 +370,38 @@ impl CassandraSinkCluster {
...
+        {
+            let mut prepare_responses = super::connection::receive(
+                self.read_timeout,
+                &self.failed_requests,
+                responses_future_prepare,
+            )
+            .await?;
+
+            if !prepare_responses.windows(2).all(|w| w[0] == w[1]) {
+                let err_str = prepare_responses
```
rukai, line 382: put the err_str in the branch its used in

rukai: Actually maybe we should use err_str in the tracing::error as well

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
rukai: Yep looks good.
The logic is a bit wonky because it doesnt take into account the message sent to the control connection, but that will be cleared up once we change the prepare logic to not send to the control connection in a follow up PR

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
conorbros: Oops, intended to use it in both.

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
rukai: Theres still a few unwraps in this function btw

conorbros: Just leaving a note here: We decided to handle the prepared metadata unwrap by returning an UNPREPARED error to force the driver to give us the metadata.

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -753,6 +878,53 @@ enum RewriteTableTy {
...
+    if let Some(Frame::Cassandra(CassandraFrame {
+        operation: CassandraOperation::Execute(execute_body),
+        version,
+        stream_id,
+        tracing_id,
+        ..
+    })) = message.frame()
+    {
+        return Some((
+            *execute_body.clone(),
```
rukai, line 903: We can avoid this clone like this:
image

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -333,26 +334,57 @@ impl CassandraSinkCluster {
...
+                                                }),
+                                            }),
+                                            stream_id: metadata.stream_id,
+                                            tracing_id: metadata.tracing_id,
+                                            version: metadata.version,
+                                            warnings: vec![],
+                                        },
+                                    ))),
+                                })
+                                .unwrap();
```
rukai, line 380: `.expect("the receive is guaranteed to be alive, so this must succeed")`

shotover-proxy/src/transforms/cassandra/sink_cluster/token_map.rs
```diff
@@ -0,0 +1,106 @@
...
+pub struct TokenMap {
+    token_ring: BTreeMap<Murmur3Token, Uuid>,
+}
+
+impl TokenMap {
+    pub fn new(nodes: &[CassandraNode]) -> Self {
+        TokenMap {
+            token_ring: nodes
+                .iter()
+                .flat_map(|node| node.tokens.iter().map(move |token| (*token, node.host_id)))
```
rukai, line 16: Does the move do anything? I removed it and it still compiles.

rukai (approved):
I havent thoroughly reviewed the copy/pasted code because the plan is to upstream most of it back to cassandra-protocol.
But the rest all looks good to me!

benbromhead (approved):
Nice one mate. This is an important bit of functionality! 

We have a bit to clean up and work on, but just wanted to say great work!

PR #825, opened 2022-09-23
Create custom cassandra image for single instance integration tests

rukai:
On my machine it now takes 5s to startup the cassandra container instead of 80s.
I have not touched multi-instance integration tests because I dont think the same speedups are possible there.

I discovered the magic speedups here: https://github.com/saidbouras/cassandra-docker-unit/blob/master/scripts/setup-config.sh
Disabling gossiping is fine because its only a single instance.
Disabling virtual nodes is more iffy, but I figure its fine because the multi-instance clusters are still running with a proper virtual node setup.

It might still be worth investigating if there is anything we can do to speedup or skip gossiping on multi-instance clusters as every node starts with a complete list of seeds.
But we would want to be very careful that it doesnt alter post-setup functionality from what a production cassandra cluster would be like.

The cassandra.yaml is from here: https://github.com/apache/cassandra/blob/cassandra-4.0/conf/cassandra.yaml
The only changes I made is to set `enable_user_defined_functions` and `enable_scripted_user_defined_functions`.

### DIFF
--- a/shotover-proxy/example-configs/docker-images/cassandra-4.0.6/cassandra.yaml
+++ b/shotover-proxy/example-configs/docker-images/cassandra-4.0.6/cassandra.yaml
@@ -24,7 +24,7 @@ cluster_name: 'Test Cluster'
 # See https://cassandra.apache.org/doc/latest/getting_started/production.html#tokens for
 # best practice information about num_tokens.
 #
-num_tokens: 16
+#num_tokens: 16
 
 # Triggers automatic allocation of num_tokens tokens for this node. The allocation
 # algorithm attempts to choose tokens in a way that optimizes replicated load over
@@ -47,7 +47,7 @@ allocate_tokens_for_local_replication_factor: 3
 # vnodes (num_tokens > 1, above) -- in which case you should provide a 
 # comma-separated list -- it's primarily used when adding nodes to legacy clusters 
 # that do not have vnodes enabled.
-# initial_token:
+initial_token: 0
 
 # May either be "true" or "false" to enable globally
 hinted_handoff_enabled: true
--- a/shotover-proxy/example-configs/docker-images/cassandra-4.0.6/cassandra.yaml
+++ b/shotover-proxy/example-configs/docker-images/cassandra-4.0.6/cassandra.yaml
@@ -1,0 +1,1419 @@
+
+# Cassandra storage config YAML
+
+# NOTE:
+#   See https://cassandra.apache.org/doc/latest/configuration/ for
+#   full explanations of configuration directives
+# /NOTE
+
+# The name of the cluster. This is mainly used to prevent machines in
+# one logical cluster from joining another.
+cluster_name: 'Test Cluster'
+
+# This defines the number of tokens randomly assigned to this node on the ring
+# The more tokens, relative to other nodes, the larger the proportion of data
+# that this node will store. You probably want all nodes to have the same number
+# of tokens assuming they have equal hardware capability.
+#
+# If you leave this unspecified, Cassandra will use the default of 1 token for legacy compatibility,
+# and will use the initial_token as described below.
+#
+# Specifying initial_token will override this setting on the node's initial start,
+# on subsequent starts, this setting will apply even if initial token is set.
+#
+# See https://cassandra.apache.org/doc/latest/getting_started/production.html#tokens for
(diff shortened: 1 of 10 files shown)


conorbros:
Can we put this in a separate repo and publish to dockerhub instead?

rukai:
IMO we are better off not doing that.
If we put dockerhub in between our checked in Dockerfile and our integration tests, anytime we need to tweak something, even just to quickly test something, we need to publish to dockerhub.

Using the approach in this PR, any change to the Dockerfile will be immediately reflected in the next test run, without any user intervention.

conorbros:
I'm hesitant to start adding the building of docker images to our integration tests. 

Anytime you need to tweak something and test you can just download the docker image repo and build it yourself. Even that is an uncommon scenario, we have barely had to edit the docker images.

And of course the added advantage that you can use it like a normal docker image/other projects can use the published one.

rukai:
> And of course the added advantage that you can use it like a normal docker image/other projects can use the published one.

The only advantage I'm hearing is that other projects will be able to make use of it too.
This is a noble goal, but, in order to better serve that goal I believe we need to test it internally first for a while to make sure we have something thats worth sharing and wont have unintended side effects.

I dont believe that building images during integration tests will introduce debt.
It's very easy to remove the building logic and transferring the images to a new repo will also be very easy when the time comes.

conorbros:
>I dont believe that building images during integration tests will introduce debt

It's another part of the CI that can fail. Plus won't it need to be rebuilt every test run? Or does caching handle that for us.

rukai:
> It's another part of the CI that can fail.

If it fails, then thats good because we've caught something thats broken?
If you think `docker build` could be flakey... well thats possible I guess but I've yet to see any evidence of that being the case.

> Plus won't it need to be rebuilt every test run? Or does caching handle that for us.

It takes ~1s to build the first time because its just adding a [very tiny layer](https://docs.docker.com/storage/storagedriver/#images-and-layers) on top of `library/cassandra:4.0.6` and subsequent runs only take ~40ms (as per the comment)

conorbros:
>If it fails, then thats good because we've caught something thats broken?
A published Docker image will never be broken once it's published, but yes I suppose `docker build` won't be likely to fail.

My main concern with this change is that it sort of sets a precedent for the Shotover repo to become a kitchen sink. For something like building/versioning/modifying Docker images, I'd expect that not to be a concern for the Shotover repo, we'd use the existing tools that manage this (Dockerhub).

shotover-proxy/example-configs/docker-images/cassandra-4.0.6/cassandra.yaml
```diff
@@ -0,0 +1,1419 @@
...
+# INFO level
+# UDFs (user defined functions) are disabled by default.
+# As of Cassandra 3.0 there is a sandbox in place that should prevent execution of evil code.
+enable_user_defined_functions: true
+
+# Enables scripted UDFs (JavaScript UDFs).
+# Java UDFs are always enabled, if enable_user_defined_functions is true.
+# Enable this option to be able to use UDFs with "language javascript" or any custom JSR-223 provider.
+# This option has no effect, if enable_user_defined_functions is false.
+enable_scripted_user_defined_functions: true
```
XA21X, line 1219: It's hard to tell what's changed in here. Would it be clearer to have the Dockerfile mutate the config instead?

```
RUN sed -i -r 's/^(enable_user_defined_functions:).*$/\1 true/' /etc/cassandra/cassandra.yaml
RUN sed -i -r 's/^(enable_scripted_user_defined_functions:).*$/\1 true/' /etc/cassandra/cassandra.yaml
```

XA21X (approved):
I agree with keeping things together for now. We can always publish later.

shotover-proxy/example-configs/docker-images/cassandra-4.0.6/cassandra.yaml
rukai: It still relies on the checked in config file actually being the default and I dont find sed particularly readable.
So IMO may as well modify it directly and then let the user diff it themselves.

That said theres definitely advantages/disadvantages both ways, so if someone felt strongly about doing it one way I would be happy to follow that.

conorbros (approved):
Fair enough, in the interest of keeping things moving forward let's go ahead with this.

PR #827, opened 2022-09-26
Remove manual redis volumes from docker-compose.yml

rukai:
I'm not sure why these were here, but removing them doesnt break anything and it makes the config a lot simpler.
It's also a possibility that volumes being kept around is causing our intermittent redis failures.

I did a bit of research and volumes are known to be faster than relying on the container layer, however removing the volumes has no observable affect on the runtime of our integration tests. And if it did help we would probably be better off just entirely avoiding the fs by disabling redis's persistence.


conorbros (approved):

XA21X (approved):
Noticed the same thing ages ago. Also no idea why it was done.

Issue #730, opened 2022-08-05
Investigate different a different lock for sharing Cassandra topology


conorbros:
Might be a candidate for https://docs.rs/arc-swap/latest/arc_swap/.

Then we can avoid the lock/drop thing.

_Originally posted by @conorbros in https://github.com/shotover/shotover-proxy/pull/723#discussion_r938426527_

PR #831, opened 2022-09-28
CassandraSinkCluster: Keep node list up to date via a tokio::watcher

rukai:
Closes https://github.com/shotover/shotover-proxy/issues/730

This PR ensures that all existing connections receive all changes to the topology tasks node list.
Previously a new connection had to be made to observe the changes to the node list.
It also simplies code in a few places that had to rely on loop+sleep to wait for changes to occur, now we can just await on the recv call.

renames:
* topology_task_nodes -> nodes_rx - we are now using a receiver type so we use the standard rx naming, `nodes_rx` also pairs nicely to the `nodes` field to which it is copied.
* local_nodes -> nodes - the `local` part no longer makes sense now that we are using a Receiver, additionally local could be confused with the `local_shotover_node` field.

# Alternative approach

I originally attempted to implement this with an enum that gets sent from the topology task to the transforms providing granular updates:
```rust
pub enum NodeChange {
    NewList(Vec<CassandraNode>),
    Add(CassandraNode),
    Remove(SocketAddr),
    Status { address: SocketAddr, is_up: bool },
}
```
It allows us to avoid copying an entire node list in many cases.
However these cases are error cases and our performance doesnt have to be amazing there.
And it comes at the cost of a large increase to complexity in both the topology task side and the transform side.
While I believe we could write a correct implementation with this approach without too much trouble, it is still harder to reason about, harder to quickly verify correctness and harder to maintain.

Additionally the main cost would be having to recreate connections, and I've added logic to this PR to copy over the connections from the old list.

So in the end I decided it wasnt worth it and went with the much simpler approach in this PR.

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
@@ -27,9 +27,8 @@ use node_pool::{GetReplicaErr, NodePool};
 use rand::prelude::*;
 use serde::Deserialize;
 use std::net::SocketAddr;
-use std::sync::Arc;
-use std::time::{Duration, Instant};
-use tokio::sync::{mpsc, oneshot, RwLock};
+use std::time::Duration;
+use tokio::sync::{mpsc, oneshot, watch};
 use topology::{create_topology_task, TaskConnectionInfo};
 use uuid::Uuid;
 use version_compare::Cmp;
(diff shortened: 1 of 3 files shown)


conorbros (approved):
LGTM but could we merge #824 first? Might be easier integrating this after. I'm happy with doing that work myself if I can avoid a merge conflict on #824

rukai:
Being a smaller PR and already reviewed having one review IMO it does make sense to land this PR first.

But I'm not that fussed, if we can get #824 in within a reasonable time then sure.

benbromhead (approved):

PR #834, opened 2022-09-29
cassandra prepared statement benches

conorbros:


shotover-proxy/benches/benches/cassandra.rs
```diff
@@ -42,13 +33,16 @@ fn cassandra(c: &mut Criterion) {
         });
         for query in &queries {
```
rukai, line 34: Can we get the list of queries from something like: `resources.queries.iter_keys()`?

conorbros: Did try that but it's a function until it gets passed to `bench_with_input`

rukai: Oh true, because we need the query type to construct the bench name really early.
Quite unfortunate but I cant think of a better approach.

rukai (approved):
lgtm

benbromhead (approved):

PR #837, opened 2022-10-04
CassandraSinkCluster: Fix system.peers rpc_address

rukai:
rpc_address is a field in system.peers but not system.peers_v2
Driver's often still use system.peers instead of system.peers_v2 for routing.
So this field turned out to actually be important but we were just setting it to null.

As expected I had to change rpc_address fields in the integration tests.
However I also had to change the system.local listen_address in cluster_multi_rack, I assume that this change is now allowing the driver to properly route between multiple shotover instances resulting in different listen_address values.


conorbros (approved):

benbromhead:
I feel like we should have an integration test that catches this / demonstrates the fix. Or is it the case that our test C* drivers don't hit this edge case?

benbromhead (approved):
Otherwise lgtm

rukai:
The integration tests do catch it, we just have to use Result::Any because we dont know which node we are talking to.
Maybe we should rewrite our assertions to assert that we receive 2 of 3 expected results.

PR #838, opened 2022-10-04
Let docker-compose handle building docker images for us

rukai:
I noticed this docker-compose `build` field which seems like a much better approach than having to manually build each image.
I carefully tested it and observed:
*  no regression in time to build the images from scratch and incrementally.
*  changes to the compose file are immediately picked up next time I run `cargo test ...` (I did have to add the `--build` flag for this to work though)

### DIFF
--- a/shotover-proxy/build/install_ubuntu_packages.sh
+++ b/shotover-proxy/build/install_ubuntu_packages.sh
@@ -10,6 +10,11 @@ sudo apt-get install -y gcc-aarch64-linux-gnu
 # Install dependencies of the cpp-driver even if they are already on CI so that we can run this locally
 sudo apt-get install -y libuv1 libuv1-dev cmake g++ libssl-dev zlib1g-dev
 
+# older docker-compose on ubuntu lacks support for docker-compose build field
+pip install pip --upgrade
+pip install pyopenssl --upgrade
+pip install docker-compose==1.29.2
+
 # set VERSION to one of the tags here: https://github.com/datastax/cpp-driver/tags
 VERSION=2.16.2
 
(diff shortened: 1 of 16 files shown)


rukai:
I suspect this might be failing due to an older version of docker-compose on ubuntu 18.04
I'm going to just wait till we upgrade ubuntu in the hope that the issue fixes itself.

rukai:
docker-compose is leaving many untagged images with unique id's around.
I think what happens is docker-compose runs once, creates an image with the tag `shotover-int-tests/cassandra:4.0.6`, then runs a second time where the old image gets untagged, a new image is created with a new id and the `shotover-int-tests/cassandra:4.0.6` tag.
I think the reason the images have unique id's is because of this:
This can be observed via `docker image inspect fc29ebdc9047` (with an image id built by this PR) and then observing the docker-compose.yaml specific labels:
```json
            "Labels": {
                "com.docker.compose.project": "cassandra-passthrough",
                "com.docker.compose.service": "cassandra-one",
                "com.docker.compose.version": "2.11.2"
            }
```

I also found this https://github.com/docker/compose/issues/7905

It needs to be investigated whether this issue is causing the cache to be ignored, or if we could just clean it up with a `docker image --prune` after creating the container and keep performance equivalent to the old implementation.

Maybe not trying to share the image name would result in better cache usage as old images wouldnt need to be pruned?

I also found this, which seems to describe a great way to achieve a security vulnerability: https://docs.docker.com/compose/compose-file/build/#consistency-with-image
I'm not sure if docker is actually that stupid though, its worded as if it may or may not be implemented like that 0.0

rukai:
Our current implementation for image building is robust enough.

I'm not opposed to ever landing this change, but its not worth putting more time into at the moment.

PR #839, opened 2022-10-05
Upgrade CI to min ubuntu 20.04

rukai:
Only upgrading to ubuntu 20.04 instead of directly to 22.04 to maintain compatibility with 20.04 in our binary builds.

I chose to create a .deb rather than directly mashing the files into /usr/lib so that:
* local users have a way to uninstall the package.
* caching just works:
   + we tell GA to cache the packages folder and then it will cache every package that gets built
        - built packages are really small, ~1MB for this one, so we dont need to worry about excessive cache usage.
        - packages are versioned so we dont have to worry about the cache accidentally pulling in an old version of cassandra-cpp-driver.

### 1
Publish packages somewhere and wget them like we do currently with the datastax packages.
I didnt take this approach as we need somewhere to host them and it is a lot of extra work to package them every time a new release occurs that we want to use.

### 2
Annoy datastax into publishing packages for a modern ubuntu.
Not sure if they care enough about the cpp driver for that, and even if they did, we would still need to reannoy them in the future when a new ubuntu comes out.

### 3
Force installation of the 18.04 package onto a modern ubuntu.
I tested this and it doesnt work at all.
We can succesfully install the package with `dpkg --force-install -i ...`.
However when we try to build the project it fails with linker errors.

## References:
* Followed this guide to create the package https://ubuntuforums.org/showthread.php?t=910717
* I reused the naming scheme of the old packages: https://downloads.datastax.com/cpp-driver/ubuntu/18.04/cassandra/v2.16.0/
* I used the AUR package as a reference for how to build the project: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=cassandra-cpp-driver


conorbros (approved):

benbromhead:
I'm so mad at this dependency right now... this seems like the easiest path forward right now. In the mean time imma bug datastax

benbromhead (approved):

PR #841, opened 2022-10-05
share prepared metadata between connections

conorbros:
#824 followup


rukai:
Will we need to use the other fields of PreparedMetadata for anything later on?
If we can reduce it to just pk_indexes then we can reduce the amount of time that we spend holding the read lock.

conorbros:
Not sure off the top of my head, I'll add that to the followups for token aware.

rukai (approved):
Alright, leaving it as a follow up is fine by me

It could be worth investigating using a concurrent hashmap instead of a `rwaccess<HashMap>` to avoid locking, but I think it's not super important as the lock isn't held for very long.
That said if we were unable to remove the other fields from the value then w concurrent hashmap would be more valuable

XA21X (approved):

PR #842, opened 2022-10-05
Create new connections concurrently for prepared statements

conorbros:


rukai (approved):
Nice!

Theres a few preexisting correctness issues here that will need to be fixed eventually, but I've filed an issue so we can do that later: https://github.com/shotover/shotover-proxy/issues/857

XA21X (approved):

PR #843, opened 2022-10-05
CassandraSinkCluster: Disallow inserting/removing the nodes list from outside of NodePool

rukai:
Better integrates the separately developed abstractions of https://github.com/shotover/shotover-proxy/pull/831 and https://github.com/shotover/shotover-proxy/pull/824


conorbros (approved):
LGTM! I think I originally had plans along the lines of the node list being immutable but it's use wasn't clear until your work on 831.

PR #845, opened 2022-10-06
fix test

conorbros:
If you run `cargo test` an error will appear because the force encode transform is behind `alpha-transforms`.


rukai (approved):
Ooops ...
Thanks for the fix.

XA21X (approved):

PR #846, opened 2022-10-06
Round Robin load balancing for CassandraSinkCluster

conorbros:


rukai (approved):
I'll leave it up to ben whether round robin is a better load balancing solution then a single random node.
But the implementation seems fine to me.

XA21X:
Could round-robin introduce any unwanted bias? We're sharing the same index between what is effectively multiple lists (nodes partitioned by rack), although they should all be the same size under normal operation. I can't imagine round-robin being any worse than random, and the more even distribution from round-robin is probably preferable. 👍

Minor note: It seems slightly odd that the round-robin always starts from index 1 (second item) though. 😂

XA21X (approved):

benbromhead:
RR is not the worst choice right now. We can improve this later.

benbromhead (approved):

Issue #768, opened 2022-08-29
Cassandra integration tests - rx_process task terminated


rukai:
This error can be seen in some cassandra integration tests (or at least it will be once https://github.com/shotover/shotover-proxy/pull/765 lands)

I believe it indicates we are doing something weird with connections, not closing things in the right order or something.
A properly closed connection should return an `Ok(None)` not an error.
I think we need to go over it carefully and have a good understanding of the order of shutdown of the read/write halves.
Referring to `server.rs`'s `spawn_read_write_tasks` should give a good reference of how to do this.

But maybe I am all wrong and things are different because this is a proxy<->server connection instead of a client<->proxy connection.

rukai:
I've narrowed this down to being caused by the network settings on the cluster docker-compose.yaml
The problem does not occur on `cassandra_int_tests::test_passthrough` but does occur if we copy over the network settings from the cluster test.

rukai:
The problem only exists when connecting directly to the containers bridge network address. e.g. `172.160.1.2:9042`
If you keep the exact same bridge network setup but do `ports: ["9043:9042"]` and set shotover to connect to `127.0.0.1:9042` then the problem disappears.

PR #850, opened 2022-10-11
Cassandra: Do not log error on TCP RST

rukai:
closes https://github.com/shotover/shotover-proxy/issues/768

My research on this topic has led me to conclude that TCP's RST is nowadays a standard way to close the connection and logging it as exceptional is incorrect.

Traditionally the protocol goes like:
```
client: FIN ACK
server: FIN ACK
client ACK
```

but cassandra appears to be trying to skip processing the final ACK by shortening the exchange to:
```
client: FIN ACK
server: RST ACK
```

In the issue I observed that some docker configurations result in shotover hitting the RST and some configurations shotover does not hit a RST.
However upon inspecting this with wireshark I found that the RST packets are always sent by cassandra regardless of the config.
I'm guessing that the change in docker config is causing the RST to be dropped by dockers networking somehow

see for example: https://osqa-ask.wireshark.org/questions/15815/all-tcp-connections-end-with-rst/


XA21X (approved):
C* uses Netty, which indeed seems to send RST by default:  https://github.com/netty/netty/issues/6818

rukai:
Handling it properly requires significant refactoring, so I've just added a comment explaining what a proper solution would look like, I'd still like to land this PR so we can get our logs back into a manageable state.

rukai:
I went ahead and rewrote this to use my suspected proper fix ... and it works! 
Probably needs to be re-reviewed from scratch as the approach to this PR is now completely different.

There are a few instances still remaining of that error, but I believe they are actual legitimate cases caused by purposely shutting down a cassandra instance for testing how shotover responds to that.

conorbros (approved):

PR #851, opened 2022-10-11
Create custom image for tls cassandra

rukai:
Cuts off ~2mins from cassandra_int_tests total runtime. (cuts ~1 min off of 2 integration tests)
also cuts 2min off the total runtime for CI benchmarks (1 TLS bench and every bench is run twice in CI)

Needed to create a separate image as we need to modify the config file to set TLS.

I also noticed the regular image cassandra.yaml had some experimental features enabled so I disabled those.


conorbros:
I thought we just needed to provide the keystore to enable TLS?

rukai:
image
We also need to set client_encryption_options to true

shotover-proxy/example-configs/docker-images/cassandra-4.0.6/cassandra.yaml
```diff
@@ -1404,16 +1404,16 @@ report_unconfirmed_repaired_data_mismatches: false
...
 
 # Enables creation of transiently replicated keyspaces on this node.
 # Transient replication is experimental and is not recommended for production use.
-enable_transient_replication: false
+#enable_transient_replication: false
 
 # Enables the used of 'ALTER ... DROP COMPACT STORAGE' statements on this node.
 # 'ALTER ... DROP COMPACT STORAGE' is considered experimental and is not recommended for production use.
-enable_drop_compact_storage: false
+#enable_drop_compact_storage: false
```
XA21X, line 1419: I see we're just switching to the implicit defaults, which are all false anyways.

It's slightly confusing that this was working for v4, given the configs were renamed:

https://cassandra.apache.org/doc/latest/cassandra/configuration/cass_yaml_file.html#materialized_views_enabled
https://cassandra.apache.org/doc/3.11/cassandra/configuration/cass_yaml_file.html#enable_materialized_views

`enable_*` to `*_enabled`

I assume C* accepts both for backwards compatibility... 🤔

XA21X (approved):

shotover-proxy/example-configs/docker-images/cassandra-4.0.6/cassandra.yaml
rukai: oh... I didnt notice they were false :sweat_smile: 
I was very confused why it was setting experimental config options by default, so this makes a lot more sense.
I'll put them back, better to keep the config as unchanged where possible, to keep diffs simpler.

rukai: Oh cassandra 3.0 does enable it by default: https://github.com/apache/cassandra/blob/dc2acba043c6978b32a9556e0d610251d5535ce6/conf/cassandra.yaml#L1294
wild.

rukai: > enable_* to *_enabled

The default config file does use ~enable_*` instead of `*_enabled`
https://github.com/apache/cassandra/blob/488c0c75a8f632f2db4e3db39f2ebcf8a489971e/conf/cassandra.yaml#L1407

conorbros (approved):

PR #853, opened 2022-10-12
Cassandra flags

conorbros:
Pre-reqs:
- https://github.com/Metaswitch/cassandra-rs/pull/147 Enables setting the tracing flag in `cassandra-cpp`
- https://github.com/krojew/cdrs-tokio/pull/118 Fixes parsing of flags in `cassandra-protocol`

At the moment Shotover is just stripping flags from messages. This PR implements support for flags. Currently this PR just tests we can send flags without error, but the followup: #854 uses this support to get tracing information from queries sent through Shotover.

- [ ] switch to shotover fork of cassandra-cpp

### DIFF
--- a/shotover-proxy/src/frame/cassandra.rs
+++ b/shotover-proxy/src/frame/cassandra.rs
@@ -98,25 +98,64 @@ pub struct CassandraMetadata {
     // missing `warnings` field because we are not using it currently
 }
 
+#[derive(PartialEq, Debug, Clone, Copy)]
+pub enum Tracing {
+    Request(bool),
+    Response(Option<Uuid>),
+}
+
+#[allow(clippy::from_over_into)]
+impl Into<Option<Uuid>> for Tracing {
+    fn into(self) -> Option<Uuid> {
+        match self {
+            Self::Request(_) => None,
+            Self::Response(uuid) => uuid,
+        }
+    }
+}
+
 #[derive(PartialEq, Debug, Clone)]
 pub struct CassandraFrame {
     pub version: Version,
     pub flags: Flags,
     pub stream_id: StreamId,
-    pub tracing_id: Option<Uuid>,
+    pub tracing: Tracing,
     pub warnings: Vec<String>,
     /// Contains the message body
     pub operation: CassandraOperation,
 }
 
 impl CassandraFrame {
+    pub fn new(
+        version: Version,
+        flags: Flags,
+        stream_id: StreamId,
+        warnings: Vec<String>,
+        operation: CassandraOperation,
+        tracing_id: Option<Uuid>,
+    ) -> Self {
+        let tracing = match operation.to_direction() {
+            Direction::Request => Tracing::Request(flags.contains(Flags::TRACING)),
+            Direction::Response => Tracing::Response(tracing_id),
+        };
+
+        Self {
+            version,
+            flags,
+            stream_id,
+            warnings,
+            operation,
+            tracing,
+        }
+    }
+
     /// Return `CassandraMetadata` from this `CassandraFrame`
     pub(crate) fn metadata(&self) -> CassandraMetadata {
         CassandraMetadata {
             version: self.version,
             flags: self.flags,
             stream_id: self.stream_id,
-            tracing_id: self.tracing_id,
+            tracing_id: self.tracing.into(),
             opcode: self.operation.to_opcode(),
         }
     }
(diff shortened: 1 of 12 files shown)


rukai:
Why do we need to carry the flags around?
Shouldnt the flag automatically get set when tracing_id is Some?

conorbros:
Because tracing_id is None for requests, but the tracing flag is set so Cassandra attaches a tracing_id to the response.

rukai:
Oooh, I see, thanks!
I'll do some research

conorbros:
Here's the relevant bit: https://github.com/apache/cassandra/blob/c5f8e63f67d105f483c56dd4f09b5ee6e46a4241/doc/native_protocol_v4.spec#L132

rukai:
A critical goal for shotover's message format is to make it so that users can modify or read fields of the message with the expectation that there is a single source of truth. And I think this design could be improved to meet that.

Sure, a transform can change the tracing_id and have the flag automatically set to the correct value at encode time.
But for transforms further down the chain they will hit an inconsistency, they might for example check the value of the tracing flag and not realize that the tracing_id is out of sync.

I can think of two alternative approaches.
Note that I'm not making any argument how cassandra-protocol should represent this, that can be decided separately.

## Alternative 1
```rust
enum Tracing {
    Request { request_tracing_id: bool }
    Response (Uuid)
}

struct CassandraFrame {
    tracing: Tracing,
    ...
}
```

This does introduce a new way to represent an invalid state: `Tracing::Request` with a response CassandraOperation.
But the existing implementation could already hit the same problem already by having `tracing_id: Some` with a request CassandraOperation.

## Alternative 2

```rust
pub enum CassandraOperationRequest {
    Query {
        query: Box<CassandraStatement>,
        params: Box<QueryParams>,
    },
    Execute(Box<BodyReqExecuteOwned>),
    Register(BodyReqRegister),
    Batch(CassandraBatch),
    ...
}

pub enum CassandraOperationResponse {
    Result(CassandraResult),
    Error(ErrorBody),
    Prepare(Vec<u8>),
    Event(ServerEvent),
    ...
}

pub enum Direction {
    Request { operation: CassandraOperationRequest, tracing: bool },
    Response { operation: CassandraOperationResponse, tracing: Uuid },
}

struct CassandraFrame {
    direction: Direction,
    ...
}
```

This completely eliminates invalid states, but it is pretty invasive so we could leave it as a possible follow up refactor if you want.

conorbros:
I decided to go with option 1 since it's much simpler to implement and doesn't block this PR anymore. I'll add note to add considering option 2 for a followup.

conorbros:
In the interest of moving this pre-req PR forward since it's blocking token aware routing, I've made the flags only public within the crate and intend on opening a followup PR with the above changes mentioned.

Can we continue with on with the reviews?

rukai:
I've made a prereq PR to unblock this https://github.com/shotover/shotover-proxy/pull/869 while avoiding adding the flags field.
Once its landed you can just set tracing_id to `Some(Uuid::new())` on the request and then at encode time the flag will get set correctly while the dummy tracing uuid is unused.

rukai:
we will also need to make a shotover/cassandra-cpp fork in order to progress this PR, since we havent heard back about https://github.com/Metaswitch/cassandra-rs/pull/147

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -92,18 +91,61 @@ pub mod raw_frame {
...
+        }
+    }
+}
+
+impl Tracing {
+    fn from_frame(frame: &RawCassandraFrame) -> Result<Self> {
+        match frame.direction {
+            Direction::Request => Ok(Self::Request(frame.flags.contains(Flags::TRACING))),
+            Direction::Response => {
+                if frame.tracing_id.is_none() && frame.flags.contains(Flags::TRACING) {
```
rukai, line 128: Cant we just rely on cassandra-protocol to give us sensible values?
Why do we need to check its work?

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -695,7 +737,7 @@ pub struct CassandraBatch {
 impl Display for CassandraFrame {
     fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
         write!(f, "{} stream:{}", self.version, self.stream_id)?;
-        if let Some(tracing_id) = self.tracing_id {
+        if let Tracing::Request(tracing_id) = self.tracing {
```
rukai, line 740: We are missing Tracing::Response here.
This is how I would go about it.
```rust
        match self.tracing {
            Tracing::Request(request) => {
                if request {
                    write!(f, " request_tracing_id:{}", request)?;
                }
            }
            Tracing::Response(response) => {
                if let Some(tracing_id) = response{
                    write!(f, " tracing_id:{}", tracing_id)?;
                }
            }
        }
````

File: shotover-proxy/tests/cassandra_int_tests/cluster.rs [hunk 3]
```diff
@@ -47,17 +47,15 @@ fn create_handshake() -> Vec<Message> {
...
+            tracing: Tracing::Request(false),
             warnings: vec![],
             operation: CassandraOperation::Startup(b"\0\x01\0\x0bCQL_VERSION\0\x053.0.0".to_vec()),
         })),
         Message::from_frame(Frame::Cassandra(CassandraFrame {
             version: Version::V4,
-            flags: Flags::default(),
             stream_id: 128,
-            tracing_id: None,
+            tracing: Tracing::Response(None),
```
rukai, line 58: AuthResponse is actually a request

benbromhead (approved):

rukai (approved):

PR #854, opened 2022-10-12
Fix and test token aware routing

conorbros:
Pre-req: 
- https://github.com/shotover/shotover-proxy/pull/853

Fixes a bug in token aware routing and tests against the Scylla driver.


rukai:
Why do we need to carry the flags around?
Shouldnt the flag automatically get set when tracing_id is Some?

conorbros:
Flags PR is #853

shotover-proxy/tests/cassandra_int_tests/routing.rs
```diff
@@ -0,0 +1,100 @@
+use crate::helpers::cassandra::{run_query, CassandraConnection, CassandraDriver};
+
+pub async fn create_keyspace(connection: &mut CassandraConnection) {
+    let create_ks: &'static str =
+        "CREATE KEYSPACE IF NOT EXISTS test_routing_ks WITH REPLICATION = { \
+                                   'class' : 'SimpleStrategy', 'replication_factor' : 1 };";
```
rukai, line 6: Why do we need so much indent here?

shotover-proxy/tests/helpers/cassandra.rs
```diff
@@ -326,6 +379,75 @@ impl CassandraConnection {
...
+            Self::Scylla { session, .. } => {
+                let mut prepared = session.prepare(query).await.unwrap();
+                prepared.set_tracing(true);
+                PreparedQuery::Scylla(prepared)
+            }
+        }
+    }
+
+    #[allow(dead_code)]
+    pub async fn execute_prepared_with_tracing(
```
rukai, line 391: either name this something like: `execute_prepared_returning_coordinator_node`
Or return something like:
```rust
struct TracingResults {
    coordinator_node: IpAddr,
}
```

Depends on if you expect to need to expose more tracing values in the future

shotover-proxy/tests/cassandra_int_tests/routing.rs
```diff
@@ -0,0 +1,97 @@
...
+        )
+        .await;
+    let mut cassandra = CassandraConnection::new(
+        cassandra_contact_points,
+        cassandra_port,
+        CassandraDriver::Scylla,
+    )
+    .await;
+
+    create_keyspace(&mut shotover).await;
```
rukai, line 43: why do we generate the keyspace through shotover and then the table directly to cassandra?

shotover-proxy/tests/helpers/cassandra.rs
conorbros: I'll go with #1, leave the second for a future improvement if required.

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -148,6 +150,7 @@ async fn test_cluster_single_rack_v3(#[case] driver: CassandraDriver) {
         };
         standard_test_suite(&connection, driver).await;
         cluster_single_rack_v3::test_dummy_peers(&connection().await).await;
+        routing::test("127.0.0.1", 9042, "172.16.1.2,172.16.1.3,172.16.1.4", 9042).await;
```
rukai, line 153: Is there a reason we need to provide every node in the cluster to scylla?
I would have thought it would derive the topology as other drivers do?
If we specify just the one node we can drop the string parsing for enable_schema_awaiter.

rukai:
I removed the bug fix and confirmed that causes the new test case to fail.

I also added some debug logs to verify that we were indeed routing to different nodes each time:
```
2022-10-23T23:56:02.855725Z ERROR lib::cassandra_int_tests::routing: delete hit: 172.16.1.3
2022-10-23T23:56:03.065194Z ERROR lib::cassandra_int_tests::routing: delete hit: 172.16.1.4
2022-10-23T23:56:03.274862Z ERROR lib::cassandra_int_tests::routing: delete hit: 172.16.1.3
2022-10-23T23:56:03.483010Z ERROR lib::cassandra_int_tests::routing: delete hit: 172.16.1.2
2022-10-23T23:56:03.691256Z ERROR lib::cassandra_int_tests::routing: delete hit: 172.16.1.2
```

The new routing.rs test case takes 84s to run on my machine.
Thats quite a hit to our total test runtime (+168s) and to the runtime of our standard cluster test case test_cluster_single_rack_v4 (84s)

With only 800 prepared queries that is 100ms per query, that seemed pretty bad to me?
But even when I skip shotover and query cassandra directly it still takes 83s, so seems cassandra is just slow :zzz: 

Can we lower the loop to something like `0..10`, yes it opens us up to a likelier chance of a false negative but that is not as likely as it first sounds since that still leaves us with `10 * 4 * 2 = 80` asserts and that number will grow as we expand our test suite to run against more scenarios.
(would also be down for `0..20` if 10 is pushing it)
We would also pretty quickly catch such cases as its bound to fail the next time we run tests.

shotover-proxy/tests/cassandra_int_tests/routing.rs
conorbros: Just a sanity check that both are working as expect.

rukai: It does mean we dont run the schema_awaiter logic.
But we havent hit any failures for that yet so maybe scylla does internal logic for schema awaiting on DDL queries?

shotover-proxy/tests/cassandra_int_tests/mod.rs
conorbros: I assumed that we might have to wait for the topology info to update in the driver but it seems that was wrong.

rukai (approved):
Good job on coming up with a way to test that our routing and then figuring out what was wrong with our implementation.
Using tracing to ensure our routing matches that of a known good implementation is really clever!

benbromhead (approved):
Nice one - clever!!

PR #856, opened 2022-10-13
Fix BenchResources drop order

rukai:
The fields of a struct are dropped in the order that they are defined: https://stackoverflow.com/questions/41053542/forcing-the-order-in-which-struct-fields-are-dropped

Reordering the BenchResources fields such that we drop the `ShotoverManager` before we drop the `DockerCompose` avoids errors like the following from being logged:
```
2022-10-13T02:14:52.562355Z ERROR connection{id=4 source="CassandraSource"}: shotover_proxy::transforms::cassandra::connection: tx_process task terminated

Caused by:
    channel closed
2022-10-13T02:14:52.562396Z ERROR connection{id=4 source="CassandraSource"}: shotover_proxy::server: connection was unexpectedly terminated

Caused by:
    0: chain failed to send and/or receive messages
    1: channel closed
```

### DIFF
--- a/shotover-proxy/benches/benches/cassandra.rs
+++ b/shotover-proxy/benches/benches/cassandra.rs
@@ -261,9 +261,9 @@ criterion_group!(benches, cassandra);
 criterion_main!(benches);
 
 pub struct BenchResources {
-    _compose: DockerCompose,
-    _shotover_manager: ShotoverManager,
     connection: CassandraConnection,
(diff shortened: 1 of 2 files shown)


conorbros (approved):

XA21X (approved):
Wow that's an interesting RFC... people started relying on the unspecified behaviour, so Rust had to keep the counter-intuitive drop order for backwards compatibility.  Huh.

PR #859, opened 2022-10-14
CassandraSinkCluster: rewrite rpc_address

rukai:
By default this field is 0.0.0.0, in that case no mapping is required as 0.0.0.0 is valid no matter what our ip is.
However if a cassandra node does have rpc_address configured properly, then we would be leaking internal cluster ip addresses by not remapping it.
So this PR maps it to shotovers ip address.

Its also worth noting that rpc_address IS the field used to set the address where the client connects to, despite `rpc` making it sound like it has to do with the old thrift stuff.
The old name is just kept for legacy reasons.


benbromhead (approved):
cc @XA21X can you double check on the C* driver operation side of this. Sounds right to me

conorbros (approved):

XA21X (approved):
LGTM. (Skimmed docs and driver code, but don't have time to go too deep.)

PR #862, opened 2022-10-17
cassandra_int_tests: reduce total table count

rukai:
This greatly reduces test runtime:
* Cuts 4s off test_passthrough (9s -> 5s)
* Cuts 54s off test_cluster_single_rack_v4 (108s -> 54s)

The wins are much larger on cluster tests because we have to wait for schema agreement after every DDL command on cluster tests.

Every test case from the old implementation is carried over into the new implementation just grouped together into the same tables.

The number of tables is now well below 150, so we no longer see any of these warnings anymore:
```
1665977080.901 [WARN] (decoder.cpp:134:bool datastax::internal::core::Decoder::decode_warnings(datastax::internal::core::WarningVec&)): Server-side warning: Cluster already contains 259 tables in 10 keyspaces. Having a large number of tables will significantly slow down schema dependent cluster operations.
```

I renamed the keyspace + table names to be more concise because the long strings were making rustfmt give up.

### DIFF
--- a/shotover-proxy/tests/cassandra_int_tests/collections.rs
+++ b/shotover-proxy/tests/cassandra_int_tests/collections.rs
@@ -2,10 +2,7 @@ use crate::helpers::cassandra::{
     assert_query_result, run_query, CassandraConnection, CassandraDriver, ResultValue,
 };
 use cassandra_protocol::frame::message_result::ColType;
-
-fn get_map_example(value: &str) -> String {
-    format!("{{0 : {}}}", value)
-}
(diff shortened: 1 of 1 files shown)


conorbros (approved):

benbromhead:
My quick read of the code is that we still have coverage on all types, but now just in a couple of tables?

rukai:
yep, thats exactly it

benbromhead (approved):
Confirmed with @rukai offline that coverage is still maintained. Approved

PR #863, opened 2022-10-17
Update deps

rukai:
`ClusterClientBuilder::open` now has deprecated warning so I replaced it with build.


rukai:
I raised https://github.com/redis-rs/redis-rs/issues/695 for the redis failure, I'll mark this as draft, maybe just remove the redis update if that takes a while to resolve.

rukai:
The redis issue is resolved, you can read details about the required change on issue but basically redis-rs now silently fails with the old generic args we were giving it.

conorbros (approved):

PR #870, opened 2022-10-19
Split CassandraFrame into CassandraFrameRequest and CassandraFrameResponse

rukai:
Waiting on: https://github.com/shotover/shotover-proxy/pull/869

Dont bother reviewing this, this still needs a lot of thought and experimentation.


rukai:
This approach would certainly be more "correct" but it might also be more painful to actually use.

Its not really worth pursuing more unless we ever get bitten by the "uncorrectness" of the current approach.

PR #873, opened 2022-10-21
existing connections handle down node

rukai:
We already cleared the node pool of dead connections but we forgot to recreate the control connection when it goes down

The insert statements were moved outside of test_connection_handles_node_down to avoid running them twice.


shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -515,6 +518,21 @@ impl CassandraSinkCluster {
         Ok(())
     }
 
+    async fn recreate_control_connection_from_node_in_pool(&mut self) -> Result<()> {
+        let random_address = self
```
conorbros, line 522: Not a random address, also could we pull out this logic into a common `create_control_connection` function and use it for creating the initial one too?

rukai: Good point, I've refactored accordingly.

conorbros (approved):

benbromhead (approved):

PR #875, opened 2022-10-21
CassandraSinkCluster: Document desired error handling

rukai:
This documents what shotover's error handling for CassandraSinkCluster should look like.
Currently we dont handle in the way described here, but I wanted us to agree on an error handling approach that seemed reasonable and that we can work towards. We can of course change things later if we find we need to though.

We could also return an overloaded error when a connection error occurs before sending the message as that would simplify shotover by letting us push retries onto the client.


docs/src/transforms.md
```diff
@@ -122,6 +122,18 @@ While `system.peers`/`system.peers_v2` will be rewritten to list the configured
   # read_timeout: 60
 ```
 
+#### Error handling
+
+If shotover sends a request to a node and never gets a response, (maybe the node went down), shotover will return a cassandra `Server` error to the client.
+This is because the message may or may not have succeded, so only the client can attempt to retry as the retry may involve checking if the original query did in fact complete succesfully.
+
+If no nodes are capable of receiving the query then shotover will return a cassandra `Overloaded` error indicating that the client should retry the query at some point.
```
conorbros, line 130: *Shotover :P

XA21X: Cassandra should be capitalised too, I think?

benbromhead (approved):

conorbros (approved):

PR #879, opened 2022-10-24
Add more context to error logs

rukai:
Got an out of context ssl error earlier that these log improvements will help with

### DIFF
--- a/shotover-proxy/src/tls.rs
+++ b/shotover-proxy/src/tls.rs
@@ -101,7 +101,10 @@ impl TlsConnector {
         let ssl = self.connector.configure()?.into_ssl("localhost")?;
 
         let mut ssl_stream = SslStream::new(ssl, tcp_stream)?;
-        Pin::new(&mut ssl_stream).connect().await?;
+        Pin::new(&mut ssl_stream)
+            .connect()
+            .await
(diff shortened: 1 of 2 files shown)


shotover-proxy/src/tls.rs
```diff
@@ -101,7 +101,10 @@ impl TlsConnector {
         let ssl = self.connector.configure()?.into_ssl("localhost")?;
 
         let mut ssl_stream = SslStream::new(ssl, tcp_stream)?;
-        Pin::new(&mut ssl_stream).connect().await?;
+        Pin::new(&mut ssl_stream)
+            .connect()
+            .await
+            .map_err(|e| anyhow!(e).context("Failed to connect to TLS connection"))?;
```
XA21X, line 107: Nitpick: Connecting to a connection... sounds odd. How about "Failed to establish TLS connection"?

XA21X (approved):

shotover-proxy/src/tls.rs
rukai: Thats a good point.
Another thing to keep in mind is the direction is important here.
We failed to connect to something rather then failing to accept a connection.

I think I'll go with "Failed to establish TLS connection to destination"

benbromhead (approved):

PR #882, opened 2022-10-25
Remove test_ prefix from integration tests

rukai:
The goal of this PR is to improve some minor papercuts when running integration tests:
* When running a specific test having to include `test_` is redundant because we already know its a test.
* when specifying a cassandra version, sometimes its `cassandra4` sometimes its `v4`, I standardized to `v4`
* I added a `_standard` postfix to the `passthrough` test case to allow easy disambiguating between `passthrough_encode`. Currently running `cargo test --all-features cassandra_int_tests::test_passthrough` will run both `passthrough_encode` and `passthrough`
* I removed the `cassandra_` or `redis_` prefix that sometimes appeared, there is always the `cassandra_int_tests::` prefix to disambiguate between redis and cassandra tests

To avoid collisions with test names I moved the cluster modules inside the cluster module.
This change inherently makes sense anyway, so seems like a good way to resolve the collision.

Some of these issues could be avoided with the `--test test_name` flag but its easier to have things just work with the default filter system.


rukai:
Lets wait a bit and see if we get an upstream fix: https://github.com/la10736/rstest/issues/165

conorbros (approved):

benbromhead (approved):

PR #885, opened 2022-10-26
include the name of driver in parametrized test results

rukai:
Results in test names like: `assandra_int_tests::test_passthrough::case_1_cdrs` instead of `assandra_int_tests::test_passthrough::case_1`

I couldnt find this properly documented, but I did find it demonstrated in: https://github.com/la10736/rstest#complete-example

Its a shame we cant remove the case_1 prefix, but this is still an improvement


conorbros (approved):
Nice find! LGTM

benbromhead (approved):

PR #886, opened 2022-10-26
cassandra_int_tests: add node lost test case

rukai:
We currently test when the node shutsdown cleanly and therefore notifies its peers and therefore shotover of the node going down.
However we didnt test the case where the node silently dissapears requiring shotover to detect the missing node itself.
This PR adds that test case.

It needs to be added as a separate test case because the existing call to `test_node_going_down` leaves the cluster in a bad state.

~~Still need to investigate and fix the problem this reveals.
But dumping this here as a draft as I'm working on something else first.~~


rukai:
This is working now, I believe it was fixed by: https://github.com/shotover/shotover-proxy/pull/873

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -199,7 +199,22 @@ async fn cluster_single_rack_v4(#[case] driver: CassandraDriver) {
...
+#[cfg_attr(feature = "cassandra-cpp-driver-tests", case(Datastax))]
+#[tokio::test(flavor = "multi_thread")]
+#[serial]
+async fn cluster_single_rack_node_lost(#[case] driver: CassandraDriver) {
+    let compose =
+        DockerCompose::new("example-configs/cassandra-cluster/docker-compose-cassandra-v4.yaml");
+
+    let shotover_manager =
+        ShotoverManager::from_topology_file("example-configs/cassandra-cluster/topology-v4.yaml");
+    cluster::single_rack_v4::test_node_going_down(compose, shotover_manager, driver, true).await;
```
conorbros, line 217: Couldn't this test just go at the end of the other tests?

benbromhead (approved):

shotover-proxy/tests/cassandra_int_tests/mod.rs
rukai: which test are you suggesting? I dont think any of the other existing tests are compatible with it?

conorbros: I was suggesting it could go the end of the `single_rack_v4` test.

rukai: image
From the PR description

conorbros: Yes but if it goes as the last test it won't matter?

rukai: There is already a test_node_going_down call at the end of single_rack_v4.
We cant have 2 test_node_going_down calls at the end of single_rack_v4 because the first call would leave the cluster in a state where the 2nd call would fail

conorbros: Ah my mistake

conorbros (approved):

PR #887, opened 2022-10-27
control connection creation error handling

rukai:
Previously shotover would give up if the node it tried to use for its control connection was down.
Now it sequentially attempts all other nodes it knows about until it either runs out of nodes reporting an error or makes a successful connection.

An error is returned if no successful connections could be made:
image

A warning is emitted if any failures occur but a successful connection could still be made in the end.
image

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
@@ -8,8 +8,8 @@ use cassandra_protocol::frame::Version;
 use cassandra_protocol::token::Murmur3Token;
 use cassandra_protocol::types::CBytesShort;
 use rand::prelude::*;
-use std::collections::HashMap;
 use std::sync::Arc;
+use std::{collections::HashMap, net::SocketAddr};
 use tokio::sync::{watch, RwLock};
 
 pub enum GetReplicaErr {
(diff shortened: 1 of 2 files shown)


benbromhead (approved):

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
```diff
@@ -73,6 +73,21 @@ impl NodePool {
         write_lock.insert(id, metadata);
     }
 
+    pub fn get_shuffled_nodes_in_dc_rack(
+        &mut self,
```
conorbros, line 77: Do we need to collect twice? This can just return a list of the shuffled addresses.

rukai: Good idea

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
```diff
@@ -73,6 +73,22 @@ impl NodePool {
         write_lock.insert(id, metadata);
     }
 
+    pub fn get_shuffled_nodes_in_dc_rack(
```
conorbros, line 76: *shuffled addresses, we say nodes in the other methods when returning a `CassandraNode`

rukai: done

conorbros (approved):

PR #891, opened 2022-10-31
Map upstream tls errors into a better format

rukai:
I originally intended on upstreaming these formatting improvements but my PR was rejected.
With some extra boilerplate we can make use of them here though.
Previously our TLS errors looked like this:
image

With this PR they look like this:
image

I've chosen to keep the library and file name in the output because TLS errors can still be quite cryptic and it can at least give us a clue in those scenarios.


conorbros (approved):

XA21X:
It's interesting... I looked the old format, and the rejected PR. Turns out I'm more familiar with that format too. 😂

XA21X (approved):

PR #895, opened 2022-11-02
Use keyspace info for replication factor

conorbros:
This PR gets the keyspace metadata from the `system_schema.keyspaces` table. When a keyspace using `NetworkTopologyStrategy` we only use the replication factor for the data center the Shotover node is in. Processing schema change events occurs the same as topology change events.

### DIFF
--- a/shotover-proxy/src/frame/cassandra.rs
+++ b/shotover-proxy/src/frame/cassandra.rs
@@ -205,38 +205,45 @@ impl CassandraFrame {
                 if let ResponseBody::Result(result) = frame.response_body()? {
                     match result {
                         ResResultBody::Rows(rows) => {
-                            let converted_rows =
-                                if rows.metadata.flags.contains(RowsMetadataFlags::NO_METADATA) {
-                                    rows.rows_content
-                                        .into_iter()
-                                        .map(|row| {
-                                            row.into_iter()
-                                                .map(|row_content| {
+                            let converted_rows = if rows
+                                .metadata
+                                .flags
+                                .contains(RowsMetadataFlags::NO_METADATA)
+                            {
+                                rows.rows_content
+                                    .into_iter()
+                                    .map(|row| {
+                                        row.into_iter()
+                                            .map(|row_content| {
+                                                if row_content.is_empty() {
+                                                    MessageValue::Null
+                                                } else {
                                                     MessageValue::Bytes(
                                                         row_content.into_bytes().unwrap().into(),
                                                     )
-                                                })
-                                                .collect()
-                                        })
-                                        .collect()
-                                } else {
-                                    rows.rows_content
-                                        .into_iter()
-                                        .map(|row| {
-                                            row.into_iter()
-                                                .enumerate()
-                                                .map(|(i, row_content)| {
-                                                    let col_spec = &rows.metadata.col_specs[i];
-                                                    MessageValue::build_value_from_cstar_col_type(
-                                                        frame.version,
-                                                        col_spec,
-                                                        &row_content,
-                                                    )
-                                                })
-                                                .collect()
-                                        })
-                                        .collect()
-                                };
+                                                }
+                                            })
+                                            .collect()
+                                    })
+                                    .collect()
+                            } else {
+                                rows.rows_content
+                                    .into_iter()
+                                    .map(|row| {
+                                        row.into_iter()
+                                            .enumerate()
+                                            .map(|(i, row_content)| {
+                                                let col_spec = &rows.metadata.col_specs[i];
+                                                MessageValue::build_value_from_cstar_col_type(
+                                                    frame.version,
+                                                    col_spec,
+                                                    &row_content,
+                                                )
+                                            })
+                                            .collect()
+                                    })
+                                    .collect()
+                            };
                             CassandraOperation::Result(CassandraResult::Rows {
                                 rows: converted_rows,
                                 metadata: Box::new(rows.metadata),
--- a/shotover-proxy/src/frame/cassandra.rs
+++ b/shotover-proxy/src/frame/cassandra.rs
@@ -205,45 +205,38 @@ impl CassandraFrame {
                 if let ResponseBody::Result(result) = frame.response_body()? {
                     match result {
                         ResResultBody::Rows(rows) => {
-                            let converted_rows = if rows
-                                .metadata
-                                .flags
-                                .contains(RowsMetadataFlags::NO_METADATA)
-                            {
-                                rows.rows_content
-                                    .into_iter()
-                                    .map(|row| {
-                                        row.into_iter()
-                                            .map(|row_content| {
-                                                if row_content.is_empty() {
-                                                    MessageValue::Null
-                                                } else {
+                            let converted_rows =
+                                if rows.metadata.flags.contains(RowsMetadataFlags::NO_METADATA) {
+                                    rows.rows_content
+                                        .into_iter()
+                                        .map(|row| {
+                                            row.into_iter()
+                                                .map(|row_content| {
                                                     MessageValue::Bytes(
                                                         row_content.into_bytes().unwrap().into(),
                                                     )
-                                                }
-                                            })
-                                            .collect()
-                                    })
-                                    .collect()
-                            } else {
-                                rows.rows_content
-                                    .into_iter()
-                                    .map(|row| {
-                                        row.into_iter()
-                                            .enumerate()
-                                            .map(|(i, row_content)| {
-                                                let col_spec = &rows.metadata.col_specs[i];
-                                                MessageValue::build_value_from_cstar_col_type(
-                                                    frame.version,
-                                                    col_spec,
-                                                    &row_content,
-                                                )
-                                            })
-                                            .collect()
-                                    })
-                                    .collect()
-                            };
+                                                })
+                                                .collect()
+                                        })
+                                        .collect()
+                                } else {
+                                    rows.rows_content
+                                        .into_iter()
+                                        .map(|row| {
+                                            row.into_iter()
+                                                .enumerate()
+                                                .map(|(i, row_content)| {
+                                                    let col_spec = &rows.metadata.col_specs[i];
+                                                    MessageValue::build_value_from_cstar_col_type(
+                                                        frame.version,
+                                                        col_spec,
+                                                        &row_content,
+                                                    )
+                                                })
+                                                .collect()
+                                        })
+                                        .collect()
+                                };
                             CassandraOperation::Result(CassandraResult::Rows {
                                 rows: converted_rows,
                                 metadata: Box::new(rows.metadata),
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
@@ -1,6 +1,7 @@
+use super::node::CassandraNode;
 use super::routing_key::calculate_routing_key;
 use super::token_map::TokenMap;
-use crate::transforms::cassandra::sink_cluster::node::CassandraNode;
+use super::KeyspaceChanRx;
 use anyhow::{anyhow, Error, Result};
 use cassandra_protocol::frame::message_execute::BodyReqExecuteOwned;
 use cassandra_protocol::frame::message_result::PreparedMetadata;
@@ -19,9 +20,15 @@ pub enum GetReplicaErr {
     Other(Error),
 }
 
+#[derive(Debug, Clone)]
+pub struct KeyspaceMetadata {
+    pub replication_factor: usize,
+}
+
 #[derive(Debug)]
 pub struct NodePool {
     prepared_metadata: Arc<RwLock<HashMap<CBytesShort, PreparedMetadata>>>,
+    keyspace_metadata: Arc<RwLock<HashMap<String, KeyspaceMetadata>>>,
     token_map: TokenMap,
     nodes: Vec<CassandraNode>,
     prev_idx: usize,
@@ -31,6 +38,7 @@ impl Clone for NodePool {
     fn clone(&self) -> Self {
         Self {
             prepared_metadata: self.prepared_metadata.clone(),
+            keyspace_metadata: self.keyspace_metadata.clone(),
             token_map: TokenMap::new(&[]),
             nodes: vec![],
             prev_idx: 0,
@@ -44,6 +52,7 @@ impl NodePool {
             token_map: TokenMap::new(nodes.as_slice()),
             nodes,
             prepared_metadata: Arc::new(RwLock::new(HashMap::new())),
+            keyspace_metadata: Arc::new(RwLock::new(HashMap::new())),
             prev_idx: 0,
         }
     }
@@ -68,6 +77,12 @@ impl NodePool {
         }
         self.nodes = new_nodes;
         self.token_map = TokenMap::new(self.nodes.as_slice());
+    }
+
+    pub async fn update_keyspaces(&mut self, keyspaces_rx: &mut KeyspaceChanRx) {
+        let updated_keyspaces = keyspaces_rx.borrow_and_update().clone();
+        let mut write_keyspaces = self.keyspace_metadata.write().await;
+        *write_keyspaces = updated_keyspaces;
     }
 
     pub async fn add_prepared_result(&mut self, id: CBytesShort, metadata: PreparedMetadata) {
@@ -121,7 +136,6 @@ impl NodePool {
         rack: &str,
         version: &Version,
         rng: &mut SmallRng,
-        rf: usize, // TODO this parameter should be removed
     ) -> Result<Option<&mut CassandraNode>, GetReplicaErr> {
         let metadata = {
             let read_lock = self.prepared_metadata.read().await;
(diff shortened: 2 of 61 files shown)


shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
```diff
@@ -131,19 +145,37 @@ impl NodePool {
                 .clone()
         };
 
+        let keyspace = {
```
conorbros, line 148: Need to handle this error better, we should just do normal routing if we don't have the keyspace info.

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
```diff
@@ -131,19 +145,37 @@ impl NodePool {
...
+                )
+                .ok_or(GetReplicaErr::NoMetadata)?
+                .clone()
+        };
+
         let routing_key = calculate_routing_key(
             &metadata.pk_indexes,
             execute.query_parameters.values.as_ref().ok_or_else(|| {
-                GetReplicaErr::Other(anyhow!("Execute body does not have query paramters"))
+                GetReplicaErr::Other(anyhow!("Execute body does not have query parameters"))
```
conorbros, line 167: Same as above.

shotover-proxy/src/transforms/cassandra/sink_cluster/token_map.rs
```diff
@@ -30,6 +30,14 @@ impl TokenMap {
             .take(replica_count)
             .map(|(_, node)| *node)
     }
+
+    // pub fn iter_replica_nodes(&self, token: Murmur3Token) -> impl Iterator<Item = Uuid> + '_ {
```
rukai, line 34: commented code

conorbros: Yeah I was just leaving that around in case we needed it but I'll remove it.

shotover-proxy/src/transforms/cassandra/sink_cluster/token_map.rs
```diff
@@ -47,7 +55,7 @@ mod test_token_map {
         vec![
             CassandraNode::new(
                 "127.0.0.1:9042".parse().unwrap(),
-                "rack1".into(),
+                "dc1".into(),
```
rukai, line 58: this arg is still for the rack right?

conorbros: Oh I don't know how I got that mixed up. Thanks

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -120,18 +150,31 @@ async fn topology_task_process(
...
                                     }
                                 }
+                                if let Err(watch::error::SendError(_)) =
+                                    nodes_tx.send(nodes.clone())
+                                {
+                                    break 'listen;
+                                }
+                            }
+                            ServerEvent::SchemaChange(change) => {
+                                tracing::warn!("{:?}", change);
```
rukai, line 160: stray log

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -106,9 +124,21 @@ async fn topology_task_process(
                                     }
 
                                     nodes = new_nodes;
+
+                                    if let Err(watch::error::SendError(_)) =
+                                        nodes_tx.send(nodes.clone())
+                                    {
+                                        break 'listen;
```
rukai, line 131: Can we just immediately `return Ok(())` instead of breaking out and then relying on the return that occurs later on?

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -183,6 +227,174 @@ async fn fetch_current_nodes(
...
+
+    pub async fn query(
+        connection: &CassandraConnection,
+        data_center: &str,
+    ) -> Result<HashMap<String, KeyspaceMetadata>> {
+        let (tx, rx) = oneshot::channel();
+
+        connection.send(Message::from_frame(
+            Frame::Cassandra(CassandraFrame{
+                version: Version::V4,
```
rukai, line 253: The topology task knows what version the protocol is using via the `let version = connection_info.connection_factory.get_version()?;` line.
We should use that version here.

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -183,6 +227,174 @@ async fn fetch_current_nodes(
...
+                stream_id: 0,
+                tracing: Tracing::Request(false),
+                warnings: vec![],
+                operation: CassandraOperation::Query{
+                    query: Box::new(
+                        parse_statement_single(
+
```
rukai, line 260: stray line?

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -183,6 +227,174 @@ async fn fetch_current_nodes(
...
+    fn into_keyspaces(
+        mut response: Message,
+        data_center: &str,
+    ) -> Result<HashMap<String, KeyspaceMetadata>> {
+        if let Some(Frame::Cassandra(frame)) = response.frame() {
+            match &mut frame.operation {
+                CassandraOperation::Result(CassandraResult::Rows { rows, .. }) => rows
+                    .iter_mut()
+                    .map(|row| build_keyspace(row, data_center))
+                    .try_collect(),
```
rukai, line 284: both of the `try_collect`s can just be `collect`s.
`try_collect` just helps with type inference sometimes and regular `collect` has its type inferred just fine here.

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -183,6 +227,174 @@ async fn fetch_current_nodes(
...
+            Frame::Cassandra(CassandraFrame{
+                version: Version::V4,
+                stream_id: 0,
+                tracing: Tracing::Request(false),
+                warnings: vec![],
+                operation: CassandraOperation::Query{
+                    query: Box::new(
+                        parse_statement_single(
+
+                            "SELECT keyspace_name, toJson(replication) AS replication FROM system_schema.keyspaces",
```
rukai, line 261: Why do we use toJson instead of just letting it send the map over the cassandra protocol?

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -183,6 +227,174 @@ async fn fetch_current_nodes(
...
+
+        Ok((name, metadata))
+    }
+
+    fn build_replication_strategy(
+        mut properties: Map<String, JsonValue>,
+    ) -> Result<ReplicationStrategy> {
+        match properties.remove("class") {
+            Some(JsonValue::String(class)) => Ok(match class.as_str() {
+                "org.apache.cassandra.locator.SimpleStrategy" | "SimpleStrategy" => {
```
rukai, line 346: Why are there 2 possible values for this?

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -183,6 +227,174 @@ async fn fetch_current_nodes(
...
+                    operation
+                )),
+            }
+        } else {
+            Err(anyhow!("Failed to parse keyspace query response"))
+        }
+    }
+
+    fn build_keyspace(
+        row: &mut Vec<MessageValue>,
```
rukai, line 296: If we take ownership of row here it alleviates the worry of "oh but what if someone tries to use row after calling this build_keyspace."

image

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -183,6 +227,174 @@ async fn fetch_current_nodes(
...
+        data_center: &str,
+    ) -> Result<(String, KeyspaceMetadata)> {
+        let metadata = if let Some(MessageValue::Varchar(string)) = row.pop() {
+            let replication: JsonValue = serde_json::from_str(&string).map_err(|error| {
+                anyhow!(format!(
```
rukai, line 301: anyhow will do its own formatting (do a quick search for other cases too)

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -183,6 +227,174 @@ async fn fetch_current_nodes(
...
+                    )))
+                }
+            };
+
+            let replication_factor = match replication_strategy {
+                ReplicationStrategy::SimpleStrategy { replication_factor } => replication_factor,
+                ReplicationStrategy::NetworkTopologyStrategy {
+                    datacenter_replication_factor,
+                } => *datacenter_replication_factor.get(data_center).unwrap_or(&0),
+                _ => 0,
```
rukai, line 322: Should this be 1 instead of 0?
I imagine 1 means the data is stored in one location?

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -183,6 +227,174 @@ async fn fetch_current_nodes(
...
+                extract_replication_factor(Some(&replication_factor))
+                    .map(move |replication_factor| (key, replication_factor))
+            })
+            .try_collect()
+    }
+
+    fn extract_replication_factor(value: Option<&JsonValue>) -> Result<usize> {
+        match value {
+            Some(JsonValue::String(replication_factor)) => {
+                let result = if let Some(slash) = replication_factor.find('/') {
```
rukai, line 380: Is there any documentation on this `/` format for the replication_factor?

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -183,6 +227,174 @@ async fn fetch_current_nodes(
...
+                    usize::from_str(&replication_factor[..slash])
+                } else {
+                    usize::from_str(replication_factor)
+                };
+
+                result.map_err(|error| {
```
rukai, line 386: This should use context 
image

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -183,18 +226,187 @@ async fn fetch_current_nodes(
...
+                        )?,
+                    }
+                }
+                "org.apache.cassandra.locator.NetworkTopologyStrategy"
+                | "NetworkTopologyStrategy" => ReplicationStrategy::NetworkTopologyStrategy {
+                    datacenter_replication_factor: extract_datacenter_replication_factor(
+                        properties,
+                    )?,
+                },
+                _ => ReplicationStrategy::Other,
```
rukai, line 358: Is this case just taken when we havent implemented support for a replication strategy yet?
If thats the case we should emit an `error!` and rename `ReplicationStrategy::Other` to `ReplicationStategy::Unknown`

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -205,38 +205,45 @@ impl CassandraFrame {
...
+                                .metadata
+                                .flags
+                                .contains(RowsMetadataFlags::NO_METADATA)
+                            {
+                                rows.rows_content
+                                    .into_iter()
+                                    .map(|row| {
+                                        row.into_iter()
+                                            .map(|row_content| {
+                                                if row_content.is_empty() {
```
rukai, line 218: Can you explain this change?
Shouldnt Null be handled at a lower level?

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
rukai: The answer is that because the replication factor is per dc, we can have a replication factor of 0 in one dc while having a non-zero replication factor in other dc's, so 0 does make semantic sense.

I''m still a little concerned about this line for other reasons but that is addressed in https://github.com/shotover/shotover-proxy/pull/895/files#r1012544156

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
conorbros: I checked driver implementations and noticed they did this, I'd imagine there is some cases where we don't get the fully qualified name.

shotover-proxy/src/frame/cassandra.rs
conorbros: The Datastax driver doesn't seem to let you create unbound prepared statements and this causes that error (tries to unwrap on a null value). I'm seeing if I can fix it by finishing off the scylla test helpers.

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -183,18 +226,147 @@ async fn fetch_current_nodes(
...
+        connection.send(Message::from_frame(
+            Frame::Cassandra(CassandraFrame{
+                version,
+                stream_id: 0,
+                tracing: Tracing::Request(false),
+                warnings: vec![],
+                operation: CassandraOperation::Query{
+                    query: Box::new(
+                        parse_statement_single(
+                            "SELECT keyspace_name, replication AS replication FROM system_schema.keyspaces",
```
rukai, line 249: `replication AS replication` -> `replication`

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
rukai: Closing and reopening on a more recent line

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -183,18 +226,147 @@ async fn fetch_current_nodes(
...
+                    KeyspaceMetadata {
+                        replication_factor: data_center_rf,
+                    }
+                }
+                "org.apache.cassandra.locator.LocalStrategy" | "LocalStrategy" => {
+                    KeyspaceMetadata {
+                        replication_factor: 0,
+                    }
+                }
+                _ => KeyspaceMetadata {
```
rukai, line 339: If strategies are exhaustive we should return an Err here instead.
If they are not exhaustive because users can make custom ones or something, then we should log an error! here and probably return a replication_factor of 1 as that sounds like a safer default?

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -183,18 +226,147 @@ async fn fetch_current_nodes(
...
+                        None => 0,
+                    };
+
+                    KeyspaceMetadata {
+                        replication_factor: data_center_rf,
+                    }
+                }
+                "org.apache.cassandra.locator.LocalStrategy" | "LocalStrategy" => {
+                    KeyspaceMetadata {
+                        replication_factor: 0,
```
rukai, line 336: a constant of 0 cant be right?

shotover-proxy/tests/cassandra_int_tests/prepared_statements.rs
```diff
@@ -102,6 +102,12 @@ async fn use_statement(session: &CassandraConnection) {
     .await;
 }
 
+async fn hardcoded_prepared(session: &CassandraConnection) {
```
rukai, line 105: This is testing a bug fix that is separate to the main goal of using replication_factor in our routing, can this test and the bug fix be moved into a separate PR?

conorbros: As part of implementing the keyspace info I added error handling which fixes that bug so I included that in this PR.

benbromhead (approved):

shotover-proxy/src/transforms/cassandra/sink_cluster/topology.rs
rukai: I tracked the source down to this https://github.com/scylladb/scylla-rust-driver/pull/327 but I cant see an appropriate justification for the change.
I also experimented locally with both cassandra and scylla, and I cant find a way to make `system_schema.keyspaces` output in that shortened form.

Unless you have any ideas as to why its there, shall we remove it?

rukai: Actually, it seems like it gets output in that format when `DESCRIBE KEYSPACE` is used.
Not sure if thats actually possible anymore, it seems to be deprecated and/or DSE specific but its probably worth keeping the shorter form around just in case the code gets repurposed for that use in the future.

shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
```diff
@@ -15,13 +16,20 @@ use tokio::sync::{watch, RwLock};
...
 
+#[derive(Debug, Clone, PartialEq)]
+pub struct KeyspaceMetadata {
+    pub replication_factor: usize,
+}
+
 #[derive(Debug)]
 pub struct NodePool {
     prepared_metadata: Arc<RwLock<HashMap<CBytesShort, PreparedMetadata>>>,
+    keyspace_metadata: Arc<RwLock<HashMap<String, KeyspaceMetadata>>>,
```
rukai, line 32: prepared_metadata uses an `Arc<RwLock<T>>` because each connection needs to be able to mutate a global state because preparation can occur on any connection.
However that same reasoning does not apply to `keyspace_metadata`.
Each connection receives the keyspace_metadata from the topology task and has no need to share this state with other connections because they all received the same state.
Changing this type to a `HashMap<String, KeyspaceMetadata>` should be sufficient to resolve this issue.

shotover-proxy/tests/cassandra_int_tests/prepared_statements.rs
```diff
@@ -119,4 +119,10 @@ where
     select_cross_connection(session, connection_creator).await;
     delete(session).await;
     use_statement(session).await;
+
+    if session.is(&[CassandraDriver::Scylla, CassandraDriver::CdrsTokio]) {
```
rukai, line 123: The tests pass just fine on my machine when I remove this check.
Did something else incidentally fix the issue we were having with the datastax driver?

rukai (approved):
LGTM

PR #896, opened 2022-11-02
Add verify_hostname tls flag

rukai:
~~Redis does not allow verify_hostname at all.
Cassandra does allow verify_hostname but some users may wish to disable it.
To support both of these uses an optional verify_hostname field is added to the tls connector config.
It has the safest default value for each protocol and extra validation to ensure that redis cannot be configured to use a setting that it is not compatible with.~~

Redis and cassandra both support verify_hostname but some users may wish to disable it if their setup does not permit it.


rukai:
Actually having tried it just now in our integration tests, redis doesnt seem to have any hard requirement forcing verify_hostname being disabled.
I'll change it from `Option<bool>` to `bool`, removing the default value and redis validation.

conorbros (approved):

XA21X (approved):

PR #897, opened 2022-11-03
Update to rust 1.65

rukai:
I decided to disable the new `clippy::bool_to_int_with_if` lint as IMO it makes the code less clear, as its not obvious what integer values correspond to true/false.
Happy to enable it if people disagree though.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.64"
+channel = "1.65"
 components = [ "rustfmt", "clippy" ]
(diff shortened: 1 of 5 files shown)


XA21X:
In the [simple example in the docs](https://rust-lang.github.io/rust-clippy/master/#bool_to_int_with_if), I do prefer what the lint suggests, but it could be situational. 🤷

XA21X (approved):

conorbros (approved):
I agree with @XA21X, but not fussed about it.

rukai:
Thanks for the feedback, I enabled the lint

PR #900, opened 2022-11-04
Refactor TCP connection logic

rukai:
prereq to https://github.com/shotover/shotover-proxy/pull/899

Every connection needs this same basic error handling so we should move it into a helper to give us consistent error messages and to ensure we never forget to implement the timeout.
It will also allow us to call this logic from within the tls connector in 899.

### DIFF
--- a/shotover-proxy/src/tcp.rs
+++ b/shotover-proxy/src/tcp.rs
@@ -1,0 +1,19 @@
+use anyhow::{anyhow, Result};
+use std::time::Duration;
+use tokio::{
+    net::{TcpStream, ToSocketAddrs},
+    time::timeout,
+};
+
+pub async fn tcp_stream<A: ToSocketAddrs + std::fmt::Debug>(destination: A) -> Result<TcpStream> {
+    timeout(Duration::from_secs(3), TcpStream::connect(&destination))
+        .await
+        .map_err(|_| {
+            anyhow!(
+                "destination {destination:?} did not respond to connection attempt within 3 seconds"
+            )
+        })?
+        .map_err(|e| {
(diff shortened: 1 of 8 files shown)


shotover-proxy/src/tcp.rs
```diff
@@ -0,0 +1,19 @@
+use anyhow::{anyhow, Result};
+use std::time::Duration;
+use tokio::{
```
conorbros, line 3: Could we rename `tls.rs` to `transport.rs` and put all the TCP/TLS stuff in there?

rukai: IMO tls has enough going on to be its own module.
Would you want to do transport/tls and trasnsport/tcp ?

conorbros: No not that fussed. Just a thought. Can merge as is.

shotover-proxy/src/tcp.rs
```diff
@@ -0,0 +1,19 @@
...
+    net::{TcpStream, ToSocketAddrs},
+    time::timeout,
+};
+
+pub async fn tcp_stream<A: ToSocketAddrs + std::fmt::Debug>(destination: A) -> Result<TcpStream> {
+    timeout(Duration::from_secs(3), TcpStream::connect(&destination))
```
benbromhead, line 9: Given this is a refactor, we should probably address making this timeout configurable with a 3 sec default. I can't imagine why we would need to longer, but shorter might be appropriate for some cases.

rukai: That is quite a large change so I've done it in a separate PR with this PR as a prereq: https://github.com/shotover/shotover-proxy/pull/901

benbromhead (approved):
LGTM

PR #902, opened 2022-11-07
Scylla Driver Tests

conorbros:
Finish implementing the Scylla driver test helpers and add it to the test suite.

### DIFF
--- a/shotover-proxy/tests/helpers/cassandra.rs
+++ b/shotover-proxy/tests/helpers/cassandra.rs
@@ -307,13 +307,7 @@ impl CassandraConnection {
                         .map(|x| {
                             x.columns
                                 .into_iter()
-                                .map(|col| {
-                                    if let Some(col) = col {
-                                        ResultValue::new_from_scylla(col)
-                                    } else {
-                                        ResultValue::Null
-                                    }
-                                })
+                                .map(ResultValue::new_from_scylla)
                                 .collect()
                         })
                         .collect(),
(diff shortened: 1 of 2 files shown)


shotover-proxy/tests/helpers/cassandra.rs
```diff
@@ -498,7 +508,23 @@ impl CassandraConnection {
 
                 Self::process_cdrs_response(response)
             }
-            Self::Scylla { .. } => todo!(),
+            Self::Scylla { session, .. } => {
+                let statement = prepared_query.as_scylla();
+                let response = session.execute(statement, (value,)).await.unwrap();
+
+                if let Ok(rows) = response.rows() {
```
rukai, line 515: We are silently ignoring the error here, while the other implementations panic on error here.

shotover-proxy/tests/helpers/cassandra.rs
```diff
@@ -498,7 +508,23 @@ impl CassandraConnection {
...
+                    rows.into_iter()
+                        .map(|row| {
+                            row.columns
+                                .into_iter()
+                                .map(|col| ResultValue::new_from_scylla(col.unwrap()))
```
rukai, line 520: we need to handle null value instead of unwrap here, I suggest changing new_from_scylla to take `Option<Value>` so that we never forget to do the null handling?

shotover-proxy/tests/helpers/cassandra.rs
conorbros: https://docs.rs/scylla/latest/scylla/transport/query_result/struct.RowsExpectedError.html

We are ignoring this error, it only occurs when the rows are empty.

rukai: Oh good point.
We can make the code less suspicious looking by directly accessing the rows field which is an Option.
image

rukai (approved):
Nice!
Its great to finally have most of our tests running against more than one driver. :tada:

benbromhead (approved):

PR #909, opened 2022-11-10
Add warning to rewritten queries with WHERE

rukai:
Supercedes https://github.com/shotover/shotover-proxy/pull/779

There has been a fair bit of confusion over the behavior of these tables with WHERE clauses and these warnings should make it obvious to the user where shotovers handling falls short.
I also made the final warnings list include all warnings from extra queries that were sent in order to generate the final query.

~~Currently this new warning produces a lot of noise in our integration tests as the drivers do use a WHERE clause for `system.local`.
I think the correct fix for this noise would be to specifically handle certain WHERE clauses that drivers are known to make.
Probably want to delay merging this until we have a fix for that though.~~
I fixed the noise by adding specific logic to ignore the common `FROM system.local WHERE key = 'local'` thing that drivers keep inexplicably doing.
e.g. here is the PR that adds the query to scylla-rust-driver https://github.com/scylladb/scylla-rust-driver/pull/223


shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -604,25 +604,37 @@ impl CassandraSinkCluster {
...
                     } else {
                         return None;
                     };
 
+                    let warnings = if Self::has_no_where_clause(ty, select) {
+                        vec![]
+                    } else {
+                        select.where_clause.clear();
+                        vec![format!(
+                            "WHERE clause on the query was ignored. Shotover does not support WHERE clauses on queries against {}",
```
XA21X, line 629: Do we know of any concrete situations (specific drivers) where this can be triggered for `system.peers`? Is it more important to try and return something, versus failing early like the TODO said, to minimise the risk of incorrect behaviour?

XA21X (approved):

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
rukai: I do not know of any concrete instances where this can be triggered for system.peers.

ben did voice the opinion that we would be better off returning everything and hope the driver gets what it needs: https://github.com/shotover/shotover-proxy/pull/779#issuecomment-1233697939

Personally I would return an error here, but I suspect that I tend to favor correctness over compatibility a bit too much.

conorbros (approved):

PR #912, opened 2022-11-14
Isolate shotovers tokio runtime in integration tests

rukai:
Makes integration tests more robust by isolating:
* Shotover async runtime from test runtime
* Shotover async runtime from other shotover runtimes created at the same time or afterwards

It also simplifies our test helper logic.

An immediate practical benefit is removing the `ERROR shotover_proxy::observability: Metrics HTTP server failed: Failed to bind to 0.0.0.0:9001` error log clutter that occurs every time we create a second Shotover instance after destroying the previous shotover instance.

### DIFF
--- a/shotover-proxy/src/runner.rs
+++ b/shotover-proxy/src/runner.rs
@@ -8,7 +8,7 @@ use clap::{crate_version, Parser};
 use metrics_exporter_prometheus::PrometheusBuilder;
 use std::env;
 use std::net::SocketAddr;
-use tokio::runtime::{self, Handle as RuntimeHandle, Runtime};
+use tokio::runtime::{self, Runtime};
 use tokio::signal::unix::{signal, SignalKind};
 use tokio::sync::watch;
 use tokio::task::JoinHandle;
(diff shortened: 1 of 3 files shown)


rukai:
Oooh the `redis_int_tests::cluster_dr` failure actually reveals a bug in shotover that was hidden before.

rukai:
There are some foundational shotover fixes we need to land this:
* pass flush flags to inner chains - needed to flush the coalesce transform
* split inner chains into buffered and non buffered variants - needed to wait for inner chain to complete before shutting down.

I have WIP branches for both of these.

rukai:
Although the plan is to eventually remove ShotoverManager, replacing it with ShotoverProcess, it makes sense to land this PR now so that in the meantime we can ensure we dont regress on any of the work that it took to make the shutdown process good enough to make the test pass.

rukai:
Superceded by https://github.com/shotover/shotover-proxy/pull/976 and https://github.com/shotover/shotover-proxy/pull/975
They will provide isolation to the majority of tests, so point in landing this anymore.

PR #913, opened 2022-11-15
Cassandra connection avoid cloning entire message

rukai:
The connection logic was previously cloning the entire message when it really needed was the stream_id and version fields.
I removed the message clone and changed the connection logic to get the values from:
* stream_id - Sent to the rx task where the original message used to be.
* version - The transform is now responsible for keeping track of the version. Its constant per connection, so we can just set it once from the first message and it will always be the same.

This is a prereq for cassandra error handling improvements.


shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -229,6 +232,18 @@ fn create_query(messages: &Messages, query: &str, version: Version) -> Result<Me
 
 impl CassandraSinkCluster {
     async fn send_message(&mut self, mut messages: Messages) -> ChainResponse {
+        if let Some(message) = messages.first() {
```
conorbros, line 235: Does this happen for every message? I think we only need to set this once. Once a Cassandra connection starts up the version won't change.

conorbros: Perhaps the version field should be an option and we check if it's none before setting it.

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
```diff
@@ -229,6 +232,18 @@ fn create_query(messages: &Messages, query: &str, version: Version) -> Result<Me
...
+            if let Ok(Metadata::Cassandra(CassandraMetadata { version, .. })) = message.metadata() {
+                self.version = version;
+            } else {
+                return Err(anyhow!(
+                    "Failed to extract cassandra version from incoming message: Not a valid cassandra message"
+                ));
+            }
+        } else {
+            return Ok(vec![]);
```
conorbros, line 244: Is this an early return if there is no messages?

rukai: yeah

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
rukai: yeah, thats reasonable, I'll change the field to an Option.

shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
conorbros: Wouldn't we want to run the checks for updating keyspaces and node list regardless? Though I can't think of a situation where a transform chain would run without any messages, seems reasonable to run the update stuff anyway.

conorbros: Ah very true, good catch. Maybe make a note of it?

conorbros (approved):

XA21X (approved):

PR #918, opened 2022-11-17
Token aware fix

conorbros:
We only want to select one replica from each rack to use as a potential choice for token aware routing.

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
@@ -8,6 +8,7 @@ use cassandra_protocol::frame::message_result::PreparedMetadata;
 use cassandra_protocol::frame::Version;
 use cassandra_protocol::token::Murmur3Token;
 use cassandra_protocol::types::CBytesShort;
+use itertools::Itertools;
 use rand::prelude::*;
 use split_iter::Splittable;
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


shotover-proxy/src/transforms/cassandra/sink_cluster/node_pool.rs
```diff
@@ -171,12 +172,14 @@ impl NodePool {
...
                 keyspace.replication_factor,
             )
+            .unique()
             .collect::<Vec<uuid::Uuid>>();
 
         let (dc_replicas, rack_replicas) = self
             .nodes
             .iter_mut()
             .filter(|node| replica_host_ids.contains(&node.host_id) && node.is_up)
+            .unique_by(|node| node.rack.clone())
```
rukai, line 182: Wouldnt this result in rack_replicas always containing 0 or 1 elements?

conorbros: Yes

rukai: In that case should we treat rack_replicas as having one value?
image

rukai:
> We only want to select one replica from each rack to use as a potential choice for token aware routing.

I think we need more justification for this statement.
Is this what other drivers do?
Is it how cassandra token aware routing is designed to work as per some spec?

Like off the top of my head routing between multiple nodes in a single rack seems like a reasonable thing to do?

conorbros:
Yes the other drivers (scylla at least) will select one replica from each rack. I think it makes more sense to do that so if we don't find one in the same rack we are spreading the load better across the other racks.

benbromhead:
How are we placed on resolving this?

conorbros:
Looking into this in the improved routing issue. (best of two random choices algorithm)

PR #921, opened 2022-11-18
cassandra_int_tests test node up

rukai:
Prereq: https://github.com/shotover/shotover-proxy/pull/917

Instead of changing tmpfs to volume, we could make a separate cassandra-cluster-v4/docker-compose.yaml for persistent tests but it doesn't seem worth it yet.

The `cassandra.yaml` changes are needed to make the cassandra nodes persistent on restart, not sure why they fix the problem but I have verified that they do not cause any regressions in test runtime.

The new integration test discovered a bug in `node_pool.rs` where we were accidentally copying over dead connections, so this PR fixes that as well.

### DIFF
--- a/test-helpers/src/docker_compose.rs
+++ b/test-helpers/src/docker_compose.rs
@@ -106,6 +106,17 @@ impl DockerCompose {
         .unwrap();
     }
 
+    /// Restarts the container with the provided service name
+    pub fn start_service(&self, service_name: &str) {
+        run_command(
+            "docker-compose",
+            &["-f", &self.file_path, "start", service_name],
+        )
+        .unwrap();
+
+        // TODO: call wait_for_containers_to_startup
+    }
+
     fn wait_for_containers_to_startup(&self) {
         match self.file_path.as_ref() {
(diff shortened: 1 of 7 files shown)


test-helpers/src/docker_compose.rs
```diff
@@ -106,6 +106,17 @@ impl DockerCompose {
...
 
+    /// Restarts the container with the provided service name
+    pub fn start_service(&self, service_name: &str) {
+        run_command(
+            "docker-compose",
+            &["-f", &self.file_path, "start", service_name],
+        )
+        .unwrap();
+
+        // TODO: call wait_for_containers_to_startup
```
conorbros, line 117: What's this TODO?

rukai: I am purposefully leaving it undone.
it needs https://github.com/shotover/shotover-proxy/pull/917 to land before it can be done.
And it turns out I didnt need it for this PR because the test logic has its own delay (receiving the cassandra UP event)
But I think the start_service method should eventually still have its own in built delay.

conorbros (approved):

XA21X (approved):

Issue #905, opened 2022-11-09
cassandra results NO_METADATA flag bugs


rukai:
I know that at least the datastax cpp driver will generate messages with this flag.

Currently there is a panic preventing us from enabling a prepared statement test case in `prepared_statements_simple.rs` due to a failure in `cassandra_int_tests::cluster_single_rack_v3::case_1_datastax` due to a null value in a no_metadata message.

~~We also completely fail to reencode no_metadata messages, everything turns into a Blob.~~
Further inspection and integration test have proved this wrong.

PR #928, opened 2022-11-21
Implement proper NO_METADATA support for cassandra protocol

rukai:
closes https://github.com/shotover/shotover-proxy/issues/905

I originally thought this would be more involved because I thought the encode logic needed to special case NO_METADATA just like we do for decoding.
But it turns out that the standard implementation for encoding null and bytes data types works just fine with the null and bytes types that we generate for NO_METADATA decoding.

To prove this I added `DebugForceEncode` to cassandra-cluster-v4.
I think its fine to add this to an existing test because we have the very similar cassandra-cluster-multi-rack test case which will test CassandraSinkCluster without the DebugForceEncode.

### DIFF
--- a/shotover-proxy/src/frame/cassandra.rs
+++ b/shotover-proxy/src/frame/cassandra.rs
@@ -224,10 +224,11 @@ impl CassandraFrame {
                                         .into_iter()
                                         .map(|row| {
                                             row.into_iter()
-                                                .map(|row_content| {
-                                                    MessageValue::Bytes(
-                                                        row_content.into_bytes().unwrap().into(),
-                                                    )
+                                                .map(|row_content| match row_content.into_bytes() {
+                                                    None => MessageValue::Null,
+                                                    Some(value) => {
+                                                        MessageValue::Bytes(value.into())
+                                                    }
                                                 })
                                                 .collect()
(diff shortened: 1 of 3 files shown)


shotover-proxy/src/frame/cassandra.rs
```diff
@@ -224,10 +224,11 @@ impl CassandraFrame {
                                         .into_iter()
                                         .map(|row| {
                                             row.into_iter()
-                                                .map(|row_content| {
-                                                    MessageValue::Bytes(
-                                                        row_content.into_bytes().unwrap().into(),
-                                                    )
+                                                .map(|row_content| match row_content.into_bytes() {
```
conorbros, line 227: This seems familiar. Did we make another change somewhere else similar to this?

conorbros (approved):

shotover-proxy/src/frame/cassandra.rs
rukai: I believe you attempted to make this change (or maybe something that was just similar?) in another PR and I asked it to be removed because I didnt yet understand the NO_METADATA scenario.

benbromhead (approved):

PR #930, opened 2022-11-22
add message_to_high_level_string debug helper

rukai:
Allows quickly debugging including high level message representations in debug logs when debugging transforms.

For example now we can just write:
```rust
tracing::error!("messages: {}", message_wrapper.messages_to_high_level_string());
```

Instead of:
```rust
tracing::error!(
    "messages: {}",
    message_wrapper.
            .messages
            .iter_mut()
            .map(|x| x.to_high_level_string())
            .collect::<Vec<_>>()
);
```

Starting the method name with `messages_` means that the method will be suggested when the user writes: 
`tracing::error!("messages: {}", message_wrapper.messages`

### DIFF
--- a/shotover/src/transforms/mod.rs
+++ b/shotover/src/transforms/mod.rs
@@ -391,6 +391,18 @@ impl<'a> Wrapper<'a> {
         }
     }
 
+    /// Helper for writing debug logs while testing
+    /// feature flagged behind alpha-transforms as this call is too expensive to ever make it into production
+    #[cfg(feature = "alpha-transforms")]
+    pub fn messages_to_high_level_string(&mut self) -> String {
+        let messages = self
+            .messages
+            .iter_mut()
+            .map(|x| x.to_high_level_string())
+            .collect::<Vec<_>>();
+        format!("{:?}", messages)
+    }
+
     pub fn reset(&mut self, transforms: &'a mut [Transforms]) {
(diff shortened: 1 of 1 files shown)


shotover-proxy/src/transforms/mod.rs
```diff
@@ -512,6 +512,15 @@ impl<'a> Wrapper<'a> {
         }
     }
 
+    pub fn messages_to_high_level_string(&mut self) -> String {
```
conorbros, line 515: Can we put `#[cfg(test)]` on it? Make it clear you shouldn't be using this outside of a debugging/test context.

conorbros (approved):
Great idea. LGTM

shotover-proxy/src/transforms/mod.rs
rukai: good idea.
done!

rukai:
Oh wait, cant use #[test] because that can only be used in actual test modules.

rukai:
I swapped it to use `#[cfg(feature = "alpha-transforms")]` instead of `#[cfg(test)]` 
`#[cfg(test)]` is no good because its inaccessible from within shotover during an integration test.
but alpha-transforms is a reasonable way to ensure we dont accidentally check in debug prints that use this

benbromhead (approved):

PR #931, opened 2022-11-22
Shutdown span

rukai:
Transform logs now get a shutdown span included to help distinguish from logs that correspond to a specific connection.
image
The remaining logs without spans are due to a bug in propogating spans across tee'd transforms.

`tokio::spawn` is internally just `Handle::current().spawn`


rukai:
Actually we should probably just delete this logic entirely, its run against the prototype chain from which real chains are cloned from, so no messages are ever sent against this transform anyway.

rukai:
Superceded by https://github.com/shotover/shotover-proxy/pull/933

PR #935, opened 2022-11-23
Split transforms into Transform and TransformBuilder stages

rukai:
We now have the following stages of transform creation

1. TransformNameConfig - raw deserialization of the yaml config
2. TransformNameBuilder - a transform that is built almost to the point of completion but anything incoming connection specific is left out.
3. TransformName - a completely built transform, messages can be sent through it now

To support this I also had to introduce a TransformChainBuilder type which is both used used to hold Builder stage transforms and used to store subchains in builder stage transforms.

Many transforms can get away with sharing the TransformName and TransformNameBuilder as one type because there is no further processing needed other than a clone.
We may wish to more strictly differentiate these stages in the future but leaving it as is for now helps keep the PR smaller.

The advantages of this approach are:
* Makes a clear separation of which chains are used as builders and which have actual messages sent through them
   + I remember when I first started I was constantly being tripped up by the fact that the transform clone implementation is effectively a `build()` method
   + heres an example of a bug caused by this confusion: https://github.com/shotover/shotover-proxy/pull/933
* Remove extra chain clones
   + only the builder types need to hold onto the `original_chain`s

Followup cleanups:
* properly move validate and is_terminating to TransformBuilder, maybe we want a TransformBuilder trait for this?
* Remove all custom transform clone implementations, instead put that logic in a Builder build method
* remove extra chain clones + remove BufferedChain::to_new_instance


benbromhead (approved):
Thanks this is a much needed improvement in clarity around chain creation

conorbros (approved):

PR #936, opened 2022-11-24
Fix propagating tracing spans into subchains

rukai:
The task for the Tee transforms BufferedChain is generated during the `.clone_with_pushed_messages_tx(pushed_messages_tx.clone()),` line (previously line 215)
By extending the async instrument call to run an async scope that includes that line, we include the task within our span, therefore improving our tracing results.

Sample without this PR:
image

Sample with this PR:
image

Still a few cases where its missing that will need to be investigated separately.


shotover-proxy/src/server.rs
```diff
@@ -183,75 +183,76 @@ impl<C: Codec + 'static> TcpCodecListener<C> {
...
+            let span = tracing::error_span!(
+                "connection",
+                id = self.connection_count,
+                source = self.source_name.as_str(),
             );
+            async {
+                // Accept a new socket. This will attempt to perform error handling.
+                // The `accept` method internally attempts to recover errors, so an
+                // error here is non-recoverable.
+                let socket = self.accept().await.unwrap();
```
rukai, line 196: Oops, I'll need to handle this

conorbros (approved):

benbromhead (approved):

PR #938, opened 2022-11-24
Timestamp tests

conorbros:
Timestamps in Cassandra can be included in the query parameters or inside the CQL. We test that they are supported in Shotover by sending a query with a timestamp included and test that this matches the last written timestamp on the row's column.

### DIFF
--- a/shotover-proxy/tests/cassandra_int_tests/timestamp.rs
+++ b/shotover-proxy/tests/cassandra_int_tests/timestamp.rs
@@ -1,0 +1,93 @@
+use crate::helpers::cassandra::{assert_query_result, run_query, CassandraConnection, ResultValue};
+use std::time::{SystemTime, UNIX_EPOCH};
+
+async fn flag(connection: &CassandraConnection) {
+    let timestamp = setup(connection).await;
+
+    let _r = connection
+        .execute_with_timestamp(
+            "UPDATE test_timestamps.test_table SET a = 'a1-modified-1' WHERE id = 0;",
+            timestamp,
+        )
+        .await;
+
+    assert_query_result(
+        connection,
+        "SELECT WRITETIME(a) FROM test_timestamps.test_table WHERE id = 0;",
+        &[&[ResultValue::BigInt(timestamp)]],
+    )
+    .await;
+}
+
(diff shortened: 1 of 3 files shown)


shotover-proxy/tests/cassandra_int_tests/timestamp.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    assert_query_result(
+        connection,
+        "SELECT WRITETIME(a) FROM test_timestamps.test_table WHERE id = 0;",
+        &[&[ResultValue::BigInt(timestamp)]],
+    )
+    .await;
+}
+
+async fn setup(connection: &CassandraConnection) -> i64 {
```
rukai, line 42: I dont really understand why we use this to generate a timestamp rather than just hardcoding a timestamp?

conorbros: Because if we send one with an earlier timestamp than Cassandra has as the last write (the insert) it will be ignored.

rukai: Oooh, thats a good reason to do it like this, can you add a comment?

Also worth considering if we can just use the current system time, the only reason I think that could go wrong is if there is a risk of it ending up in what cassandra considers the future.

Lets also rename `setup` -> `get_timestamp`

rukai: Changes look good, my only remaining question is why not just use the current system time?

shotover-proxy/tests/cassandra_int_tests/timestamp.rs
```diff
@@ -0,0 +1,77 @@
+use crate::helpers::cassandra::{assert_query_result, run_query, CassandraConnection, ResultValue};
+use std::time::{SystemTime, UNIX_EPOCH};
+
+async fn flag(connection: &CassandraConnection) {
+    let timestamp = get_timestamp();
+
+    let _r = connection
```
rukai, line 7: The `let _r = ` can be removed

rukai (approved):
nice!

benbromhead (approved):

PR #943, opened 2022-11-29
cassandra_int_tests: Refactor CassandraConnection execute methods

rukai:
CassandraConnection implementation changes:
* improves support for the scylla and cdrs-tokio driver backends, I think the only thing they are missing now is TLS support.
* greatly reduces duplicated logic across: batch, prepared and regular queries. image :eyes: 

CassandraConnection API changes:
* _fallible methods return more useful types:
    + On success they return the generic Vec of results instead of some cassandra_cpp specific type.
    + On failure they return the cdrs-tokio errors instead of cassandra_cpp errors because cdrs-tokio errors are nicer to construct and read from
* _expect_err methods are replaced with _fallible methods, this gives us a more composable API.
* The execute method is kept instead of just using execute_fallible because it provides a better error message on failure then just unwrap.
* The execute_batch method is kept instead of just using execute_batch_fallible because it provides better assertions.


conorbros (approved):

conorbros:
LGTM. Should probably rename those methods at some point, execute and execute prepared are a bit confusing.

benbromhead (approved):

PR #945, opened 2022-11-29
fix prepared results bug

conorbros:
The entire frame was being compared to check if all the prepared responses were the same and having the tracing flag set meant that all frames had different tracing ids. This PR only checks that the prepared metadata part of the response is used for comparison.


rukai (approved):
LGTM!

benbromhead (approved):

PR #960, opened 2022-12-13
Do not clone transforms when running chain in reverse

rukai:
The dangerous line is `let mut chain: Vec<_> = self.chain.iter().cloned().rev().collect();` in `transforms/chain.rs`.
This causes pushed messages to be run against a clone of the transforms so any state set when processing pushed messages is lost.
Currently I'm not aware of any bugs this has caused but its clearly dangerous.

This PR should also be a nice speedup for pushed messages as we no longer need to clone every transform for every message.
Performance for regular messages will not be affected in any meaningful way, just an extra branch per transform.


conorbros (approved):

benbromhead (approved):
Nice!

PR #973, opened 2022-12-20
Fix prepared metadata sharing between connections

rukai:
Previously prepared metadata was not being shared between incoming connections.
This was because we were not cloning the `Arc<Mutex<..>>` and instead creating a new one for each connection.
This PR fixes that by cloning the NodePool that contains the mutex in CassandraSinkClusters constructor.


benbromhead (approved):

rukai:
Superceded by https://github.com/shotover/shotover-proxy/pull/972

PR #988, opened 2023-01-17
Options rewrite

conorbros:
Shotover does not support compression in Cassandra yet.  Changes in this PR:

- Reject STARTUP messages if they attempt to use compression.
-  Add a transform for rewriting the SUPPORTED response to OPTIONS requests so users can disable compression

### DIFF
--- a/shotover-proxy/src/codec/cassandra.rs
+++ b/shotover-proxy/src/codec/cassandra.rs
@@ -63,7 +63,7 @@ impl Decoder for CassandraCodec {
                     );
 
                     let version = Version::try_from(bytes[0])
-                        .expect("Gauranteed because check_envelope_size only returns Ok if the Version will parse");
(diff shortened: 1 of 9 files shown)


rukai:
This PR contains two separate large new features, they should be landed in two separate PRs.

For the second feature I'm not quite sure what the use case is?

conorbros:
I'll split these into 2 PR's but address those comments here. 

Here's an example of a driver checking the OPTIONS response before connecting. 

https://github.com/scylladb/scylla-rust-driver/blob/ce9654473f713310290e42321d289daa6b3766ed/scylla/src/transport/connection.rs#L1210

Regardless it makes sense to not break the protocol, we should not be reporting compression supported if it's not. This is a temporary fix, the plan is to introduce full compression support.

conorbros:
Opened #989, #990, #991

PR #989, opened 2023-01-24
startup body support

conorbros:
Parsing Cassandra startup messages


rukai (approved):
This does add a bunch of extra parsing work when creating a new cassandra connection, but creating cassandra connections is so expensive the client needs to be conservative with them anyway.

Looks good!

benbromhead (approved):

PR #990, opened 2023-01-24
Startup compression reject

conorbros:
Reject STARTUP messages if they attempt to use compression.

### DIFF
--- a/shotover-proxy/src/codec/cassandra.rs
+++ b/shotover-proxy/src/codec/cassandra.rs
@@ -235,8 +235,10 @@ mod cassandra_protocol_tests {
         ColSpec, ColType, ColTypeOption, ColTypeOptionValue, RowsMetadata, RowsMetadataFlags,
         TableSpec,
     };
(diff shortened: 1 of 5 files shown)


benbromhead (approved):

shotover-proxy/src/codec/cassandra.rs
```diff
@@ -196,6 +205,26 @@ fn reject_protocol_version(version: u8) -> CodecReadError {
...
+fn reject_compression(compression: &String) -> CodecReadError {
+    info!(
+        "Rejecting compression option {} (configure the client to use no compression)",
+        compression
+    );
+
+    CodecReadError::RespondAndThenCloseConnection(vec![Message::from_frame(Frame::Cassandra(
+        CassandraFrame {
+            version: Version::V4,
+            stream_id: 0,
```
rukai, line 217: can we take this from the incoming message? The spec makes no guarantee that the first message has stream_id 0

rukai (approved):

PR #991, opened 2023-01-24
Options rewrite transform

conorbros:
Add a transform for rewriting the SUPPORTED response to OPTIONS requests so users can disable compression

### DIFF
--- a/shotover-proxy/src/codec/cassandra.rs
+++ b/shotover-proxy/src/codec/cassandra.rs
@@ -235,8 +235,10 @@ mod cassandra_protocol_tests {
         ColSpec, ColType, ColTypeOption, ColTypeOptionValue, RowsMetadata, RowsMetadataFlags,
         TableSpec,
     };
+    use cassandra_protocol::frame::message_startup::BodyReqStartup;
     use cassandra_protocol::frame::Version;
     use hex_literal::hex;
+    use std::collections::HashMap;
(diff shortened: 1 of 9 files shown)


benbromhead (approved):

rukai:
I dont think disabling compression is sufficient reason to introduce an entire new transform into our API.
And even then we cant expect the user to know that we dont support compression and so they must configure this undocumented transform in order to force the client to avoid compression.

If we must urgently implement this functionality before implementing proper compression support it should go into the protocol level or into our sink transforms.
I suspect we would be better off just implementing compression support though.

test-helpers/src/connection/cassandra.rs
```diff
@@ -179,6 +179,7 @@ impl CassandraConnection {
                     )
                     .user("cassandra", "cassandra")
                     .default_consistency(Consistency::One)
+                    .compression(Some(scylla::transport::Compression::Snappy))
```
rukai, line 182: I'm confused how this works?
Does the driver take this as a suggestion only and drop down to no compression when the DB doesnt support compression?

conorbros: Yes, if the cluster does not support any compression options the driver will default to none.

rukai: sounds good!

rukai (approved):
LGTM!

Issue #993, opened 2023-01-25
BinProcess gives poor errors when errors occur early during shotover startup.


rukai:
An easy way to reproduce is to make the config.yaml invalid:
![image](https://user-images.githubusercontent.com/5120858/214480462-82f2557e-e20a-4a18-bc51-5049b632aced.png)

And then run `cargo test cassandra_int_tests::passthrough`

The issue is caused by errors being bubbled all the way up past `fn main` and then relying on rusts post-main error handling logic to write the error to stdout.

BinProcess cant handle this kind of error, it can only process logs that have been emitted through tokio tracing.

To fix the issue we should write the error to `error!()` in main just before terminating and then return an `Err(())` such that we have a negative exit code but nothing is printed to stdout.
Might need to create a tokio runtime just to process the `error!`, but as long as we keep it simple it should be pretty hard to fail to start the runtime.

PR #999, opened 2023-01-30
Remove config.yaml

conorbros:
Removes the need for a config.yaml file

### DIFF
--- a/shotover-proxy/benches/benches/chain.rs
+++ b/shotover-proxy/benches/benches/chain.rs
@@ -5,8 +5,8 @@ use criterion::{criterion_group, BatchSize, Criterion};
 use hex_literal::hex;
 use shotover_proxy::frame::cassandra::{parse_statement_single, Tracing};
 use shotover_proxy::frame::RedisFrame;
-use shotover_proxy::frame::{CassandraFrame, CassandraOperation, Frame, MessageType};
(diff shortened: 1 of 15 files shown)


test-helpers/Cargo.toml
```diff
@@ -11,6 +11,7 @@ license = "Apache-2.0"
 cassandra-cpp-driver-tests = ["cassandra-cpp"]
 
 [dependencies]
+clap = { version = "4.0.4", features = ["cargo", "derive"] }
```
rukai, line 14: what is this for?

rukai:
this change alone wont be enough to resolve https://github.com/shotover/shotover-proxy/issues/993
The CLI args will still fail to parse before the tokio runtime is started.
So I dont think moving topology generation to runtime makes sense.

Instead I think we should have an error path that looks like:
```rust
fn Runner::new() -> Result<()> {
  if let Err(err) = Runner::new_inner() {
    create_trivial_tokio_runtime_just_for_tracing();
    error!("{:#?}", err.context("failed to init"));
    return Err(()); // or possibly just directly call exit(1)
  }
}
```

conorbros:
I'll update this PR to just remove the config file.

conorbros:
Hmm, will need to do a new release for this one. It breaks the docker example using the published DockerHub image.

PR #1005, opened 2023-01-31
prevent recompiling by properly specifying build.rs

rukai:
## Wins
This is a decent improvement to our local development cycle.
Previously running `cargo test` after modifying a test took 5.5s to build on my machine, with this PR it takes 3s to build.
I expect the improvement to be even more on machines with lower core counts than mine.

## Investigation
By looking at `cargo test --no-run --timings` before and after the PR we can see what changed:

Before:

After:

The build script i.e. build.rs was rerunning and this was forcing all of shotover-proxy to be rebuilt and any example binaries.
Afterwards only the test binary that we modified is being rebuilt.

## Solution
The problem with our current setup is that unless instructed otherwise cargo will rerun the build.rs everytime a single file is changed in the package.
The `shotover-proxy/build.rs` is no longer needed, I assume the user of the env var it emits was moved into `test-helpers` or `tokio-bin-process`, so I just removed it.
The other remaining build.rs files have been changed to declare the env var that they take in so that cargo doesnt default to rerunning every time.
see: https://doc.rust-lang.org/cargo/reference/build-scripts.html#change-detection


benbromhead (approved):
nice

conorbros (approved):

Issue #977, opened 2022-12-21
shotover is executing queries on nodes in other racks


yuzhouchen-instaclustr:
When running the Cassandra testing tool cassandra-stress, I can see shotover sending out EXECUTE requests to nodes in racks other than its own, which is not supposed to happen by design. 

The table I'm using has a composite primary key, so I also tested it with a single primary key, I can confirm I'm seeing the same phenomenon in both cases.

I've also tested this with my own Java client and I can see it happening as well, which means this unlikely a cassandra-stress issue, the Cassandra node in other racks being hit isn't necessarily a replica either.

This appears to be quite random, repeating the same query with the same PK does not always yield the same result (i.e. same trace)

rukai:
Assuming a properly generated token distribution: if the keyspace being queried has a replica count less than the number of nodes per rack then sometimes the destination will end up outside of the rack.
Can you confirm that the keyspace you are querying has a replica count higher than or equal to the number of nodes per rack?

rukai:
closed by https://github.com/shotover/shotover-proxy/pull/1384

PR #1012, opened 2023-02-02
Add counter for when CassandraSinkCluster routes execute message outside of rack

rukai:
My understanding of our routing model is:
Cassandra tokens are unique per cassandra instance, replicas are determined by proceeding to the next token along the token ring.
Shotover instances report their token list as the list of all tokens in their rack.
The client should then see this and always route to a shotover instance that contains the destination replica node in its rack by just performing token aware routing.
From there shotover just needs to find the correct destination node in its rack via token aware routing and it will have routed directly to a replica node that exists within its rack.

Following this model, the warning I have added should only occur according to the reasons listed in the warning.
I think its reasonable to report for the first reason, shotover is now routing in a degraded manner, and a node going down should be a rare event.
However if it triggers for the second reason then we have observed a bug that needs to be fixed.

I'm hoping this warning will help track down https://github.com/shotover/shotover-proxy/issues/977
But if you can spot a flaw in my reasoning then that will help track it down even faster.

### DIFF
--- a/shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
+++ b/shotover-proxy/src/transforms/cassandra/sink_cluster/mod.rs
@@ -133,7 +133,7 @@ impl CassandraSinkClusterBuilder {
         connect_timeout_ms: u64,
         timeout: Option<u64>,
     ) -> Self {
-        let failed_requests = register_counter!("failed_requests", "chain" => chain_name, "transform" => "CassandraSinkCluster");
+        let failed_requests = register_counter!("failed_requests", "chain" => chain_name.clone(), "transform" => "CassandraSinkCluster");
         let receive_timeout = timeout.map(Duration::from_secs);
         let connect_timeout = Duration::from_millis(connect_timeout_ms);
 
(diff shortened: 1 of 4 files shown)


benbromhead:
A node going down does happen on a regular basis, especially when you factor in maintenance events (upgrades, rolling restarts etc). This log message is likely to generate a large number of logging events when a C* node goes down (but doesn't leave the cluster). E.g. a 3 node cluster with 21k writes per second, would result in roughly 7k logging events per second when restarting a C* node.

We should either look to make this a counter, rather than a logging event, or if it needs to be a logged event, make it trace. My feel here would be a metrics counter, that way we can still determine if we are constantly routing outside a rack during certain times. We are also not logging any details about the routing error, so a count also makes sense in that regard... we just need to know its happening and how often.

rukai:
Thanks for the feedback, that makes a lot of sense, I think using a metric is the best solution so I've redone this PR to do that.
Users can still access the data when investigating and we can still still assert on the metric never activating in our tests. (even if we have to do so manually)

Previously with the `warn!` our test infrastructure would fail a test if that warn ever occurred in an integration test.
We dont have any such infrastructure for our metrics so I've instead added a single assert in our multi_rack integration test.

benbromhead (approved):

conorbros (approved):

PR #1013, opened 2023-02-02
Initial kafka integration test

rukai:
Sets up a trivial kafka integration test, it doesnt touch shotover yet, just creates a kafka instance and then sends a message to a topic.
The goal is just to give us a basic setup for experimenting with.

I chose the [rdkafka](https://github.com/fede1024/rust-rdkafka) library for our int tests which uses the upstream kafka C++ library for similar reasons to the cassandra datastax c++ library we've been using.
When we go to actually implement kafka, [rskafka](https://github.com/influxdata/rskafka/) would be a better starting point as it is pure rust.

There is also the [kafka](https://github.com/kafka-rust/kafka-rust) crate but I would just ignore that, its not async.

I chose to use kraft (internal kafka cluster management) instead of zookeeper (external server that performs cluster management).
kraft brings our startup time down by 3s (8s -> 5s), and it wont affect how shotover interacts with kafka, so it seems like an obvious win for us.


benbromhead (approved):
Great start, looks good to me - Agree with the logic on the libs and Kafka deployment mode

conorbros (approved):

PR #1015, opened 2023-02-05
Update to serde_yaml 0.9 (via renaming Null -> NullSink)

rukai:
An alternative approach to https://github.com/shotover/shotover-proxy/pull/733

There are effectively two ways to parse yaml.

## The loose way
* default in serde_yaml <= 0.8
* can be reverted back to via `serde_yaml::with::singleton_map_recursive::deserialize(deserializer)` in serde_yaml 0.9
* no exclamation marks
* enum variant names will be misinterpreted as special yaml values such as null.
* Implemented in this PR by renaming Null transform to NullSink to avoid collision.
    + NullSink is technically not a sink by our definition but it is used in place of a Sink so I think the name still makes sense.

## The strict way:
* default in serde_yaml >= 0.9
* enum variants are always declared with an exclamation mark, this serves to disambiguate between enum variants and special yaml values such as null.
* implemented in https://github.com/shotover/shotover-proxy/pull/733

So ultimately the decision between this and #733 comes down to whether its more important to avoid the user having to put `!` everywhere or more important to avoid the possibility of yaml special value collisions with enum variant names.


conorbros (approved):

benbromhead:
having to put `!` everywhere is going to suck... and we'll just have to cope with a Null collision if it comes up (though I'm hoping due to technical carcinisation we'll all just use None)

benbromhead (approved):

rukai:
Sounds good, thanks for weighing in

Issue #985, opened 2023-01-05
set tcp nodelay for outgoing connections?


rukai:
I was reading a blog post during my holiday and realized that we set tcp nodelay for client<->shotover connections but we dont set it for shotover <-> DB connections.
Should we?

benbromhead:
Yes we should!!

rukai:
I wrote a benchmark that puts shotover as the bottleneck between the client and database.
I observed that:
Removing the client<->shotover nodelay halves our throughput. 
Adding a shotover<->DB nodelay drops our throughput by about ~10%

So seems we should just leave it as is.
I am quite puzzled by these results though.

benbromhead:
What does it do to latency though?

rukai:
Tried enabling outgoing tcpnodelay again today, it seems to improve cassandra latency and regress redis latency.
I set a cap to operations per second to avoid the noise I have not yet eliminated from windsock.

redis (left nodelay=false, right nodelay=true)
![image](https://github.com/shotover/shotover-proxy/assets/5120858/46854187-d7e2-4f20-8a41-bc6cd263ffbd)

cassandra (left nodealy=false, right nodelay=true)
![image](https://github.com/shotover/shotover-proxy/assets/5120858/6a95a788-ed4f-4c51-aa5f-9d8acb825c23)

rukai:
We should retry this since the recent performance improvements

rukai:
I got nagle'd on a hobby project just now and it gave me the insight I needed to know how to proceed here.

We wont see performance improvement on our current integration benchmarks because they are all stress tests.
We are stuffing as much data through 1 or 100 connections as possible.
This wont hit issues with tcp_nodelay enabled since constantly pushing more data in will force the packet out into the network and no delay will occur.

However if we set up a benchmark where we send 1 request every second on a single connection, the latency results of that benchmark should demonstrate tcp_nodelay causing us issues!

rukai:
This morning I spent some time looking into this.
I set up a benchmark as I described above with 1 small request per second.
However this did not reproduce the issues of nagle's algorithm at all.
Whether i had tcp_nodelay enabled did not affect this benchmark.

And on closer analysis:
* Each redis operation has a latency of ~0.6ms.
* Each network hop between the 3 EC2 instances takes ~0.15ms.
* There are 4 network hops that must be taken: client -> shotover -> redis -> shotover -> client
[...]

PR #1016, opened 2023-02-06
Cassandra mock bench

rukai:
## cassandra_bench_mocked.rs

We want a bench that can better show small changes to performance without it getting lost in noise.

I achieved this by replacing cassandra with a custom cassandra implementation where everything is mocked out and it only supports the absolute bare minimum to perform the benchmark.
An important insight I had while working on this was that we want to assign the cores of the machine such that shotover is the bottleneck. By doing so any improvements/regression to shotover will be much more visible.

To do this I experimented with different core counts and found that giving the client and DB 3 cores and shotover 1 core made shotover sufficiently bottlenecking while allowing it to run on just an 8 core machine.
If we dont have 8 cores available then we fall back to 1 core each which is not ideal but is still better than overassigning cores on a machine that doesnt have the cores.

## context of this work
I originally wrote this a while back to confirm that https://github.com/shotover/shotover-proxy/issues/985 improves performance as expected. It ended up demonstrating degraded performance so I've put 985 on hold for now, but I would still like to get this checked in for future use.

I additionally think this approach could give us more consistent results in our criterion benchmarks as I suspect the majority of the instability of results is from cassandra itself.

But for now lets just evaluate it as an `--example` benchmark that compares sending messages with and without shotover in between.


benbromhead (approved):
Fastest known protocol correct C* server known to humankind

conorbros (approved):

PR #1018, opened 2023-02-06
remove Transform::_prep_transform_chain

rukai:
We talked about removing this 2 years ago but still to this day no transforms use it and the chain never calls it.
So I think its time to remove it.
If the transform needs this functionality they can very easily implement it themselves by storing their state in an Option or using a boolean flag, which is IMO easier to reason about for transform authors than yet another function to implement.


conorbros (approved):

benbromhead (approved):
👍

PR #1020, opened 2023-02-06
Cassandra test speed

conorbros:
An idea for Cassandra test speedup. 

- Add a suffix to our test keyspaces with the name of the driver
- Create the `DockerCompose` once and run all of each driver's tests against it
- Involves moving away from `rstest` because their `#[once]` fixture does not drop values so the containers would hang

This also gives us the ability to potentially run the driver test suites concurrently.

We lose the convenience of `rstest` telling us which driver failed what test. If we improve the logging we could fix this but it does mean when a driver fails a test that entire test stops.

### DIFF
--- a/shotover-proxy/tests/cassandra_int_tests/mod.rs
+++ b/shotover-proxy/tests/cassandra_int_tests/mod.rs
@@ -4,7 +4,7 @@ use cdrs_tokio::frame::events::{
 };
 use futures::future::join_all;
 use futures::Future;
-use rstest::rstest;
+use rstest::*;
 use serial_test::serial;
 #[cfg(feature = "cassandra-cpp-driver-tests")]
 use test_helpers::connection::cassandra::CassandraDriver::Datastax;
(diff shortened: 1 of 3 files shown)


rukai:
I agree that something like this would bring a huge speedup to running tests on CI.
However, normally my dev iteration process involves rerunning just one test with one driver, so having to run every driver would actually slow down that iteration process.
And as you've noted we run into into issues where losing the granularity of each driver being its own test introduces a bunch of other pain points too.

A solution I had been thinking of would be a custom cargo test runner (fork or upstream to nextext?) that allows each test to specify an external dependency that needs to be initialized for it to run. Then the test runner can search through all the tests and run them in the optimal order, avoiding re-initializing external dependencies where possible. A test generated uuid would be sufficient for avoiding collisions in test keyspaces.

Personally I had put this in the not-worth-it basket because of all of the complexity and the chance of realizing too late that some of our tests are not compatible with this approach.

A likely simpler but just as effective way to improve CI times would be to just employ nextext's CI parallelization feature where it automatically spreads tests across as many CI instances as possible.
Currently we just split by cassandra and non-cassandra but using nextest to automate an even distribution across more nodes would be faster and less brittle than our current test distribution approach.

benbromhead:
A property based testing framework (e.g. quick theories / QuickCheck from java, but for rust), might also be a possible solution here (e.g. the driver name enum would be part of the properties that get tested / iterated over). However I'm not sure of that state of that for rust or if there are suitable rust property testing libraries whether they would give the level of control required to make this work

conorbros:
Closing as it's not worth losing the ability to know what test failed on which driver.

PR #1021, opened 2023-02-06
Cassandra compression

conorbros:
Implements support for compression in Cassandra.

## Changes

- Wrap the compression field in a mutex so the state of the selected compression can be shared between the read and write tasks for the socket.
- Catch startup messages in the `encode`/`decode` methods of the codec to set the compression.
- Add a new method  to the `Codec` trait in Shotover to clone without state (i.e. when a new `Handler` is created).
- Modifications to `frame/cassandra` because the tracing uuid and warnings need to be compressed with the body.
- Move the `Codec` trait to the `codec` module as it makes more sense for it to be there. 
- Refactor the `CassandraConnection` test helper methods with `new_inner` to re-use code where possible.

## Testing

Not convinced covering each compression type with passthrough/encoding on each driver is necessary, if reviewers agree I can make the tests smaller.

### DIFF
--- a/shotover-proxy/src/codec/cassandra.rs
+++ b/shotover-proxy/src/codec/cassandra.rs
@@ -1,23 +1,25 @@
+use crate::codec::{Codec, CodecReadError};
 use crate::frame::cassandra::{CassandraMetadata, CassandraOperation, Tracing};
 use crate::frame::{CassandraFrame, Frame, MessageType};
 use crate::message::{Encodable, Message, Messages, Metadata};
-use crate::server::CodecReadError;
 use anyhow::{anyhow, Result};
 use bytes::{Buf, BufMut, BytesMut};
 use cassandra_protocol::compression::Compression;
 use cassandra_protocol::frame::message_error::{ErrorBody, ErrorType};
-use cassandra_protocol::frame::message_supported::BodyResSupported;
+use cassandra_protocol::frame::message_startup::BodyReqStartup;
 use cassandra_protocol::frame::{
-    CheckEnvelopeSizeError, Envelope as RawCassandraFrame, Opcode, Version,
+    CheckEnvelopeSizeError, Envelope as RawCassandraFrame, Flags, Opcode, Version,
 };
 use cql3_parser::cassandra_statement::CassandraStatement;
 use cql3_parser::common::Identifier;
+use std::sync::Arc;
+use std::sync::RwLock;
 use tokio_util::codec::{Decoder, Encoder};
 use tracing::info;
 
 #[derive(Debug, Clone)]
 pub struct CassandraCodec {
-    compression: Compression,
+    compression: Arc<RwLock<Compression>>,
     messages: Vec<Message>,
     current_use_keyspace: Option<Identifier>,
 }
@@ -31,20 +33,58 @@ impl Default for CassandraCodec {
 impl CassandraCodec {
     pub fn new() -> CassandraCodec {
         CassandraCodec {
-            compression: Compression::None,
+            compression: Arc::new(RwLock::new(Compression::None)),
             messages: vec![],
             current_use_keyspace: None,
         }
     }
 }
 
-impl CassandraCodec {
-    fn encode_raw(&mut self, item: CassandraFrame, dst: &mut BytesMut) {
-        let buffer = item.encode(self.compression);
-        if buffer.is_empty() {
-            info!("trying to send 0 length frame");
+impl Codec for CassandraCodec {
+    fn clone_without_state(&self) -> Self {
+        Self {
+            compression: Arc::new(RwLock::new(Compression::None)),
+            messages: self.messages.clone(),
+            current_use_keyspace: self.current_use_keyspace.clone(),
         }
-        dst.put(buffer.as_slice());
+    }
+}
+
+impl CassandraCodec {
+    fn check_compression(&mut self, bytes: &BytesMut) -> Result<bool> {
+        if bytes.len() < 9 {
+            return Err(anyhow!("Not enough bytes for cassandra frame"));
+        }
+        let opcode = Opcode::try_from(bytes[4])?;
+
+        let compressed = Flags::from_bits_truncate(bytes[1]).contains(Flags::COMPRESSION);
+
+        // check if startup message and set the codec's selected compression
+        if Opcode::Startup == opcode {
+            if let CassandraFrame {
+                operation: CassandraOperation::Startup(startup),
+                ..
+            } = CassandraFrame::from_bytes(bytes.clone().freeze(), Compression::None)?
+            {
+                self.set_compression(&startup);
+            };
+            return Ok(false);
+        }
+
+        Ok(compressed)
+    }
+
+    fn set_compression(&mut self, startup: &BodyReqStartup) {
+        if let Some(compression) = startup.map.get("COMPRESSION") {
+            let mut write = self.compression.as_ref().write().unwrap();
+
+            *write = match compression.as_str() {
+                "snappy" | "SNAPPY" => Compression::Snappy,
+                "lz4" | "LZ4" => Compression::Lz4,
+                "" | "none" | "NONE" => Compression::None,
+                _ => panic!(),
+            };
+        }
     }
 }
 
@@ -72,26 +112,16 @@ impl Decoder for CassandraCodec {
                         return Err(reject_protocol_version(version.into()));
                     }
 
-                    let mut message = Message::from_bytes(bytes.freeze(), MessageType::Cassandra);
+                    let compressed = self.check_compression(&bytes).unwrap();
 
-                    // if is startup message, reject compression because shotover does not support
-                    if let Some(Frame::Cassandra(frame)) = message.frame() {
-                        if let CassandraOperation::Startup(startup) = &mut frame.operation {
-                            if let Some(compression) = startup.map.get("COMPRESSION") {
-                                return Err(reject_compression(frame.stream_id, compression));
-                            }
-                        }
-
-                        if let CassandraOperation::Supported(BodyResSupported { data }) =
-                            &mut frame.operation
-                        {
-                            if let Some(value) = data.get_mut("COMPRESSION") {
-                                *value = vec![];
-                            }
-
-                            message.invalidate_cache();
-                        }
-                    }
+                    let mut message = if !compressed {
+                        Message::from_bytes(bytes.freeze(), MessageType::Cassandra)
+                    } else {
+                        Message::from_bytes_with_compression(
+                            bytes.freeze(),
+                            *self.compression.read().unwrap(),
+                        )
+                    };
 
                     if let Ok(Metadata::Cassandra(CassandraMetadata {
                         opcode: Opcode::Query | Opcode::Batch,
@@ -216,26 +246,6 @@ fn reject_protocol_version(version: u8) -> CodecReadError {
     ))])
 }
 
-fn reject_compression(stream_id: i16, compression: &String) -> CodecReadError {
-    info!(
-        "Rejecting compression option {} (configure the client to use no compression)",
-        compression
-    );
-
-    CodecReadError::RespondAndThenCloseConnection(vec![Message::from_frame(Frame::Cassandra(
-        CassandraFrame {
-            version: Version::V4,
-            stream_id,
-            operation: CassandraOperation::Error(ErrorBody {
-                message: format!("Unsupported compression type {}", compression),
-                ty: ErrorType::Protocol,
-            }),
-            tracing: Tracing::Response(None),
-            warnings: vec![],
-        },
-    ))])
-}
-
 impl Encoder<Messages> for CassandraCodec {
     type Error = anyhow::Error;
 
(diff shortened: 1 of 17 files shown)


shotover-proxy/src/codec/cassandra.rs
```diff
@@ -31,20 +33,58 @@ impl Default for CassandraCodec {
...
+        }
+    }
+}
+
 impl CassandraCodec {
-    fn encode_raw(&mut self, item: CassandraFrame, dst: &mut BytesMut) {
-        let buffer = item.encode(self.compression);
-        if buffer.is_empty() {
-            info!("trying to send 0 length frame");
+    fn check_compression(&mut self, bytes: &BytesMut) -> Result<bool> {
```
benbromhead, line 54: Does it make sense to store the compression state in the codec so we can just assume subsequent messages are all encoded?

shotover-proxy/src/codec/cassandra.rs
```diff
@@ -1,23 +1,25 @@
...
 use cql3_parser::common::Identifier;
+use std::sync::Arc;
+use std::sync::RwLock;
 use tokio_util::codec::{Decoder, Encoder};
 use tracing::info;
 
 #[derive(Debug, Clone)]
 pub struct CassandraCodec {
-    compression: Compression,
+    compression: Arc<RwLock<Compression>>,
```
benbromhead, line 22: Given we would likely just set this once in a session (unless the spec says otherwise), do we need the Arc<RwLock<>>? This is just my gut feel from an eyeball read, the compiler sounds like it says something different otherwise?

shotover-proxy/src/codec/cassandra.rs
conorbros: From the spec: 

> a server may only compress frames above a certain size at its
  discretion

https://github.com/apache/cassandra/blob/5f8270eb8670b7d2f68c8b1a1d791a8b2cb849a0/doc/native_protocol_v4.spec#L825

The compression used is agreed on in the startup message but subsequent messages won't always be encoded so we need to check whether the message is compressed each time, while saving the agreed upon compression.

shotover-proxy/src/codec/cassandra.rs
conorbros: The compression won't be changed but this is needed because the agreed compression needs to be shared to both the read and write codec.

benbromhead (approved):

shotover-proxy/src/message/mod.rs
```diff
@@ -74,6 +77,18 @@ impl Message {
                 message_type,
             }),
             meta_timestamp: None,
+            compression: None,
+        }
+    }
+
+    pub fn from_bytes_with_compression(bytes: Bytes, compression: Compression) -> Self {
```
rukai, line 84: We should always be using this for cassandra message creation, otherwise we might forget to include the compression, so can we name it `from_cassandra_bytes`?

shotover-proxy/src/message/mod.rs
```diff
@@ -60,6 +61,8 @@ pub struct Message {
     // TODO: Not a fan of this field and we could get rid of it by making TimestampTagger an implicit part of ConsistentScatter
     // This metadata field is only used for communication between transforms and should not be touched by sinks or sources
     pub meta_timestamp: Option<i64>,
+
+    pub compression: Option<Compression>,
```
rukai, line 65: Making this an Option is duplicating the Compression::None state.

shotover-proxy/src/message/mod.rs
rukai: Maybe we should make `MessageType::Cassandra` into `MessageType::Cassandra { compression: Compression }`.
That way we can just do away with this method entirely.
Semantically that makes a lot of sense since each compression type is effectively its own message type as you cant decode a compressed cassandra message without the knowledge of what compression type was used.
But maybe we'll likely need to fork the MessageType enum as there are probably uses of this enum that dont care about the compression type.


This would also help cleanup Frame as well!

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -385,27 +386,43 @@ impl CassandraFrame {
...
             } else {
                 // TODO: While compression is obviously going to cost more than no compression, I suspect it doesnt have to be quite this bad
                 let mut body_buf = Vec::with_capacity(128);
                 let mut body_cursor = Cursor::new(&mut body_buf);
+
+                if let Tracing::Response(Some(uuid)) = self.tracing {
+                    body_cursor.write_all(uuid.as_bytes()).ok();
+                }
+
+                if !self.warnings.is_empty() {
```
rukai, line 415: can we avoid duplicating all of this by pulling it into a function?

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -385,27 +386,43 @@ impl CassandraFrame {
         cursor.write_all(&[self.operation.to_opcode().into()]).ok();
 
         serialize_with_length_prefix(&mut cursor, |cursor| {
-            if let Tracing::Response(Some(uuid)) = self.tracing {
-                cursor.write_all(uuid.as_bytes()).ok();
-            }
+            if Compression::None == compression || self.operation.to_opcode() == Opcode::Startup {
```
rukai, line 389: For this bit `|| self.operation.to_opcode() == Opcode::Startup`, I'm not sure where this help, say the client accidentally compressed the startup message, trying to interpret it as non-compressed isnt going to save us.

rukai (changes requested):
I think that I'm generally happy with this approach.
It sucks the cassandra protocol is stateful like this but it still makes sense to support this feature for users who benefit from compressing their messages, and it at least pushes us to make shotovers abstractions better support even worse stateful protocols in the future.

I've left feedback for the implementation, once thats addressed it should be in a pretty good state.
I'll take a look at the integration tests in a bit.

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -662,6 +663,70 @@ async fn request_throttling(#[case] driver: CassandraDriver) {
...
+    for topology in [
+        "example-configs/cassandra-passthrough/topology.yaml",
+        "example-configs/cassandra-passthrough/topology-encode.yaml",
+    ] {
+        for compression in [Compression::Lz4, Compression::Snappy] {
+            test(driver, topology, compression).await;
+        }
+    }
+}
+
```
rukai, line 694: > Not convinced covering each compression type with passthrough/encoding on each driver is necessary, if reviewers agree I can make the tests smaller.

On my machine it only takes 3mins total to run these new tests.
But it seems to take 20mins on CI which is pretty bad!
So I agree, lets trim it down a bit, I would probably get rid of cdrs on both tests (I dont think testing cdrs is critical) and get rid of LZ4 on compression_cluster (compression_cluster is slower, so theres a bigger win to remove from there)

test-helpers/src/connection/cassandra.rs
```diff
@@ -153,36 +229,54 @@ impl CassandraConnection {
...
                     )
                     .user("cassandra", "cassandra")
-                    .default_consistency(Consistency::One)
-                    .compression(Some(scylla::transport::Compression::Snappy))
-                    .build()
-                    .await
-                    .unwrap();
+                    .default_consistency(Consistency::One);
+
+                if let Some(compression) = compression {
```
rukai, line 266: We can write this directly in the builder chain like:
```rust
                    .compression(compression.map(|x| match x {
                        Compression::Snappy => scylla::transport::Compression::Snappy,
                        Compression::Lz4 => scylla::transport::Compression::Lz4,
                    }))
                    ```

File: shotover-proxy/src/message/mod.rs (hunk 4 above)
Comment (conorbros) [reply shotover-proxy/src/message/mod.rs:65]: It's to indicate that the message has no compression set. e.g. a redis message has no concept of compression, whereas a Cassandra messages does. I'd think of it like compression = null rather than compression = compression::none.

File: shotover-proxy/src/message/mod.rs (hunk 3 above)
Comment (conorbros) [reply shotover-proxy/src/message/mod.rs:84]: I had initially done that but methods like `get_type` and `into_encodable` become messy.

File: shotover-proxy/src/frame/cassandra.rs [hunk 9]
```diff
@@ -387,27 +388,25 @@ impl CassandraFrame {
...
-            }
-            if let Compression::None = compression {
-                // Special case None to avoid large copies
                 self.operation.serialize(cursor, self.version)
             } else {
                 // TODO: While compression is obviously going to cost more than no compression, I suspect it doesnt have to be quite this bad
                 let mut body_buf = Vec::with_capacity(128);
                 let mut body_cursor = Cursor::new(&mut body_buf);
+
+                if let Tracing::Response(Some(uuid)) = self.tracing {
```
rukai, line 404: can this go in write_warnings? (obviously need to rename write_warnings then)

shotover-proxy/src/message/mod.rs
rukai: Can get_type and into_encodable have their own separate MessageType enum? Not sure what you would call it, maybe get_type/into_encodable have MessageType and from_bytes has a ProtocolType? or maybe a MessageTypeFull?

shotover-proxy/src/message/mod.rs
rukai: If None indicates a non-cassandra message then I think thats a reasonable attempt at future proofing our logic, but we should make that clear by making the field `pub codec_state: CodecState` and having a:
```rust
enum CodecState {
    Cassandra { compression: Compression },
    Redis,
}
```

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -387,27 +388,25 @@ impl CassandraFrame {
...
+                self.write_warnings(cursor);
 
-                for warning in &self.warnings {
-                    let warning_len = warning.len() as i16;
-                    cursor.write_all(&warning_len.to_be_bytes()).ok();
-                    cursor.write_all(warning.as_bytes()).ok();
-                }
-            }
-            if let Compression::None = compression {
-                // Special case None to avoid large copies
```
rukai, line 405: can we keep this comment?

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -387,27 +388,19 @@ impl CassandraFrame {
...
                 self.operation.serialize(cursor, self.version)
+
+            // Special case None to avoid large copies
```
rukai, line 396: this looks like the comment is applying to the else block now, we want the comment to apply to the initial if block.

shotover-proxy/src/frame/cassandra.rs
```diff
@@ -378,6 +378,7 @@ impl CassandraFrame {
             u8::from(self.version) | u8::from(self.operation.to_direction());
 
         let mut flags = Flags::default();
+        flags.set(Flags::COMPRESSION, compression != Compression::None);
```
rukai, line 381: The compression flag should be unmodified right?

shotover-proxy/src/message/mod.rs
```diff
@@ -81,6 +103,7 @@ impl Message {
     /// Providing both the raw bytes and Frame results in better performance if they are both already available.
     pub fn from_bytes_and_frame(bytes: Bytes, frame: Frame) -> Self {
```
rukai, line 104: I think we need to take a ProtocolType here in order to properly set codec_state

shotover-proxy/src/message/mod.rs
```diff
@@ -91,6 +114,7 @@ impl Message {
     /// Providing just the Frame results in better performance when only the Frame is available.
     pub fn from_frame(frame: Frame) -> Self {
```
rukai, line 115: I think we need to take a ProtocolType here in order to properly set codec_state

shotover-proxy/src/frame/cassandra.rs
conorbros: I think I'm misunderstanding this comment, what do you mean by unmodified? Here we are setting it using the message's codec state.

shotover-proxy/src/message/mod.rs
conorbros: If you're creating a new message from a frame you can assume that it's not compressed and we can use Compression::None as a default. The Cassandra protocol allows for messages to be compressed/uncompressed, even if compression was agreed on at startup.

rukai: You are right, we can safely assume that :+1:

shotover-proxy/src/frame/cassandra.rs
rukai: yeah I was just mistaken, sorry about that.

One thing worth noting here is that we are actually disregarding whether the client had originally requested this specific message to be compressed or not and are actually compressing every single message that goes through.
But I'm not worried about that, I doubt the fine-grained control of compressing or not compressing messages was that useful in the first place. If we do hit performance issues for it we should look to optimize our compression code path first (theres a TODO about that below)

shotover-proxy/src/frame/mod.rs
```diff
@@ -14,16 +14,61 @@ pub enum MessageType {
...
 pub enum Frame {
     Cassandra(CassandraFrame),
     Redis(RedisFrame),
 }
 
 impl Frame {
-    pub fn from_bytes(bytes: Bytes, message_type: MessageType) -> Result<Self> {
+    pub fn from_bytes(
+        bytes: Bytes,
+        message_type: MessageType,
```
rukai, line 47: Can we take a ProtocolType to avoid duplicating these methods?

shotover-proxy/src/codec/cassandra.rs
```diff
@@ -31,20 +33,58 @@ impl Default for CassandraCodec {
...
+        // check if startup message and set the codec's selected compression
+        if Opcode::Startup == opcode {
+            if let CassandraFrame {
+                operation: CassandraOperation::Startup(startup),
+                ..
+            } = CassandraFrame::from_bytes(bytes.clone().freeze(), Compression::None)?
+            {
+                self.set_compression(&startup);
+            };
+            return Ok(false);
```
rukai, line 71: What is this early return for?
Cant we just rely on the compressed flag like we usually do?

shotover-proxy/src/codec/cassandra.rs
```diff
@@ -72,26 +112,23 @@ impl Decoder for CassandraCodec {
...
-                            &mut frame.operation
-                        {
-                            if let Some(value) = data.get_mut("COMPRESSION") {
-                                *value = vec![];
-                            }
-
-                            message.invalidate_cache();
-                        }
-                    }
+                    let mut message = if !compressed {
```
rukai, line 117: can we flip this so its `if compressed {`?

rukai: oh how about we do:
```rust
                    let mut message = Message::from_bytes(
                        bytes.freeze(),
                        crate::message::ProtocolType::Cassandra {
                            compression: if compressed {
                                *self.compression.read().unwrap()
                            } else {
                                Compression::None
                            },
                        },
                    );
```

shotover-proxy/src/codec/cassandra.rs
conorbros: Startup should never be compressed.

rukai: But even if it is compressed, pretending that it isn't won't save us.

shotover-proxy/src/frame/mod.rs
conorbros: I think we can just remove `from_bytes_with_compression`

rukai: yep, sounds good!

rukai (approved):
Alright, I think we are good to land this!

PR #1029, opened 2023-02-09
split codecs into builders, encoders and decoders

rukai:
The Codec trait is replaced with a CodecBuilder trait.
This is used in all the server.rs level usage where codecs were passed around in order to clone more codecs off it.

Now we use the `.build()` method which will generate a encoder and decoder halves.
We could just have these encoder/decoder codecs as separate generic types in server.rs etc, but we instead use a builder as that reduces the generic soup and allows for sharing state between the decoder/encoder codec pairs like we want to do in https://github.com/shotover/shotover-proxy/pull/1021

RedisCodec is left pretty much as is, demonstrating that if we have a simple codec we can just use a single struct as the builder, decoder and encoder.
To demonstrate the actual functionality I have gone ahead and split up CassandraCodec into its builder, decoder and encoder.
We could land this separately though.

Splitting up CassandraCodec allows us to keep only the fields relevant to decoding or encoding.
* This improves struct byte count (which I doubt really matters for codecs anyway)
* Makes the implementations easier to follow, only the fields actually used by decoding or encoding are in the struct.
* Better matches to how we use codecs, we only ever use a codec for decoder or encoder, never both, so it makes sense to reflect that in our types.

## Future work
* Splitup KafkaCodec
* Splitup RedisCodec

### DIFF
--- a/shotover-proxy/benches/benches/codec.rs
+++ b/shotover-proxy/benches/benches/codec.rs
@@ -4,7 +4,8 @@ use cassandra_protocol::frame::message_result::{
 };
 use cassandra_protocol::frame::Version;
 use criterion::{black_box, criterion_group, BatchSize, Criterion};
-use shotover_proxy::codec::cassandra::CassandraCodec;
+use shotover_proxy::codec::cassandra::CassandraCodecBuilder;
+use shotover_proxy::codec::CodecBuilder;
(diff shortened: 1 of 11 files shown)


benbromhead:
Let's land the compression PR first

benbromhead (approved):

conorbros (approved):

PR #1042, opened 2023-02-10
Add kafka codec, sink and source

rukai:
A trivial codec, sink and source implementation.
It successfully sends and receives kafka messages without performing any parsing on the messages.


conorbros (changes requested):
Can we put this behind alpha-transforms and add docs for it?

rukai:
I put it behind alpha-transforms an hour or so ago, maybe you were reviewing a slightly old version?

as for docs, the idea of alpha-transforms is that they dont have any docs yet, docs get added once they are moved into the public API.

conorbros (approved):

benbromhead (approved):
Good start! 

The one shot todo makes sense, and something to revisit when we start looking at perf

PR #1046, opened 2023-02-14
remove Frame::None

conorbros:
I think this was originally intended to mean "null" of redis frame but it's prevalence has gone beyond the scope of that. We can just use `RedisFrame::Null`.


rukai (approved):
Nice catch!

I think that, semantically, `Frame::None` makes a bit more sense than `Frame::Redis(RedisFrame::Null)` in `redis/sink_cluster.rs`
As it conveys the lack of a message rather than a null redis message.
But `Frame::Redis(RedisFrame::Null)` does work just fine here, so I'm more than happy to make the swap here to allow us to remove `Frame::Null`

benbromhead (approved):

PR #1049, opened 2023-02-19
Remove PCAP based tests

rukai:
These tests have a large maintenance upkeep:
* a large amount of implementation code on our side
* a lot of dependencies to keep updated
* relies on a magic pcap file on an AWS bucket somewhere

In contrast I don't think they bring much value:
* They only test the codec which does very minimal parsing. The codec aims to only do enough parsing to split each message off the incoming buffer. Sometimes we have to do some extra parsing but most of the time messages are only full parsed within the transforms. (when the transform calls `.frame()` )
* We have a vast test suite of cassandra integration tests which already covers these cases and much more.

## Alternatives
We could improve the cassandra pcap tests by actually calling `.frame()` on the returned messages so that we actually parse each message rather than just split them up.
But I still dont see anywhere that these tests provide coverage that isnt already provided by our integration tests.


conorbros:
We can also remove the need for `libpcap-dev`.

rukai:
Oh true, I've taken advantage of that in the PR now.

conorbros (approved):
LGTM, just update the contributing docs too.

benbromhead:
I'm happy to remove this given the coverage on the integration tests

benbromhead (approved):

Issue #1006, opened 2023-02-01
Add support for the Kafka protocol


benbromhead:
### Structure
Kafka protocol has multiple messages per publish call

### Performance concerns
How do we do allocations when parsing the Kafka message. 
Do we do what we do with C* and guess if we need to parse it or not... or do we do something smarter and allocate all at once (bump allocator?), or just point into the existing bytes (index into the byte array).

Protocol should be pretty simple in terms of needing to parse. 

How does parsing / zero copy bytes from the network all interact with call into IO accelerator code

benbromhead:
Versions!!!! (support every version of the protocol)

rukai:
Our kafka protocol support needs to be fast for our intended use case and I've been researching the tools that rust gives us to make this happen.

I originally thought we would be able to do a quick hack of  "just directly manipulate the bytes" for our fast message rewriting needs but it turns out that publish messages have 3-4 layers of nested list types so that would be quite a challenging hack.

Usually when writing zero-copy deserialization you dont need to support the user modifying the data structure and then serializing it again.
That ideal scenario means that all the existing options i've found fall just short of what we need, but I dont think its actually impossible to get what we want.

## serde approach

There is a bunch of resources for kafka zerocopy via serde:
* serde can already use `std::borrow::Cow` to zerocopy deserialize a string.
* [zerovec](https://docs.rs/zerovec/latest/zerovec/) provides `Vec` and `HashMap` equivalents that deserialize from serde and refer to the original buffer to get their values, unless they are modified in which case they take ownership of their data and heap allocate. So they are effectively `std::borrow::Cow` for `Vec`.
* https://github.com/DataDog/kafka-serde allows us to define kafka messages that can be serde serialized/deserialized, however it does not provide the message definitions itself.
* This can then be combined with [yoke](https://docs.rs/yoke/latest/yoke/index.html) to allow the deserialized message to be passed around without lifetimes getting in the way.

If we were to try to combine these into a solution we would hit these issues:
* Yoke gets in the way of us modifying the message after its been created - https://github.com/unicode-org/icu4x/issues/3134
* kafka-serde does not provide message definitions so we will have to do that ourselves, hopefully every kafka message is expressible in serde.

## custom approach

If the serde approach doesnt work out, either due to serdes overhead or an unexpected hidden obstacle, we can implement what the serde approach would be doing directly for kafka, i.e. come up with our own zerocopy kafka protocol implementation, including our own kafka specific alternative to zerovec

[...]

PR #1053, opened 2023-02-21
Initial kafka protocol

rukai:
as per my [writeup](https://github.com/shotover/shotover-proxy/issues/1006#issuecomment-1437669604), this PR gives an initial protocol implementation that we can start with.
I have no idea what the performance is like internally in the `kafka-protocol` crate but the code in the PR itself tries to be performant.

## kafka message overview

The spec does not give a good overview, so I'll give a quick one.
Every kafka message consists of: `size header` -> `message header` -> `message body`

* size header: every message starts with this simple header https://kafka.apache.org/protocol.html#protocol_common
* message header: A different header is used depending on if its a request or response - https://kafka.apache.org/protocol.html#protocol_messages
* message body: A different format is used depending on if its a request or response and on the api_key in the message_header: e.g. https://kafka.apache.org/protocol.html#The_Messages_Produce

## choices

unlike requests, kafka responses do not contain the message api_key or version.
So in order to get api_key+version for parsing a response you need to match it up with its corresponding request.
This is easily done because kafka does not support out of order messaging, so we just rely on matching up the order of messages.
To perform this matching up in shotover I've setup the codec to send the api_key+version from encoder to the decoder when its used as a sink. (when its a sink the encoder always has the requests and the decoder always has the responses)
Then the decoder includes the api_key+version in the messages codec state field so it can be used by the `.frame()` parser later on.

I've kept the logic for stripping the message size in the frame because every other protocol deals with that in the frame, but the kafka protocol would uniquely allow for us to actually handle that at the codec stage if we wanted to.
It doesn't particularly matter either way though.

api_keys that we havent hooked up parsing for yet will just get stored as a `RequestBody::Unknown` / `ResponseBody::Unknown`, allowing them to silently and correctly pass through shotover.

## weird bits

I'm not sure whats up with those header versions, we basically have to assume everyone is using the latest version, there's no way to detect what header versions are in use (the version in the header is for the body not the header itself), seems like a bug in the spec.

There is some duplication in state between the RequestHeader and the RequestBody: the `request_api_key` and the RequestBody variant.
But I've chosen to leave that for now because this protocol implementation is likely to need vast changes in the future in order to get the performance we need.

## Future work

In the next PR I'll add parsing for receive messages, I havent added them in this PR because they'll need integration test changes.

### DIFF
--- a/shotover-proxy/src/frame/kafka.rs
+++ b/shotover-proxy/src/frame/kafka.rs
@@ -1,0 +1,129 @@
+use crate::codec::kafka::RequestHeader as CodecRequestHeader;
+use anyhow::{anyhow, Context, Result};
+use bytes::{BufMut, Bytes, BytesMut};
+use kafka_protocol::messages::{
+    ApiKey, ProduceRequest, ProduceResponse, RequestHeader, ResponseHeader,
+};
+use kafka_protocol::protocol::{Decodable, Encodable};
+
+// No way to know which version to use, just have to guess
+const REQUEST_HEADER_VERSION: i16 = 2;
+const RESPONSE_HEADER_VERSION: i16 = 1;
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum KafkaFrame {
+    Request {
+        header: RequestHeader,
+        body: RequestBody,
+    },
+    Response {
+        version: i16,
+        header: ResponseHeader,
+        body: ResponseBody,
+    },
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum RequestBody {
+    Produce(ProduceRequest),
+    Unknown { api_key: ApiKey, message: Bytes },
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum ResponseBody {
+    Produce(ProduceResponse),
+    Unknown { api_key: ApiKey, message: Bytes },
+}
+
+impl KafkaFrame {
+    pub fn from_bytes(
+        mut bytes: Bytes,
+        request_header: Option<CodecRequestHeader>,
+    ) -> Result<Self> {
+        // remove length header
+        let _ = bytes.split_to(4);
+
+        match request_header {
+            Some(request_header) => KafkaFrame::parse_response(bytes, request_header),
+            None => KafkaFrame::parse_request(bytes),
+        }
+    }
+
+    fn parse_request(mut bytes: Bytes) -> Result<Self> {
+        let header = RequestHeader::decode(&mut bytes, REQUEST_HEADER_VERSION)
+            .context("Failed to decode request header")?;
+
+        let api_key = ApiKey::try_from(header.request_api_key)
+            .map_err(|_| anyhow!("unknown api key {}", header.request_api_key))?;
+        let body = match api_key {
+            ApiKey::ProduceKey => RequestBody::Produce(
+                ProduceRequest::decode(&mut bytes, header.request_api_version)
+                    .context("Failed to decode request body")?,
+            ),
+            api_key => RequestBody::Unknown {
+                api_key,
+                message: bytes,
+            },
+        };
+
+        Ok(KafkaFrame::Request { header, body })
+    }
+
+    fn parse_response(mut bytes: Bytes, request_header: CodecRequestHeader) -> Result<Self> {
+        let header = ResponseHeader::decode(&mut bytes, RESPONSE_HEADER_VERSION)
+            .context("Failed to decode response header")?;
+
+        let body = match request_header.api_key {
+            ApiKey::ProduceKey => ResponseBody::Produce(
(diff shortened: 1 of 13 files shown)


shotover-proxy/src/frame/mod.rs
```diff
@@ -111,6 +130,7 @@ impl Display for Frame {
         match self {
             Frame::Cassandra(frame) => write!(f, "Cassandra {}", frame),
             Frame::Redis(frame) => write!(f, "Redis {:?})", frame),
+            Frame::Kafka(frame) => write!(f, "Redis {:?})", frame),
```
conorbros, line 133: *Kafka

shotover-proxy/src/message/mod.rs
```diff
@@ -195,6 +202,7 @@ impl Message {
             MessageInner::Modified { frame } | MessageInner::Parsed { frame, .. } => match frame {
                 Frame::Cassandra(frame) => frame.cell_count()?,
                 Frame::Redis(_) => nonzero!(1u32),
+                Frame::Kafka(_) => todo!(),
```
conorbros, line 205: Couldn't this just be `nonzero(1u32)` same as Redis?

https://dattell.com/data-architecture-blog/load-balancing-with-kafka/#:~:text=What%20is%20Kafka%20loading%20balancing,partitions%20while%20preserving%20message%20ordering.

From a quick read I think it's safe to just assume 1  kafka message = 1 throttling cell

rukai: in cassandra we count each message within a batch as a separate cell, kafka produce messages allow for [batching many messages together](https://kafka.apache.org/protocol.html#The_Messages_Produce) so it feels equivalent to a cassandra batch.
Maybe its closer to just having multiple rows on an insert though?
Its a complicated question which I considered it out of scope and left as a todo!()

conorbros: Ah good point, didn't know that. Thanks

shotover-proxy/src/frame/kafka.rs
```diff
@@ -0,0 +1,129 @@
+use crate::codec::kafka::RequestHeader as CodecRequestHeader;
+use anyhow::{anyhow, Context, Result};
+use bytes::{BufMut, Bytes, BytesMut};
+use kafka_protocol::messages::{
+    ApiKey, ProduceRequest, ProduceResponse, RequestHeader, ResponseHeader,
+};
+use kafka_protocol::protocol::{Decodable, Encodable};
+
+// No way to know which version to use, just have to guess
+const REQUEST_HEADER_VERSION: i16 = 2;
```
benbromhead, line 10: _Kafka has a "bidirectional" client compatibility policy. In other words, new clients can talk to old servers, and old clients can talk to new servers. This allows users to upgrade either clients or servers without experiencing any downtime.

Since the Kafka protocol has changed over time, clients and servers need to agree on the schema of the message that they are sending over the wire. This is done through API versioning.

Before each request is sent, the client sends the API key and the API version. These two 16-bit numbers, when taken together, uniquely identify the schema of the message to follow.

The intention is that clients will support a range of API versions. When communicating with a particular broker, a given client should use the highest API version supported by both and indicate this version in their requests.

The server will reject requests with a version it does not support, and will always respond to the client with exactly the protocol format it expects based on the version it included in its request. The intended upgrade path is that new features would first be rolled out on the server (with the older clients not making use of them) and then as newer clients are deployed these new features would gradually be taken advantage of._

So it looks like we pick the highest we support, and then have to negotiate down if the server comes back a different protocol format (e.g. in the form of the two 16 bit numbers)

benbromhead: From https://kafka.apache.org/protocol.html#protocol_compatibility

benbromhead: Also supports and ApiVersionsRequest - https://kafka.apache.org/protocol.html#api_versions

benbromhead:
I'm impressed at `kafka-protocol-rs` and it's coverage of protocol versions (trying to hand roll all that would be a terrible task).

Given the support of protocol negation, if you want to do that in another PR that's fine and I'm happy to let this land as is.

shotover-proxy/src/frame/kafka.rs
rukai: > So it looks like we pick the highest we support, and then have to negotiate down if the server comes back a different protocol format (e.g. in the form of the two 16 bit numbers)


That is correct and we already handle that correctly, or at least we dont get in the way of the client/server handling that correctly.
What my code comment is talking about is the version used for the header which doesnt seem to have any negotiation in the protocol. The driver I was referring to `rskafka` just hardcoded it to the versions I am hardcoding here.

benbromhead (approved):

conorbros (approved):

PR #1055, opened 2023-02-22
CassandraConnectionBuilder

conorbros:
Now that we have a bunch of different config options for the `CassandraConnection` helper it makes sense to use the builder pattern. This will make testing different combinations of protocol versions, compression and TLS easier.

### DIFF
--- a/test-helpers/src/connection/cassandra.rs
+++ b/test-helpers/src/connection/cassandra.rs
@@ -43,6 +43,79 @@ use std::sync::Arc;
 use std::time::Duration;
 use tokio::time::timeout;
 
+pub struct CassandraConnectionBuilder {
+    contact_points: String,
+    port: u16,
+    driver: CassandraDriver,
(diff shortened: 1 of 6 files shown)


test-helpers/src/connection/cassandra.rs
```diff
@@ -118,66 +198,13 @@ pub enum CassandraConnection {
...
-    }
-
-    async fn new_inner(
+    pub async fn new(
         contact_points: &str,
         port: u16,
         driver: CassandraDriver,
         compression: Option<Compression>,
         tls: Option<Tls>,
+        protocol: ProtocolVersion,
```
rukai, line 207: the protocol version is unused by scylla.
When compression is unused by a backend we panic, should we do the same for the protocol?

rukai (approved):
Nice! Thanks for doing this

benbromhead (approved):

Issue #1056, opened 2023-02-23
kafka flexible versions support


rukai:
Currently we hardcode our header versions, currently this seems fine as our rdkafka driver is a little old but it will cause issues when a more modern driver is used.
The correct way to version headers is not included in the kafka protocol documentation.

I found this which is very enlightening:
https://stackoverflow.com/questions/71794315/questions-about-the-documentation-of-the-kafka-protocol
They link to source generator, but their link is broken, I've fixed it here: https://github.com/apache/kafka/blob/698319b8e2c1f6cb574f339eede6f2a5b1919b55/generator/src/main/java/org/apache/kafka/message/ApiMessageTypeGenerator.java#L324

I then found that the header versioning disaster was introduced by something called flexible versions:
https://cwiki.apache.org/confluence/display/KAFKA/KIP-482%3A+The+Kafka+Protocol+should+Support+Optional+Tagged+Fields
Each message type has its own flexibleVersions range: https://github.com/apache/kafka/blob/db6beb9b0302ce99d0a768d90a125f6a084aa48b/clients/src/main/resources/common/message/FindCoordinatorResponse.json#L28

I think the best way to support this would be to add support to kafka-protocol crate.
It should generate functions:
```rust
fn request_header_version(api_key: u16, api_version: u16) -> u16 {
   /// return 1 or 2 depending on if the api_key+api_version is a flexibleVersion
}

fn response_header_version(api_key: u16, api_version: u16) -> u16 {
   /// return 0 or 1 depending on if the api_key+api_version is a flexibleVersion
}
```

PR #1057, opened 2023-02-23
Implement FindCoordinator rewriting (fixes clients connecting directly to kafka)

rukai:
KafkaSinkSingle works by having a shotover instance sit on the same address as every kafka instance.
The kafka instances all use one consistent port for communication to the client and shotover uses another consistent port for communication to the client.
It is up to shotover to rewrite any messages that refer to the kafka port to instead refer to the shotover port.

FindCoordinator is one such message, it contains ip address + ports of other kafka brokers that are acting as a coordinator
So this PR introduces parsing and rewriting of these messages which fixes our issue of the client connecting directly to kafka.

You can observe more messages are now flowing through shotover by setting the `tracing::debug`s to `tracing::info` in `codec/kafka.rs`

## a secondary issue
The hardcoded header versions were too high and causing parse failures, so I lowered them.
A proper solution to these headers is tracked by https://github.com/shotover/shotover-proxy/issues/1056
These bad header versions were getting past our tests because we werent actually parsing any messages before because the produce messages were bypassing shotover and going directly to kafka.

### DIFF
--- a/shotover-proxy/src/transforms/kafka/sink_single.rs
+++ b/shotover-proxy/src/transforms/kafka/sink_single.rs
@@ -1,5 +1,7 @@
 use crate::codec::kafka::{Direction, KafkaCodecBuilder};
 use crate::error::ChainResponse;
+use crate::frame::kafka::{KafkaFrame, ResponseBody};
+use crate::frame::Frame;
 use crate::message::Messages;
 use crate::tcp;
 use crate::transforms::util::cluster_connection_pool::{spawn_read_write_tasks, Connection};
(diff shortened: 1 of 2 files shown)


benbromhead (approved):

shotover-proxy/src/transforms/kafka/sink_single.rs
```diff
@@ -111,11 +113,28 @@ impl Transform for KafkaSinkSingle {
...
 
         // TODO: since kafka will never send requests out of order I wonder if it would be faster to use an mpsc instead of a oneshot or maybe just directly run the sending/receiving here?
-        if let Some(read_timeout) = self.read_timeout {
+        let mut responses = if let Some(read_timeout) = self.read_timeout {
             timeout(read_timeout, read_responses(responses)).await?
         } else {
             read_responses(responses).await
+        }?;
+
+        for response in &mut responses {
```
conorbros, line 122: Can we call this `message`, I thought it was a ResponseBody type instead.

rukai: it is a response from the list of responses.
Naming `response` -> `message` seems strange, should we also do `responses` -> `messages` as well?

conorbros, line 122: I'd add a comment here, just so it's clear of the side effect in Kafka sink single.

conorbros: Just used to seeing `message.frame()`, not too fussed though.

conorbros (approved):

Issue #951, opened 2022-12-08
TransformBuilder cleanups


rukai:
Followup cleanups for https://github.com/shotover/shotover-proxy/pull/935:
* properly move validate and is_terminating to TransformBuilder, maybe we want a TransformBuilder trait for this? TODO
* Remove all custom transform clone implementations, instead put that logic in a Builder build method :heavy_check_mark: 
* remove extra chain clones + remove BufferedChain::to_new_instance https://github.com/shotover/shotover-proxy/pull/1026

PR #1059, opened 2023-02-24
Make TransformBuilder a trait

rukai:
closes: https://github.com/shotover/shotover-proxy/issues/951

There are 3 stages to a transform: config, builder and transform
Currently we use seperate enums to represent all 3, I am experimenting with changing all of these enums into traits so that we can provide shotover as a library for purposes of users implementing their own transforms.

In this PR I have changed just TransformBuilder into a trait.

I skipped the other enums for now because:
* The transforms themselves are performance sensitive so I wanted to be careful changing them to dynamic dispatch.
* The configs are complicated to make into a trait due to needing to be deserialized via serde.

So this left TransformBuilder as a good, non-controversial, first step because:
It lacks the previously mentioned downsides and making TransformBuilder into a trait gives us:
* Less central configuration of transforms in transform.rs
* Cleans up the duplication of `fn validate` and `fn is_terminating()` across both the builder and transform.

dyn-clone dependency was added to allow us to clone the TransformBuilder trait.
We were previously cloning the TransformBuilder enum and its possible we could avoid doing that in the future, but thats out of scope of this PR.

Previously we were using IntoStaticStr to get the name of the TransformBuilder enum variant as the transform name.
That is no longer possible with it as a trait, but I think its reasonable to just have the user define it instead. We could possibly work out some macro solution for this in the future.

### DIFF
--- a/shotover-proxy/src/config/topology.rs
+++ b/shotover-proxy/src/config/topology.rs
@@ -120,8 +120,7 @@ impl Topology {
 
 #[cfg(test)]
 mod topology_tests {
-    use tokio::sync::watch;
-
+    use crate::config::topology::{Topology, TopologyConfig};
     use crate::transforms::coalesce::CoalesceConfig;
     use crate::{
         sources::{redis_source::RedisConfig, Sources, SourcesConfig},
@@ -132,8 +131,7 @@ mod topology_tests {
         },
     };
     use std::{collections::HashMap, fs};
-
-    use super::{Topology, TopologyConfig};
+    use tokio::sync::watch;
 
     async fn run_test_topology(chain: Vec<TransformsConfig>) -> anyhow::Result<Vec<Sources>> {
         let mut chain_config = HashMap::new();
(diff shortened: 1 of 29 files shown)


benbromhead:
I'm happy to make the TransformBuilder and Config into traits with all the fun it entails. 

As for the Transform itself:

- 100% there are perf concerns with dynamic dispatch (esp for things like redis, where it can push millions of op/s).
- Practically I don't think Rust has landed async functions in traits yet as well, which would also be required.

So without a pretty convincing reason to head to traits for the transform itself, I would be a -1 if we went down that path.

shotover-proxy/src/config/topology.rs
```diff
@@ -355,7 +353,7 @@ redis_chain:
     async fn test_validate_chain_invalid_subchain_redis_cache() {
         let expected = r#"Topology errors
 redis_chain:
-  SimpleRedisCache:
+  RedisCache:
```
benbromhead, line 356: nit: let's avoid renames in this PR

rukai: The transform is already called RedisCache, I'm fixing a bug where shotover's error output referred to RedisCache as SimpleRedisCache.

benbromhead (approved):

conorbros (approved):

PR #1061, opened 2023-02-27
Avoid storing ConsistencyBehaviour::SubchainOnMismatch config in the final tee transform

rukai:
This PR:
* avoids invalid states by storing the chain in the ConsistencyBehavior enum itself
* avoids cloning the mismatch chain config into every built tee

To achieve this a unique ConsistencyBehavior enum was created for every stage of the transform


benbromhead (approved):
Good catch! Definitely better from an encapsulation and data ownership perspective

conorbros (approved):

PR #1064, opened 2023-03-07
Make TransformConfig into a trait

rukai:
To achieve this, we sprinkle in some dtolnay magic crates: 
* typetag - we will always need this to magically hook up our traits to serde
* async_trait - the rust team seem pretty intent on landing some form of async traits one day so we should eventually be able to remove this

And finally we need a custom deserialize implementation for our chain config.
The problem was that typetagged structs with no fields would not deserialize from a yaml string.
e.g.
this would fail to deserialize:
```
chain_config:
  redis_chain:
    - NullSink
```

but this would work fine:
```
chain_config:
  redis_chain:
    - NullSink: {}
```

I do not really understand what the `vec_transform_config` custom deserializer is doing, but it clearly does fix the issue, allowing us to write our config as `NullSink` instead of `NullSink: {}`.
I can do a thorough investigation to better understand it if needed.
The implementation was a suggestion from dtolnay: https://github.com/dtolnay/typetag/pull/40#issuecomment-1454961686
This solution does add a bunch of complexity but it is at least hidden away in its own module and doesn't affect users of the API.

### DIFF
--- a/shotover-proxy/benches/benches/chain.rs
+++ b/shotover-proxy/benches/benches/chain.rs
@@ -12,11 +12,12 @@ use shotover_proxy::transforms::chain::{TransformChain, TransformChainBuilder};
 use shotover_proxy::transforms::debug::returner::{DebugReturner, Response};
 use shotover_proxy::transforms::filter::QueryTypeFilter;
 use shotover_proxy::transforms::null::NullSink;
+#[cfg(feature = "alpha-transforms")]
 use shotover_proxy::transforms::protect::{KeyManagerConfig, ProtectConfig};
 use shotover_proxy::transforms::redis::cluster_ports_rewrite::RedisClusterPortsRewrite;
 use shotover_proxy::transforms::redis::timestamp_tagging::RedisTimestampTagger;
 use shotover_proxy::transforms::throttling::RequestThrottlingConfig;
-use shotover_proxy::transforms::Wrapper;
+use shotover_proxy::transforms::{TransformConfig, Wrapper};
 
 fn criterion_benchmark(c: &mut Criterion) {
     let rt = tokio::runtime::Runtime::new().unwrap();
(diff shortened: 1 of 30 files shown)


shotover-proxy/benches/benches/chain.rs
```diff
@@ -272,6 +275,7 @@ fn criterion_benchmark(c: &mut Criterion) {
         });
     }
 
+    #[cfg(feature = "alpha-transforms")]
```
rukai, line 278: These are needed now because ProtectConfig is only available when alpha-transforms is enabled

shotover-proxy/src/config/chain.rs
```diff
@@ -0,0 +1,89 @@
...
+    pub async fn get_builder(&self, name: String) -> Result<TransformChainBuilder> {
+        let mut transforms: Vec<Box<dyn TransformBuilder>> = Vec::new();
+        for tc in &self.0 {
+            transforms.push(tc.get_builder(name.clone()).await?)
+        }
+        Ok(TransformChainBuilder::new(transforms, name))
+    }
+}
+
+fn vec_transform_config<'de, D>(deserializer: D) -> Result<Vec<Box<dyn TransformConfig>>, D::Error>
```
benbromhead, line 25: I think it would be worth a short comment explaining why we need to do our own Deserialiser stuff here.

shotover-proxy/src/config/topology.rs
```diff
@@ -95,22 +93,27 @@ impl Topology {
...
             distributed::tuneable_consistency_scatter::TuneableConsistencyScatterConfig,
             parallel_map::ParallelMapConfig, redis::cache::RedisConfig as RedisCacheConfig,
-            TransformsConfig,
         },
     };
     use std::collections::HashMap;
     use tokio::sync::watch;
 
-    async fn run_test_topology(chain: Vec<TransformsConfig>) -> anyhow::Result<Vec<Sources>> {
+    async fn run_test_topology(
```
conorbros, line 112: Could this function just box the vec's elements itself. Save a bunch of boilerplate `Box::new()`

shotover-proxy/src/transforms/debug/force_parse.rs
```diff
@@ -6,20 +6,27 @@
 /// It could also be used to ensure that messages round trip correctly when parsed.
 use crate::error::ChainResponse;
 use crate::transforms::{Transform, TransformBuilder, Transforms, Wrapper};
-use anyhow::Result;
 use async_trait::async_trait;
 use serde::Deserialize;
```
conorbros, line 11: Delete this for proper formatting

conorbros (approved):

shotover-proxy/src/config/topology.rs
rukai: It cant because we need to erase the full type in order to store different types within the same vec, and its the Box that does the type erasure

benbromhead (approved):

PR #1065, opened 2023-03-07
Cassandra protocol v5 support

conorbros:
Just opening to get some eyes on the code, still a WIP.


shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -823,3 +823,28 @@ async fn events_keyspace(#[case] driver: CassandraDriver) {
...
+#[rstest]
+#[case::cdrs(CdrsTokio)]
+#[tokio::test(flavor = "multi_thread")]
+#[serial]
+async fn test_protocol_v5(#[case] driver: CassandraDriver) {
+    let _docker_compose =
+        DockerCompose::new("example-configs/cassandra-passthrough/docker-compose.yaml");
+
+    let shotover = ShotoverProcessBuilder::new_with_topology(
+        "example-configs/cassandra-passthrough/topology.yaml",
```
rukai, line 836: we should use topology-encode.yaml here to get better coverage of our v5 support.

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -823,3 +823,28 @@ async fn events_keyspace(#[case] driver: CassandraDriver) {
...
+    )
+    .start()
+    .await;
+
+    let connection = || {
+        CassandraConnectionBuilder::new("127.0.0.1", 9042, driver)
+            .with_protocol_version(ProtocolVersion::V5)
```
rukai, line 843: Can we make variants of this test for v3 and v4?
Currently V4 is the default for cdrs so we do get coverage that way but there is no gaurantee that V4 will remain the default in the future.

PR #1066, opened 2023-03-07
codec direction logging

conorbros:
Pulled this out of #1065. Found it pretty difficult to read the debugging logs with all the concurrent connections and figure out what messages paired up without this change.


rukai:
The improved logging is a good idea but why move the direction arg from `new()` to `build()`?

rukai (approved):
LGTM!

benbromhead (approved):

PR #1073, opened 2023-03-09
Add ShotoverProcessBuilder::with_observability_address

rukai:
Allows us to correctly configure shotover in integration tests with multiple shotover instances.
Previously we had to tell BinProcess to ignore the resulting error.

Generating config for this is not ideal, but we currently dont expose this as a CLI flag, and I dont want to block this improvement on figuring out what our CLI/config.yaml split should look like.

### DIFF
--- a/test-helpers/src/shotover_process.rs
+++ b/test-helpers/src/shotover_process.rs
@@ -1,16 +1,15 @@
-use std::net::SocketAddr;
 use std::time::Duration;
+use uuid::Uuid;
 
 pub use tokio_bin_process::event::{Event, Level};
 pub use tokio_bin_process::event_matcher::{Count, EventMatcher, Events};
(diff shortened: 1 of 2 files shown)


test-helpers/src/shotover_process.rs
```diff
@@ -29,11 +33,32 @@ impl ShotoverProcessBuilder {
         self
     }
 
+    pub fn with_observability_address(mut self, address: SocketAddr) -> Self {
```
conorbros, line 36: This could just take a port number. Saves the `.parse().unwrap()` in the method calls.

conorbros (approved):

test-helpers/src/shotover_process.rs
rukai: yep, good call

benbromhead (approved):

PR #1075, opened 2023-03-10
remove unneeded clippy allows

conorbros:
Went through all our clippy allows, at some point we didn't need these ones anymore.

### DIFF
--- a/test-helpers/src/connection/cassandra.rs
+++ b/test-helpers/src/connection/cassandra.rs
@@ -715,8 +715,6 @@ impl CassandraConnection {
         assert_eq!(result.len(), 0, "Batches should never return results");
     }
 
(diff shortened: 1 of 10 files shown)


benbromhead (approved):

test-helpers/src/connection/cassandra.rs
```diff
@@ -716,7 +716,6 @@ impl CassandraConnection {
     }
 
     // allow reason: micro performance doesnt matter in a test and boxing ErrorBody makes matches unergonomic
```
rukai, line 718: we should remove the `allow reason` comment too

rukai: I think you missed the one I actually commented on? :sweat_smile:

rukai (approved):
Very nice!

PR #1090, opened 2023-03-17
Cassandra connection pass multiple messages to codec in sink single

conorbros:


shotover-proxy/src/transforms/cassandra/connection.rs
```diff
@@ -145,24 +145,51 @@ impl CassandraConnection {
     ///
     /// If an internal invariant is broken the internal tasks may panic and external invariants will no longer be upheld.
```
benbromhead, line 146: Does this comment still apply?

PR #1097, opened 2023-03-24
CassandraSinkCluster: pull table rewriting into its own module

rukai:
I should go over the exact boundary on monday but this seems promising.


conorbros (approved):

benbromhead:
I'm happy to merge as is, but let's pull `ShotoverNodes` out of it as it's a weird breaking of encapsulation (even if rewriter is the only thing that uses it atm).

benbromhead (approved):

PR #1099, opened 2023-03-26
Update scylla dep

rukai:
scylla now sets the default consistency via an "execution profile" instead of directly on the "session builder".

The extra `match compression { .. }` seems like some kind of merge error so I just removed it

### DIFF
--- a/shotover-proxy/tests/cassandra_int_tests/routing.rs
+++ b/shotover-proxy/tests/cassandra_int_tests/routing.rs
@@ -3,21 +3,16 @@ use test_helpers::connection::cassandra::{CassandraConnectionBuilder, CassandraD
 mod single_key {
     use test_helpers::connection::cassandra::{run_query, CassandraConnection, ResultValue};
 
-    pub async fn create_keyspace(connection: &CassandraConnection) {
+    pub async fn setup_schema(connection: &CassandraConnection) {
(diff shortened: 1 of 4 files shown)


conorbros:
Will leave as a draft until there's a resolution on that scylla issue.

shotover-proxy/tests/cassandra_int_tests/routing.rs
```diff
@@ -204,7 +229,13 @@ mod composite_key {
 
     async fn simple_test(shotover: &CassandraConnection, cassandra: &CassandraConnection) {
         create_keyspace(shotover).await;
+        if shotover.is(&[CassandraDriver::Scylla]) {
+            tokio::time::sleep(std::time::Duration::from_secs(20)).await;
+        }
         create_table(shotover).await;
+        if shotover.is(&[CassandraDriver::Scylla]) {
```
conorbros, line 236: Are we better off not introducing a workaround just to update? Might as well just wait for the fix

rukai: Oh, I agree, I just wanted to confirm that was the problem.

Also handy to have around in case we do need to land this early for some other scylla fix we need.

conorbros (approved):

benbromhead (approved):

PR #1069, opened 2023-03-07
statically linked custom transforms demo

rukai:
Not for reviewing or merging, just for demonstrating a working demo of statically linked custom transforms.

PR #1101, opened 2023-03-27
Allow for custom transforms via using shotover as a crate

rukai:
Implements custom transforms in the least invasive way possible: we simply add a single `Custom` variant to the `Transforms` enum, sidestepping any discussion about dynamic dispatch for internal transforms.

An example/integration test is provided at `custom-transforms-example`.

We could split redis_get_rewrite into its own crate like we did in the [demo](https://github.com/shotover/shotover-proxy/pull/1069) but I didnt do that in order to avoid adding extra noise to the root of the workspace for an example.

I would however like to create a demo that points at the crates.io version, in its own git repo, to act as a template. In that demo it would be good to demonstrate the crate split.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -853,7 +853,6 @@ version = "0.0.1"
 dependencies = [
  "anyhow",
  "async-trait",
- "clap 4.2.1",
  "redis",
  "serde",
  "shotover",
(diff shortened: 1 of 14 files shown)


conorbros (approved):

benbromhead (approved):
LGTM

PR #1102, opened 2023-03-27
Split shotover library/binary into separate crates

rukai:
Naming is up for discussion but I am proposing:
* shotover the library is called `shotover`
   + Shorter to write when referring to in the application/custom transform code
* shotover the binary is called `shotover-proxy`
   + Name is unique from the library, allows for distinguishing between lib and bin in both human discussion and also within cargo. For example users of custom transforms will now be able to name their binary `shotover-proxy` to match upstream if they desire, previously that was impossible because of the collision with the library name.
   
The `shotover` library will contain all of shotover, the `shotover-proxy` binary is just there to statically link in custom transforms.

## Alternatives:

Shotover the product is called just "shotover" instead of "shotover proxy" these days.
So the `-proxy` postfix doesnt mean as much, maybe we should go for something more descriptive like `shotover-bin`?


conorbros:
+1 for `shotover-bin`

conorbros (approved):

benbromhead (approved):
+1 shotover-bin

rukai:
The people have spoken!
Once this lands I'll do a follow up PR for shotover-proxy -> shotover-bin

PR #1103, opened 2023-03-28
Cleanup shotover::Runner api

rukai:
To give custom transform users a nicer setup process.

List of changes:
* User no longer has to provide ConfigOpts
* The user no longer has to manually set `with_observability_interface`, this method was originally added as a hack for specific integration tests but that hack is no longer needed so the function can just be removed and the observability interface always enabled.
* `Runner::new()` now internally performs the backup tracing reporter logic that was previously in `shotover-proxy/src/main.rs`
* run_block now returns `!`
* `Runner` renamed to `Shotover` as from the users perspective that type is shotover.

We could have just done a single `run_shotover()` method but `Shotover::new().run_block()` adds barely any more complication and gives us room to expand without breaking changes in the future.

### DIFF
--- a/shotover-proxy/src/main.rs
+++ b/shotover-proxy/src/main.rs
@@ -1,33 +1,5 @@
-#![warn(rust_2018_idioms)]
-#![recursion_limit = "256"]
+use shotover::runner::Shotover;
 
-use anyhow::{Context, Result};
-use clap::Parser;
-use shotover::runner::{ConfigOpts, Runner, TracingState};
-use tokio::runtime::Runtime;
-
-fn main() -> Result<()> {
-    let opts = ConfigOpts::parse();
-    let log_format = opts.log_format.clone();
-    match run(opts) {
-        Ok(()) => Ok(()),
-        Err(err) => {
-            let rt = Runtime::new()
-                .context("Failed to create runtime while trying to report {err:?}")
-                .unwrap();
-            let _guard = rt.enter();
-            let _tracing_state = TracingState::new("error", log_format)
-                .context("Failed to create TracingState while trying to report {err:?}")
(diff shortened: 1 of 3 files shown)


shotover-proxy/src/runner.rs
```diff
@@ -64,44 +65,74 @@ impl Default for ConfigOpts {
...
     runtime: Runtime,
     topology: Topology,
     config: Config,
     tracing: TracingState,
 }
 
-impl Runner {
-    pub fn new(params: ConfigOpts) -> Result<Self> {
+impl Shotover {
+    #[allow(clippy::new_without_default)]
```
conorbros, line 76: Why no default?

rukai: Its just a strange thing to want to do (just call new, its shorter!)
And I didnt want to have to commit to having default in our API.

shotover-proxy/src/runner.rs
```diff
@@ -64,44 +65,74 @@ impl Default for ConfigOpts {
...
+        Ok(Shotover {
             runtime,
             topology,
             config,
             tracing,
         })
     }
 
-    pub fn with_observability_interface(self) -> Result<Self> {
+    pub fn start_observability_interface(
```
conorbros, line 119: This can be private?

conorbros (approved):
LGTM. Just that one comment

shotover-proxy/src/runner.rs
rukai: good catch

benbromhead (commented):
As this is a clean up for an external consumer of the api, do you want to add some doc strings to run_block?

benbromhead (approved):

PR #1104, opened 2023-03-28
Cassandra protocol v5 

conorbros:
Will supersede the previous v5 PR. This PR implements support for the Cassandra v5 protocol without support for compression or non-self contained frames, which will be completed in followups.


conorbros:
Oops, I'll pull out that flamegraph change.

shotover-proxy/src/codec/cassandra.rs
```diff
@@ -67,96 +153,267 @@ impl CassandraDecoder {
...
+        let version: Version = self.version.load(Ordering::Relaxed).into();
+        let compression: Compression = self.compression.load(Ordering::Relaxed).into();
+        let handshake_complete = self.handshake_complete.load(Ordering::Relaxed);
+
+        loop {
+            match self.check_size(src, version, compression, handshake_complete) {
+                Ok(frame_len) => {
+                    let mut messages = self
+                        .decode_frame(src, frame_len, version, compression, handshake_complete)
+                        .unwrap(); // TODO
```
conorbros, line 389: convert this to a `CodecReadError`

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -867,23 +867,23 @@ async fn test_protocol_v4(#[case] driver: CassandraDriver) {
...
 #[tokio::test(flavor = "multi_thread")]
 #[serial]
-async fn test_protocol_v5(#[case] _driver: CassandraDriver) {
+async fn test_protocol_v5(#[case] driver: CassandraDriver) {
     let _docker_compose =
         DockerCompose::new("example-configs/cassandra-passthrough/docker-compose.yaml");
 
     let shotover = ShotoverProcessBuilder::new_with_topology(
-        "example-configs/cassandra-passthrough/topology-encode.yaml",
+        "example-configs/cassandra-passthrough/topology.yaml",
```
rukai, line 875: why are we losing `-encode`?
Does it not support encoding?

conorbros: Mistake, I must've turned off during dev

shotover-proxy/src/codec/cassandra.rs
```diff
@@ -278,54 +544,136 @@ impl Encoder<Messages> for CassandraEncoder {
...
-                    if let Frame::Cassandra(CassandraFrame {
-                        operation: CassandraOperation::Startup(startup),
-                        ..
-                    }) = &frame
-                    {
-                        set_compression(&mut self.compression, startup);
-                    };
 
-                    let buffer = frame.into_cassandra().unwrap().encode(compression);
+                put3b(&mut dst[header_start..], payload_len as i32);
```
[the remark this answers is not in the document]
rukai: I was curious so I checked the assembly of both and the copy_from_slice approach seems to produce better assembly, I assume because it doesnt have to do a bounds check on every index into buffer in put3b
image

rukai (commented):
Heres some feedback so far, I'll continue the review tomorrow.

shotover-proxy/src/codec/cassandra.rs
```diff
@@ -67,96 +153,267 @@ impl CassandraDecoder {
...
+    ) -> Result<usize, CheckFrameSizeError> {
+        match (version, handshake_complete) {
+            (Version::V5, true) => match compression {
+                Compression::None => {
+                    let buffer_len = src.len();
+                    if buffer_len < UNCOMPRESSED_FRAME_HEADER_LENGTH {
+                        return Err(CheckFrameSizeError::NotEnoughBytes);
+                    }
+
+                    let header = if buffer_len >= 8 {
```
rukai, line 308: why are we branching on the buffer len here? that doesnt sound right.

shotover-proxy/src/codec/cassandra.rs
```diff
@@ -67,96 +153,267 @@ impl CassandraDecoder {
...
+        version: Version,
+        compression: Compression,
+        handshake_complete: bool,
+    ) -> Result<Vec<Message>> {
+        match (version, handshake_complete) {
+            (Version::V5, true) => match compression {
+                Compression::None => {
+                    let mut frame_bytes = src.split_to(frame_len);
+
+                    let header = if frame_bytes.len() >= 8 {
```
rukai, line 189: same as in the encode

rukai:
I've realized that none of our benchmarks are actually running in protocol v5:
* our criterion benches use casandra-cpp which uses v4 by default (but supposedly supports v5)
* our latte benches uses scylla driver under the hood which only supports protocol v4

We will need to get access to some benchmarks before we can land this as modern drivers will enable v5 by default.

A way avoid this unbenchmarked protocol support from hitting production while still landing this PR would be setting a feature flag (that is enabled in tests) such that we reject any attempts to connect with v5 for now.

shotover-proxy/src/codec/cassandra.rs
```diff
@@ -278,54 +544,136 @@ impl Encoder<Messages> for CassandraEncoder {
...
-                    }) = &frame
-                    {
-                        set_compression(&mut self.compression, startup);
-                    };
 
-                    let buffer = frame.into_cassandra().unwrap().encode(compression);
+                put3b(&mut dst[header_start..], payload_len as i32);
+                put3b(
+                    &mut dst[header_start + 3..],
+                    cassandra_protocol::crc::crc24(&payload_len.to_le_bytes()[..3]),
```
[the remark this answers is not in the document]
conorbros: I'm going to add this as something to checkout on the issue.

shotover-proxy/src/codec/cassandra.rs
```diff
@@ -32,25 +101,42 @@ impl CodecBuilder for CassandraCodecBuilder {
...
+                handshake_complete.clone(),
+            ),
+            CassandraEncoder::new(version, compression, self.direction, handshake_complete),
         )
     }
 }
 
 pub struct CassandraDecoder {
-    compression: Arc<RwLock<Compression>>,
+    version: Arc<AtomicVersionState>,
```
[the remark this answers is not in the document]
conorbros: I'll also add this to the issue.

shotover-proxy/src/codec/cassandra.rs
```diff
@@ -67,96 +153,267 @@ impl CassandraDecoder {
...
+                            break;
+                        }
+
+                        let body_len =
+                            i32::from_be_bytes(payload[5..9].try_into().unwrap()) as usize;
+
+                        let envelope_len = ENVELOPE_HEADER_LEN + body_len;
+
+                        if envelope_len > payload.len() {
+                            break;
```
rukai, line 250: This should be treated as an error as its a violation of the cassandra protocol for the body of a frame to be longer than the bytes of the message.
Also it should probably be:
```rust
if envelope_len > payload.len() - CRC32_LEN {
```

shotover-proxy/src/codec/cassandra.rs
```diff
@@ -67,96 +153,267 @@ impl CassandraDecoder {
...
-                        crate::message::ProtocolType::Cassandra {
-                            compression: if compressed {
-                                *self.compression.read().unwrap()
-                            } else {
-                                Compression::None
+                    let mut envelopes: Vec<Message> = vec![];
+
+                    loop {
+                        if payload.len() < ENVELOPE_HEADER_LEN {
+                            break;
```
rukai, line 241: We should combine this with the loop making it into a while loop.
I think the logic is also slightly off.

```rust
while payload.len() >= CRC32_LEN {
```

conorbros: I've split off to the end of the payload, not the end of the frame.

shotover/src/codec/cassandra.rs
```diff
@@ -67,96 +153,246 @@ impl CassandraDecoder {
...
+    ) -> Result<usize, CheckFrameSizeError> {
+        match (version, handshake_complete) {
+            (Version::V5, true) => match compression {
+                Compression::None => {
+                    let buffer_len = src.len();
+                    if buffer_len < UNCOMPRESSED_FRAME_HEADER_LENGTH {
+                        return Err(CheckFrameSizeError::NotEnoughBytes);
+                    }
+
+                    let header = i64::from_le_bytes(src[..8].try_into().unwrap()) & 0xffffffffffff; // convert to 6 byte int
```
rukai, line 297: The buffer_len check uses UNCOMPRESSED_FRAME_HEADER_LENGTH which is 6
However here we are exceeding that by reading 8 bytes.

I think the best fix here would be:
```rust
let payload_length = (u32::from_le_bytes(src[..4].try_into().unwrap()) & 0x1ffff) as usize;
```

conorbros: Yeah I'm fine with that, and I assume you mean use that in both `check_size` and `decode_frame`?

shotover/src/codec/cassandra.rs
```diff
@@ -67,96 +153,246 @@ impl CassandraDecoder {
...
-                    let bytes = src.split_to(frame_len);
-                    tracing::debug!(
-                        "{}: incoming cassandra message:\n{}",
-                        self.direction,
-                        pretty_hex::pretty_hex(&bytes)
-                    );
-
-                    let version = Version::try_from(bytes[0])
-                        .expect("Gauranteed because check_envelope_size only returns Ok if the Version will parse");
-                    if let Version::V3 | Version::V4 = version {
```
rukai, line 116: Because new versions can be added without a breaking bump to `cassandra-protocol`s we should keep these:
```rust
if let Version::V3 | Version::V4 | Version::V5 = version {
```

Also we probably want to make the Version::V5 dependent on `alpha-transforms` unless we want to land benchmarks first.

conorbros: We can't know what the version byte will be in new versions though

shotover/src/codec/cassandra.rs
```diff
@@ -67,96 +153,246 @@ impl CassandraDecoder {
...
+                        ));
+                    }
+
+                    Ok(envelopes)
+                }
+                _ => unimplemented!(),
+            },
+            (_, _) => {
+                let bytes = src.split_to(frame_len);
+                tracing::debug!(
```
rukai, line 258: Can we put this in a `for message in messages {` after the loop so that it still runs for v5?

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -877,13 +877,96 @@ async fn test_protocol_v5(#[case] _driver: CassandraDriver) {
...
+    };
 
-    // standard_test_suite(&connection, driver).await;
+    standard_test_suite(&connection, driver).await;
 
     shotover.shutdown_and_then_consume_events(&[]).await;
 }
+
+#[rstest]
+#[case::cdrs(CdrsTokio)]
```
rukai, line 892: Can we do something like `#[case::cdrs_v5(CdrsTokio, V5)` to avoid duplicating this test?

conorbros: I'll actually just remove this test and leave the pass through ones for now. Need to come up with a better testing plan for v5, once the compression stuff is done that may become more clear.

rukai: sounds good, lets land this

rukai (approved):
Thankyou for all your efforts in getting this through.
Lets land this!

benbromhead (approved):
Good work on this one!

PR #1108, opened 2023-03-30
Replace openssl with rustls

rukai:
The implementation does end up a bit more complicated than with openssl, but the replacement of a C dependency with a rust one is definitely a win for memory safety and ease of building.

### DIFF
--- a/custom-transforms-example/src/main.rs
+++ b/custom-transforms-example/src/main.rs
@@ -1,6 +1,7 @@
 use shotover::runner::Shotover;
 
 mod redis_get_rewrite;
+shotover::import_transform!(redis_get_rewrite::RedisGetRewriteConfig);
 
 fn main() {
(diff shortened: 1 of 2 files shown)


conorbros (approved):

benbromhead (approved):

shotover/Cargo.toml
```diff
@@ -78,3 +78,7 @@ chacha20poly1305 = { version = "0.10.0", features = ["std"] }
 generic-array = { version = "0.14", features = ["serde"] }
 dyn-clone = "1.0.10"
 kafka-protocol = "0.6.0"
+rustls = { version = "0.21.0", features = ["dangerous_configuration"] }
```
benbromhead, line 81: Might be worth a comment about why we allow dangerous config (e.g. some DBs we support do bad things (tm))

PR #1110, opened 2023-03-31
Cleanup gen_certs.sh

rukai:
* The truststore is unused so no need to generate it 
* No need to restore the current dir, unix shells (unlike powershell) do that automatically


conorbros (approved):

rukai:
superceded by https://github.com/shotover/shotover-proxy/pull/1111

PR #1111, opened 2023-03-31
Replace gen_certs.sh with rust

rukai:
supercedes https://github.com/shotover/shotover-proxy/pull/1110

Completely replaces cassandra gen_certs.sh with the same logic used for generating redis certs + extra 1 openssl command to generate the p12 file.

I've renamed all the generated certs from `redis` -> `localhost` so we dont have cassandra certs named redis :P

There was a stray `generate_cassandra_test_certs();` that was causing issues because it was regenerating the certs mid test run so I removed it.

## Motivation

The certificate generated by gen_certs.sh is not a valid x509 v3 cert (its probably v1 or v2), although I think the ca cert is v3.
This is causing issues with rustls which accepts only x509 v3 rejecting v1/v2.
Trying to figure out what magic flags and settings I need to give openssl to make the cert v3 proved impossible for me.
So instead I opted to just reuse what we already have for redis + keep the p12 generation command which is working fine.

I've verified that this change fixes the issue I was hitting with rustls.


conorbros (approved):
I still get a peer cert not found on this branch with a fresh clone, only on the Datastax driver. Must be something specific to my machine since it's passing on the CI.

rukai:
The datastax TLS tests have been failing on main for me so I've been unable to test them :/

benbromhead:
Welcome to the wild world of openssl compatibility

benbromhead (approved):

PR #1113, opened 2023-04-02
Cassandra protocol v5 benches

conorbros:


conorbros:
After a bit of digging around it seems the only driver supporting v5 is cdrs-tokio. Even the native cpp driver and node drivers don't support it yet so these benchmarks will have to be written with cdrs-tokio.

rukai:
I'm going to go ahead and close this as it has been superceded by https://github.com/shotover/shotover-proxy/pull/1180

Issue #160, opened 2021-08-30
Finish Cassandra Protocol support


benbromhead:
Implement better version support (3.X and 4.0)
Support for things like:
- [x] prepared statements,
- [x] events
- [ ] v5 protocol support
    - [x] basic support #1104 
    - [x] v5 benchmarks https://github.com/shotover/shotover-proxy/pull/1113
    - [x] v5 compression https://github.com/shotover/shotover-proxy/pull/1118
    - [x] non-self contained frames https://github.com/shotover/shotover-proxy/pull/1141
    - [ ] re-encode frames with multiple envelopes when applicable  https://github.com/shotover/shotover-proxy/pull/1090

rukai:
@conorbros can this be closed now that https://github.com/shotover/shotover-proxy/pull/662 is merged?

rukai:
Oh, it does mention version support as well which still isnt done

conorbros:
https://github.com/shotover/shotover-proxy/pull/1104#discussion_r1150102407

conorbros:
https://github.com/shotover/shotover-proxy/pull/1141#discussion_r1179927697

PR #1118, opened 2023-04-04
Cassandra protocol v5 compression

conorbros:


shotover/src/transforms/cassandra/connection.rs
```diff
@@ -145,24 +145,51 @@ impl CassandraConnection {
     ///
     /// If an internal invariant is broken the internal tasks may panic and external invariants will no longer be upheld.
     /// But this indicates a bug within CassandraConnection and should be fixed here.
+    pub fn send_multiple(
```
rukai, line 148: Can this be pulled into a separate PR?

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -856,3 +856,31 @@ async fn test_protocol_v5_single(#[case] driver: CassandraDriver) {
...
+#[case::cdrs(CdrsTokio)]
+#[tokio::test(flavor = "multi_thread")]
+#[serial]
+async fn test_protocol_v5_compression(#[case] driver: CassandraDriver) {
+    {
+        let _docker_compose =
+            docker_compose("example-configs/cassandra-passthrough/docker-compose.yaml");
+
+        let shotover = ShotoverProcessBuilder::new_with_topology(
+            "example-configs/cassandra-passthrough/topology-encode.yaml",
```
rukai, line 870: We also need to test `cassandra-passthrough/topology.yaml`, in order to properly test the encoder.

shotover/src/codec/cassandra.rs
```diff
@@ -561,46 +659,106 @@ impl CassandraEncoder {
...
+                            &cassandra_protocol::crc::crc32(&dst[payload_start..]).to_le_bytes(),
+                        );
+                    }
+                    Compression::Lz4 => {
+                        let header_start = dst.len();
+
+                        dst.extend_from_slice(&[0, 0, 0, 0, 0, 0, 0, 0]);
+                        let payload_start = dst.len();
+
+                        let uncompressed_len = self.encode_compressed_payload(dst, m)?;
```
rukai, line 703: we probably need logic to avoid compressing for small messages.
Its a shame that information is lost when we split the frame into separate messages :/ 

We should add a TODO to the tracking issue to investigate this once we have proper benchmarks.
What we have is fine for now.
There are a few possible solutions:
* always compress or never compress
* transform configuration determines if we compress
* reimplement similar logic to the driver for determining if we compress.
* store a `compress: bool` somewhere so we compress messages that came in compressed. (maybe we could store it in the deprecated v4 compression flag :eyes: )

conorbros: Yep, we will also need something similar for non self-contained.

conorbros: https://github.com/shotover/shotover-proxy/issues/160 Added a todo on the issue

shotover/src/codec/cassandra.rs
```diff
@@ -7,18 +7,21 @@ use anyhow::{anyhow, Result};
 use atomic_enum::atomic_enum;
 use bytes::{Buf, BufMut, BytesMut};
 use cassandra_protocol::compression::Compression;
+use cassandra_protocol::crc::{crc24, crc32};
```
rukai, line 10: since we have this now, it would be good to go through and ensure we always use the imported version in this file instead of fully specifying it.

shotover/src/codec/cassandra.rs
```diff
@@ -343,6 +393,53 @@ impl CassandraDecoder {
...
+
+fn header_crc_mismatch_error(computed_crc: i32, header_crc24: i32) -> anyhow::Error {
+    anyhow!(format!(
+        "Header CRC mismatch - expected {header_crc24}, found {computed_crc}."
+    ))
+}
+
+fn payload_crc_mismatch_error(computed_crc: u32, payload_crc32: u32) -> anyhow::Error {
+    anyhow!(format!(
+        "Payload CRC mismatch - read {payload_crc32}, computed {computed_crc}."
```
rukai, line 441: lets make the expected/found and read/computed terminology consistent between these errors

shotover/src/codec/cassandra.rs
```diff
@@ -221,43 +219,75 @@ impl CassandraDecoder {
...
+
+                    let payload = if uncompressed_length == 0 {
+                        // protocol spec 2.2:
+                        // An uncompressed length of 0 signals that the compressed payload should be used as-is
+                        // and not decompressed.
+                        frame_bytes.split_to(compressed_len)
+                    } else {
+                        BytesMut::from(
+                            decompress(&frame_bytes.split_to(compressed_len), uncompressed_length)
+                                .unwrap() // TODO
```
rukai, line 281: We should return an error instead of panicking

shotover/src/codec/cassandra.rs
```diff
@@ -221,43 +219,75 @@ impl CassandraDecoder {
...
+                                .as_slice(),
```
rukai, line 282: This forces the memory to be reallocated and copied.

shotover/src/codec/cassandra.rs
```diff
@@ -221,43 +219,75 @@ impl CassandraDecoder {
...
+                        )
+                    };
+
+                    let envelopes = self.extract_envelopes_from_payload(payload)?;
+
                     Ok(envelopes)
                 }
-                _ => unimplemented!(),
+                _ => unimplemented!("Only Lz4 compression is supported for v5"),
```
rukai, line 290: This should return an error instead of panicking

shotover/src/codec/cassandra.rs
```diff
@@ -561,46 +659,107 @@ impl CassandraEncoder {
...
     }
 
-    fn encode_envelope(&mut self, dst: &mut BytesMut, m: Message) -> Result<()> {
-        let message_compression = m.codec_state.as_cassandra();
+    fn encode_compressed_payload(&mut self, dst: &mut BytesMut, m: Message) -> Result<usize> {
+        // TODO: always check if cassandra message
+        Ok(match m.into_encodable(MessageType::Cassandra)? {
+            Encodable::Bytes(bytes) => {
+                let uncompressed_len = bytes.len();
+                let compressed = compress(&bytes);
```
rukai, line 743: https://docs.rs/lz4_flex/latest/lz4_flex/fn.compress_into.html we should use compress_into to directly compress into dst, that will avoid the extra allocation + copy.

shotover/src/codec/cassandra.rs
```diff
@@ -561,46 +659,107 @@ impl CassandraEncoder {
...
+                    }
+                    Compression::Lz4 => {
+                        let header_start = dst.len();
+
+                        dst.extend_from_slice(&[0, 0, 0, 0, 0, 0, 0, 0]);
+                        let payload_start = dst.len();
+
+                        // TODO we should not be encoding small frames
+                        let uncompressed_len = self.encode_compressed_payload(dst, m)?;
+                        let compressed_len = dst.len() - payload_start;
```
rukai, line 705: we should return an error when compressed_len or uncompressed_len are longer than supported by the protocol.
Trying to debug that case would be a nightmare because the bits would overlap, an error here would save us from that possibility.

shotover/src/codec/cassandra.rs
rukai: ooooh my bad.
I thought the .as_slice() was on the output of decompress

shotover/src/codec/cassandra.rs
conorbros: In that case we should be sending the envelope as a non-self contained frame

rukai: thats fair, we should still add at least a:
```rust
if uncompressed_len < MAX_UNCOMPRESSED_LEN {
    todo!("non self contained frames not yet implemented.")
}
```

so that we dont run the risk of mangling the bits

shotover/src/codec/cassandra.rs
```diff
@@ -561,46 +657,108 @@ impl CassandraEncoder {
...
+                .into_cassandra()
+                .unwrap()
+                .encode(Compression::None)
+                .into(),
+        };
+
+        let uncompressed_len = bytes.len();
+        dst.resize(payload_start + get_maximum_output_size(uncompressed_len), 0);
+        let compressed_len = compress_into(&bytes, &mut dst[payload_start..])?;
+        dst.truncate(payload_start + compressed_len);
```
rukai, line 752: Does this have any effect? I would think that the earlier resize has already set it to this size.

shotover/src/codec/cassandra.rs
```diff
@@ -221,43 +219,73 @@ impl CassandraDecoder {
...
+                    frame_bytes.advance(COMPRESSED_FRAME_HEADER_LENGTH);
+
+                    let payload = if uncompressed_length == 0 {
+                        // protocol spec 2.2:
+                        // An uncompressed length of 0 signals that the compressed payload should be used as-is
+                        // and not decompressed.
+                        frame_bytes.split_to(compressed_len)
+                    } else {
+                        let mut payload = BytesMut::zeroed(uncompressed_length);
+                        decompress_into(&frame_bytes.split_to(compressed_len), &mut payload)?;
```
rukai, line 280: I figured out my confusion around vec conversions.
While you cant convert a Vec into a BytesMut, you can convert a Vec into a Bytes.
So we can simplify this back into a decompress call:

image

shotover/src/codec/cassandra.rs
rukai: ooooh right, because `get_maximum_output_size` only provides a maximum, makes sense :+1:

shotover/src/codec/cassandra.rs
```diff
@@ -561,46 +659,110 @@ impl CassandraEncoder {
...
+                        let header_start = dst.len();
+
+                        dst.extend_from_slice(&[0, 0, 0, 0, 0, 0, 0, 0]);
+                        let payload_start = dst.len();
+
+                        // TODO we should not be encoding small frames
+                        let (uncompressed_len, compressed_len) =
+                            self.encode_compressed_payload(dst, m, payload_start)?;
+
+                        if compressed_len > PAYLOAD_SIZE_LIMIT {
```
rukai, line 704: we also need to check that uncompressed_len is within its limit.
It could overflow into the self contained flag

conorbros: Won't `as u64` stop that?

shotover/src/codec/cassandra.rs
```diff
@@ -221,43 +219,75 @@ impl CassandraDecoder {
...
+                        )?
+                        .into()
+                    };
+
+                    let envelopes = self.extract_envelopes_from_payload(payload)?;
+
                     Ok(envelopes)
                 }
-                _ => unimplemented!(),
+                _ => Err(anyhow!("Only Lz4 compression is supported for v5")),
```
rukai, line 290: We should do the same for the other "Only Lz4 compression is supported for v5" panics

rukai (approved):
nice work!

benbromhead (approved):

PR #1125, opened 2023-04-11
Remove error module

rukai:
I feel that `ChainResponse` is not pulling its weight as an alias so I just removed it and used `Result<Messages>` directly instead.


conorbros (approved):
Agreed

benbromhead (approved):

PR #1126, opened 2023-04-12
Make DockerCompose image waiter independent of image version

rukai:
I'm going to think a bit more if I am happy with this approach


rukai:
The reason I needed this has been resolved by https://github.com/shotover/shotover-proxy/pull/1127

I'm not convinced the complexity provides enough convenience to be worth it, so I'm going to close this PR.

PR #1141, opened 2023-04-21
Non self contained frames support

conorbros:
<s>dependent on #1118</s>

First commit of this PR demonstrates that the test is sending a non-self contained frame to Shotover, once this PR is completed the test will pass.


shotover/src/codec/cassandra.rs
```diff
@@ -768,15 +878,11 @@ impl CassandraEncoder {
...
-        dst: &mut BytesMut,
-        m: Message,
-        envelope_compresson: Compression,
-    ) -> Result<()> {
+    fn encode_envelope(&mut self, m: Message, envelope_compresson: Compression) -> Result<Bytes> {
         // TODO: always check if cassandra message
-        match m.into_encodable(MessageType::Cassandra)? {
+        Ok(match m.into_encodable(MessageType::Cassandra)? {
             Encodable::Bytes(bytes) => {
+                // TODO this is a weird side effect to keep in this function
```
conorbros, line 885: Going to handle this in a followup PR, but this method has some weird side effects and needs to be called in a magic order for the codec to work that isn't clear at a first glance.

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -914,3 +914,102 @@ async fn test_protocol_v5_compression_encode(#[case] driver: CassandraDriver) {
...
+    for id in 0..1000 {
+        let statement = format!(
+            "INSERT INTO keyspace_name.my_table (id, data) VALUES ({}, 0x{})",
+            id,
+            hex::encode(random_bytes)
+        );
+        queries.push(statement);
+    }
+
+    connection.execute_batch(queries).await;
```
rukai, line 962: how did you verify this actually triggers the PAYLOAD_SIZE_LIMIT?
Also can this just be a test case in `standard_test_suite` so that we always test and dont have to pay for more database startup/shutdowns?

shotover/src/codec/cassandra.rs
```diff
@@ -673,65 +705,128 @@ impl CassandraEncoder {
...
+                                let (mut uncompressed_len, mut compressed_len) = self
+                                    .encode_compressed_payload_into_buffer(
+                                        dst,
+                                        &payload_bytes,
+                                        payload_start,
+                                    )?;
 
-                        let mut header =
-                            (compressed_len) as u64 | ((uncompressed_len as u64) << 17);
+                                if compressed_len >= PAYLOAD_SIZE_LIMIT {
```
rukai, line 775: I dont understand the logic in this branch, can you explain what its for?

It seems like: If the compressed payload needs to be split up fall back to an uncompressed payload.
But I dont understand why we would want to do that.

shotover-proxy/tests/cassandra_int_tests/mod.rs
conorbros: Adding the test to the current main branch shows that the test fails because it is above the payload max size, and the driver has sent it as a non self contained frame.


image

image

shotover/src/codec/cassandra.rs
conorbros: https://github.com/shotover/shotover-proxy/pull/1143

This is checking the same thing as this PR but I will just check against the uncompressed length to make it more clear.

shotover-proxy/tests/cassandra_int_tests/batch_statements.rs
```diff
@@ -43,132 +64,137 @@ async fn use_statement(connection: &CassandraConnection) {
     }
 }
 
-pub async fn test(connection: &CassandraConnection) {
-    // setup keyspace and table for the batch statement tests
-    {
-        run_query(connection, "CREATE KEYSPACE batch_keyspace WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 };").await;
-        run_query(connection, "CREATE TABLE batch_keyspace.batch_table (id int PRIMARY KEY, lastname text, firstname text);").await;
+async fn insert_batch(connection: &CassandraConnection) {
```
conorbros, line 67: pulled out the restructuring changes in this file to https://github.com/shotover/shotover-proxy/pull/1144

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -670,8 +670,6 @@ async fn request_throttling(#[case] driver: CassandraDriver) {
 
     tokio::time::sleep(std::time::Duration::from_secs(1)).await; // sleep to reset the window
 
-    batch_statements::test(&connection).await;
```
conorbros, line 673: The new `large_batch` test in this module triggers the request throttling so I've removed it. We are already testing that the rate limiter tests batches both below and above the threshold so it should be fine to remove.

shotover/src/codec/cassandra.rs
```diff
@@ -674,65 +706,115 @@ impl CassandraEncoder {
...
-                        let (uncompressed_len, compressed_len) =
-                            self.encode_compressed_payload(dst, m, payload_start)?;
+                                dst.extend_from_slice(&[0, 0, 0, 0, 0, 0, 0, 0]);
+                                let payload_start = dst.len();
 
-                        if compressed_len > PAYLOAD_SIZE_LIMIT {
-                            todo!("non self contained frames not yet implemented.")
-                        }
+                                let payload_bytes = envelope_bytes
+                                    .split_to(envelope_bytes.len().min(PAYLOAD_SIZE_LIMIT - 1));
```
rukai, line 767: If we take a section of payload of length PAYLOAD_SIZE_LIMIT and compress it, the final compressed output may be larger than PAYLOAD_SIZE_LIMIT.

If we apply some algebra to get_maximum_output_size we can derive:
```rust
fn maximum_input_size(output_size: usize) -> usize {
    ((output_size - 20) as f32 / 1.1) as usize
}
```

And then we can do `maximum_input_size(PAYLOAD_SIZE_LIMIT)` to get the maximum size we can use here.

conorbros: `encode_compressed_payload_into_buffer` checks if it's larger than `uncompressed_len` which is going to be at maximum `PAYLOAD_SIZE_LIMIT -1`

rukai: Oh, I missed that!
Yeah, that looks good, maybe just add a comment to make it clear that logic is not just for effeciency but also important for correctness?

```diff
diff --git a/shotover/src/codec/cassandra.rs b/shotover/src/codec/cassandra.rs
index 57bf38f2..7fb4b3bd 100644
--- a/shotover/src/codec/cassandra.rs
+++ b/shotover/src/codec/cassandra.rs
@@ -841,6 +841,8 @@ impl CassandraEncoder {
         let mut compressed_len = compress_into(bytes, &mut dst[payload_start..])?;
 
         // fallback to uncompressed data if its more efficient
+        // this is also important for correctness, if the compressed len is larger than the uncompressed len,
+        // the compressed payload could overflow the max message size that was checked against the uncompressed size
         if compressed_len > uncompressed_len {
             dst[payload_start..(payload_start + uncompressed_len)]
                 .copy_from_slice(&bytes[..uncompressed_len]);
```

rukai (approved):
nice work! Seems robust to me.

benbromhead (approved):

PR #1142, opened 2023-04-24
update deps

conorbros:
Fixes the Scylla driver connection helper with the solution described in https://github.com/scylladb/scylla-rust-driver/issues/681#issuecomment-1517815975


benbromhead (approved):

shotover-proxy/tests/cassandra_int_tests/routing.rs
```diff
@@ -17,6 +17,8 @@ mod single_key {
     pub async fn test(shotover: &CassandraConnection, cassandra: &CassandraConnection) {
         create_keyspace(shotover).await;
         create_table(shotover).await;
+        shotover.await_schema_agreement().await;
```
rukai, line 20: I think a simpler approach would be to just create the `cassandra` connection after we have created the keyspace and table.

Some issues with the current approach:

We already awaited schema agreement automatically due to the logic in `CassandraConnection`:
```rust
        if query.starts_with("CREATE") || query.starts_with("ALTER") || query.starts_with("DROP") {
            self.await_schema_agreement().await;
        }
```
In these cases the driver has automatically refreshed the metadata for us because it knows its needed.
So we are effectively refreshing the metadata twice every time we alter the schema.

It hides the fix in an unsuspecting function: `await_schema_agreement` says nothing of the internal state of the connection, just that the databases on the other side all agree on what the schema currently is.

piodul hinted that there was a way to fix this issue within scylla-rust-driver but that it wouldn't be robust due to a timing issue.
If thats the case then I suspect the other drivers lack robustness in the same way and we might find rare intermittent failures there one day unless we can refresh their metadata too.
Creating the second connection after the schema is setup would completely sidestep this issue.

conorbros: Sounds good to me

PR #1143, opened 2023-05-01
fallback to no compression if compressed len is longer

conorbros:
Lz4 compression replaces repeated occurrences of data with references to a single copy of that data existing earlier in the input - https://en.wikipedia.org/wiki/LZ77_and_LZ78#LZ77. The compressed size can be larger than the uncompressed in cases where there is random text without any repeating sequences. We should fallback to no compression when this occurs. This is how the java driver behaves: https://github.com/datastax/native-protocol/blob/78e62fd7c93afb2500c7db103f4bf7c9dfd6662d/src/main/java/com/datastax/oss/protocol/internal/SegmentCodec.java#L54

### DIFF
--- a/shotover/src/codec/cassandra.rs
+++ b/shotover/src/codec/cassandra.rs
@@ -761,9 +761,18 @@ impl CassandraEncoder {
                 .into(),
         };
 
-        let uncompressed_len = bytes.len();
+        let mut uncompressed_len = bytes.len();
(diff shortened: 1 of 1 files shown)


rukai (approved):
Nice improvement!
Thanks for the helpful writeup as well.

Maybe a small `// fallback to uncompressed data if its more efficient` in front of the `if ..` would make the logic more immediately obvious.

benbromhead (approved):

PR #1145, opened 2023-05-01
Add windsock library, make use of it for our examples benchmarks

rukai:
This PR introduces a new library called windsock that we will use for our full integration benchmarks that hit a database (named because it gives a highlevel measurement lacking precision)
It follows the idea I described under `2. realistic integration benchmarks` in https://github.com/shotover/shotover-proxy/issues/544

## this PR

I have big dreams for this library but for now it just solves the exponentially growing database `examples/` benchmarks.
This PR replaces our numerous `examples/` benchmarks with a single windsock example benchmark
This will also give us a nice speedup to our release build times, as building individual binaries is very expensive in release.

There is a mismatch between how the database benchmarks are done currently and how windsock plans to do them.
The existing benches compare with and without shotover, so I have kept that functionality within a single case to avoid regressing functionality.
But windsock will handle that generically in the future allowing comparison between any two benchmarks. e.g. we could compare with and without shotover like we do now but we could also do comparisons at different OPS or even entirely different databases.
That said I still think its good to convert them over early since we are able to do so without regressing anything.

I've tried to choose abstractions that will better suit later features.
Even though we currently only use tags to generate a name for the test, in the future we will use them for filtering which benches to run and also which results to include in a 1 on 1 numerical comparison or in a graph.

There are two ways for a bench to be configured.
* Internally - If its something that should show up as individual new bench, then multiple benches should be made, with the difference listed in the benches tags.
* Externally - If its something that should not show up as individual test but alters the tests purpose, it should be given as an arg to the `run` method  e.g. `flamegraph` which allows for investigation or `local` which allows for running the test locally during development.

This way the user can pass in a flag like `--flamegraph` to instruct the bench which functionality is desired but at the same time the user can easily run every possible bench without having to list out every bench combination that makes semantic sense.

The report sender will be used to allow separation of report generation and benchmarking logic.
The user can implement benchmarking logic for any desired database, sending reports back to windsock which will handle report generation for the user in the background.

## Initial development

During initial development of windsock we will keep it in the shotover repo so that we can move quickly without having to organize PRs and releases across repos.
Then at a later stage it will be moved to its own repo, to make that transition easier I kept shotover specific logic outside of windsock and have included its own example.

## Usage:
There is an equivalent for all of the removed benchmarks in windsock.
Usually with benches we want to run a single bench, so to do that the usage is:
First list all the available benches:
```
shotover-proxy> cargo run --example windsock --release -- --list
    Finished release [optimized] target(s) in 0.14s
     Running `target/release/examples/windsock --list`
Benches:
kafka,OPS=10000000,connections=TODO,message_type=write1000bytes,shotover=with_and_without,topology=single
cassandra,OPS=10000000,connections=TODO,message_type=write1000bytes,shotover=with_and_without,topology=single
cassandra,OPS=10000000,connections=TODO,message_type=write1000bytes,shotover=with_and_without,topology=cluster3
cassandra-mocked,OPS=10000000,connections=TODO,message_type=write1000bytes,shotover=with_and_without,topology=single
```
Then pick out a bench case and run it like:
```
shotover-proxy> cargo run --example windsock --release -- --name kafka,OPS=10000000,connections=TODO,message_type=write1000bytes,shotover=with_and_without,topology=single

    Finished release [optimized] target(s) in 0.14s
     Running `target/release/examples/windsock -- --file kafka,OPS=10000000,connections=TODO,message_type=write1000bytes,shotover=with_and_without,topology=single`
bench output...
```

In the future we will also be able to do tag level filtering to run multiple benches.
To run every bench against kafka that runs at an OPS of 1000 would be:
```
shotover-proxy> cargo run --example windsock --release -- name=kafka OPS=1000
```


conorbros (approved):

benbromhead (approved):
Looks super interesting!!!

PR #1148, opened 2023-05-04
Empty source chain mapping handling

### LINKED ISSUE #302: Shotover will shutdown immediately if the source to chain mapping is empty
If Shotover somehow gets to this point and the source to chain mapping is empty Shotover won't build any sources or chains and will immediately exit.

https://github.com/shotover/shotover-proxy/blob/4cca9e34e47691d8f1dc443d1934815e54b1aa5c/shotover-proxy/src/config/topology.rs#L122

We should be returning an error instead of silently shutting down.

conorbros:
closes #302


rukai (approved):
Nice! Thanks for doing this

benbromhead (approved):

PR #1149, opened 2023-05-04
Replace latte with windsock

rukai:
## vs latte

The goal of this PR is to reach feature parity with latte (for the features we care about) and replace it.
I have tried to recreate the exact benches we had setup in latte, not because they were particularly amazing benches but because it at least lets us compare windsock's results against lattes results as a basic sanity check.
Here are our results from windsock:
image

And here are the results from latte:

### cassandra,topology=single
shotover=none on left, shotover=standard on right
image
The shotover=standard results looks about equivalent to windsock, but without shotover windsock has about 2.5x the throughput.
latte also has worse latency
A bit unexpected, but given the scenario is equivalent I guess windsock can just achieve higher throughput than latte?
Maybe latte is hitting some performance cost due to its rune scripting?

### cassandra,topology=cluster3
shotover=none on left, shotover=standard on right
image
similar results to the topology=single case, windsock has 3x more throughput than latte without shotover and 2x more throughput that latte with shotover
latte also has worse latency

## comparison table

This PR introduces both a results table `--results-by-name` as shown above and a compare table `--compare-by-name` which compares all columns of results against a base column (the first column):
image

This comparison table shows the results of 3 different related benchmarks against the shotover=none,topology=single benchmark.

## Various notes
I have not used our `test-helpers/connection/cassandra.rs` abstractions because they were not written with performance in mind and are not suitable for benchmarking.

Kafka benches are untouched, the original bench results are still reported to stdout and their windsock results are just reported as 0 or NaN.

latte employs some intelligent statistics to determine if bench results are significant or not. In latte the red/green color coding is only employed when the results are significant.
windsock currently has no such logic, instead we just assume comparisons of less than +-5% are insignificant, but it would be nice to add it eventually.
I dont think its a blocker for replacing latte though.

The mock responses in test-helpers/src/mock_cassandra.rs had to be changed because latte uses an older version of scylla driver and the new version sends different messages to obtain its metadata.
Same as when that code was originally written: no need to review it, its just hacked together to get the driver past the init stage so that benchmarking can begin.


benbromhead:
I don't think we've paid a lot of attention to significance and at the moment perf is a smell rather than a science. 

LGTM

benbromhead (approved):

conorbros (approved):

PR #1152, opened 2023-05-05
cassandra protocol version metric

conorbros:
Add metrics for which protocol versions clients are using to connect with, this lets us observe what protocol versions clients are using. We could also expand this to driver's used and so on but Ben just mentioned protocol version for now so we can have an idea of v4 vs v5 adoption. The results can be viewed with: 

Run the cassandra passthrough example
`docker-compose -f tests/test-configs/cassandra-passthrough/docker-compose.yaml up -d`
`cargo run --all-features -- -t tests/test-configs/cassandra-passthrough/topology.yaml`

Open the metrics page: http://localhost:9001/metrics

Connect with cqlsh using protocol v4:
`cqlsh --protocol-version 4`

And protocol v5:
`cqlsh --protocol-version 5`


conorbros:
Ah wait, it's incrementing twice because I've put the counter in the both sides of the codec.

shotover/src/codec/cassandra.rs
```diff
@@ -465,6 +473,15 @@ fn set_startup_state(
         );
     }
 
+    if direction == Direction::Source {
+        match version {
+            Version::V3 => increment_counter!("client_protocol_version", "version" => "v3"),
```
rukai, line 478: Is it possible to use a metric handle here for better performance?

conorbros: Good point, ending up finding some optimisations where `CassandraCodecBuilder` was being created unnecessarily.

rukai (approved):
I'll defer to ben on the exact needs of this feature, but the implementation seems reasonable.

benbromhead:
To allow auditing of protocol versions in production... will allow us to evaluate whether we need to spend more time on improving V5 perf

benbromhead (approved):

PR #1164, opened 2023-05-10
cassandra compression benches

conorbros:
Add compression to Cassandra windsock benches


shotover-proxy/examples/windsock/main.rs
```diff
@@ -24,31 +24,73 @@ fn main() {
...
                 CassandraDb::Mocked,
                 Topology::Single,
                 Shotover::None,
+                Compression::None,
+            )),
+            Box::new(CassandraBench::new(
+                CassandraDb::Mocked,
+                Topology::Single,
+                Shotover::None,
+                Compression::Lz4,
```
rukai, line 81: I dont think this could work because the mocked cassandra doesnt support compression?

rukai:
The failure is due to benches these names no longer existing https://github.com/shotover/shotover-proxy/blob/d5541d15dfa30aa8e87c33716a7110b56af69e4a/.github/workflows/build_and_test.yaml#L77
The immediate fix is to just change them to use the new names, but once https://github.com/shotover/shotover-proxy/pull/1165 lands it wont be so brittle anymore

benbromhead (approved):

rukai (approved):
lgtm!

PR #1167, opened 2023-05-11
windsock: Operations each second

rukai:
Adds an operations each second metric which records how many operations occurred within a second of operation.
image

This is an important metric as it gives insight into how stable the bench is.
As you can see in the screenshot our existing benches are passably stable, however I was experimenting with some changes that had very unstable results across multiple runs and im hoping this insight into stability within a single run will help with my investigation.


benbromhead:
nice!

benbromhead (approved):

conorbros (approved):

PR #1165, opened 2023-05-10
windsock: Implement tag filtering

rukai:
closes https://github.com/shotover/shotover-proxy/issues/1153

This PR implements tag filtering for running benches (`[FILTER]`) and for displaying results (`--results-by-tag` and `--compare-by-tag`).

This PR provides the most common filter we need (a simple key/value match) and hooks it up to various locations that need a filter.
We can improve on the exact features we provide in our filtering later on but I do intend to keep it lean, maybe just a key level match that ignores the value?

This PR also improves some general error handling that was left in a poor state from the initial windsock PR.

The cli flags should be properly documented in `cargo run --release --example windsock -- --help`

## Some expected workflows
```shell
cargo run --release --example windsock # run all benches
cargo run --release --example windsock -- --results-by-tags name=cassandra # show a table of all cassandra benches
cargo run --release --example windsock -- --compare-by-tags "cassandra,OPS=unlimited,connections=128,message_type=read_bigint,shotover=standard,topology=single name=cassandra shotover=standard" # show a comparison of one cassandra bench against some others
```

```shell
cargo run --release --example windsock name=kafka # run all kafka benches
cargo run --release --example windsock -- --results-by-tags name=kafka # show all the benches we just ran
```

PR #1172, opened 2023-05-14
windsock: Tweak table format

rukai:
in response to: https://github.com/shotover/shotover-proxy/pull/1165#issuecomment-1545266596

before:
image

after:
image

@conorbros thoughts?

### DIFF
--- a/windsock/src/tables.rs
+++ b/windsock/src/tables.rs
@@ -107,6 +107,9 @@ fn base(reports: &[ReportArchive], table_type: &str, comparison: bool) {
         .into_iter()
         .collect();
     nonintersecting_keys.sort();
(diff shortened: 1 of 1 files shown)


benbromhead:
Just to bike shed for a hot second.... could the tables get shifted right a couple of tabs? 

Having "Operation Time Percentiles" crossing into the columns looks odd. IMHO it should either be column values, or the separator "=" characters that run through the columns

rukai:
Yep, thats a good point, how about this?

image

I shortened Operations to Ops to avoid losing too much horizontal space.

benbromhead (approved):

conorbros (approved):

Issue #1166, opened 2023-05-10
windsock: message rate handling


rukai:
This is a particularly finicky part of windsocks design but at the same time feels super important, so probably needs discussion before work begins on it.
Or maybe we'll just build a prototype which can be iterated on.

I dont expect to start working on this soon as its so complex and we can still get plenty of value from windsock without it.

## The problem

When measuring performance of databases we care about much more than just how many messages can I yeet through this thing. We want to know the performance characteristics of the database at different loads: does latency skyrocket? do responses start to fail?
I think a good starting point would be breaking these down into 4 load categories:
* basic load - highest load before negative characteristics observed
* medium load - lets define this as `average(basic_load, maximum_load)` - just gives us an idea of how it performs at a middle ground.
* maximum load - highest load before throughput stops increasing
* stress load - the maximum amount of messages the bencher can send, the goal is to see if the db falls over under extreme load

These are the points at which its useful to perform benchmarks because they give meaning to the results.
If we just have benchmarks at various hardcoded OPS say `ops=4000` and `ops=4000000` then just from reading those numbers its not clear what kind of a load they represent.
Additionally those numbers will need to be changed on every machine that they run on in order to hit the load point it is supposed to be testing.

## A proposed solution

windsock benches are not responsible for defining these load points.
Instead windsock will drive the bench implementation at different loads to explore the databases performance and locate these points itself.
Maybe that could look like windsock starting at a load of say 10 OPS and then doubling the load every second, if the characteristics match one of the definitions of our 4 load points then note that load down.

We will need a whole bunch of heuristics to:
* increase accuracy (maybe binary search within the current and previous load for a more accurate value)
* increase speed, jumping by more than double at a time or starting at a larger base value.

Latte's significance measurements could be really useful here.

But how do we go about actually integrating this into the windsock workflow.
If we are not careful we would end up with benches that do not compare properly because they are running at customized loads.
It would also be time consuming to find these values every run so that should be avoided.

A possible approach would be:
* benches have a configurable OPS, by default its unlimited but the user can pass in a `--operations-per-second 4000` flag to manually set the rate at which all benches run, this allows the user to manually bench at different rates depending on what is useful on their machine while also being able to compare the results sanely across different benches.
* The user can generate 3 extra benches (for basic, medium and maximum load points ) by running `--set-benches-at-load-points` and `--clear-benches-at-load-points` similar to the UX for https://github.com/shotover/shotover-proxy/issues/1158
[...]

PR #1175, opened 2023-05-15
windsock: add --operations-per-second flag

rukai:
first step towards: https://github.com/shotover/shotover-proxy/issues/1166
No matter what complete solution we end up with there, the user will still want to be able to manually set the OPS somehow.

This PR introduces the first helper to be used within a windsock bench implementation: `BenchTask`
Previously it was purely up to the user to implement required bench logic.
I think this is a good starting point but we can pull logic in and out of BenchTask as we start using it.
I think that it is important that everything in BenchTask can also be done manually because some people will have specific needs that it cant meet.

Ignore the "Ops per sec" results in the screenshots, it is broken and will be fixed by: https://github.com/shotover/shotover-proxy/pull/1174
Instead refer to the "Ops each second" results.

## --operations-per-second 100
image

## --operations-per-second 1000
image

## --operations-per-second 10000
image

The "Ops Each Second" instability appears to be due to the timing of these reports being sent being off due to tokio::time::sleep being best effort only:
```rust
            for _ in 0..15 {
                let second = Instant::now();
                tokio::time::sleep(Duration::from_secs(1)).await;
                reporter
                    .send(Report::SecondPassed(second.elapsed()))
                    .unwrap();
            }
```
This is an inaccuracy in the way we measure operation counting rather than an inaccuracy in the timing of running actual operations.
We could improve precision here by moving the Report::SecondPassed timestamp into `Report::QueryCompletedIn` but I'm worried about the performance implications of that. Might experiment with that approach later.

### DIFF
--- a/shotover-proxy/examples/windsock/cassandra.rs
+++ b/shotover-proxy/examples/windsock/cassandra.rs
@@ -1,5 +1,8 @@
 use async_trait::async_trait;
-use scylla::{transport::Compression as ScyllaCompression, SessionBuilder};
+use scylla::{
+    prepared_statement::PreparedStatement, transport::Compression as ScyllaCompression, Session,
+    SessionBuilder,
+};
 use std::{
     collections::HashMap,
     sync::Arc,
@@ -12,7 +15,9 @@ use test_helpers::{
     shotover_process::ShotoverProcessBuilder,
 };
 use tokio::sync::mpsc::UnboundedSender;
-use windsock::{Bench, Report};
+use windsock::{Bench, BenchTask, Report};
+
+const ROW_COUNT: usize = 1000;
 
 pub enum Compression {
     None,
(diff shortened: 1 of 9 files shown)


benbromhead (approved):
Throttling and exact operations per second stuff is very tricky - could always look at the leaky bucket approach or the token bucket approach.

For now this looks good, but check out the two above approaches

shotover-proxy/examples/windsock/cassandra.rs
```diff
@@ -90,10 +95,6 @@ impl Bench for CassandraBench {
                     Shotover::ForcedMessageParsed => "forced-message-parsed".to_owned(),
                 },
             ),
-            // TODO: run with different OPS values
-            ("OPS".to_owned(), "unlimited".to_owned()),
-            // TODO: run with different connection count
-            ("connections".to_owned(), "128".to_owned()),
```
conorbros, line 96: What was the plan for connections?

rukai: how many concurrent connections the operations are distributed over.
Might end up bringing it back eventually, not quite sure yet how to best decide the number of connections.

conorbros (approved):

PR #1170, opened 2023-05-12
Cassandra benches operations

conorbros:
Add `Operation::Read` and `Operation::Blob`

PR #1180, opened 2023-05-18
Windsock cassandra - add cdrs-tokio benches

conorbros:


shotover-proxy/examples/windsock/cassandra_protocols.rs
```diff
@@ -0,0 +1,210 @@
...
+        .iter()
+        .cloned()
+        .collect()
+    }
+
+    async fn run(
+        &self,
+        flamegraph: bool,
+        _local: bool,
+        _runtime_seconds: u32,
```
rukai, line 80: same as https://github.com/shotover/shotover-proxy/pull/1170/files#r1203455859
we also want to be handling `operations-per-second` arg which the BenchTask may be able to help with (or maybe BenchTask abstractions doesnt work with cdrs-tokio? lets discuss.)

shotover-proxy/examples/windsock/cassandra_protocols.rs
```diff
@@ -0,0 +1,219 @@
...
+                TcpSessionBuilder::new(RoundRobinLoadBalancingStrategy::new(), cluster_config)
+                    .build()
+                    .await
+                    .unwrap(),
+            );
+
+            let row_count = 1000usize;
+
+            session.query("CREATE KEYSPACE IF NOT EXISTS ks WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 }").await.unwrap();
+            // session.await_schema_agreement().await.unwrap();
```
rukai, line 141: do we need to replace these with a cdrs equivalent or does cdrs do this automatically?

shotover-proxy/examples/windsock/cassandra_protocols.rs
```diff
@@ -0,0 +1,219 @@
...
+            ),
+            (
+                "topology".to_owned(),
+                match self.topology {
+                    Topology::Single => "single".to_owned(),
+                    Topology::Cluster3 => "cluster3".to_owned(),
+                },
+            ),
+            (
+                "shotover".to_owned(),
```
rukai, line 64: lets use to_tag

shotover-proxy/examples/windsock/cassandra_protocols.rs
```diff
@@ -0,0 +1,219 @@
...
+                .await
+                .unwrap();
+            // session.await_schema_agreement().await.unwrap();
+            session
+                .query("CREATE TABLE ks.bench(id int PRIMARY KEY)")
+                .await
+                .unwrap();
+            // session.await_schema_agreement().await.unwrap();
+            tokio::time::sleep(Duration::from_secs(1)).await; //TODO: this should not be needed >:[
+                                                              //
```
rukai, line 153: ?

shotover-proxy/examples/windsock/cassandra_protocols.rs
conorbros: I'm not sure if it does it automatically but waiting for one second seems to work, I'd imagine it receives the keyspace change event in time.

rukai:
results on my machine:
image

As expected the scylla driver vastly outperforms cdrs, particularly when going through shotover.

V5 seems to regress performance both with and without shotover, this could just be a reflection of cdrs's v5 performance but from my memory of the protocol spec, the v5 protocol seemed difficult to implement efficiently.
I reran just the cdrs benches twice more and got similar results both times:
image


It would be nice if these benches were integrated with the other cassandra benches, under a tag of "driver=cdrs". The existing benches would then be "driver=scylla"
But I wont block this PR on that, we can easily do that in a cleanup and its not needed to bring value.

shotover-proxy/examples/windsock/cassandra_protocols.rs
```diff
@@ -0,0 +1,208 @@
...
+
+            for _ in 0..runtime_seconds {
+                let second = Instant::now();
+                tokio::time::sleep(Duration::from_secs(1)).await;
+                reporter
+                    .send(Report::SecondPassed(second.elapsed()))
+                    .unwrap();
+            }
+
+            tokio::time::sleep(Duration::from_secs(15)).await;
```
rukai, line 169: oh.
so thats why the benches felt slower.

conorbros:
I'll move them under the other Cassandra benches in this PR

shotover-proxy/examples/windsock/main.rs
```diff
@@ -1,4 +1,5 @@
 mod cassandra;
+//mod cassandra_protocols;
```
rukai, line 2: stray comment

rukai:
so we ended up recreating a cassandra connection similar to the one in test_helpers, what does this one do differently to justify not just using the one in test_helpers?

rukai (approved):
This approach seems reasonable, we can land it once you fix the stray comment

benbromhead (approved):

PR #1181, opened 2023-05-18
routing bug fix

conorbros:
The result to a prepare query in Cassandra has no `pk_indexes`: https://github.com/apache/cassandra/blob/2ff1ad4788a1e29b99f81f75b2966b7951ba8250/doc/native_protocol_v3.spec#L638

This was added in v4: https://github.com/apache/cassandra/blob/2ff1ad4788a1e29b99f81f75b2966b7951ba8250/doc/native_protocol_v4.spec#L1204

This means we aren't routing v3 execute requests properly and actually never had the ability to do so which is unfortunate but we should at least not be blowing up when it happens.

### DIFF
--- a/shotover/src/transforms/cassandra/sink_cluster/node_pool.rs
+++ b/shotover/src/transforms/cassandra/sink_cluster/node_pool.rs
@@ -27,6 +27,7 @@ pub enum GetReplicaErr {
     NoKeyspaceMetadata,
     NoNodeAvailable(anyhow::Error),
     Other(Error),
+    NoRoutingKey,
 }
 
 #[derive(Debug, Clone, PartialEq)]
@@ -198,7 +199,7 @@ impl NodePool {
             })?,
             version,
         )
-        .unwrap();
+        .ok_or(GetReplicaErr::NoRoutingKey)?;
 
         let replica_host_ids = self
             .token_map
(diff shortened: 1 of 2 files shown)


shotover/src/transforms/cassandra/sink_cluster/node_pool.rs
```diff
@@ -198,7 +199,7 @@ impl NodePool {
             })?,
             version,
         )
-        .unwrap();
+        .ok_or(GetReplicaErr::NoRoutingKey)?;
```
rukai, line 202: CassandraSinkCluster keeps track of the version, lets make use of that version to log a `tracing::error` if NoRoutingKey occurs and we are on a version other than v3.
Otherwise we could start silently losing routing without ever realizing.

conorbros: Yeah, I'll make that change. This is unlikely though, we were unwrapping before and hit no issues.

rukai: Right, i'm more concerned about accidentally breaking things in the future.
Say a new version of cassandra changes how this works or we change some of our routing implementation in shotover.

rukai (approved):
LGTM

benbromhead (approved):

PR #1182, opened 2023-05-19
server.rs refactor

rukai:
There are a few things going on here but no logic changes should have occurred in this PR.
1. our connection limiting logic was overly complicated because it was working around the lack of owned permits. Since then tokio has implemented [owned permits](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#method.acquire_owned) so we should just use those instead. The now deleted comments did a pretty good job of explaining the workaround so you should refer to them.
2. I moved as much logic as possible out of the socket acceptor loop and into the Handle task. We want to be able to accept many connections quickly so the string generation and nodelay setting is better off occurring in the async task.
3. I combined the `conn_details` and `client_details` strings into a single string, I figured there was no reason for `conn_details` to exist we may as well just always use `client_details`. `client_details` is important, it gets used in our metrics but `conn_details` only gets used in `tracing::debug` logs which are very rarely examined.
4. terminate_tasks_rx/terminate_rx now just relies on terminate_tx being dropped instead of a message being sent.

### DIFF
--- a/shotover/src/server.rs
+++ b/shotover/src/server.rs
@@ -13,7 +13,7 @@ use std::sync::Arc;
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::net::{TcpListener, TcpStream};
 use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
-use tokio::sync::{mpsc, watch, Semaphore};
+use tokio::sync::{mpsc, watch, OwnedSemaphorePermit, Semaphore};
 use tokio::task::JoinHandle;
 use tokio::time;
 use tokio::time::timeout;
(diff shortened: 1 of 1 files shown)


shotover/src/server.rs
```diff
@@ -627,26 +574,6 @@ impl<C: CodecBuilder + 'static> Handler<C> {
...
-        // This is done in a `Drop` implementation in order to guarantee that
-        // the permit is added even if the task handling the connection panics.
-        // If `add_permit` was called at the end of the `run` function and some
-        // bug causes a panic. The permit would never be returned to the
-        // semaphore.
-
-        self.limit_connections.add_permits(1);
-
-        if let Some(terminate_tasks) = &self.terminate_tasks {
-            terminate_tasks.send(()).ok();
```
rukai, line 646: oh whoops, forgot to finish this bit

conorbros (approved):
LGTM, getting rid of the some of the leftovers from the mini-redis codebase

benbromhead (approved):

Issue #1188, opened 2023-05-25
kafka: handle produce with acks=0


rukai:
When a kafka produce message has the ack field set to 0, the server does not send a response.
This breaks shotovers assumption that every request has a corresponding response.

A possible solution is having `Request { return_chan: None, .. }` make the tx/rx tasks skip waiting for a response. In this case they would immediately return with a dummy message.

PR #1189, opened 2023-05-25
kafka: fix produce acks0

rukai:
closes https://github.com/shotover/shotover-proxy/issues/1188

When a kafka produce message has the ack field set to 0, the server does not send a response.
This breaks shotovers assumption that every request has a corresponding response.

shotover's transforms rely on every request having a corresponding response.
The responses must also come back in the same order that the requests went out in.
The transforms rely on these properties in order to match up a request and its corresponding response.
But also the outgoing tx/rx tokio tasks tend to rely on this as well.

The return_chan in `Request { return_chan: Option<oneshot::Sender<Response>>, ..}` was previously never set to None.
So I took that value and used it to represent the case where a request does not expect a response and therefore the rx task should skip processing that response.
This required some changes in cluster_connection_pool.rs to change the rx task to follow this new logic.

In order to maintain the transform invariants, the PR introduces a dummy message type that the codecs just ignore but allows transforms to maintain proper response ordering.

This implementation as a whole should allow us to maintain the lower latency provided by acks0: shotover sends the acks0 produce to the server but does not await its response and instead immediately returns the dummy message so it can begin processing the next requests.


benbromhead (approved):
Nice, glad to the solution didn't require too much of a rework

conorbros (approved):

PR #1192, opened 2023-05-26
windsock: add redis benches

rukai:
The redis benches largely follows the same structure as the cassandra benches.

I introduced a new redis driver (fred) because the `redis` crate is harder to drive from a benchmark due to requiring `&mut T` access to the driver to send a message. Fred however only requires `&T` to send messages making it easy to send messages from many tasks.
We might be able to wrap the redis crate in an MPSC receiving tokio task but it smelt like trouble so I just went with fred for now.
Might be worth doing a full comparison of the two crates one day but for now lets just land with fred.

Because there are so many bench combinations for redis I opted to use `itertools::iproduct` which conor discovered would be useful for generating all possible combinations of benches.

The results seem quite consistent, heres one run:
image

However there are some strange results, the TLS benches perform way better with shotover than without shotover:
image

### DIFF
--- a/shotover-proxy/examples/windsock/redis.rs
+++ b/shotover-proxy/examples/windsock/redis.rs
@@ -1,0 +1,293 @@
+use crate::common::Shotover;
+use anyhow::{anyhow, Context, Result};
+use async_trait::async_trait;
+use fred::{
+    prelude::*,
+    rustls::{Certificate, ClientConfig, PrivateKey, RootCertStore},
+};
+use rustls_pemfile::{certs, Item};
+use std::{
+    collections::HashMap,
+    fs::File,
+    io::BufReader,
+    sync::Arc,
+    time::{Duration, Instant},
+};
+use test_helpers::{
+    docker_compose::docker_compose,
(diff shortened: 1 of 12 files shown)


benbromhead (approved):

shotover-proxy/examples/windsock/redis.rs
```diff
@@ -0,0 +1,293 @@
...
+                .await;
+        }
+
+        if let Some(perf) = perf {
+            perf.flamegraph();
+        }
+    }
+}
+
+fn load_certs(path: &str) -> Vec<Certificate> {
```
conorbros, line 225: What's the reason for these `inner` functions?

rukai: avoids duplicating the filename in the error handling

rukai: If we end up adding more drivers that also expose rustls like this then we can pull this logic into a test-helpers module.

conorbros (approved):

PR #1204, opened 2023-06-01
Update to rust 1.70 🦀🦀🦀

rukai:


conorbros (approved):
When are we switching to crablang? :p

benbromhead (approved):

PR #1215, opened 2023-06-13
Cassandra Websocket support

conorbros:
The objective of this PR is to add support for CQL over Websockets through Shotover. That means having a Websocket listener on Shotover that a compatible driver can send CQL messages to and have Shotover pass these messages to Cassandra through the traditional TCP socket, receive the responses and return it over the Websocket protocol to the client.

At the moment the only instance of this in the wild is an experiment from last year by the Scylla team that is not in production: https://www.scylladb.com/2022/08/11/scylladb-student-projects-cql-over-websocket/.

I am using the `tokio-tungstenite` library which handles all the websocket internals and lets us just feed the data of a binary Websocket message (which will be raw Cassandra protocol bytes) into our codecs. I also make use of the `Sec-WebSocket-Protocol` header to specify `"cql"` as the sub-protocol being used. Some examples of other database technologies using this: https://www.iana.org/assignments/websocket/websocket.xml.

To test this I began work on a simple Cassandra driver: https://github.com/conorbros/cql-ws

The design for configuring the Websocket listener in the `topology.yaml` is as follows, open to other suggestions as I just went with the simplest one to start with. 

```YAML
---
sources:
  cassandra_prod:
    Cassandra:
      listen_addr: "127.0.0.1:9042"
      websocket: true
chain_config:
  main_chain:
    - CassandraSinkSingle:
        remote_address: "127.0.0.1:9043"
        connect_timeout_ms: 3000
source_to_chain_mapping:
  cassandra_prod: main_chain
```

Moving forward: 

 - make it robust, handle the shutdown logic properly
 - flesh out the test driver more and get more coverage
 - investigate performance and what can be done to improve it


benbromhead:
Good start! We can flesh out test coverage and perf in a different PR - Getting the shutdown logic sorted will be good to land this patch

conorbros:
Alright, got the shutdown flow finished, just a couple error handling todos

conorbros:
I think now the shutdown flow is handled this can merged as is, I've started modifying the test driver to let us send different types of Websocket messages that will let us test the error handling for that. But as it stands this PR has the MVP functionality.

shotover-proxy/tests/test-configs/cassandra-passthrough-websocket/topology-encode.yaml
```diff
@@ -0,0 +1,15 @@
+---
```
rukai, line 1: I think this file is unused

shotover/src/server.rs
```diff
@@ -279,7 +412,8 @@ fn spawn_read_write_tasks<
     // 3. The writer task detects that the last out_tx is dropped by out_rx returning None and terminates
     //
     // client closes connection:
-    // 1. The reader task detects that the client has closed the connection via reader returning None and terminates, dropping in_tx and the first out_tx
+    // 1. The reader task detects that the client has closed the connection via reader returning None and terminates,
```
rukai, line 415: can you pull all these comment formatting changes into a separate PR?

rukai:
in terms of API, we could make it a field like `transport: WebSocket` that would let us extend to more types of transports in the future, e.g. `transport: Quic`

shotover/src/server.rs
```diff
@@ -255,6 +262,132 @@ pub struct Handler<C: CodecBuilder> {
...
+
+    // read task
+    tokio::spawn(async move {
+        loop {
+            tokio::select! {
+                result = reader.next() => {
+                    if let Some(ws_message) = result {
+                        match ws_message {
+                            Ok(WsMessage::Binary(ws_message_data)) => {
+                                let message = decoder.decode(&mut BytesMut::from(ws_message_data.as_slice()));
```
rukai, line 283: This an unfortunate clone, is there anything we can do about it?

shotover/src/server.rs
```diff
@@ -255,6 +262,132 @@ pub struct Handler<C: CodecBuilder> {
...
+                                match message {
+                                    Ok(Some(message)) => {
+                                        if in_tx.send(message).is_err() {
+                                            // main task has shutdown down, this task is no longer needed
```
rukai, line 287: `shutdown down`

shotover/src/server.rs
rukai: Hmmm... The only way this could happen is one of:
* we fork tungstenite to use a BytesMut instead of a Vec.
* BytesMut one day decides to make public its from_vec method.

For now lets just add a comment.
```
// Entire message is reallocated and copied here due to incompatibility between tokio codecs and tungstenite.
```

shotover/src/server.rs
```diff
@@ -255,6 +263,133 @@ pub struct Handler<C: CodecBuilder> {
...
+                                        unreachable!("CodecReadError::Io should not occur because we are reading from a newly created BytesMut")
+                                    }
+                                }
+                            }
+                            Ok(_ws_message) => {
+                                // TODO we need to tell the client about a protocol error
+
+                                todo!();
+                            }
+                            Err(err) => panic!("{err}") // TODO
```
benbromhead, line 322: Could we get rid of these todo / panics on the match? even if it just bails on the connection altogether and logs it instead of killing the whole shotover process

shotover/src/server.rs
```diff
@@ -255,6 +263,133 @@ pub struct Handler<C: CodecBuilder> {
...
+                    match encoder.encode(message, &mut bytes) {
+                        Err(err) => {
+                            error!("failed to encode message destined for client: {err:?}")
+                        }
+                        Ok(_) => {
+                            let message = WsMessage::binary(bytes);
+                            match writer.send(message).await {
+                                Ok(_) => {}
+                                Err(err) => {
+                                    panic!("{err}"); // TODO
```
benbromhead, line 353: Same as previous comments on panics

shotover/src/server.rs
```diff
@@ -255,6 +263,138 @@ pub struct Handler<C: CodecBuilder> {
...
+                                        return;
+                                    }
+                                    Err(CodecReadError::Parser(err)) => { // TODO we need to send a
+                                                                          // close message to the
+                                                                          // client, protocol error
+                                        warn!("failed to decode message: {:?}", err);
+                                        return;
+                                    }
+                                    Err(CodecReadError::Io(_err)) => {
+                                        unreachable!("CodecReadError::Io should not occur because we are reading from a newly created BytesMut")
```
rukai, line 313: I dont understand this unreachable.
What is the BytesMut it is referring to?
An IO error sounds very reachable to me.

shotover/src/server.rs
```diff
@@ -255,6 +263,138 @@ pub struct Handler<C: CodecBuilder> {
...
+                                error!("{err}");
+                                return;
+                            }
+                        }
+                    } else {
+                        return;
+                    }
+                }
+                _ = in_tx.closed() => {
+                    // main task has shutdown down, this task is no longer needed
```
rukai, line 332: `shutdown down`

shotover/src/server.rs
```diff
@@ -255,6 +263,138 @@ pub struct Handler<C: CodecBuilder> {
...
+                                            // main task has shutdown, this task is no longer needed
+                                            return;
+                                        }
+                                    }
+                                    Ok(None) => {
+                                        // websocket client has closed the connection
+                                        return;
+                                    }
+                                    Err(CodecReadError::RespondAndThenCloseConnection(messages)) => {
+                                        if let Err(err) = out_tx.send(messages) { // TODO we need
```
rukai, line 298: can this comment go on its own line to avoid being formatted like this?

shotover/src/server.rs
conorbros: line 285, we put the read data from the websocket binary message into a BytesMut.

rukai: oooh I see, the actual IO error would happen on line 321 :+1:

shotover/src/server.rs
```diff
@@ -255,6 +263,138 @@ pub struct Handler<C: CodecBuilder> {
...
+                    }
+                } else {
+                    // Main task has ended.
+                    // First flush out any remaining messages.
+                    // Then end the task thus closing the connection by dropping the write half
+                    while let Ok(message) = out_rx.try_recv() {
+                        let mut bytes = BytesMut::new();
+                        match encoder.encode(message, &mut bytes) {
+                            Err(err) => {
+                                error!("failed to encode message destined for client: {err:?}")
```
rukai, line 371: this should terminate the connection, if we miss a response then the connection is no longer usable.

shotover/src/server.rs
```diff
@@ -255,6 +263,138 @@ pub struct Handler<C: CodecBuilder> {
...
+
+    // write task
+    tokio::spawn(
+        async move {
+            loop {
+                if let Some(message) = out_rx.recv().await {
+                    let mut bytes = BytesMut::new();
+                    match encoder.encode(message, &mut bytes) {
+                        Err(err) => {
+                            error!("failed to encode message destined for client: {err:?}")
```
rukai, line 349: This should terminate the connection, if we miss a response then the connection is no longer usable.

shotover/src/server.rs
conorbros: I'll just fix it in this?

rukai: go ahead if you want

rukai:
This is pretty cool work!
Just fix the remaining comments and I'll be able to approve.

In order to make use of this in a meaningful way I imagine we'll want to modify an existing driver to get all the driver features one would expect.
Do we have any plans for that?

Also, have you considered wrapping websockets such that they can be exposed as a tokio AsyncRead/AsyncWrite?
I imagine that kind of abstraction would let us share the websockets logic between shotover and the driver and would make for an easier drop in replacement of TCP for shotover or any driver that uses AsyncRead/AsyncWrite.
I imagine it would also come at a performance cost though.

rukai (approved):
Lgtm!

conorbros:
https://github.com/snapview/tokio-tungstenite/issues/32

benbromhead (approved):
Nice!!

Issue #1214, opened 2023-06-13
KafkaSinkCluster


rukai:
We need a kafka sink that will allow us to run shotover on a different host from kafka.
For this use case performance is a critical concern.

### message rewriting

In order to ensure clients connect to shotover instead of the upstream kafka instances we need to rewrite messages to refer to shotover addresses instead.
But we need a way to tell which shotover instance's are assigned to which kafka instances.

For [CassandraSinkCluster](https://docs.shotover.io/shotover-blog/docs/transforms.html#cassandrasinkcluster) we defined each shotover instance as being responsible for a single cassandra rack.
This allows us to nicely assign a shotover node to a subset of the cluster.
However kafka has no such "rack" or "dc" level separation between nodes.
In order to load balance between multiple shotover nodes we could have the user set a list of known shotover nodes in the topology.yaml and then rewrite kafka addresses by `kafka_address = shotover_addresses[hash(kafka_address) % shotover_addresses.len()]`.
That way we have a consistent shotover address per kafka node in case the client relies on kafka nodes being consistent.

The transform config would look like:
```yaml
- KafkaSinkCluster:
    shotover_nodes:
      - "1.1.1.1:9092"
      - "1.1.1.2:9092"
      - "1.1.1.3:9092"
```
The list would include the nodes own address to simplify generating config.

### message routing
We also need to perform routing of messages to the kafka cluster.
These articles are useful background here: https://hevodata.com/learn/kafka-clusters/ https://www.oreilly.com/library/view/kafka-the-definitive/9781491936153/ch04.html
A produce message must go to the partition leader.
To achieve this we must keep track of messages sent by the client to build up a topology of the kafka cluster.
A consume message must go to the broker that contains the replica/original partition that the client was assigned when joining the group
JoinGroup messages must be routed to the "group coordinator broker".

There is quite likely more routing requirements that will be uncovered during development.

## Alternative approach - extend CassandraSinkSingle

No CassandraSinkCluster, just use CassandraSinkSingle and one shotover per kafka node

This has the benefits that:
* Avoids the complex code and performance cost of routing.
* Shotover does not need to know about other shotover nodes it just rewrites addresses from kafka to shotover using a consistent scheme. This custom address rewriting could easily go in a custom transform.

However there is a critical downside:
* Shotover and kafka clusters must be given addresses in some kind of consistent scheme.
   + One such scheme could be add one to the kafka ip address. e.g. kafka of 1.1.1.1 has a shotover of  1.1.1.2

PR #1219, opened 2023-06-15
Initial KafkaSinkCluster

rukai:
Completes the first chunk of work listed in https://github.com/shotover/shotover-proxy/issues/1214

Because the transform lacks routing it can only be used with a 1 node kafka cluster.
This still brings some value as we can host shotover on a different IP which KafkaSinkSingle does not allow.

The logic in the new `send_requests` method has been pulled out of the two sinks.

Introduces the xxhash-rust crate under the BSL license which should pose no issue as its just a more lenient version of the MIT license.

### DIFF
--- a/shotover-proxy/tests/test-configs/kafka/cluster/docker-compose.yaml
+++ b/shotover-proxy/tests/test-configs/kafka/cluster/docker-compose.yaml
@@ -1,0 +1,56 @@
+version: "3"
+networks:
+  cluster_subnet:
+    name: cluster_subnet
+    driver: bridge
+    ipam:
+      driver: default
+      config:
+        - subnet: 172.16.1.0/24
+          gateway: 172.16.1.1
+services:
+  kafka0:
+    image: 'bitnami/kafka:3.4.0-debian-11-r22'
+    networks:
+      cluster_subnet:
+        ipv4_address: 172.16.1.2
+    environment:
+      - KAFKA_CFG_LISTENERS=PLAINTEXT://:9092,CONTROLLER://:9093
+      - KAFKA_CFG_ADVERTISED_LISTENERS=PLAINTEXT://172.16.1.2:9092
+      - ALLOW_PLAINTEXT_LISTENER=yes
+      - KAFKA_KRAFT_CLUSTER_ID=abcdefghijklmnopqrstuv
+      - KAFKA_CFG_CONTROLLER_QUORUM_VOTERS=0@kafka0:9093,1@kafka1:9093,2@kafka2:9093
+      - KAFKA_CFG_NODE_ID=0
+    volumes:
+      - type: tmpfs
+        target: /bitnami/kafka
+  kafka1:
+    networks:
+      cluster_subnet:
+        ipv4_address: 172.16.1.3
+    image: 'bitnami/kafka:3.4.0-debian-11-r22'
+    environment:
+      - KAFKA_CFG_LISTENERS=PLAINTEXT://:9092,CONTROLLER://:9093
+      - KAFKA_CFG_ADVERTISED_LISTENERS=PLAINTEXT://172.16.1.3:9092
+      - ALLOW_PLAINTEXT_LISTENER=yes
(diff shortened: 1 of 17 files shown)


benbromhead:
BSL is good. 

Personally I like to just use the ASF categories as described in https://www.apache.org/legal/resolved.html to mentally check compatibility without too much brain power (we can obviously make decisions different to the ASF, but it's a good starting position).

shotover/src/transforms/kafka/sink_cluster.rs
```diff
@@ -0,0 +1,234 @@
...
+            )
+            .await?;
+            let (rx, tx) = tcp_stream.into_split();
+            self.outbound = Some(spawn_read_write_tasks(&codec, rx, tx));
+        }
+
+        let outbound = self.outbound.as_mut().unwrap();
+        let responses = send_requests(message_wrapper.messages, outbound)?;
+
+        // TODO: since kafka will never send requests out of order I wonder if it would be faster to use an mpsc instead of a oneshot or maybe just directly run the sending/receiving here?
```
benbromhead, line 137: Sounds like a good idea worth some investigation.

benbromhead (approved):

shotover-proxy/tests/test-configs/kafka/cluster/docker-compose.yaml
```diff
@@ -0,0 +1,56 @@
...
+      - KAFKA_CFG_NODE_ID=0
+    volumes:
+      - type: tmpfs
+        target: /bitnami/kafka
+  kafka1:
+    networks:
+      cluster_subnet:
+        ipv4_address: 172.16.1.3
+    image: 'bitnami/kafka:3.4.0-debian-11-r22'
+    environment:
```
conorbros, line 32: can use docker anchor for these ones

shotover/src/transforms/kafka/sink_cluster.rs
```diff
@@ -0,0 +1,234 @@
...
+
+fn hash_address(host: &str, port: i32) -> u64 {
+    let mut hasher = xxhash_rust::xxh3::Xxh3::new();
+    hasher.write(host.as_bytes());
+    hasher.write(&port.to_be_bytes());
+    hasher.finish()
+}
+
+fn rewrite_address(shotover_nodes: &[KafkaAddress], host: &mut StrBytes, port: &mut i32) {
+    //tracing::info!("rewriting {} {}", host, port);
```
conorbros, line 224: :p

shotover/src/transforms/kafka/sink_cluster.rs
```diff
@@ -0,0 +1,234 @@
...
+        // TODO: since kafka will never send requests out of order I wonder if it would be faster to use an mpsc instead of a oneshot or maybe just directly run the sending/receiving here?
+        let mut responses = if let Some(read_timeout) = self.read_timeout {
+            timeout(read_timeout, read_responses(responses)).await?
+        } else {
+            read_responses(responses).await
+        }?;
+
+        // Rewrite responses to use shotovers port instead of kafkas port
+        for response in &mut responses {
+            tracing::info!("{:?}", response.to_high_level_string());
```
conorbros, line 146: :p

rukai: ah yes, thanks

shotover/src/transforms/kafka/common.rs
```diff
@@ -0,0 +1,49 @@
...
+use tokio::sync::oneshot;
+
+pub fn send_requests(
+    messages: Vec<Message>,
+    outbound: &mut Connection,
+) -> Result<Vec<oneshot::Receiver<Response>>> {
+    messages
+        .into_iter()
+        .map(|mut message| {
+            let acks0 = if let Some(Frame::Kafka(KafkaFrame::Request {
```
conorbros, line 19: Could you add a comment with a small explanation for this bit

conorbros (approved):
LGTM

PR #1221, opened 2023-06-20
use iproduct for kafka benches aswell

conorbros:
Going to use `iproduct` for the Cassandra ones too, so made sense to switch the Kafka ones over in a pre-req PR.


rukai (approved):
nice!

benbromhead (approved):

PR #1222, opened 2023-06-20
KafkaSinkCluster: implement produce routing

rukai:
progress towards: https://github.com/shotover/shotover-proxy/issues/1214

With this PR we are now able to send produce messages to the kafka cluster.
Previously, due to incorrect routing, produce messages would likely end up at the wrong instance and be rejected.

I added a temporary `produce_only` test case to test the part of the test suite that we can pass so far.

I had to rewrite `common::send_requests` into `common::produce_channel` because the KafkaSinkCluster ended up needing more control over this logic then I expected.

The KafkaSinkCluster routing works by reading metadata messages that the client sends and using them to build up an internal topology structure (The `nodes` and `topics` fields) which are later used for routing messages. This is pretty similar to what we do for cassandra but we are relying purely on messages sent by the client without generating any ourselves.

In order to avoid complex message splitting I have assumed that all records in a message can be routed to the same kafka broker.
However this is not necessarily true since we rewrite messages, the client may end up putting two records that need to go to different brokers on the same message. We may have to revisit this logic later if this proves to be an issue.

I had to change the `cluster/docker-compose.yaml` because I realized it was misconfigured and not listening for clients.

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster.rs
+++ b/shotover/src/transforms/kafka/sink_cluster.rs
@@ -1,16 +1,19 @@
 use crate::codec::{kafka::KafkaCodecBuilder, CodecBuilder, Direction};
-use crate::frame::kafka::{strbytes, KafkaFrame, ResponseBody};
+use crate::frame::kafka::{strbytes, KafkaFrame, RequestBody, ResponseBody};
 use crate::frame::Frame;
-use crate::message::Messages;
+use crate::message::{Message, Messages};
 use crate::tcp;
-use crate::transforms::kafka::common::send_requests;
 use crate::transforms::util::cluster_connection_pool::{spawn_read_write_tasks, Connection};
-use crate::transforms::util::Response;
+use crate::transforms::util::{Request, Response};
 use crate::transforms::{Transform, TransformBuilder, Transforms, Wrapper};
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use async_trait::async_trait;
 use kafka_protocol::protocol::StrBytes;
+use rand::rngs::SmallRng;
+use rand::seq::SliceRandom;
+use rand::SeedableRng;
 use serde::Deserialize;
+use std::collections::HashMap;
 use std::hash::Hasher;
 use std::net::SocketAddr;
 use std::time::Duration;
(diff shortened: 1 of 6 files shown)


benbromhead:
> In order to avoid complex message splitting I have assumed that all records in a message can be routed to the same kafka broker. However this is not necessarily true since we rewrite messages, the client may end up putting two records that need to go to different brokers on the same message. We may have to revisit this logic later if this proves to be an issue.

Is this because of the way we do things, or is this actually allowed in the spec?

rukai:
> Is this because of the way we do things, or is this actually allowed in the spec?

Kafka brokers reject messages that they cant handle, the result is that all records in a message must be intended for the recipient.

However we are hiding the kafka cluster behind our shotover cluster.
This means the client may send a message to a shotover instance that it is told will be able to handle all the records, but in actual fact that message needs to be forwarded onto a kafka instance that may or may not be able to handle all of those records.

shotover/src/transforms/kafka/sink_cluster.rs
```diff
@@ -171,12 +181,56 @@ impl Transform for KafkaSinkCluster {
...
+                        }
                         rewrite_address(
                             &self.shotover_nodes,
                             &mut broker.1.host,
                             &mut broker.1.port,
                         )
                     }
+
+                    for topic in &metadata.topics {
+                        self.topics.insert(
```
benbromhead, line 215: This could _potentially_ grow to a significant size over the course of a long enough connection with sufficient topics / changes to topics? 

Is it worth clearing out old / stale entries based on the new metadata? Or is this a problem only in theory?

rukai: the metadata request specifies a topic to retrieve metadata on, so we cant use a metadata response to determine which topics no longer exist.

I think the best approach would be to remove entries if a metadata response returns whatever error indicates that the topic does not exist. This might not be sufficient though.

We could implement `self.topics` with some kind of LRU eviction.
But that would mean we have to fetch metadata on demand if we evict something that the client expects to still be there.

benbromhead: Ok thanks for the clarification I think just for the moment maybe evicting on the error would be sufficient - It will get cleared when a connection drops as well.

rukai: huh, turns out its a bit more complicated than that.
if kafka is configured with `auto.create.topics.enable=true` (the default setting) then it will return error code 3 but also immediately create the topic.

We could determine if the server is configured with auto.create.topics.enable=false` and then in that case use error code 3 to clear `self.topics` values.
But considering it doesnt even help us in the default config it doesnt feel worth it.

I think for now lets rely on closing the connection to clear the cache, and then we can investigate alternatives if the memory growth turns out to be too large.

conorbros (approved):

benbromhead (approved):

PR #1226, opened 2023-06-27
Websocket subprotocol header

conorbros:
The WebSocket protocol defines a header that can be used in the handshake to specifiy an application level subprotocol to use for communication for the reminder of the connection.

https://datatracker.ietf.org/doc/html/rfc6455#section-1.9

This PR adds the use of this header with the value "cql" for our Cassandra source. 

<s>Before this can be merged some upstream changes need to be merged, I will unmark as a draft once that happens.</s> Since the dependencies are in the test driver this can be merged as is.

### DIFF
--- a/shotover/src/sources/mod.rs
+++ b/shotover/src/sources/mod.rs
@@ -11,10 +11,10 @@ pub mod cassandra;
 pub mod kafka;
 pub mod redis;
 
-#[derive(Deserialize, Debug, Copy, Clone)]
+#[derive(Deserialize, Debug, Clone)]
 pub enum Transport {
     Tcp,
-    WebSocket,
+    WebSocket(#[serde(skip)] String),
 }
 
 #[derive(Debug)]
--- a/shotover/src/sources/mod.rs
+++ b/shotover/src/sources/mod.rs
@@ -11,10 +11,10 @@ pub mod cassandra;
 pub mod kafka;
 pub mod redis;
 
-#[derive(Deserialize, Debug, Clone)]
+#[derive(Deserialize, Debug, Clone, Copy)]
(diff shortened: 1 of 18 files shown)


rukai:
Why is specifying the protocol an optional flag on the driver?
Shouldnt the driver just always specify the protocol?

shotover/src/sources/mod.rs
```diff
@@ -11,10 +11,10 @@ pub mod cassandra;
 pub mod kafka;
 pub mod redis;
 
-#[derive(Deserialize, Debug, Copy, Clone)]
+#[derive(Deserialize, Debug, Clone)]
 pub enum Transport {
     Tcp,
-    WebSocket,
+    WebSocket(#[serde(skip)] String),
```
rukai, line 17: Rather than smuggling the protocol name in via the Transport Config, why dont we have the codec specify what protocol it would use for websockets.
```rust
pub trait CodecBuilder: Clone + Send {
    type Decoder: DecoderHalf;
    type Encoder: EncoderHalf;
    fn build(&self) -> (Self::Decoder, Self::Encoder);

    fn new(direction: Direction) -> Self;
    
    fn websocket_subprotocol(&self) -> &'static str;
}
```

rukai:
Is the only point of this to provide a better error when a websocket connection is made to an open websocket port for the wrong database?

conorbros:
Sort of? It's more just a common convention when using a subprotocol over WebSockets so it makes sense for Shotover to follow it. e.g. https://www.iana.org/assignments/websocket/websocket.xml

benbromhead (approved):

rukai (approved):
seems reasonable

Issue #1218, opened 2023-06-13
shotover raft investigation


rukai:
WIP.

There are two places shotover would benefit from cross-instance communication:
* Improved load balancing
* Simplified horizontal scaling

This issue addresses only "Simplified horizontal scaling" but our design should leave open the possibility for "Improved load balancing" to be added later

An investigation into what cross-instance communication could look like for shotover
A quick look at paxos vs raft says that raft is the simpler one of the two, so I'll start there.
I found this crate https://github.com/ritelabs/riteraft which builds higher level abstractions upon this crate https://github.com/tikv/raft-rs
Both seem to be unmaintained though :/

## Transform API design

Transforms will need to send their own custom data across raft.
Say for example CassandraSinkCluster needs to know the data_center, rack and host_id of all shotover nodes in the cluster. Currently this is all hardcoded into the config.

```yaml
    - CassandraSinkCluster:
        first_contact_points: ["172.16.1.2:9044", "172.16.1.3:9044"]
        local_shotover_host_id: "2dd022d6-2937-4754-89d6-02d2933a8f7a"
        shotover_nodes:
          - address: "127.0.0.1:9042"
            data_center: "dc1"
            rack: "rack1"
            host_id: "2dd022d6-2937-4754-89d6-02d2933a8f7a"
        connect_timeout_ms: 3000
```

As such I propose that transforms should be given an internal shotover API from which they can communicate over raft.
TODO: writeup internal API design.

## User facing design

From the users perspective, the transforms would have to be responsible for configuring the use of raft.
For example, we have the following topology.yaml

```yaml
sources:
  cassandra_prod:
    Cassandra:
      listen_addr: "127.0.0.1:9042"
chain_config:
  main_chain:
    - CassandraSinkCluster:
        first_contact_points: ["172.16.1.2:9042", "172.16.1.3:9042"]
        local_shotover_host_id: "2dd022d6-2937-4754-89d6-02d2933a8f7a"
        shotover_nodes:
          - address: "127.0.0.1:9042"
            data_center: "dc1"
            rack: "rack1"
            host_id: "2dd022d6-2937-4754-89d6-02d2933a8f7a"
          - address: "127.0.0.2:9042"
            data_center: "dc1"
            rack: "rack2"
            host_id: "3c3c4e2d-ba74-4f76-b52e-fb5bcee6a9f4"
          - address: "127.0.0.3:9042"
            data_center: "dc1"
            rack: "rack3"
[...]

PR #1228, opened 2023-06-28
KafkaSinkCluster fetch routing

rukai:
Progress towards https://github.com/shotover/shotover-proxy/issues/1214

Implements routing for Fetch, JoinGroup, OffsetFetch, Heartbeat and SyncGroup message types.
This is enough to get a single shotover instance cluster to route messages to a 3 node kafka cluster.
This is demonstrated by `cluster_single_shotover` test now running the full kafka test suite.
`cluster_fake` can now be deleted as `cluster_single_shotover` is now providing the same coverage.

However.
The current design can not be made to work with a multi-node shotover cluster.
The problem is that the client will retrieve metadata from one shotover node and then send messages relying on that metadata to a second shotover node, however that shotover node lacks that metadata and cannot route the message.
We will have to change the design to do one of:
* Generate extra messages in shotover to populate missing metadata.
* Communicate kafka metadata between shotover nodes https://github.com/shotover/shotover-proxy/issues/1218

I still think we should land this PR as is because:
* Its a good starting point towards a full solution and demonstrates where the current design falls short.
* It should be enough to unblock the acceleration transform work for now.

The fork of rdkafka is needed because the transform triggers a bug in the underlying C driver where it gets stuck in a loop sending endless metadata requests. That issue is resolved in a later release of the C rdkafka driver, so updating that resolves the issue. Unfortunately the rust rdkafka developers are holding off updating the C driver until the next release so we will have to depend on a fork for a while.

I bumped the timeouts to 30s because I observed that the initial fetch can take multiple seconds on my machine which could end up going over the 10s timeout in a degenerate CI scenario.

I added a bunch more message types to `Frame::Kafka`, some because I need access to their data, others just because it made it easier to debug having access to the parsed data. I dont expect this to have any meaningful difference to performance since the real work all happens in Produce/Fetch messages.

The actual routing implementation builds upon what was setup in https://github.com/shotover/shotover-proxy/pull/1222 but also changes the metadata collection to be shared between all connections, since the client will open multiple connections to shotover and not issue metadata requests it already knows.

I determined that JoinGroup, OffsetFetch, Heartbeat and SyncGroup need to go the coordinator broker by observing that they all returned the below error when using the default random routing.
image


benbromhead (approved):
Thanks for working through this one. I agree on landing this for now and we can address the multi-shotover scenario with a bit more thought.

conorbros:
In `CassandraSinkCluster` we just use a `RwLock<HashMap>` for concurrent access to a hashmap, did you find the performance benefit from this new dependency `DashMap` to be worth it?

shotover/src/transforms/kafka/sink_cluster.rs
```diff
@@ -54,6 +56,9 @@ pub struct KafkaSinkClusterBuilder {
     shotover_nodes: Vec<KafkaAddress>,
     connect_timeout: Duration,
     read_timeout: Option<Duration>,
+    coordinator_broker_id: Arc<AtomicI32>,
+    topics: Arc<DashMap<StrBytes, Topic>>,
```
conorbros, line 60: In CassandraSinkCluster we just use a RwLock<HashMap> for concurrent access to a hashmap, did you find the performance benefit from this new dependency DashMap to be worth it?

conorbros (approved):

shotover/src/transforms/kafka/sink_cluster.rs
rukai: Well using a concurrent hashmap instead of a RwLock *should* be faster, but I cant measure anything till we get windsock kafka benchmarks to be stable enough.
Definitely something to experiment with once we get better benchmarks.

PR #1229, opened 2023-06-28
Tiny comment fix in server.rs

rukai:
Noticed while reviewing https://github.com/shotover/shotover-proxy/pull/1215


rukai:
We ended up resolving in https://github.com/shotover/shotover-proxy/pull/1229

PR #1230, opened 2023-06-30
CassandraSinkCluster: switch to dashmap

conorbros:
I found the benchmarks too noisy to determine if this actually results in a speedup but it's reasonable to assume it does at least make it a bit faster. And since we brought in the dep for Kafka sink cluster might as well use it here.


rukai:
I'll review this once I have windsock cassandra benches running on AWS, should take a few days.

rukai:
I ran benches for this 3 times and all 3 times the results were clearly saying this PR regressed performance.
image

I would still like to keep this PR open until I can resolve the intermittent benching noise though.
Feel free to mark as draft and stop maintaining it in the mean time though.

conorbros:
Thanks, interesting results.

conorbros:
I checked this again, it's still a clear regression.

Issue #1236, opened 2023-07-05
Redesign KafkaSinkCluster to handle multiple shotover instances


rukai:
There are two possible approaches:
* Send a metadata message before routing a fetch/produce if we dont know about that topic yet.
* Communicate kafka metadata between shotover nodes i.e. use raft

However I can already see some problems with a raft solution.
* race conditions - the client might query a shotover node before it receives the metadata it requires from another shotover node
* it doesnt handle deletion of metadata - currently topic metadata will grow unboundedly, we arent sure if this is a problem yet, but a raft solution wont fix this while sending our own metadata messages will fix it by allowing us to refetch cleared data if we need it again.

PR #1243, opened 2023-07-10
KafkaSinkCluster: send metadata/findcoordinator messages when missing metadata

rukai:
closes https://github.com/shotover/shotover-proxy/issues/1236

The `produce_only` test case is deleted as we can now run the full test suite against `cluster_multi_shotover`.

The basic process stays the same, but this PR introduces the following:
* if a request is routed based on a group id we first send a findcoordinator message to ensure the metadata for that group is populated.
* if a request is routed based on a topic we first send a metadata message to ensure the metadata for that topic is populated.
* some logic is pulled into `update_local_nodes` and `process_metadata` methods to avoid duplicating some logic.

I could batch the findcoordinator messages into a single message instead of one per group_id, but that requires a higher protocol version (seems to be introduced around 2020, which is probably fine?), so I just left it as is for now.

I have left clearing outdated metadata as a TODO for a follow up PR.
I also added a bullet point to https://github.com/shotover/shotover-proxy/issues/1214 for it.
It should be possible to do by observing errors sent from the server back to the client.

## versioning notes
I attempted to use the ApiVersions message to determine what minimum version to use for the generated metadata and findcoordinator messages but this turned out to be pointless since min_version is always 0.

Strangely a metadata message of version 0 makes the connection time out, maybe they are dropping support for messages but just not changing the min_version due to ossification? Or maybe its just a bug in the kafka protocol crate on very old versions of the protocol?

So I ended up just hardcoding a low version number that has all the features we need and that doesnt mysteriously kill the connection.


conorbros (dismissed):

conorbros:
Oops, approved the wrong one by accident

rukai:
I think if you do a review as "request changes" it should replace the passed review.

conorbros:
Is it possible to test it in the same way that we test Cassandra routing- find which node the client would hit if routed correctly and check which one Shotover hit?

shotover/src/transforms/kafka/sink_cluster.rs
```diff
@@ -186,11 +182,197 @@ impl Transform for KafkaSinkCluster {
...
+                })) => {
+                    for topic in &fetch.topics {
+                        self.store_topic(&mut topics, topic.topic.clone());
+                    }
+                }
+                Some(Frame::Kafka(KafkaFrame::Request {
+                    body: RequestBody::Heartbeat(heartbeat),
+                    ..
+                })) => {
+                    self.store_group(&mut groups, heartbeat.group_id.clone());
```
conorbros, line 241: We can join these match arms together since they all do the same thing.

image

rukai:
> Is it possible to test it in the same way that we test Cassandra routing- find which node the client would hit if routed correctly and check which one Shotover hit?

For Cassandra token aware routing we need to do that since Cassandra will silently handle routing mustakes for us. (But at a cost of performance)
However for the Kafka routing we are doing here if we route to the wrong node, the node will return an error, failing the integration test.

benbromhead (approved):

shotover/src/transforms/kafka/sink_cluster.rs
rukai: oh that is very nice, thanks!

conorbros: Oh nice, even better

conorbros (approved):

PR #1245, opened 2023-07-11
KafkaSinkCluster: fix FindCoordinator handling

rukai:
I discovered that our handling of FindCoordinator messages is wrong.
I thought there was a single coordinator node in the cluster that coordinates all groups.
It turns out a different node can coordinate each group.
so to properly store this, this PR changes `coordinator_broker_id`  from a `Arc<AtomicI32>` to a `Arc<DashMap<StrBytes, i32>>`.

In protocols v0-3:
The request specifies the group name and we need to match up the request with the response to determine what group name to insert into the map

In protocol v4:
The response includes the group name with each coordinator, we can just use it directly.

There is also a `key_type` field in the request, I believe this is `0` to indicate group coordinators and I believe we dont currently have any reason to store non-group coordinators.


conorbros (approved):

shotover/src/transforms/kafka/sink_cluster.rs
```diff
@@ -55,7 +54,7 @@ pub struct KafkaSinkClusterBuilder {
     shotover_nodes: Vec<KafkaAddress>,
     connect_timeout: Duration,
     read_timeout: Option<Duration>,
-    coordinator_broker_id: Arc<AtomicI32>,
+    coordinator_broker_id: Arc<DashMap<StrBytes, i32>>,
```
benbromhead, line 57: Why the change from an Atomic int to a regular one? Does Dashmap give what we need?

rukai: An `Arc<Atomic<I32>>` is used instead of an `Arc<i32>` because `Arc<Atomic<I32>>` is read/write while `Arc<i32> is read only.
`Arc<DashMap<StrBytes, i32>` is already read/write, so making it an `Arc<DashMap<StrBytes, AtomicI32>` would be redundant.

benbromhead (approved):

PR #1249, opened 2023-07-13
update rust version and deps

conorbros:
`with_single_cert` -> `with_client_auth_cert` change is because the former was deprecated.


rukai (approved):
🦀

benbromhead (approved):

PR #1250, opened 2023-07-14
Shotover 0.1.11

conorbros:
Will get a few of the open PRs in before merging this and releasing a new Shotover version.


rukai (approved):
This should trigger the first release to crates.io
We previously attempted a release to crates.io for `0.10.0` but it failed due to a bug in our release GA workflow.

As previously mentioned for `0.10.0`, since we are now on crates.io its a good time to commit to semver.
There are a lot of breaking changes we'll need to do in the future, butno need to delay the release over that since theres nothing scary about releasing a `0.2.0`

rukai:
For some reason docs.rs failed for https://github.com/shotover/aws-throwaway, give me some time to investigate before we merge this. I'm worried shotover might suffer the same fate.

rukai:
This can go ahead, it should be fine since aws-sdk-ec2 is a dev dep and therefore not required for `cargo doc --no-deps`

conorbros:
Cool, once @benbromhead approves this we can decide what we'd like to get in before releasing and go ahead.

benbromhead (approved):

rukai:
Can you add something like this to the changelog.md in this PR?
```
## 0.1.11

### topology.yaml

* No recorded changes

### shotover rust api

* `shotover::message_value` is now `shotover::frame::value`
* `shotover::message_value::MessageValue` is now `shotover::frame::value::GenericValue`
```

There might be more changes I missed but I couldnt see any when I took a quick look just now.

PR #1251, opened 2023-07-14
Websocket docs

conorbros:
- Add info for using WebSocket transport to the `Cassandra` source docs.
- Some small fixes I noticed after going through the docs.


docs/src/source-types.md
```diff
@@ -33,6 +33,10 @@ Cassandra:
  
   # Timeout in seconds after which to terminate an idle connection. This field is optional, if not provided, idle connections will never be terminated.
   # timeout: 60
+
+  # Transport option, either TCP, being the standard Cassandra protocol or use the Cassandra protocol over WebSockets, with a compatible
```
rukai, line 37: The actual value needed is `Tcp` right? Or is it actually case insensitive?

conorbros: Hopefully that clears it up, I was using the correct form of the names for "WebSocket" and "TCP". With the actual value needed in the configuration file below.

rukai: > Hopefully that clears it up

Hopefully what clears it up? Did you forget to push a code change?

I'm thinking that we should continue to use the format we already use elsewhere for documenting enum variants.
```yaml
  # The transport that cassandra communication will occur over.
  # TCP is the only cassandra protocol conforming transport.
  transport: Tcp
  
  # alternatively:
  #
  # Use websockets as a transport using a shotover specific variation of the cassandra protocol.
  # transport: WebSocket
```

conorbros: Clears up what case to use in the enum variant, anyway I'll switch to that recommendation given it's convention.

rukai (approved):
nice!

benbromhead (approved):

PR #1255, opened 2023-07-19
opensearch int tests

conorbros:
Just opensearch client -> cluster for now.

### DIFF
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -52,6 +52,9 @@ aws-throwaway.workspace = true
 windsock = { path = "../windsock" }
 regex = "1.7.0"
 cql-ws = { git = "https://github.com/conorbros/cql-ws" }
+opensearch = "2.1.0"
+serde_json = "1.0.103"
+reqwest.workspace = true
 
 [[bench]]
 name = "benches"
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -54,7 +54,6 @@ regex = "1.7.0"
 cql-ws = { git = "https://github.com/conorbros/cql-ws" }
 opensearch = "2.1.0"
 serde_json = "1.0.103"
-reqwest.workspace = true
 
 [[bench]]
 name = "benches"
--- a/shotover-proxy/tests/opensearch_int_tests/mod.rs
+++ b/shotover-proxy/tests/opensearch_int_tests/mod.rs
@@ -23,25 +23,23 @@ pub async fn index_documents(client: &OpenSearch) -> Response {
         .await
         .unwrap();
 
-    if exists_response.status_code() == StatusCode::NOT_FOUND {
-        let mut body: Vec<BulkOperation<_>> = vec![];
-        for i in 1..=10 {
-            let op = BulkOperation::index(json!({"title":"OpenSearch"}))
-                .id(i.to_string())
-                .into();
-            body.push(op);
-        }
+    assert_eq!(exists_response.status_code(), StatusCode::NOT_FOUND);
 
-        client
-            .bulk(BulkParts::Index(index))
-            .body(body)
-            .refresh(Refresh::WaitFor)
-            .send()
-            .await
-            .unwrap()
-    } else {
-        exists_response
+    let mut body: Vec<BulkOperation<_>> = vec![];
+    for i in 1..=10 {
+        let op = BulkOperation::index(json!({"title":"OpenSearch"}))
+            .id(i.to_string())
+            .into();
+        body.push(op);
     }
+
+    client
+        .bulk(BulkParts::Index(index))
+        .body(body)
+        .refresh(Refresh::WaitFor)
+        .send()
+        .await
+        .unwrap()
 }
 
 #[tokio::test(flavor = "multi_thread")]
@@ -57,7 +55,8 @@ async fn passthrough_standard() {
         .unwrap();
     let client = OpenSearch::new(transport);
 
-    let _ = index_documents(&client).await;
+    index_documents(&client).await;
+
     let response = client
         .search(SearchParts::None)
         .body(json!({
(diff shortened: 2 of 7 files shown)


shotover-proxy/tests/opensearch_int_tests/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn passthrough_standard() {
+    // This setting changes the maximum number of memory map areas a process may have.
+    // OpenSearch requires vm.max_map_count to be set higher to run.
+    // This command sets it temporarily higher (for the length of the user session), it is reset after rebooting.
+    if &run_command("sysctl", &["vm.max_map_count"]).unwrap() != "vm.max_map_count = 262144\n" {
+        run_command("sudo", &["sysctl", "-w", "vm.max_map_count=262144"]).unwrap();
```
rukai, line 54: Is this is in response to an actual failure you encountered on your machine?
Or did you just do it because the docs recommended it?

shotover-proxy/Cargo.toml
```diff
@@ -52,6 +52,9 @@ aws-throwaway.workspace = true
 windsock = { path = "../windsock" }
 regex = "1.7.0"
 cql-ws = { git = "https://github.com/conorbros/cql-ws" }
+opensearch = "2.1.0"
+serde_json = "1.0.103"
+reqwest.workspace = true
```
rukai, line 57: is this actually used?

shotover-proxy/tests/opensearch_int_tests/mod.rs
conorbros: Docs recommended it and a failure on my machine. Did it occur for you?

rukai: The pr as is hung on my machine, I'll investigate more tomorrow

shotover-proxy/tests/test-configs/opensearch-passthrough/docker-compose.yaml
```diff
@@ -0,0 +1,56 @@
+version: '3'
+services:
+  opensearch-node1:
+    image: opensearchproject/opensearch:latest
```
rukai, line 4: we should specify a complete tag instead of latest, otherwise our tests will suddenly break one day.

shotover-proxy/tests/test-configs/opensearch-passthrough/docker-compose.yaml
```diff
@@ -0,0 +1,56 @@
...
+      - "OPENSEARCH_JAVA_OPTS=-Xms512m -Xmx512m"
+    ulimits:
+      memlock:
+        soft: -1
+        hard: -1
+      nofile:
+        soft: 65536
+        hard: 65536
+    volumes:
+      - opensearch-data1:/usr/share/opensearch/data
```
rukai, line 21: we can just define the volume directly where we use it:
```
    volumes:
      - type: volume
        target: /usr/share/opensearch/data
```

shotover-proxy/tests/test-configs/opensearch-passthrough/docker-compose.yaml
```diff
@@ -0,0 +1,56 @@
...
+    volumes:
+      - opensearch-data2:/usr/share/opensearch/data
+    networks:
+      - opensearch-net
+
+volumes:
+  opensearch-data1:
+  opensearch-data2:
+
+networks:
```
rukai, line 54: no need to define a network unless we actually do anything custom with it.

shotover-proxy/tests/opensearch_int_tests/mod.rs
rukai: We can avoid needing root by running opensearch in a non-clustered way that disables production checks.
This should be all the config needed to enable this:
image
(And also remove the second instance)

On my machine without the above changes the container fails to start without vm.max_map_count set.
However when using the above changes it just works.

Having our tests halt and wait for a sudo password is really problematic.
In the future we may want to test clustered opensearch but we can deal with that problem when we need to.
A possible fix I'm thinking of could be to have our CI setup script set vm.max_map_count and then have the test just fail locally (with a nice instructive error message) if they dont have the correct vm.max_map_count set.

test-helpers/src/lib.rs
```diff
@@ -17,7 +17,7 @@ use subprocess::{Exec, Redirection};
 /// # Arguments
 /// * `command` - The system command to run
 /// * `args` - An array of command line arguments for the command
-pub(crate) fn run_command(command: &str, args: &[&str]) -> Result<String> {
```
rukai, line 20: restore this to pub(crate)

shotover-proxy/tests/opensearch_int_tests/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+    };
+
+    if let Some(c) = response.content_length() {
+        assert!(c > 0)
+    };
+
+    assert_eq!(response.url(), &expected_url);
+    assert_eq!(response.status_code(), StatusCode::OK);
+    assert_eq!(response.method(), opensearch::http::Method::Post);
+    let debug = format!("{:?}", &response);
```
rukai, line 88: Why do we assert on the debug string of response?
Shouldnt the above assertions on its methods be enough?

shotover-proxy/tests/opensearch_int_tests/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    let expected_url = {
+        let addr = "https://localhost:9200/";
+        let mut url = Url::parse(addr).unwrap();
+        url.set_username("").unwrap();
+        url.set_password(None).unwrap();
+        url.join("_search?allow_no_indices=true").unwrap()
+    };
+
+    if let Some(c) = response.content_length() {
```
rukai, line 81: If we expect this to return Some, lets unwrap it

shotover-proxy/tests/opensearch_int_tests/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+        }))
+        .allow_no_indices(true)
+        .send()
+        .await
+        .unwrap();
+
+    let expected_url = {
+        let addr = "https://localhost:9200/";
+        let mut url = Url::parse(addr).unwrap();
+        url.set_username("").unwrap();
```
rukai, line 76: I removed the username and password lines here and the assertion still passes, do we need them for anything?

rukai: In fact this whole block can be simplified to: `let expected_url = Url::parse("https://localhost:9200/_search?allow_no_indices=true").unwrap();`

shotover-proxy/tests/opensearch_int_tests/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+pub async fn index_documents(client: &OpenSearch) -> Response {
+    let index = "posts";
+    let exists_response = client
+        .indices()
+        .exists(IndicesExistsParts::Index(&[index]))
+        .send()
+        .await
+        .unwrap();
+
+    if exists_response.status_code() == StatusCode::NOT_FOUND {
```
rukai, line 26: branches like this in test code are problematic.
Is it reasonable to assert that the status_code is NOT_FOUND instead?

shotover-proxy/tests/opensearch_int_tests/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    assert_eq!(response.url(), &expected_url);
+    assert_eq!(response.status_code(), StatusCode::OK);
+    assert_eq!(response.method(), opensearch::http::Method::Post);
+    let debug = format!("{:?}", &response);
+    assert!(debug.contains("method"));
+    assert!(debug.contains("status_code"));
+    assert!(debug.contains("headers"));
+    let response_body = response.json::<Value>().await.unwrap();
+    assert!(response_body["took"].as_i64().is_some());
```
rukai, line 93: can we assert that hits contains all the objects that we expected?

conorbros: I tested that it has the 10 records we expected aswell.

rukai (approved):
nice!

benbromhead (approved):

PR #1256, opened 2023-07-20
message_value -> message/value

conorbros:
Small refactor


rukai:
The hierarchy goes: `Message` contains `Frame` creates `MessageValue`
so it seems like MessageValue belongs more with Frame than with Message?
Maybe we should rename MessageValue -> FrameValue and store it in a `frame::value` module?

rukai (approved):
Seems reasonable, can you do the rename from `MessageValue` to `FrameValue` in a follow up PR?
Now that I'm thinking about it, 'GenericValue` sounds like a more useful name as it conveys that this is a value that can be converted between all protocols.

benbromhead (approved):

PR #1257, opened 2023-07-20
ec2-cargo: run rsync before testing.

rukai:
I searched for a rust rsync implementation that we can statically link but there appears to be none: https://github.com/dropbox/fast_rsync/issues/24
So instead I implemented this by shelling out to rsync.
If we ever want to support windows then we will need to write that rsync implementation missing from the ecosystem.
Otherwise, the implementation in this PR can quite happily support just mac os (has rsync installed by default) and linux (easily installed from package manager) 

This PR also changes ec2-cargo to run tests under nextest instead of `cargo test` as per the reasons in https://github.com/shotover/shotover-proxy/pull/1270


rukai:
hmm, I should change this to use nextest before review

rukai:
done

conorbros (approved):

cjrolo (approved):
I used to write a ton of code like this 10+ years ago. It's good to know some things never change. :)

PR #1260, opened 2023-07-24
Use nextest in CI to partition test runs

rukai:
nextest provides very powerful [test sharding capabilities](https://nexte.st/book/partitioning)
This PR uses those capabilities to distribute our tests across 8 different GA job instances.
This gives us a time to build and run tests of 35m without a cache.
Previously this was ~1h 15m without a cache and 50m with a cache.
Once this is merged and we can make use of the cache we should be able to expect times of 25m

This is clear win, lets land this.

## future plans

Currently windsock is bottlenecking our CI, taking 1 hour to complete.
I have plans to expose windsock benches as tests so that they can also take advantage of this sharding, but that will have to be a follow up PR.

In another (optional) follow up PR we can swap to using nextest locally as it has a better UX than `cargo test`.

## Implementation

I had to configure nextest to run our tests properly in `.config/nextest.toml`
This is needed since nextest runs every test in its own isolated process which ignores the `#[serial]` attributes we currently use to avoid resource conflicts in tests.
The configuration sets up an integration test [test-group](https://nexte.st/book/test-groups.html).
We specify `filter = 'package(shotover-proxy) | package(custom-transforms-example)'` to put all our integration tests into that test group.
And then we set the test-group to `max-threads = 1 ` so that tests in the test-group are never run concurrently with each other. (But tests outside of the group can still be run concurrently with tests in the group)

I used [this example](https://github.com/nextest-rs/reuse-build-partition-example/blob/main/.github/workflows/ci.yml) closely when implementing the new GA workflow.

I had to fix a bug in `shotover_process` where the `assert_fails_to_start` did not make use of the provided direct binary and was still attempting to invoke cargo to find the binary.
This is a problem because the sharded test jobs do not have access to cargo as everything is already built for them in the build job.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,11 @@
+[profile.default]
+fail-fast = false
+slow-timeout = { period = '5m', terminate-after = 2 }
+
+# Overwrites profile.default when the filter matches
+[[profile.default.overrides]]
+filter = 'package(shotover-proxy) | package(custom-transforms-example)'
+test-group = 'integration-tests'
+
+[test-groups]
+integration-tests = { max-threads = 1 }
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -16,29 +16,20 @@ env:
   RUST_BACKTRACE: 1
 
 jobs:
-  build_and_test:
+  build_check_and_upload:
     strategy:
+      fail-fast: false
       matrix:
         include:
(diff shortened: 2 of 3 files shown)


rukai:
wow!
image
35 mins on an uncached build

conorbros (approved):
Looks like you will need to change the merge checklist to remove those old jobs.

rukai:
Yep I'll do that once this has approval to merge

benbromhead (approved):

PR #1262, opened 2023-07-26
windsock: profiling support in AWS

rukai:
This almost works, just need to investigate whats going wrong with the flamegraph download.
Possibly I want to abandon this approach though.

I think having local benches go through `run_service` could be really nice in the future.
At the moment it makes things way more complicated and brittle without much benefit.
If we are able to generate our topology.yaml entirely within run_service then it might be worth it.
We should explore topology.yaml generation for local benches first.

We might want to land this with only cloud runs using `run_service` and the local runs remaining as is. But there is still some weirdness on the cloud side: shotover log handling is wonky.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5800,6 +5800,7 @@ dependencies = [
  "console",
  "copy_dir",
  "docker-compose-runner",
+ "nix",
  "scylla",
  "serde",
  "strum 0.25.0",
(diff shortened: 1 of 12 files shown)


rukai:
I am abandoning this approach in favor of https://github.com/shotover/shotover-proxy/pull/1316

Issue #1198, opened 2023-05-31
Optimize kafka encoding


rukai:
From inspecting the flamegraph generated from `cargo windsock --profilers flamegraph --name kafka,shotover=standard,size=100KB,topology=single` it seems like the kafka encode process is a bottleneck (15% of execution time)
We should be able to improve things by using the kafka-protocol's `Encodable::compute_size` method, bubbling up the results to calculate the full size required for a message and then preallocating the space required for all messages.
Any improvements should be verified by `cargo windsock --name kafka,shotover=standard,size=100KB,topology=single`

rukai:
The core problem here was generating errors in the success path, resolved by https://github.com/shotover/shotover-proxy/pull/1281
I will have to evaluate if there is any value in https://github.com/shotover/shotover-proxy/pull/1263 but I'll close this issue since its currently not showing up in profiling.

PR #1263, opened 2023-07-26
KafkaEncoder: precompute message sizes to avoid allocations

rukai:
closes https://github.com/shotover/shotover-proxy/issues/1198

This PR implements the suggestions from the issue: All messages in a batch of kafka messages have their required space reserved in the destination buffer before encoding begins.

The microbenchmarks demonstrate improvements from this PR:
image

Edit:
Looking deeper into this I think the microbenchmarks are misleading.
The buffer is actually preallocated to 8K by tokio: https://github.com/tokio-rs/tokio/blob/6cb106c3538cf527495ef5491c088d1365b14c8e/tokio-util/src/codec/framed_impl.rs#L53
From reading the implementation it seems this internal buffer is reused forever. Once the split off BytesMut's are rejoined the capacity is reused?
So we can probably just allow it to grow and stabilize to reasonable number without having to worry about preallocations?

Things to do from here:
* Add a bench case with preallocated buffer.
* abandon this PR?

Waiting on https://github.com/shotover/shotover-proxy/pull/1281 to land so we can properly evaluate this.

### DIFF
--- a/shotover/src/codec/kafka.rs
+++ b/shotover/src/codec/kafka.rs
@@ -135,18 +135,26 @@ impl KafkaEncoder {
 impl Encoder<Messages> for KafkaEncoder {
     type Error = CodecWriteError;
 
-    fn encode(&mut self, item: Messages, dst: &mut BytesMut) -> Result<(), Self::Error> {
-        item.into_iter().try_for_each(|m| {
+    fn encode(&mut self, messages: Messages, dst: &mut BytesMut) -> Result<(), Self::Error> {
+        let mut to_reserve = 0;
+        for message in &messages {
+            to_reserve += message.bytes_len().map_err(CodecWriteError::Encoder)?;
+        }
+        dst.reserve(to_reserve);
+        let expected_len = dst.len() + to_reserve;
+
+        for m in messages {
             let start = dst.len();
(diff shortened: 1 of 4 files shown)


benbromhead:
I think you would want to run an strace on shotover to count the number of buffer allocations happening (or count it some other way) during a Kafka session. I had a look at the code you linked to on the implementation and it was hard for me to guess at what the end buffer allocation would be in the various cases (e.g. a frame larger than 8k, a frame smaller than 8k that is still being processed by shotover elsewhere while the next frame is being decoded etc such that there are multiple frames with bytes from that buffer in flight, batch frames etc). 

What does a flame graph look like at the moment as well? Are we spending a ton of time doing allocations? lot's of little ones will show up as a good amount of time on the graph right?

rukai:
Here is a flamegraph of `kafka,shotover=message-parsed,size=100KB,topology=single`: 
image
Interactive link: https://share.firefox.dev/47Vvy10


It looks like we are spending ~9% (3.8% + 5.6%) of our time in encoding, but all of that time appears to be in memcpy, it doesnt show any alloc occurring there.

rukai:
Going to close this, as per the discussion earlier, the micro benchmarks were misleading

PR #1265, opened 2023-07-28
update governor

conorbros:
https://docs.rs/governor/latest/governor/struct.RateLimiter.html#method.check_n

https://docs.rs/governor/0.5.0/governor/struct.RateLimiter.html#method.check_n

Updates the `governor` dependency for our request throttling. The links above demonstrate the changes between versions that caused the test failures described in https://github.com/shotover/shotover-proxy/pull/1261. The changes in this PR keep the behaviour the same. 

The new method succeeds in one way and can now fail in two ways, compared to just one previously. `Ok(Ok())` occurs if all cells can be accommodated and `Ok(Err(_))` occurs if not all cells can be accommodated.

A failure occurs when the batch can never go through, meaning the rate limiter's quota's burst size is too low for the given number of cells to be ever allowed through.

### DIFF
--- a/shotover-proxy/tests/cassandra_int_tests/mod.rs
+++ b/shotover-proxy/tests/cassandra_int_tests/mod.rs
@@ -655,7 +655,14 @@ async fn request_throttling(#[case] driver: CassandraDriver) {
 
     tokio::time::sleep(std::time::Duration::from_secs(1)).await; // sleep to reset the window
 
-    shotover.shutdown_and_then_consume_events(&[EventMatcher::new().with_level(Level::Warn).with_target("shotover::transforms::throttling").with_message(r#"A message was received that could never have been successfully delivered since it contains more sub messages than can ever be allowed through via the `RequestThrottling` transforms `max_requests_per_second` configuration."#)]).await;
+    let event_watcher_message = r#"A message was received that could never have been successfully delivered since it contains more sub messages than can ever be allowed through via the `RequestThrottling` transforms `max_requests_per_second` configuration."#;
+
+    shotover
+        .shutdown_and_then_consume_events(&[EventMatcher::new()
+            .with_level(Level::Warn)
+            .with_target("shotover::transforms::throttling")
+            .with_message(event_watcher_message)])
+        .await;
 }
 
 #[cfg(feature = "alpha-transforms")]
--- a/shotover-proxy/tests/cassandra_int_tests/mod.rs
+++ b/shotover-proxy/tests/cassandra_int_tests/mod.rs
@@ -655,7 +655,7 @@ async fn request_throttling(#[case] driver: CassandraDriver) {
 
     tokio::time::sleep(std::time::Duration::from_secs(1)).await; // sleep to reset the window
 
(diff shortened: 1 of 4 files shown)


shotover/src/transforms/throttling.rs
```diff
@@ -67,15 +67,20 @@ impl Transform for RequestThrottling {
...
-                    Some((message, i))
-                } else {
-                    None
+                    Ok(Ok(())) => None,
+                    Ok(Err(_)) => {
+                        let message = requests_wrapper.requests.remove(i);
+                        Some((message, i))
+                    }
+                    Err(e) => {
+                        tracing::info!("{e}");
```
rukai, line 80: accidentally left in?

conorbros: No, that error will occur when a batch is too big to go through, I figured we would want to log when that happens.

rukai: My understanding here is that this `info` log will trigger when shotover receives a cassandra batch which contains more queries than the `max_requests_per_second` value in the topology.yaml
Such a message would never be able to make it through so I agree its probably worth communicating that it in some way.

Ideally we would include this information in the error message of the `ErrorType::Overloaded` response that goes back to the client so that the developer can easily match up the message sent to the problem.

However that would require some refactoring, so instead, for now, we could just clean up this log a bit.
I think bumping it to a warning makes sense because it indicates a misconfiguration somewhere as such a message could never have succeeded.
Lets also include some extra context since the display impl for that error wont make much sense to the user https://docs.rs/governor/latest/src/governor/errors.rs.html#10
So something like:
```rust
tracing::warn!("A message was received that could never have been successfully delivered since it contains more sub messages than can ever be allowed through via the `RequestThrottling` transforms `max_requests_per_second` configuration.")
```

shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -655,7 +655,7 @@ async fn request_throttling(#[case] driver: CassandraDriver) {
 
     tokio::time::sleep(std::time::Duration::from_secs(1)).await; // sleep to reset the window
 
-    shotover.shutdown_and_then_consume_events(&[]).await;
+    shotover.shutdown_and_then_consume_events(&[EventMatcher::new().with_level(Level::Warn).with_target("shotover::transforms::throttling").with_message(r#"A message was received that could never have been successfully delivered since it contains more sub messages than can ever be allowed through via the `RequestThrottling` transforms `max_requests_per_second` configuration."#)]).await;
```
rukai, line 658: maybe split this into multiple lines?

rukai (approved):
Nice! Thanks for write up, made it a lot easier to understand what was going on here.

shotover/src/transforms/throttling.rs
cjrolo, line 1: I had a bit of trouble going through this code.
`filter_map` with a confusing `match` (due to the `Ok(Ok(()))` and `Ok(Err(_))`, could we get one or two comments around the `match` block?

cjrolo (approved):

PR #1267, opened 2023-08-01
cassandra_int_tests: add cqlsh smoke test

rukai:
Run a very basic test for cqlsh against a single instance and a single TLS instance.
This is part of my investigation into why cqlsh is failing to connect to shotover since the switch to rustls

The `need_to_reinstall` logic is possibly a little overengineered, but it should make us robust to both changes to CQLSH_VERSION and to venvs in a broken state, possibly due to the system level python being upgraded.
Possibly we would be better off just nuking the venv every time we run the test but I think its worth cutting off the 5s cost of recreating it.

a python venv is an isolated python install where you can install python packages isolated from the rest of the system.
Here i've created and used a venv inside the target directory to ensure we are not fiddling with or relying on system state in our tests.

### DIFF
--- a/ec2-cargo/src/main.rs
+++ b/ec2-cargo/src/main.rs
@@ -56,7 +56,7 @@ until sudo apt-get update -qq
 do
   sleep 1
 done
-sudo apt-get install -y cmake pkg-config g++ libssl-dev librdkafka-dev uidmap
+sudo apt-get install -y cmake pkg-config g++ libssl-dev librdkafka-dev uidmap python3-venv
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 
 curl -sSL https://get.docker.com/ | sudo sh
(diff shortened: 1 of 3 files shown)


conorbros:
Pretty hesitant to add automatically installing cqlsh this to our test workflow, for example I use the Aur version of cqlsh. Plus if the test passes this doesn't reveal anything about the problem.

rukai:
> Pretty hesitant to add automatically installing cqlsh this to our test workflow, for example I use the Aur version of cqlsh.
If your concern is that it would mess with your environment: The setup logic does not override any existing cqlsh installation outside of the integration test. The venv is entirely self contained.

The value I see in this approach is:
* easy to jump between versions of cqlsh by just changing `CQLSH_VERSION`
* All machines use the same version of cqlsh avoiding "works on my machine" problems

> Plus if the test passes this doesn't reveal anything about the problem.

That is very true, the fact that it hasnt revealed any issues yet does indicate that its maybe not testing anything of value.
Once https://github.com/shotover/shotover-proxy/pull/1257 lands I wont need to be hacking together unmerged PRs in order to test cqlsh on aws, making merging this less important.
I'll mark this as a draft, and figure out how we want to approach this once we've gotten to the bottom of the cassandra TLS issue.

rukai:
Given that the problem I was trying to solve ended up being an internal configuration issue and nothing to do with cqlsh compatibility I'll go ahead and close this PR.
Its probably not worth the extra complexity and python dependency.

PR #1268, opened 2023-08-01
Opensearch support

conorbros:


shotover-proxy/tests/opensearch_int_tests/mod.rs
```diff
@@ -87,3 +69,183 @@ async fn passthrough_standard() {
         10
     );
 }
+
+// TODO this test fails even if we are waiting for the health status to go green
+// async fn test_health(client: &OpenSearch) {
```
rukai, line 74: If this is a test case we want to get working eventually, lets not leave the code commented out. (otherwise it'll bit rot)
And instead just remove the failing assert and include a `TODO: add assert that X once Y functionality is fixed/implemented`

If this test case was a misstep and its not possible to get working then just delete it entirely.

shotover/src/codec/opensearch.rs
```diff
@@ -0,0 +1,323 @@
...
+                    Ok(())
+                }
+            };
+
+            result.map_err(CodecWriteError::Encoder)
+        })
+    }
+}
+
+// #[cfg(test)]
```
rukai, line 279: same here, either:
* get this working
* leave out a failing assert + add TODO
* delete entirely

shotover/src/frame/opensearch.rs
```diff
@@ -0,0 +1,30 @@
...
+pub enum HttpHead {
+    Response(ResponseParts),
+    Request(RequestParts),
+}
+
+#[derive(Debug, Clone, Derivative)]
+#[derivative(PartialEq)]
+pub struct OpenSearchFrame {
+    #[derivative(PartialEq = "ignore")]
+    pub headers: Arc<HttpHead>,
```
rukai, line 18: why is this an Arc?

test-helpers/src/docker_compose.rs
```diff
@@ -56,24 +56,34 @@ pub fn get_image_waiters() -> &'static [Image] {
             log_regex_to_wait_for: r"Startup complete",
             timeout: 120,
         },
+        Image {
```
rukai, line 59: bad merge

shotover/src/transforms/opensearch/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+        Ok(result)
+    }
+}
+
+struct Connection {
+    outbound_tx: FramedWrite<WriteHalf<PinStream>, OpenSearchEncoder>,
+    response_messages_rx: mpsc::UnboundedReceiver<Message>,
+}
+
+async fn server_response_processing_task(
```
rukai, line 168: why cant we just use `crate::transforms::util::cluster_connection_pool::spawn_read_write_tasks`?

shotover/src/frame/opensearch.rs
```diff
@@ -0,0 +1,30 @@
...
+    pub body: Bytes,
+}
+
+impl OpenSearchFrame {
+    pub fn new(headers: Arc<HttpHead>, body: Bytes) -> Self {
+        Self { headers, body }
+    }
+
+    pub fn from_bytes(_bytes: &Bytes) -> Result<Self> {
```
rukai, line 27: lets implement and use from_bytes instead of directly constructing via `new`?

shotover/src/frame/opensearch.rs
```diff
@@ -0,0 +1,30 @@
...
+#[derive(Debug, Clone, Derivative)]
+#[derivative(PartialEq)]
+pub struct OpenSearchFrame {
+    #[derivative(PartialEq = "ignore")]
+    pub headers: Arc<HttpHead>,
+    pub body: Bytes,
+}
+
+impl OpenSearchFrame {
+    pub fn new(headers: Arc<HttpHead>, body: Bytes) -> Self {
```
rukai, line 23: this method is unused

shotover/src/codec/opensearch.rs
```diff
@@ -0,0 +1,323 @@
...
+        let mut headers = [httparse::EMPTY_HEADER; 16];
+        let mut request = httparse::Request::new(&mut headers);
+
+        let amt = match request.parse(&src[0..src.len()]).unwrap() {
+            httparse::Status::Complete(amt) => amt,
+            httparse::Status::Partial => return Ok(None),
+        };
+        match request.version.unwrap() {
+            1 => (),
+            _version => panic!("error!"), // TODO version error
```
rukai, line 58: considering this method returns a `Result`, all of these `TODO error handling` should just return an error instead of being left as a TODO.

shotover/src/sources/opensearch.rs
```diff
@@ -0,0 +1,100 @@
...
+use tokio::sync::{watch, Semaphore};
+use tokio::task::JoinHandle;
+use tracing::{error, info};
+
+#[derive(Deserialize, Debug, Clone)]
+pub struct OpenSearchConfig {
+    pub listen_addr: String,
+    pub connection_limit: Option<usize>,
+    pub hard_connection_limit: Option<bool>,
+    pub tls: Option<TlsAcceptorConfig>,
```
rukai, line 18: either add an integration test for both opensearch Sink and Source TLS support or remove TLS support from this PR.

shotover/src/codec/opensearch.rs
```diff
@@ -0,0 +1,323 @@
...
+                    if let Some((amt, parts, cl)) = decode_res {
+                        self.state = State::ReadingBody(parts, cl);
+                        src.advance(amt);
+                    } else {
+                        return Ok(None);
+                    };
+                }
+                State::ReadingBody(parts, cl) => {
+                    if src.is_empty() {
+                        // sometimes messages will have a content-length but no
```
rukai, line 174: is this behavior you've encountered or something that you are guessing might happen?

shotover/src/codec/opensearch.rs
```diff
@@ -0,0 +1,323 @@
...
+                }
+                Encodable::Frame(frame) => {
+                    let opensearch_frame = frame.into_opensearch().unwrap();
+                    match opensearch_frame.headers.as_ref() {
+                        HttpHead::Request(request_parts) => {
+                            dst.extend_from_slice(
+                                format!("{} ", request_parts.method.as_str()).as_bytes(),
+                            );
+                            dst.extend_from_slice(format!("{} ", request_parts.uri).as_bytes());
+                            dst.extend_from_slice(b"HTTP/1.1");
```
rukai, line 237: what if we received a request that was a different version of HTTP?
Cant we use a crate for HTTP encoding?

shotover/src/codec/opensearch.rs
```diff
@@ -0,0 +1,323 @@
...
+            httparse::Status::Partial => return Ok(None),
+        };
+        match request.version.unwrap() {
+            1 => (),
+            _version => panic!("error!"), // TODO version error
+        }
+
+        let mut builder = Request::builder();
+
+        builder = builder
```
rukai, line 63: We can merge these:
```rust
        let mut builder = Request::builder()
            .method(request.method.unwrap().to_owned().as_str())
            .uri(request.path.unwrap().to_owned().as_str());
```

shotover/src/codec/opensearch.rs
```diff
@@ -0,0 +1,323 @@
...
+            direction,
+            state: State::ParsingResponse,
+        }
+    }
+
+    fn decode_request(&self, src: &mut BytesMut) -> Result<Option<(usize, HttpHead, usize)>> {
+        let mut headers = [httparse::EMPTY_HEADER; 16];
+        let mut request = httparse::Request::new(&mut headers);
+
+        let amt = match request.parse(&src[0..src.len()]).unwrap() {
```
rukai, line 52: this is equivalent to just  `let amt = match request.parse(src).unwrap() {`

Also, what is `amt`? Can you abbreviate it?

shotover/src/codec/opensearch.rs
```diff
@@ -0,0 +1,323 @@
...
+        };
+        match request.version.unwrap() {
+            1 => (),
+            _version => panic!("error!"), // TODO version error
+        }
+
+        let mut builder = Request::builder();
+
+        builder = builder
+            .method(request.method.unwrap().to_owned().as_str())
```
rukai, line 64: These can be just:
```rust
        let mut builder = Request::builder()
            .method(request.method.unwrap())
            .uri(request.path.unwrap());
```

shotover/src/codec/opensearch.rs
```diff
@@ -0,0 +1,323 @@
...
+        let builder_headers = builder.headers_mut().unwrap();
+
+        let request_headers: Vec<(String, Vec<u8>)> = request
+            .headers
+            .iter()
+            .take_while(|h| !h.name.is_empty() && !h.value.is_empty())
+            .map(|h| (h.name.to_owned(), h.value.to_owned()))
+            .collect();
+
+        for (name, value) in request_headers {
```
rukai, line 76: These can be combined to avoid reallocating all of those strings.
```rust
        for header in request.headers {
            if header.name.is_empty() && header.value.is_empty() {
                break;
            }
            builder_headers.insert(
                HeaderName::from_bytes(header.name.as_bytes()).unwrap(),
                HeaderValue::from_bytes(header.value).unwrap(),
            );
        }
```

shotover-proxy/tests/opensearch_int_tests/mod.rs
```diff
@@ -87,3 +69,183 @@ async fn passthrough_standard() {
...
+    assert_eq!(
+        results["hits"].as_object().unwrap()["hits"]
+            .as_array()
+            .unwrap()
+            .len(),
+        0
+    );
+}
+
+async fn test_delete_index(client: &OpenSearch) {
```
rukai, line 199: can we combine with test_create_index into a test_create_and_delete_index?
That way we are not at risk of breaking things by reordering or introducing new test cases.
e.g.
```rust
async fn test_create_and_delete_index(client: &OpenSearch) {
    client
        .indices()
        .create(IndicesCreateParts::Index("test-index"))
        .send()
        .await
        .unwrap();

    let exists_response = client
        .indices()
        .exists(IndicesExistsParts::Index(&["test-index"]))
        .send()
        .await
        .unwrap();

    assert_eq!(exists_response.status_code(), StatusCode::OK);
    
    client
        .indices()
        .delete(IndicesDeleteParts::Index(&["test-index"]))
        .send()
        .await
        .unwrap();

    let exists_response = client
        .indices()
        .exists(IndicesExistsParts::Index(&["test-index"]))
        .send()
        .await
        .unwrap();

    assert_eq!(exists_response.status_code(), StatusCode::NOT_FOUND);
}
```

shotover-proxy/tests/test-configs/opensearch-passthrough/docker-compose.yaml
```diff
@@ -9,6 +9,7 @@ services:
       - bootstrap.memory_lock=true
       - "OPENSEARCH_JAVA_OPTS=-Xms512m -Xmx512m"
       - discovery.type=single-node
+      - plugins.security.disabled=true
```
rukai, line 12: our integration tests used to work without disabling authentication, what changed?

shotover/src/codec/opensearch.rs
conorbros: We only accept HTTP/1 in the decode side of the codec. OpenSearch only started supporting HTTP/2 last year so I think it's safe to only support HTTP/1 for the moment.

shotover-proxy/tests/test-configs/opensearch-passthrough/docker-compose.yaml
conorbros: This disables SSL, which we don't have support for yet.

shotover/src/codec/opensearch.rs
conorbros: Behaviour I encountered with OpenSearch

shotover/src/codec/opensearch.rs
```diff
@@ -0,0 +1,316 @@
...
+        let builder_headers = builder.headers_mut().unwrap();
+
+        let response_headers: Vec<(String, Vec<u8>)> = response
+            .headers
+            .iter()
+            .take_while(|h| !h.name.is_empty() && !h.value.is_empty())
+            .map(|h| (h.name.to_owned(), h.value.to_owned()))
+            .collect();
+
+        for (name, value) in response_headers {
```
rukai, line 112: same feedback from request can be applied here

shotover-proxy/tests/test-configs/opensearch-passthrough/docker-compose.yaml
rukai: ahh, makes sense :+1:

test-helpers/src/docker_compose.rs
```diff
@@ -58,22 +58,27 @@ pub fn get_image_waiters() -> &'static [Image] {
...
-            log_regex_to_wait_for: r"Startup complet",
+            log_regex_to_wait_for: r"Startup complete",
             timeout: 120,
         },
         Image {
             name: "shotover/cassandra-test:3.11.13-r1",
             log_regex_to_wait_for: r"Startup complete",
             timeout: 120,
         },
+        Image {
```
rukai, line 69: and here

shotover/src/transforms/opensearch/mod.rs
conorbros: Because HTTP is a synchronous protocol, we want to send a message, wait for a response and then send the next message.

rukai: spawn_read_write_tasks should work for both synchronous and asynchronous protocols.
we use it for kafka which is synchronous.

shotover-proxy/tests/opensearch_int_tests/mod.rs
```diff
@@ -87,3 +63,152 @@ async fn passthrough_standard() {
...
+                    "name": "John",
+                }
+            }
+        }))
+        .send()
+        .await
+        .unwrap();
+
+    let results = response.json::<Value>().await.unwrap();
+    assert!(results["took"].as_i64().is_some());
```
rukai, line 113: These can all be: `assert!(results["took"].is_i64());`

shotover-proxy/tests/opensearch_int_tests/mod.rs
```diff
@@ -87,3 +63,152 @@ async fn passthrough_standard() {
...
+        .unwrap();
+
+    let exists_response = client
+        .indices()
+        .exists(IndicesExistsParts::Index(&["test-index"]))
+        .send()
+        .await
+        .unwrap();
+
+    assert_eq!(exists_response.status_code(), StatusCode::OK);
```
rukai, line 82: If we assert that we get status ok, then the error reported to the developer is not very useful, all they know is that the status wasnt 200.
Instead, lets use a helper like this:
```rust
async fn assert_ok_and_get_json(response: Result<Response, Error>) -> Value {
    let response = response.unwrap();

    let status = response.status_code();
    let json = response.json::<Value>().await.unwrap();
    if status != StatusCode::OK {
        panic!("Opensearch query failed: {json:#?}");
    }
    json
}
```
Every query that we make to opensearch should use this helper, it will assert that the query succeeded, giving us the json if it did succeed.
If it fails then we get the error message from opensearch displayed in a panic.
e.g. I edited a query which was invalid, was I wrote this helper I got the following message which helped me understand what went wrong:

shotover-proxy/tests/opensearch_int_tests/mod.rs
```diff
@@ -1,30 +1,27 @@
...
-    BulkOperation, BulkParts, OpenSearch, SearchParts,
+    BulkOperation, BulkParts, DeleteParts, IndexParts, OpenSearch, SearchParts,
 };
 use serde_json::{json, Value};
 use test_helpers::docker_compose::docker_compose;
 
-pub async fn index_documents(client: &OpenSearch) -> Response {
-    let index = "posts";
-    let exists_response = client
+pub async fn test_bulk(client: &OpenSearch) {
```
rukai, line 17: If we sort our search results then we can assert that we get the exact list of items that we expect:
```rust
pub async fn test_bulk(client: &OpenSearch) {
    assert_ok_and_get_json(
        client
            .indices()
            .create(IndicesCreateParts::Index("posts"))
            .send()
            .await,
    )
    .await;

    let mut body: Vec<BulkOperation<_>> = vec![];
    for i in 0..10 {
        let op = BulkOperation::index(json!({"title": "OpenSearch", "i": i}))
            .id(i.to_string())
            .into();
        body.push(op);
    }

    assert_ok_and_get_json(
        client
            .bulk(BulkParts::Index("posts"))
            .body(body)
            .refresh(Refresh::WaitFor)
            .send()
            .await,
    )
    .await;

    let results = assert_ok_and_get_json(
        client
            .search(SearchParts::None)
            .body(json!({
                "query": {
                    "match_all": {}
                },
                "sort": [
                    {
                        "i": {
                            "order": "asc"
                        }
                    }
                ]
            }))
            .allow_no_indices(true)
            .send()
            .await,
    )
    .await;

    assert!(results["took"].is_i64());
    let hits = results["hits"]["hits"].as_array().unwrap();
    assert_eq!(
        hits.iter().map(|x| &x["_source"]).collect::<Vec<_>>(),
        vec!(
            &json!({ "title": "OpenSearch", "i": 0 }),
            &json!({ "title": "OpenSearch", "i": 1 }),
            &json!({ "title": "OpenSearch", "i": 2 }),
            &json!({ "title": "OpenSearch", "i": 3 }),
            &json!({ "title": "OpenSearch", "i": 4 }),
            &json!({ "title": "OpenSearch", "i": 5 }),
            &json!({ "title": "OpenSearch", "i": 6 }),
            &json!({ "title": "OpenSearch", "i": 7 }),
            &json!({ "title": "OpenSearch", "i": 8 }),
            &json!({ "title": "OpenSearch", "i": 9 }),
        )
    );
}
```

shotover-proxy/tests/opensearch_int_tests/mod.rs
rukai, line 113: a few improvements here:
```rust
    assert!(results["took"].is_i64());
    let hits = results["hits"]["hits"].as_array().unwrap();
    assert_eq!(
        hits.iter().map(|x| &x["_source"]).collect::<Vec<_>>(),
        vec!(&json!({
            "name": "John",
            "age": 30,
        }))
    );
    hits[0]["_id"].as_str().unwrap().to_owned()
```

* We can use `is_i64()` instead of `as_i64().is_some()`
* We can assert a list of expected `_sources` rather than just the length.
* We can use `["hits]` directly on a `Value` instead of calling `.as_object()`

shotover/src/codec/opensearch.rs
```diff
@@ -0,0 +1,332 @@
...
+                    if src.is_empty() {
+                        // sometimes messages will have a content-length but no
+                        // body ?
+                        return Ok(Some(vec![Message::from_frame(Frame::OpenSearch(
+                            OpenSearchFrame::new(parts, bytes::Bytes::new()),
+                        ))]));
+                    }
+
+                    if src.len() < content_length {
+                        self.state = State::ReadingBody(parts, content_length);
```
rukai, line 199: this looks like a noop to me. 
If we reach this line then we are already in the ReadingBody state.

shotover/src/codec/opensearch.rs
```diff
@@ -0,0 +1,332 @@
...
+
+impl OpenSearchDecoder {
+    pub fn new(direction: Direction) -> Self {
+        Self {
+            direction,
+            state: State::ParsingResponse,
+        }
+    }
+
+    fn decode_request(&self, src: &mut BytesMut) -> Result<Option<(usize, HttpHead, usize)>> {
```
rukai, line 50: This should return a struct, its not clear what all the fields of this tuple mean.

shotover/src/codec/opensearch.rs
rukai: This sounds like a bug to me, we should remove this hack and investigate whats going on.

shotover/src/codec/opensearch.rs
```diff
@@ -0,0 +1,332 @@
...
+                                dst.extend_from_slice(v.as_bytes());
+                                dst.extend_from_slice(b"\r\n");
+                            }
+                        }
+                    }
+
+                    dst.extend_from_slice(b"\r\n");
+                    dst.extend_from_slice(&opensearch_frame.body);
+
+                    tracing::debug!(
```
rukai, line 275: this should go outside of the match so that it runs for both Encodable::Bytes and Encodable::Frame

shotover/src/codec/opensearch.rs
rukai: I found that the issue occurs on:
```rust
        client
            .indices()
            .exists(IndicesExistsParts::Index(&["test-index"]))
            .send()
            .await,
```

And that if I do:
```rust
    assert_ok_and_get_json(
        client
            .indices()
            .exists(IndicesExistsParts::Index(&["test-index"]))
            .send()
            .await,
    )
    .await;
```

Then I get a panic in `assert_ok_and_get_json`:
```
thread 'opensearch_int_tests::passthrough_standard' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Http(reqwest::Error { kind: Decode, source: Error("EOF while parsing a value", line: 1, column: 0) }) }', shotover-proxy/tests/opensearch_int_tests/mod.rs:22:47
```

Both with and without shotover.
So seems to be a bug in either opensearch or the rust opensearch driver.

rukai: Ah! figured it out.
The client sends an HTTP HEAD request
And as per https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD a head request, requests for just the headers without the body.
So we should change this to check for HEAD instead of `src.is_empty()`

shotover-proxy/tests/opensearch_int_tests/mod.rs
rukai: Now that we know about `HEAD` requests, we should write this function like:
```rust
async fn assert_ok_and_get_json(response: Result<Response, Error>) -> Value {
    let response = response.unwrap();
    let status = response.status_code();

    if response.method() == Method::Head {
        if status != StatusCode::OK {
            panic!("Opensearch HEAD query returned status code {status}");
        }
        Value::Null
    } else {
        let json = response.json().await.unwrap();
        if status != StatusCode::OK {
            panic!("Opensearch query failed: {json:#?}");
        }
        json
    }
}
```

shotover/src/codec/opensearch.rs
```diff
@@ -0,0 +1,332 @@
...
+        let mut response = httparse::Response::new(&mut headers);
+
+        let body_start = match response.parse(src).unwrap() {
+            httparse::Status::Complete(body_start) => body_start,
+            httparse::Status::Partial => return Ok(None),
+        };
+        match response.version.unwrap() {
+            1 => (),
+            version => {
+                return Err(anyhow!(format!(
```
rukai, line 116: anyhow! already does interpolation, no need for the format macro.

rukai:
Once we land this PR I think we should investigate using hyper instead of reimplementing so much http logic.
This example demonstrates a simple proxy with hyper https://github.com/hyperium/hyper/blob/master/examples/gateway.rs
We would be replacing our typical TcpCodecListener that we use in our Source implementations with `http1::Builder::new().serve_connection(io, service).await`.
And then our hyper service implementation would put the request in a shotover `Message` and send it down the transform chain.
I suspect that using hyper will gives us HTTP2 support for free.

I still think its a good idea to land this PR first though as it gives us a baseline to compare against.

shotover/src/codec/opensearch.rs
conorbros: That would require some shared state between the encode and decode sides of the codecs, as the response to a HEAD request has nothing indicating it is that, it just responds the same as a GET request, but only the headers.

shotover-proxy/tests/opensearch_int_tests/mod.rs
```diff
@@ -1,89 +1,240 @@
...
+        vec!(&json!({
+            "name": "John",
+            "age": 30,
+        }))
+    );
+    hits[0]["_id"].as_str().unwrap().to_owned()
+}
+
+async fn test_delete_and_search_document(client: &OpenSearch, id: String) {
+    client
```
rukai, line 162: We should use `assert_ok_and_get_json` on all queries that we expect to succeed.

shotover/src/codec/opensearch.rs
rukai: sounds like we'll need to add that shared state then.
checking `src.is_empty()` is no good as if we happen to perform the check just before receiving the bytes of the body we will ignore the body entirely.

conorbros: Hmm, I had imagined that case would be very unlikely. Anyway the state wasn't difficult to implement just thought that it was better avoided if possible.

shotover/src/transforms/opensearch/mod.rs
conorbros: Looking into this more, the current state of the sink is closer to `RedisSinkSingle` than the cluster implementation. I think keeping it simplest would be the best path forward for the time being.

rukai: is this supposed to be a sink cluster or a single cluster?
We should name it either OpenSearchSinkSingle or OpenSearchSinkCluster to be consistent with our other sinks.

rukai: I swapped this over to `spawn_read_write_tasks` and the tests still pass:
```diff
diff --git a/shotover/src/transforms/opensearch/mod.rs b/shotover/src/transforms/opensearch/mod.rs
index a91727a6..d7d8ca7d 100644
--- a/shotover/src/transforms/opensearch/mod.rs
+++ b/shotover/src/transforms/opensearch/mod.rs
@@ -1,13 +1,16 @@
-use crate::codec::{
-    opensearch::{OpenSearchCodecBuilder, OpenSearchDecoder, OpenSearchEncoder},
-    CodecBuilder, Direction,
-};
 use crate::message::Message;
 use crate::tcp;
 use crate::tls::AsyncStream;
 use crate::transforms::{
     Messages, Transform, TransformBuilder, TransformConfig, Transforms, Wrapper,
 };
+use crate::{
+    codec::{
+        opensearch::{OpenSearchCodecBuilder, OpenSearchDecoder, OpenSearchEncoder},
+        CodecBuilder, Direction,
+    },
+    transforms::util::cluster_connection_pool::spawn_read_write_tasks,
+};
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
 use futures::{FutureExt, SinkExt, StreamExt};
@@ -15,10 +18,13 @@ use serde::Deserialize;
 use std::pin::Pin;
 use std::time::Duration;
 use tokio::io::{ReadHalf, WriteHalf};
-use tokio::sync::mpsc;
+use tokio::sync::{mpsc, oneshot};
 use tokio_util::codec::{FramedRead, FramedWrite};
 use tracing::{trace, Instrument};
 
+use super::util::cluster_connection_pool::Connection;
+use super::util::Request;
+
 type PinStream = Pin<Box<dyn AsyncStream + Send + Sync>>;
 
 #[derive(Deserialize, Debug)]
@@ -96,22 +102,9 @@ impl Transform for OpenSearchSink {
             trace!("creating outbound connection {:?}", self.address);
 
             let tcp_stream = tcp::tcp_stream(self.connect_timeout, self.address.clone()).await?;
-            let stream = Box::pin(tcp_stream) as Pin<Box<dyn AsyncStream + Send + Sync>>;
-
-            let (decoder, encoder) = self.codec_builder.build();
-            let (stream_rx, stream_tx) = tokio::io::split(stream);
-            let outbound_tx = FramedWrite::new(stream_tx, encoder);
-            let outbound_rx = FramedRead::new(stream_rx, decoder);
-            let (response_messages_tx, response_messages_rx) = mpsc::unbounded_channel();
-
-            tokio::spawn(
-                server_response_processing_task(outbound_rx, response_messages_tx)
-                    .in_current_span(),
-            );
-            self.connection = Some(Connection {
-                response_messages_rx,
-                outbound_tx,
-            })
+            let (rx, tx) = tcp_stream.into_split();
+
+            self.connection = Some(spawn_read_write_tasks(&self.codec_builder, rx, tx));
         }
 
         let connection = self.connection.as_mut().unwrap();
@@ -120,61 +113,18 @@ impl Transform for OpenSearchSink {
 
         let mut result = Vec::with_capacity(messages_len);
         for message in requests_wrapper.requests {
+            let (tx, rx) = oneshot::channel();
             connection
-                .outbound_tx
-                .send(vec![message])
-                .await
-                .map_err(|err| {
-                    anyhow!("Failed to send messages to OpenSearch destination: {err:?}")
-                })?;
-
-            let message = connection
-                .response_messages_rx
-                .recv()
-                .await
-                .ok_or_else(|| anyhow!("Failed to receive message because OpenSearchSink response processing task is dead"))?;
+                .send(Request {
+                    message,
+                    return_chan: Some(tx),
+                })
+                .map_err(|_| anyhow!("Failed to send"))?;
+
+            let message = rx.await?.response?;
             result.push(message);
         }
 
         Ok(result)
     }
 }
-
-struct Connection {
-    outbound_tx: FramedWrite<WriteHalf<PinStream>, OpenSearchEncoder>,
-    response_messages_rx: mpsc::UnboundedReceiver<Message>,
-}
-
-async fn server_response_processing_task(
-    mut outbound_rx: FramedRead<ReadHalf<PinStream>, OpenSearchDecoder>,
-    response_messages_tx: mpsc::UnboundedSender<Message>,
-) {
-    loop {
-        tokio::select! {
-            responses = outbound_rx.next().fuse() => {
-                match responses {
-                    Some(Ok(messages)) => {
-                        for message in messages {
-                            if let Err(mpsc::error::SendError(_)) = response_messages_tx.send(message) {
-                                // RawSinkSingle dropped after a message was received from server,
-                                // request processor task shutting down
-                                return;
-                            }
-                        }
-                    }
-                    Some(Err(err)) => {
-                        tracing::error!("encountered error in OpenSearchSink stream: {err:?}");
-                        return;
-                    }
-                    None => {
-                        return;
-                    }
-                }
-            },
-            _ = response_messages_tx.closed() => {
-                // sink stream ended, shutdown the response processing task
-                return;
-            },
-        }
-    }
-}

```

conorbros: I'm happy to switch it over to avoid adding more lines of code, I just found the initial one easier to reason about.

rukai: yep I agree that spawn_read_write_tasks has a very complicated implementation. :/
In the future we should investigate if we can simplify it, I'm thinking we might be able to change it to use a single task without a performance hit.

In the meantime we are better off using a known good implementation rather than reimplementing spawn_read_write_tasks.
Either we do end up simplifying it and then all sinks benefit from the simplification or we dont because the complexity was actually important.

shotover/src/transforms/opensearch/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+use tracing::trace;
+
+#[derive(Deserialize, Debug)]
+pub struct OpenSearchSinkSingleConfig {
+    #[serde(rename = "remote_address")]
+    address: String,
+    connect_timeout_ms: u64,
+}
+
+#[typetag::deserialize(name = "OpenSearchSink")]
```
rukai, line 26: This is the name used by deserialization.
This needs to be OpenSearchSinkSingle.

shotover/src/transforms/opensearch/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+        Transforms::OpenSearchSinkSingle(OpenSearchSinkSingle {
+            address: self.address.clone(),
+            connect_timeout: self.connect_timeout,
+            codec_builder: OpenSearchCodecBuilder::new(Direction::Sink),
+            connection: None,
+        })
+    }
+
+    fn get_name(&self) -> &'static str {
+        "OpenSearchSink"
```
rukai, line 66: We should also update this string as its user facing.

rukai (approved):
I think we are ready to land this.

shotover-proxy/tests/opensearch_int_tests/mod.rs
```diff
@@ -1,52 +1,242 @@
...
+    let status = response.status_code();
 
-    assert_eq!(exists_response.status_code(), StatusCode::NOT_FOUND);
+    if response.method() == Method::Head {
+        if status != StatusCode::OK {
+            panic!("Opensearch HEAD query returned status code {status}");
+        }
+        Value::Null
+    } else {
+        let json = response.json().await.unwrap();
```
cjrolo, line 27: We always get a valid json even if the status is not ok? Shouldn't we unwrap after testing the response status?

rukai: Yep, opensearch returns valid json even when the http code is not successful.
We include the json in our panic message because it includes a detailed error message.

cjrolo (approved):
This was a long review, dived into some of the shotover internals, but for sure missed a lot of edges (and non-edges too!).  This should be more of a "non-binding approval" since I lack the knowledge about the internal system. BUT at the same time, it is a new part, so I get to know this one better.

cjrolo (approved):

rukai:
Thanks!

PR #1271, opened 2023-08-03
windsock: aws cassandra topology=single + shotover/cassandra-test

rukai:
With this PR:
* cassandra topology=single benches are now working (previously only clustered was supported)
* cassandra-test images are used for a faster bench run time.
    + time taken to run every cassandra cluster bench was previously 19m now its 10mins

In order to use cassandra-test we need to manually set the tokens, so I had to refactor a bunch of things to set those tokens.


conorbros (approved):

cjrolo (approved):
Approved.

PR #1272, opened 2023-08-04
Add user facing changelog

rukai:
After spending a lot of time back and forth internally on an issue that was caused by us changing an API without documenting that change in an easily accessible place I have decided to introduce a changelog that we and users can refer to for information on such changes.

We already have the changelogs generated on releases such as https://github.com/shotover/shotover-proxy/releases/tag/v0.1.10 but they have so much noise from all the unrelated PRs and the PR descriptions are not directly useful when trying to upgrade to a new version or diagnose issues from an upgrade.

I think both systems can coexist, the automated changelogs are a complete list of every PR that went into the release while this changelog will be a hand curated reference to refer to when upgrading to new releases.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,0 +1,17 @@
+# User Facing Changelog
+
+Any breaking changes to the `topology.yaml` or `shotover` rust API should be documented here.
+This assists us in knowing when to make the next release a breaking release and assists users with making upgrades to new breaking releases.
+
+## 0.1.10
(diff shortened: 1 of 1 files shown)


conorbros (approved):

benbromhead:
Good idea, top level apache projects do something similar.

benbromhead (approved):

Issue #1274, opened 2023-08-04
windsock: get stable results


rukai:
Running the same bench twice does not give stable results, neither locally nor on AWS
But I'm focusing on AWS atm because that feels more important.

network and disk IO does not go above 5MB, so it feels unlikely that we are hitting limits there.

I've noticed that shotover benches will tend to be either +100% or -100% for the entire bench
While non shotover benches will vary around 0.0% rising up to +100% or down to -100% before returning to 0.0%

So it looks like shotover is introducing far worse noise than without.
So we should first address the noise without shotover.

cassandra benches have the bencher setup to only use 1 thread.
using 2 threads on a m6a.large instance seems to make noise worse
But maybe using more threads on a larger instance will help?

rukai:
This PR has shown promise: https://github.com/shotover/shotover-proxy/pull/1360

However, I think the next step is to add functionality to windsock to allow reusing EC2 instances.
This will eliminate the noise caused by differences in EC2 instances.
I am thinking an API like this:
```shell
> # Create the resources required to run the benches specified in FILTER and then store the information required to access those instances to disk
> cargo windsock --store-cloud-resources-to-disk FILTER
Creating AWS resources: CloudResourcesRequired {
    shotover_instance_count: 1,
    docker_instance_count: 3,
    include_shotover_in_docker_instance: false,
}
> # Run the benches once, using the instances created in the previous command.
> cargo windsock --use-cloud-resources-from-disk FILTER
Running "kafka,shotover=standard,size=100KB,topology=cluster3"
...
> # Run the benches a second time reusing the same instances
> cargo windsock --use-cloud-resources-from-disk FILTER
Running "kafka,shotover=standard,size=100KB,topology=cluster3"
...
> cargo windsock --cleanup-cloud-resources
```



After that is implemented it should be easier to evaluate #1360

PR #1276, opened 2023-08-08
scylla driver: disable topology refresh

rukai:
progress towards: https://github.com/shotover/shotover-proxy/issues/1274

We cant disable topology refresh but setting an absurdly long refresh interval is close enough.

This resolves the dip in OPS that can be observed in cassandra benches between 42-45s into the bench:
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4202,8 +4202,7 @@ dependencies = [
 [[package]]
 name = "scylla"
 version = "0.9.0"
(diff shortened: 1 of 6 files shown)


conorbros (approved):
LGTM. I would just add a comment explaining the reason, i.e. topology refresh adds alot of noise to the benches so we put the interval at a very high value so it doesn't occur

rukai:
good point

benbromhead (approved):

PR #1277, opened 2023-08-09
windsock: add sys monitor profiler - archive support

rukai:
## Goals

Something I've been missing for a while now in windsock is the ability to record system usage metrics.
I can ssh into the instances and run htop but:
* That is a lot of manual setup for each benchmark run, making it infeasible when performing many bench runs which teardown and recreate ec2 instances for each run.
* I need to be able to view a log of system usage, its hard to analyze anything while it flys by on htop
* I need to be able to correlate system usage to particular benchmarks
* I need to be able to correlate system usage to per second benchmark results.

Being able to do those things will allow us to:
* Use changes in system usage metrics as another data point when evaluating shotover changes
* Evaluate how well our benchmarks are designed.
    + Allow us to track down the cause of noisy benchmarks
    + Make sure shotover is the bottleneck so that changes to shotover show up in the results.

## Implementation

Implementation wise, I've decided to achieve this by running the unix `sar` command on each instance and then processing its output.
We could swap out `sar` for something else in the future but it seems reasonable enough and sar also supports disk and network usage so we will be able to make use of that in the future.

The process for sys monitoring looks like this:
1. We begin running sar before the bench runs.
2. The bench completes and writes the bench archive to disk.
3. we read all the remaining sar output, parse it and terminate sar
4. Load the bench archive from disk
5. Use the parsed sar metrics to insert metrics into the bench archive
6. Write the bench archive back to disk.
7. Then when the bench archive is displayed the user generated metrics are included in the results tables.

## Example

I ran:
```shell
cargo windsock --cloud --profilers sys_monitor "name=cassandra topology=single driver=scylla"
cargo windsock --results-by-tags "name=cassandra topology=single driver=scylla"
```

And the output is screenshotted below.

image
image

If an instance was not used for a certain bench then it just shows up as empty in the table, as can be seen by shotover in the above screenshot.

Future work:
* record disk and network usage metrics
* Add another table type to windsock for displaying all the results of a single bench in a table.
    * Currently when displaying a single bench the results are effectively a list as we only have one column. We should instead display a single bench with each "per second" results as their own column.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5803,6 +5803,7 @@ dependencies = [
  "scylla",
  "serde",
  "strum 0.25.0",
+ "time 0.3.25",
  "tokio",
 ]
 
--- a/shotover-proxy/examples/windsock/profilers/sar.rs
+++ b/shotover-proxy/examples/windsock/profilers/sar.rs
@@ -8,15 +8,90 @@ use tokio::{
     process::Command,
     sync::mpsc::{unbounded_channel, UnboundedReceiver},
 };
+use windsock::{Goal, Metric, ReportArchive};
 
 /// Reads the bench archive for `bench_name` from disk.
 /// Inserts the passed sar metrics for `instance_name`.
 /// Then writes the resulting archive back to disk
-pub fn insert_sar_results_to_bench_archive(bench_name: &str, instance_name: &str, sar: ParsedSar) {
-    // TODO: Insert to archive instead of dumping to stdout
-    println!("sys_monitor results for {bench_name} - {instance_name}");
-    println!("started at: {}", sar.started_at);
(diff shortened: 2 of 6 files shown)


conorbros (approved):

benbromhead:
Happy to merge this as is. One question I have now is given the amount of info we are now generating, do we have a simple mechanism for visualisation? E.g. a quick html page or something with a graph?

benbromhead (approved):

rukai:
> do we have a simple mechanism for visualisation? E.g. a quick html page or something with a graph?

Not yet, but I'm very excited to jump onto graph generation when I get the chance.

On the meantime, the extra data generated by the sys_monitor profiler is off by default so we aren't hiding the main results in the default case.

PR #1278, opened 2023-08-10
windsock: add os monitor profiler - direct to stdout

rukai:
A chunk of implementation split out from https://github.com/shotover/shotover-proxy/pull/1277
Please refer to 1277 to better understand what the final feature will look like.

image


conorbros (approved):

cjrolo (approved):
Approving but leaving 2 suggestions:

1. SAR can output to JSON/XML/CSV, should be easy to parse (https://github.com/sysstat/sysstat tool called sadf)
"sadf displays data collected by sar in multiple formats (CSV, XML, JSON, etc.) and can be used for data exchange with other programs. This command can also be used to draw graphs for the various activities collected by sar using SVG (Scalable Vector Graphics) format."

2. Shouldn't we look into pulumi to make the instance creation cloud agnostic (Not really sure if it would work here)

rukai:
> 1. SAR can output to JSON/XML/CSV ...
I did explore a few options here.
First I planned on having a windsock specific binary that gets uploaded and sends back messages, this is nice from the perspective that we can share message structures between windsock and the binary.
Once I discovered sar I figured I was better off using a tool I can just install via package manager to avoid logistics of building and uploading this extra binary.

If SAR's stdout format ends up brittle then I will abandon this approach and use sadf or another tool entirely.
However the current approach does provide the advantage that metrics are streamed live which would be really handy if I ever refactor windsock to display data live as the benches run.

2.
I dont believe pulumi has a rust client.
Since the goal is to use windsock to evaluate the performance of changes to shotover, being cloud agnostic isnt currently a goal.
That said, it would be really nice to be cloud agnostic one day for the purpose of publishing graphs of shotover's performance on different cloud providers.
But even for that purpose just AWS will likely be enough.

PR #1283, opened 2023-08-11
Shotover release v0.1.11

conorbros:


shotover-proxy/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "shotover-proxy"
-version = "0.1.10"
+version = "0.1.11"
```
rukai, line 3: ah oops

.github/workflows/build_and_test.yaml
```diff
@@ -56,6 +56,8 @@ jobs:
       run: cargo install cargo-hack --version 0.5.8
     - name: Check `cargo fmt` was run
       run: cargo fmt --all -- --check
+    - name: Check that the project is releaseable
+      run: shotover-proxy/build/is_releasable.sh
```
rukai, line 60: you wont be able to run this script on every PR since the logic of the script depends on the release tag being on the current commit.

rukai:
I think the solution here is to land this PR to fix the version and then tag this PR as the release instead.

.github/workflows/build_and_test.yaml
conorbros: Yep, just realised.

conorbros:
Yep, I should be able to rename the previous commit too.

Cargo.lock
```diff
@@ -66,9 +66,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.0.2"
```
rukai, line 69: we should not be bumping dependencies in a release commit

rukai (approved):
Context for second reviewer:
We previously attempted to make a release but it failed because while we bumped the `shotover` crate version, we forgot to bump the `shotover-proxy` crate.
This PR fixes this issue by bumping `shotover-proxy`.
we will then tag this commit as `v0.1.11`

benbromhead (approved):

PR #1287, opened 2023-08-15
KafkaSinkCluster: refactor

rukai:
Its been a while since I wrote this code and I can see now that its quite hard to understand.
This PR is an attempt to fix that by:
* rename `coordinator_broker_id` to `group_to_coordinator_broker`
    + it is a map so its important to convey what values maps to what.
* Swap to kafka-protocols newtype wrappers e.g. `StrBytes` -> `GroupId`, `StrBytes` -> `TopicName`, `i32` -> `BrokerId`
    + This improves type safety by ensuring that we cant mixup a group id with a topic name.
    + Also documents the purpose of the type making the code easier to understand.
    + Doesnt end up complicating things much at all since we were often extracting the inner StrBytes from the newtype only to reconstruct the newtype again later on when we passed the value back to `kafka-protocol`
* Pull large chunks of logic into new functions e.g. `find_coordinator_of_group` and `get_metadata_of_topics`

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster.rs
+++ b/shotover/src/transforms/kafka/sink_cluster.rs
@@ -12,8 +12,9 @@ use async_trait::async_trait;
 use dashmap::DashMap;
 use kafka_protocol::messages::metadata_request::MetadataRequestTopic;
 use kafka_protocol::messages::{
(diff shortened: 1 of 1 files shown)


conorbros (approved):

benbromhead (approved):
Nice, great readability improvement

Issue #1289, opened 2023-08-15
unknown topology.yaml fields are ignored and startup succeeds


rukai:
We could solve this by adding `#[serde(deny_unknown_fields)]` everywhere, but im hoping to find a more global solution.

PR #1291, opened 2023-08-15
shotover config: Fix ignored fields

rukai:
closes https://github.com/shotover/shotover-proxy/issues/1289

Before this PR an unknown field would in a config.yaml or topology.yaml would just be ignored.
This led to issues in our internal use of shotover.
This PR changes shotover to fail to startup if an unknown field is found in configuration.

Unfortunately I could not find a way to achieve this globally without an easily forgotten serde config per struct.
I guess this is just another argument towards migrating to something better designed for configuration instead of serde_yaml.


conorbros (approved):

benbromhead (approved):
Can we improve serde_yaml with a PR?

rukai:
Unfortunately not, this is by design, they swapped over to ignore by default back in 2016

PR #1310, opened 2023-08-27
Move rdkafka back to crates.io

rukai:
We previously had to use git here due to a bug in the driver, now the new version has been released we can go back to proper versioning.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3654,8 +3654,9 @@ dependencies = [
 
 [[package]]
 name = "rdkafka"
-version = "0.31.0"
-source = "git+https://github.com/shotover/rust-rdkafka?branch=updating_librdkafka_to_v2.1.1#1c6a3160ff7f3f33f00cb395397c256395f39e74"
+version = "0.34.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "053adfa02fab06e86c01d586cc68aa47ee0ff4489a59469081dc12cbcde578bf"
 dependencies = [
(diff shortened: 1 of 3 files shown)


rukai:
will investigate failure

rukai:
The cause is the same as the original issue that caused us to switch to the git dep in https://github.com/shotover/shotover-proxy/pull/1228

Current theory why this happened:

The branch that we depended on never made it to main, instead a different PR to upgrade the underlying C library was merged. Maybe there was some extra changes in that branch that fixed the issue somehow.
Maybe its some spooky UB issue caused by slight change in compilation.

rukai:
I was never able to fully understand what the issue is in the rdkafka driver, but I was able to fix the issue by ensuring that shotover never returned duplicate addresses in its list of brokers.

It feels a bit unfortunate that we need to perform deduplication as it adds a bunch of complexity.
The drivers should probably be able to handle this case, but its also not something that kafka itself would ever do so it seems unreasonable to request that they do handle it.
Additionally say if we fixed it in the C driver similar issues could appear in other drivers.

conorbros (approved):

cjrolo (approved):
Is this MetadataResponse structure small? Otherwise we might iterate over a large object several times.

rukai:
I assume you mean MetadataResponseBroker? It looks like this:
image

Looking at this with fresh eyes though, I think using MetadataResponseBroker is wrong because we dont want to include other fields in the comparison, I'll replace it with KafkaAddress.

PR #1314, opened 2023-08-30
ShotoverProcessBuilder: fix with_profile

rukai:
blocked on: https://github.com/shotover/tokio-bin-process/pull/22

`with_profile` should overwrite `with_bin`.
`with_bin` is a hint to `ShotoverProcessBuilder` that there is a precompiled binary available.
`with_profile` is demanding that we use a specific profile which has priority over `with_bin`.

This bug was causing windsock to profile shotover binaries not suitable for profiling.

### DIFF
--- a/test-helpers/src/shotover_process.rs
+++ b/test-helpers/src/shotover_process.rs
@@ -30,21 +30,27 @@ impl ShotoverProcessBuilder {
         }
     }
 
+    /// Hint that there is a precompiled shotover binary available.
+    /// This binary will be used unless a profile is specified.
     pub fn with_bin(mut self, bin_path: &Path) -> Self {
         self.bin_path = Some(bin_path.to_owned());
         self
     }
 
+    /// Prefix forwarded shotover logs with the provided string.
+    /// Use this when there are multiple shotover instances to give them a unique name in the logs.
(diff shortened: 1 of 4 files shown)


rukai:
will need to investigate this failure

test-helpers/src/shotover_process.rs
```diff
@@ -30,21 +30,27 @@ impl ShotoverProcessBuilder {
...
         self
     }
 
+    /// Run shotover with the specified number of cores
     pub fn with_cores(mut self, cores: u32) -> Self {
         self.cores = Some(cores.to_string());
         self
     }
 
+    /// Force shotover to recompiled with the specified profile
```
rukai, line 53: `to be recompiled`

rukai:
The mysterious GA failures were resolved by making tokio-bin-process only compile the exact binary it needs rather than the entire workspace: https://github.com/shotover/tokio-bin-process/pull/22

conorbros (approved):

rukai:
Nix version needed to be updated for compatibility with the new tokio-bin-process version.

cjrolo (approved):
Thanks for the comments!

PR #1315, opened 2023-08-30
Add samply as a windsock profiler

rukai:
Samply uses a custom rust sampling profiler and then displays the results in the firefox profiler UI.
The firefox profiler UI can display the results in a flamegraph and a few other formats.

I renamed the tokio thread names to `shotover-worker` as the previous name was so long it gets truncated on linux.

### samply
`cargo windsock --profilers samply --name kafka,shotover=message-parsed,size=100KB,topology=single`
image
Interactive: https://share.firefox.dev/47Vvy10

### perf + inferno
`cargo windsock --profilers flamegraph --name kafka,shotover=message-parsed,size=100KB,topology=single`
Screen Shot 2023-08-30 at 16 23 13-fullpage

For some reason perf places TcpCodecListener in an `[unknown]` function apart from the rest of the tasks running in tokio.
However samply correctly places this in `tokio::runtime::scheduler::multi_thread::worker::Context::run` along with the rest of the tasks called by tokio.
So samply seems more accurate to me.

## Usage
1. run `cargo windsock --profilers samply --name kafka,shotover=message-parsed,size=100KB,topology=single`
2. run `samply load samply.json` in `shotover/target/windsock_data/profiler_results/kafka,shotover=message-parsed,size=100KB,topology=single`

## Future work

I'll add support for running samply on cloud benches once https://github.com/shotover/shotover-proxy/pull/1309 lands and I can run cloud benches again.

I'll keep the existing perf+inferno profiler around for now, it might be useful to compare results if samply does anything strange.
However we could delete it in the future to avoid maintaining it and improve compile time.

### DIFF
--- a/shotover/src/runner.rs
+++ b/shotover/src/runner.rs
@@ -182,7 +182,7 @@ impl Shotover {
         let mut runtime_builder = runtime::Builder::new_multi_thread();
         runtime_builder
             .enable_all()
-            .thread_name("Shotover-Proxy-Thread")
+            .thread_name("shotover-worker")
             .thread_stack_size(stack_size);
         if let Some(worker_threads) = worker_threads {
             runtime_builder.worker_threads(worker_threads);
--- a/shotover/src/runner.rs
+++ b/shotover/src/runner.rs
@@ -182,7 +182,7 @@ impl Shotover {
         let mut runtime_builder = runtime::Builder::new_multi_thread();
         runtime_builder
             .enable_all()
-            .thread_name("shotover-worker")
+            .thread_name("Shotover-Proxy-Thread")
             .thread_stack_size(stack_size);
         if let Some(worker_threads) = worker_threads {
             runtime_builder.worker_threads(worker_threads);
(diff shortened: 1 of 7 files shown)


shotover/src/runner.rs
```diff
@@ -182,7 +182,7 @@ impl Shotover {
         let mut runtime_builder = runtime::Builder::new_multi_thread();
         runtime_builder
             .enable_all()
-            .thread_name("Shotover-Proxy-Thread")
+            .thread_name("shotover-worker")
```
conorbros, line 185: Does this have any other side effects?

rukai: Afaik it just changes the process name in htop and similar

conorbros: That could cause scripts/other tools that rely on the name being "Shotover-Proxy-Thread" to break.

rukai: hmm fair enough, I'll pull it into its own PR so it can be considered seperately

conorbros (approved):

cjrolo (approved):
In the conversation the screenshots don't seem to provide the exact same result for the same benchmark (Is it expected?). So, comparison between doesn't seems perfect. Given my less experience with the codebase, I'm approving because it does seem to provide more value/clarity than the previous profiler.

PR #1316, opened 2023-08-31
windsock aws: samply support

rukai:
blocked on https://github.com/shotover/shotover-proxy/pull/1315

This PR runs samply on an AWS instance and retrieves the results back to the local machine.
However for some reason the results contain only a single sample which is clearly broken, I will need to investigate whats going wrong.

### DIFF
--- a/.github/workflows/windsock_benches.yaml
+++ b/.github/workflows/windsock_benches.yaml
@@ -32,6 +32,10 @@ jobs:
         save-if: ${{ github.ref == 'refs/heads/main' }}
     - name: Ensure that custom benches run
       run: |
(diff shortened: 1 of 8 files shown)


rukai:
ah, AWS does not support perf events on shared hosts.
The largest instances do support it (excluding burstable instances) but they are very expensive...)
https://stackoverflow.com/questions/69395912/pmu-x86-64-performance-counters-not-showing-in-perf-under-aws

rukai:
This feature is not practical as we cant actually run sampling profilers on shared aws instances. (and bare metal is too expensive)

PR #1318, opened 2023-08-31
fix tokio thread_name

rukai:
pulled out of and continuing discussion from https://github.com/shotover/shotover-proxy/pull/1315#discussion_r1311109557

After enabling "show custom thread names" in htop (its [disabled by default](https://askubuntu.com/questions/987957/how-to-see-thread-name-in-htop)) we get output like this:
image
It looks truncated because linux does not support thread names that long.

This PR fixes the issue by using a shorter thread name.
It now looks like this:
image

While there could theoretically be some obscure case out there, at our current scale I dont believe this will cause any issues due to the change of name.
Scripts that invoke things like `killall shotover-proxy` will be unaffected because the main thread is still called `shotover-proxy`.

### DIFF
--- a/shotover/src/runner.rs
+++ b/shotover/src/runner.rs
@@ -182,7 +182,7 @@ impl Shotover {
         let mut runtime_builder = runtime::Builder::new_multi_thread();
         runtime_builder
             .enable_all()
-            .thread_name("Shotover-Proxy-Thread")
+            .thread_name("shotover-worker")
             .thread_stack_size(stack_size);
         if let Some(worker_threads) = worker_threads {
             runtime_builder.worker_threads(worker_threads);


conorbros (approved):
>Scripts that invoke things like killall shotover-proxy will be unaffected because the main thread is still called shotover-proxy.

I hadn't considered this when I raised that original point, thanks for pulling out though. LGTM.

cjrolo (approved):
Approving, but relying on htop as the go-to tool to identify processes is not always the best. I mostly rely on `ps` for that if working on a server\server tools. For those tools I don't thing we would have limitations on naming size. But I do prefer the shorter name.

rukai:
It is the Linux API that limits to 16 characters not htop https://man7.org/linux/man-pages/man3/pthread_setname_np.3.html

PR #1320, opened 2023-08-31
Tee changes

conorbros:
Some changes that I found myself wanting while developing OpenSearch.

<s>- The ability to set the fail message on `DebugReturner`</s>
- `LogMismatch` behaviour for `Tee`


conorbros:
Ah, will update the docs too before merging.

rukai:
What benefit did you get from setting the fail message on `DebugReturner`?

shotover/src/transforms/tee.rs
```diff
@@ -97,6 +99,7 @@ pub struct Tee {
 
 pub enum ConsistencyBehavior {
     Ignore,
+    LogMismatch,
```
rukai, line 102: Maybe this should be `LogWarningOnMismatch`?

shotover/src/transforms/tee.rs
```diff
@@ -198,6 +203,24 @@ impl Transform for Tee {
...
+                        .process_request(requests_wrapper.clone(), self.timeout_micros),
+                    requests_wrapper.call_next_transform()
+                );
+
+                let tee_response = tee_result?;
+                let chain_response = chain_result?;
+
+                if !chain_response.eq(&tee_response) {
+                    warn!(
+                        "Tee mismatch: chain response: {:?} tee response: {:?}",
```
rukai, line 218: A new line should help readability:
`Tee mismatch: chain response: {:?}\ntee response: {:?}`

rukai: We should probably also generate a better message, the message inner stuff should not be exposed to the user.
But if we just call to_high_level_string that would be expensive for messages that arent already parsed.
Maybe we should just log the raw message bytes or maybe we should avoid including the message at all?

For now I would say use to_high_level_string, if this is a performance issue for anyone they can always use `ConsistencyBehaviour::Ignore`

conorbros: >to_high_level_string that would be expensive for messages

This will only occur on mismatches so I think the performance hit is fine, assuming a mismatch is an error and not part of some expected behaviour when using the Tee transform

docs/src/transforms.md
```diff
@@ -524,6 +524,9 @@ The response from the down-chain transform is returned back up-chain but various
     # If the responses returned by the sub chain do not equal the responses returned by down-chain then return an error.
     # behavior: FailOnMismatch
     #
+    # If the responses returned by the sub chain do not equal the responses returned by down-chain then log the mismatch at the warning level.
+    # behavior: LogOnMismatch
```
rukai, line 528: should be LogWarningOnMismatch

shotover/src/transforms/tee.rs
rukai: If errors are significantly more expensive to handle than regular messages, if say 50% of messages start to error that could be enough to cause shotover to fail to serve the remaining 50% of successful messages.
Thats what I want to be careful of here.

That said since this is one of multiple configuration modes here its probably fine to take the performance hit for more meaningful warnings.
If the user wants no performance hit on mismatch they can use `Ignore`
I'm happy to go ahead with this as is. (just need to fix the small other comment I made)

rukai:
> What benefit did you get from setting the fail message on DebugReturner?

I would still like to know the use case for this before proceeding.

conorbros:
Oops, I thought I had removed that. I thought it would be useful for debugging different subchains but not that much.

rukai (approved):
looks good!

> I thought it would be useful for debugging different subchains but not that much.

That does sound reasonable actually.
Feel free to make a separate PR for it.

cjrolo (approved):

PR #1321, opened 2023-09-01
run_command refactor

rukai:
The core problem driving this PR is I observed that build files for windsock-cloud-docker was taking up 200MB in debug.
This is absurd for a small tool like this so I investigated how to fix it.
The problem was that test_helpers is huge and pulling in all sorts of stuff when all we wanted was a helper to run commands.
Once I moved `run_command` into its own minimal crate and windsock-cloud-docker was reduced to 8MB in debug.

So this PR:
* pulls run_command into its own crate.
    + This helper crate doesnt see as much use as it could because we often want to run commands asynchronously and I wanted to avoid including the tokio dependency here.
    + Alternatively we could just copy run_command into windsock-cloud-docker to avoid creating an entire new crate, I couldnt really decide either way so just went with this.
* While I was thinking about these abstractions I noticed an instance in `ec2-cargo/src/main.rs` where we didnt need to shell out so I removed it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5727,7 +5727,7 @@ dependencies = [
 name = "windsock-cloud-docker"
 version = "0.1.0"
 dependencies = [
- "test-helpers",
+ "subprocess",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


conorbros:
>copy run_command into windsock-cloud-docker to avoid creating an entire new crate 

To me that's the better option since its just one small function.

rukai:
Yep, thats reasonable, I removed the new crate and just copied run_command directly into windsock-cloud-docker + changed it to panic instead of return Result since we always unwrap here.

conorbros (approved):

PR #1323, opened 2023-09-04
Add debug log to FailOnMismatch tee behaviour

conorbros:


rukai (approved):
I know that the debug log only occurs if the application is configured to log at the debug level.
However I wasnt confident that the expressions here wont wont be evaluated when not at debug level.
So I replaced the `chain_response...` expression with `panic!()` and observed that the panic was not called proving that the expressions here arent evaluated unless debug is enabled.

So this change seems quite useful for debugging mismatches without affecting performance in production.

cjrolo (approved):

PR #1325, opened 2023-09-05
Update deps

rukai:
I've patched our bigdecimal version as they are unresponsive to a request for a new version.
It is urgent that we work around this as the new version fixes a flaw in their build.rs that causes most of our dependencies to be rebuilt the second time shotover is compiled (particularly annoying in CI and ec2-cargo) 
see: https://github.com/akubera/bigdecimal-rs/pull/114

The other dependency upgrades are straightforward, the new version of nix requires that we set the `signals` feature to get access to the signals API that we use.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -193,13 +193,13 @@ checksum = "9338790e78aa95a416786ec8389546c4b6a1dfc3dc36071ed9518a9413a542eb"
 
 [[package]]
 name = "async-recursion"
-version = "1.0.4"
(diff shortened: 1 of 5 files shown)


rukai:
ahh blocked on https://github.com/shotover/tokio-bin-process/pull/24 due to nix being part of tokio-bin-process's api

rukai:
datastax cpp cassandra driver + TLS tests are segfaulting :/

conorbros (approved):

PR #1334, opened 2023-09-13
windsock: apply shell escaping where needed

rukai:
Fixes a TODO and a previously unknown case where shell escaping was missing.
This fix allows the running of `cargo run --package windsock-cloud-docker --cloud "name=redis encryption=none"` which previously failed due to `"name=redis encryption=none"` losing its quotes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4314,6 +4314,12 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


conorbros (approved):

cjrolo (approved):
Good module to know about

PR #1338, opened 2023-09-14
add zlib to allowed licenses

conorbros:
The CI on #1298 is failing due to https://github.com/shotover/shotover-proxy/actions/runs/6180379036/job/16776751233?pr=1298#step:4:11

A dependency uses the zlib license which we have not allowed in our `deny.toml`.

Zlib is described as a permissive free software license and has been approved by the Free Software Foundation as a free software license and by the Open Source Initiative as an open source license. It is compatible with the GNU General Public License. [refs](https://en.wikipedia.org/wiki/Zlib_License)

I also removed `pktparse` from our exceptions as we no longer use it as a dependency.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -76,7 +76,8 @@ allow = [
     "BSD-3-Clause",
     "BSD-2-Clause",
     "Unicode-DFS-2016",
-    "BSL-1.0"
+    "BSL-1.0",
+    "Zlib"
 ]
 # List of explictly disallowed licenses
 # See https://spdx.org/licenses/ for list of possible licenses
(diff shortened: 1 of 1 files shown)


rukai:
I cant find anything that clearly says the zlib license is compatible with apache license.
It could be fine, but I dont really know enough to say.

conorbros:
https://www.apache.org/legal/resolved.html#category-a

>Category A: Licenses in Category A may be included in Apache Software Foundation products. They are said to be "Apache-like".

zlib is included in that list

rukai (approved):
oooh nice, thanks for finding definitive proof, approved

cjrolo (approved):

Issue #1168, opened 2023-05-12
remove source_to_chain_mapping in topology.yaml format


rukai:
Currently the format looks like this:
```yaml
---
sources:
  kafka_source:
    Kafka:
      listen_addr: "127.0.0.1:9192"
chain_config:
  main_chain:
    - KafkaSinkSingle:
        remote_address: "127.0.0.1:9092"
        connect_timeout_ms: 3000
source_to_chain_mapping:
  kafka_source: main_chain
```
source_to_chain_mapping has no reason to exist in its current form.

Being parsed into a HashMap means that its a 1->n relationship between chains and sources.
So writing it as a mapping is misleading as it implies an n<->n relationship and makes topologies harder to follow.

## Minimal fix

This could be more effectively represented as an extra field on the chain:
```yaml
---
sources:
  kafka_source:
    Kafka:
      listen_addr: "127.0.0.1:9192"
chain_config:
  main_chain:
    - sources:
       - kafka_source
    - KafkaSinkSingle:
        remote_address: "127.0.0.1:9092"
        connect_timeout_ms: 3000
```
I think at a minimum we should take this approach.

## An extra step

But I propose we go one step further and do away with the 1->n mapping.
The sources can live directly in the chain, pretending to be the first transform.
```yaml
---
main_chain:
    - KafkaSource:
        listen_addr: "127.0.0.1:9192"
    - KafkaSinkSingle:
        remote_address: "127.0.0.1:9092"
        connect_timeout_ms: 3000
```

I am yet to see a shotover production deployment with multiple chains so it seems safe to assume that it is most common for there to be a single chain.
And this format makes the single chain case far easier to understand.
Treating the Source as just another transform greatly reduces cognitive load.
Along with this change to format we would also integrate the source documentation into the transform documentation, although under the hood they will still be implemented as completely different things.

If the user needs a few duplicate chains mapped to different sources then I think its easy enough to manage.
If the user needs 100 duplicate chains mapped to different sources then they are surely generating the config programmatically anyway.

PR #1339, opened 2023-09-14
Move source config inline with transforms

rukai:
closes https://github.com/shotover/shotover-proxy/issues/1168

After discussion in the canberra office we decided this was the direction we want to take the shotover topology configuration.

TODO:
* write up design decisions
* cleanup implementation


rukai:
Superceded by https://github.com/shotover/shotover-proxy/pull/1342

Issue #1324, opened 2023-09-05
Improve tracing logs


conorbros:
It would be helpful to know the chain name (and parent chain if a sub-chain) in tracing spans. Otherwise all messages log as part of the same source but they are logged together so it is difficult to figure out which message is from which chain.

rukai:
related: https://github.com/shotover/shotover-proxy/issues/337

PR #1342, opened 2023-09-18
move root chains into sources

rukai:
closes https://github.com/shotover/shotover-proxy/issues/1168
supercedes https://github.com/shotover/shotover-proxy/pull/1339

After discussion in the canberra office we decided this was the direction we want to take the shotover topology configuration.

## The change
Configs like this:
```yaml
---
sources:
  redis_prod:
    Redis:
      listen_addr: "127.0.0.1:6379"
chain_config:
  redis_chain:
    - RedisSinkSingle:
        remote_address: "127.0.0.1:1111"
        connect_timeout_ms: 3000
source_to_chain_mapping:
  redis_prod: redis_chain
```

becomes:

```yaml
---
sources:
  - RedisSource:
      name: "redis_prod"
      listen_addr: "127.0.0.1:6379"
      chain:
        - RedisSinkSingle:
            remote_address: "127.0.0.1:1111"
            connect_timeout_ms: 3000
```

The root level chains are now part of the source configuration.

## alternative approach

In the draft PR https://github.com/shotover/shotover-proxy/pull/1339 I proposed this:

```yaml
---
chain_config:
  redis_chain:
    - RedisSource:
        listen_addr: "127.0.0.1:6379"
    - RedisSinkSingle:
        remote_address: "127.0.0.1:1111"
        connect_timeout_ms: 3000
```

This is actually a lot clearer to read for simple cases.
Its very clear that RedisSource flows into RedisSinkSingle and we have reduced the rightwards drift.
In order to allow this sources are now a subtype of transform called source transforms that must come at the start of the chain.
This is similar to how transforms can be terminating or non-terminating and a transform chain must end in a terminating transform.
So it feels worth the complexity of intermingling transforms/sources to get this clean configuration.

However as soon as we consider subchains the semantics become strange.
```yaml
---
chain_config:
  redis_chain:
    - RedisSource:
        listen_addr: "127.0.0.1:6379"
        connection_limit: 3000000
    - Tee:
        behavior: FailOnMismatch
        buffer_size: 10000
        chain:
          - QueryTypeFilter:
              filter: Read
          - DebugReturner:
              Redis: "42"
    - DebugReturner:
        Redis: "42"
```

In root level chains we must start with a source while in a subchain we must never start with a source.
I considered this complexity to push past the point of being worth it and so I abandoned that approach in favor of this PR.

## Naming

Previously the source and root level chain were each given their own name.
Now the chain name is derived from the source name since its 1-1.
This does leave the naming a little wonky but will be refined in https://github.com/shotover/shotover-proxy/issues/337 and https://github.com/shotover/shotover-proxy/issues/1324

## Other Notes

I've left a TODO in the TLS error handling to keep scope of this PR down, pulling on that string will yield many other issues.


conorbros:
Should we add a version field to `topology.yaml` to have a safeguard against using incompatible versions of Shotover with topology file formats? We don't need this at the moment but I'd imagine it will save us from some headaches in the future.

rukai:
Usually a version field means being backwards compatible across previous versions which would add extra maintenance burden.
But it could be worth it.
I agree, something to consider later on.

conorbros (approved):

benbromhead (approved):

PR #1345, opened 2023-09-21
add rstest_reuse

conorbros:
Small helper library that removes duplications from our rstest setup for the Cassandra integration tests.


shotover-proxy/tests/lib.rs
```diff
@@ -1,3 +1,7 @@
+#[cfg(test)]
```
rukai, line 1: why #[cfg(test)] ? isnt this always a test?

rukai (approved):
seems reasonable

cjrolo (approved):

PR #1346, opened 2023-09-22
Filter AllowList & DenyList

conorbros:
Expands Filter transform functionality to let you specify either an allow list or deny list of messages types.


rukai:
This seems like a nice improvement to the API!
It is both more flexible and makes the transform self documenting as it has AllowList or DenyList written right there, removing ambiguity about whether the filter is an allow or deny kind of filter.

The one last thing we need is an entry in the changelog describing how to upgrade to the new format. No need to describe how to use the new features. Just need to give the equivalent of the old usage in the new usage.

changelog.md
```diff
@@ -3,6 +3,21 @@
 Any breaking changes to the `topology.yaml` or `shotover` rust API should be documented here.
 This assists us in knowing when to make the next release a breaking release and assists users with making upgrades to new breaking releases.
 
+## 0.2.1
```
rukai, line 6: 0.2.0 has not been released yet so this should go under the 0.2.0 heading

rukai: lets also remove 0.2.1 as its a bit confusing to have it when 0.2.0 still doesnt exist yet.

rukai (approved):
nice!

PR #1350, opened 2023-10-03
Tee switch chain

conorbros:
This change allows you to switch the subchain used by Tee with a HTTP API.

In a follow up PR I want to add the ability to swap between returning responses from the main chain and the tee subchain

### DIFF
--- a/shotover-proxy/Cargo.toml
+++ b/shotover-proxy/Cargo.toml
@@ -53,6 +53,7 @@ opensearch = "2.1.0"
 serde_json = "1.0.103"
 time = { version = "0.3.25" }
 inferno = { version = "0.11.15", default-features = false, features = ["multithreaded", "nameattr"] }
+hyper.workspace = true
 shell-quote = "0.3.0"
 
 [features]
--- a/shotover/src/observability/mod.rs
+++ b/shotover/src/observability/mod.rs
@@ -86,7 +86,7 @@ impl LogFilterHttpExporter {
                                                 format!("{:?}", error),
                                             )
                                         }
-                                        Ok(()) => rsp(StatusCode::NO_CONTENT, Body::empty()),
+                                        Ok(()) => rsp(StatusCode::OK, Body::empty()),
                                     },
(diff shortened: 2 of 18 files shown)


shotover-proxy/Cargo.toml
```diff
@@ -53,6 +53,9 @@ opensearch = "2.1.0"
 serde_json = "1.0.103"
 time = { version = "0.3.25" }
 inferno = { version = "0.11.15", default-features = false, features = ["multithreaded", "nameattr"] }
+# workaround segfault in cassandra-cpp, seems to be caused by recent upgrade of openssl-src from 1.1.1 to 3.0.0
+openssl-sys = { version = "=0.9.92" }
```
conorbros, line 57: Oops, merge error I think

shotover/src/observability/mod.rs
```diff
@@ -86,7 +86,7 @@ impl LogFilterHttpExporter {
                                                 format!("{:?}", error),
                                             )
                                         }
-                                        Ok(()) => rsp(StatusCode::NO_CONTENT, Body::empty()),
+                                        Ok(()) => rsp(StatusCode::OK, Body::empty()),
```
rukai, line 89: what does this do?

rukai:
Could you do a write up of how the user will use this new functionality and your reasoning behind the user facing parts?

From a quick look at this I'm thinking it would be simpler for the user to just have it always be Multi and just let them specify a single chain for the Single case.

shotover/src/observability/mod.rs
conorbros: I'll revert this, turns out NO_CONTENT was correct.

conorbros:
After a bit of thought, I decided we should only have the ability to switch between the main and tee subchain, not the ability to switch between Tee subchains. 

The reasons being: I can't think of any good usecases and the potential for breakage with stateful connections.

I originally thought of it as a nice-to-have in addition to the main chain switching.

I'll pull those changes out and just have the main chain switching changes.

PR #1352, opened 2023-10-05
switch to shotover org cql-ws repo

conorbros:


rukai (approved):
I was hoping to get the cql-ws repo in a better state before merging this, but seeing we are blocked on ben who seems to be out today, lets go ahead and land it as its still an improvement over the previous state of things.

cjrolo (approved):

PR #1356, opened 2023-10-15
Update deps

rukai:
We can return scylla back to the crates.io release since the fix we needed is in the new 0.10 release.
fred update removes the dependency on sha1 which will help from scratch build times a little.


conorbros (approved):

cjrolo (approved):

rukai:
bigdecimal 0.4.2 was released last night, so I removed its cargo patch to a git repo.

PR #1359, opened 2023-10-18
Update aws-throwaway + aws-sdk-kms

rukai:
I'd like to be able to make use of the placement group changes in aws-throwaway.
The code changes needed here all stem from: https://github.com/shotover/aws-throwaway/pull/14

### DIFF
--- a/shotover-proxy/benches/windsock/profilers.rs
+++ b/shotover-proxy/benches/windsock/profilers.rs
@@ -1,5 +1,6 @@
 use self::samply::Samply;
 use crate::common::Shotover;
+use anyhow::Result;
 use aws_throwaway::Ec2Instance;
 use perf_flamegraph::Perf;
 use std::{collections::HashMap, path::PathBuf};
@@ -19,7 +20,7 @@ pub struct ProfilerRunner {
     results_path: PathBuf,
     perf: Option<Perf>,
     samply: Option<Samply>,
-    sys_monitor: Option<UnboundedReceiver<String>>,
+    sys_monitor: Option<UnboundedReceiver<Result<String>>>,
 }
 
 impl ProfilerRunner {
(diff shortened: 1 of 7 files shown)


conorbros:
Why not do a new release for aws-throwaway?

rukai:
I have a few reasons but they arent very good ones, I'll make a release.

conorbros (approved):

shotover-proxy/benches/windsock/profilers/sar.rs
```diff
@@ -121,7 +122,7 @@ pub fn parse_sar(rx: &mut UnboundedReceiver<String>) -> ParsedSar {
         };
     };
     let started_at =
-        OffsetDateTime::from_unix_timestamp_nanos(started_at.parse().unwrap()).unwrap();
+        OffsetDateTime::from_unix_timestamp_nanos(started_at.unwrap().parse().unwrap()).unwrap();
```
cjrolo, line 125: unwrap() the unwrap() the unwrap() :)

shotover-proxy/benches/windsock/profilers.rs
cjrolo, line 1: I think `Option<UnboundedReceiver<Result<String>>>,` is getting kind of large, but if it is stable as is, keep it, otherwise probably building a type around it?

cjrolo (approved):
Minor comments in case we get here in the future, but good to go!

shotover-proxy/benches/windsock/profilers/sar.rs
rukai: not sure if you were just amused at all the unwraps but this is test code so unwrap is the most useful tool for the job here since it immediately terminates when something goes wrong and gives us a nice stack trace to investigate.

Sometimes we can give better diagnostics to the developer by handling errors so that we can, for example, report multiple things that went wrong instead of just one.
but I dont see that as being the case here.

shotover-proxy/benches/windsock/profilers.rs
rukai: It is a bit unwieldy yeah.
Maybe we could make an `StdoutLineReceiver` type alias for `UnboundedReceiver<Result<String>>`, and use that everywhere in a follow up PR.

PR #1361, opened 2023-10-19
tee switch main chain

conorbros:
This change allows the user to switch which chain to "read" responses from within Tee. The purpose in mind for this is switching from the source cluster to the target cluster during OpenSearch dual writing. More details can be found in `transforms.md`.

### DIFF
--- a/docs/src/transforms.md
+++ b/docs/src/transforms.md
@@ -508,9 +508,9 @@ The response from the down-chain transform is returned back up-chain but various
 
 Tee also exposes an optional HTTP API to switch which chain to use as the "result source", that is the chain to return responses from.
 
-`GET` `/result-source` will return `"regular-chain"` or `"tee-chain"` indicating what chain is being used for the result source.
+`GET` `/transform/tee/result-source` will return `regular-chain` or `tee-chain` indicating which chain is being used for the result source.
 
-`PUT` `/result-source` with the body content as either `regular-chain` or `tee-chain` to set the result source.
+`PUT` `/transform/tee/result-source` with the body content as either `regular-chain` or `tee-chain` to set the result source.
 
 ```yaml
 - Tee:
@@ -534,8 +534,10 @@ Tee also exposes an optional HTTP API to switch which chain to use as the "resul
     #         filter: Read
     #     - NullSink
 
-    # Optional port to for the HTTP chain switcher API to listen on
-    # switch_port: 1234
+    # The port that the HTTP API will listen on.
+    # When this field is not provided the HTTP API will not be run.
+    # http_api_port: 1234
+    #
     # Timeout for sending to the sub chain in microseconds
     timeout_micros: 1000
     # The number of message batches that the tee can hold onto in its buffer of messages to send.
--- a/docs/src/transforms.md
+++ b/docs/src/transforms.md
@@ -506,11 +506,11 @@ This is mainly used in conjunction with the `TuneableConsistencyScatter` transfo
 This transform sends messages to both the defined sub chain and the remaining down-chain transforms.
 The response from the down-chain transform is returned back up-chain but various behaviours can be defined by the `behaviour` field to handle the case when the responses from the sub chain and down-chain do not match.
 
-Tee also exposes an optional HTTP API to switch which chain is the "sub-chain" and which chain is the "primary chain", that being the chain which responses are returned to the client from.
+Tee also exposes an optional HTTP API to switch which chain to use as the "result source", that is the chain to return responses from.
 
-`GET` `/switched` will return `"true"` or `"false"` indicating whether the chain has been switched from what was ooriginally defined in the topology file.
+`GET` `/result-source` will return `"regular-chain"` or `"tee-chain"` indicating what chain is being used for the result source.
 
-`PUT` `/switch` with the body content as either `true` or `false` will either switch the chain or revert it to the definition in the topology file.
+`PUT` `/result-source` with the body content as either `regular-chain` or `tee-chain` to set the result source.
 
 ```yaml
 - Tee:
--- a/docs/src/transforms.md
+++ b/docs/src/transforms.md
@@ -506,6 +506,12 @@ This is mainly used in conjunction with the `TuneableConsistencyScatter` transfo
 This transform sends messages to both the defined sub chain and the remaining down-chain transforms.
 The response from the down-chain transform is returned back up-chain but various behaviours can be defined by the `behaviour` field to handle the case when the responses from the sub chain and down-chain do not match.
 
+Tee also exposes an optional HTTP API to switch which chain is the "sub-chain" and which chain is the "primary chain", that being the chain which responses are returned to the client from.
+
+`GET` `/switched` will return `"true"` or `"false"` indicating whether the chain has been switched from what was ooriginally defined in the topology file.
+
+`PUT` `/switch` with the body content as either `true` or `false` will either switch the chain or revert it to the definition in the topology file.
+
 ```yaml
 - Tee:
     # Ignore responses returned by the sub chain
@@ -528,6 +534,8 @@ The response from the down-chain transform is returned back up-chain but various
     #         filter: Read
     #     - NullSink
 
+    # Optional port to for the HTTP chain switcher API to listen on
+    # switch_port: 1234
     # Timeout for sending to the sub chain in microseconds
     timeout_micros: 1000
     # The number of message batches that the tee can hold onto in its buffer of messages to send.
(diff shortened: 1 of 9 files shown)


shotover/src/transforms/tee.rs
```diff
@@ -153,26 +175,73 @@ impl TransformConfig for TeeConfig {
             buffer_size,
             behavior,
             self.timeout_micros,
+            self.switch_port,
+            Some(Arc::new(Mutex::new(false))),
```
rukai, line 179: This is an Option but it is always Some.

shotover/src/transforms/tee.rs
```diff
@@ -97,6 +117,7 @@ pub struct Tee {
     pub behavior: ConsistencyBehavior,
     pub timeout_micros: Option<u64>,
     dropped_messages: Counter,
+    switched_chain: Option<Arc<Mutex<bool>>>,
```
rukai, line 120: This should be an `Arc<AtomicBool>` otherwise we'll have lock contention when there are multiple active connections at the same time.

shotover-proxy/Cargo.toml
```diff
@@ -54,6 +54,7 @@ serde_json = "1.0.103"
 time = { version = "0.3.25" }
 inferno = { version = "0.11.15", default-features = false, features = ["multithreaded", "nameattr"] }
 shell-quote = "0.3.0"
+hyper.workspace = true
```
rukai, line 57: Can we also change `shotover/Cargo.toml` to use `hyper.workspace = true` ?

shotover/src/transforms/tee.rs
rukai: Its not clear from the name + type what `switched_chain` does.

Maybe we should use an atomic enum like this?
```rust
Tee {
    result_source: ResultSource
    ...
}

enum ResultSource {
     RegularChain,
     TeeChain,
}
```

shotover/src/transforms/tee.rs
```diff
@@ -38,12 +49,21 @@ impl TeeBuilder {
...
+            switch_port,
         }
     }
 }
 
 impl TransformBuilder for TeeBuilder {
     fn build(&self) -> Transforms {
+        let result_source = Arc::new(AtomicResultSource::new(ResultSource::RegularChain));
+
+        if let Some(switch_port) = self.switch_port {
```
rukai, line 61: I think the task and arc need to be shared between all tee instances from the same builder, like you had before.
Otherwise it wont be able to handle 2 incoming connections at once!

shotover/src/transforms/tee.rs
```diff
@@ -242,12 +322,115 @@ impl Transform for Tee {
...
+                        };
+                        Ok::<_, Infallible>(response)
+                    }
+                }))
+            }
+        });
+
+        let address = self.address;
+        Server::try_bind(&address)
+            .with_context(|| format!("Failed to bind to {}", address))?
```
rukai, line 427: format! -> anyhow!

shotover/src/transforms/tee.rs
```diff
@@ -242,12 +322,115 @@ impl Transform for Tee {
...
+                                            body,
+                                            result_source.clone(),
+                                        )
+                                        .await
+                                        {
+                                            Err(error) => {
+                                                error!(?error, "setting result source failed");
+                                                Self::rsp(
+                                                    StatusCode::BAD_REQUEST,
+                                                    "setting result source failed",
```
rukai, line 402: can we include the error in the message we send back to the user?

shotover/src/transforms/tee.rs
```diff
@@ -242,12 +322,115 @@ impl Transform for Tee {
...
+
+    async fn async_run_inner(self, result_source: Arc<AtomicResultSource>) -> Result<()> {
+        let make_svc = make_service_fn(move |_| {
+            let result_source = result_source.clone();
+            async move {
+                Ok::<_, Infallible>(service_fn(move |req: Request<Body>| {
+                    let result_source = result_source.clone();
+                    async move {
+                        let response = match (req.method(), req.uri().path()) {
+                            (&Method::GET, "/result-source") => {
```
rukai, line 384: Its quite likely that we will want to combine this webservice with our metrics webservice.
So I wonder if we should namespace this like `/transform/tee/result-source` in an attempt to make the transition to such a setup just a matter of changing ports.
Trying to guess what exactly we will need the API to be is futile but this is at least a bit closer to what we will need.

rukai:
Implementing this first as a separate webservice seems reasonable to me as it will give us a reference point of what we need when we go to implement a combined metrics/config webservice.
Shotover can issue breaking changes on 0.x releases as needed, so while we want to minimize breaking changes I think its worth it in this case.

shotover/src/transforms/tee.rs
rukai: I thought about it more, lets just leave it as is, we'll need some kind of naming system to address which tee we actually want to configure.

docs/src/transforms.md
```diff
@@ -506,6 +506,12 @@ This is mainly used in conjunction with the `TuneableConsistencyScatter` transfo
 This transform sends messages to both the defined sub chain and the remaining down-chain transforms.
 The response from the down-chain transform is returned back up-chain but various behaviours can be defined by the `behaviour` field to handle the case when the responses from the sub chain and down-chain do not match.
 
+Tee also exposes an optional HTTP API to switch which chain to use as the "result source", that is the chain to return responses from.
+
+`GET` `/transform/tee/result-source` will return `"regular-chain"` or `"tee-chain"` indicating what chain is being used for the result source.
```
rukai, line 511: `"regular-chain"` -> `regular-chain`, otherwise people might try to send the string "\\"regular-chain\\""

rukai, line 511: I'm pretty sure it returns `regular-chain` not `"regular-chain"`, right?

rukai: "what chain" -> "which chain"

docs/src/transforms.md
```diff
@@ -528,6 +534,8 @@ The response from the down-chain transform is returned back up-chain but various
     #         filter: Read
     #     - NullSink
 
+    # Optional port to for the HTTP chain switcher API to listen on
```
rukai, line 537:
```
# The port that the HTTP API will listen on.
# When this field is not provided the HTTP API will not be run.
```

docs/src/transforms.md
```diff
@@ -528,6 +534,8 @@ The response from the down-chain transform is returned back up-chain but various
...
+    # switch_port: 1234
```
rukai, line 538: Maybe give this a generic name like "http_api_port" to allow it to be extended with other configurations in the future?

shotover/src/transforms/tee.rs
```diff
@@ -153,15 +191,38 @@ impl TransformConfig for TeeConfig {
...
+        self.transform_inner(requests_wrapper).await
+    }
+}
+
+impl Tee {
+    async fn return_response(
+        &mut self,
+        tee_result: Messages,
+        chain_result: Messages,
+    ) -> Result<Messages> {
```
rukai, line 211: We Return a Result but only ever return Ok

shotover/src/transforms/tee.rs
```diff
@@ -153,15 +191,38 @@ impl TransformConfig for TeeConfig {
...
         )))
     }
 }
 
 #[async_trait]
 impl Transform for Tee {
     async fn transform<'a>(&'a mut self, requests_wrapper: Wrapper<'a>) -> Result<Messages> {
-        match &mut self.behavior {
-            ConsistencyBehavior::Ignore => {
+        self.transform_inner(requests_wrapper).await
```
rukai, line 202: Is there a reason to pull this function out?

shotover/src/transforms/tee.rs
```diff
@@ -153,15 +191,38 @@ impl TransformConfig for TeeConfig {
...
+        chain_result: Messages,
+    ) -> Result<Messages> {
+        let result_source: ResultSource = self.result_source.load(Ordering::Relaxed);
+        match result_source {
+            ResultSource::RegularChain => Ok(chain_result),
+            ResultSource::TeeChain => Ok(tee_result),
+        }
+    }
+
+    async fn ignore_behaviour_inner<'a>(
```
rukai, line 219: ignore_behaviour_inner -> ignore_behaviour

shotover-proxy/config/config.yaml
```diff
@@ -1,3 +1,3 @@
 ---
-main_log_level: "info,shotover_proxy=info"
+main_log_level: "debug,shotover_proxy=debug"
```
rukai, line 2: I think this was left in by accident?

rukai (approved):
LGTM!
Lets land this

cjrolo (approved):
Really nice one!

PR #1362, opened 2023-10-30
fix opensearch codec bug

conorbros:
This path was never being run and I didn't notice the error until a large enough message went through. 

The `match std::mem::replace(&mut self.state, State::ParsingResponse)` part is a bit misleading, it replaces `self.state` with `State::ParsingResponse` and returns what was there before, we want to set it back to `State::ReadingBody` if there is still more body to read. i.e. if `src.len() < content_length`


rukai (approved):
ah yep, that mem::replace makes correctness hard, the fix seems reasonable to me.

cjrolo (approved):

PR #1369, opened 2023-11-16
windsock: fix rare cloud shotover shutdown race condition

rukai:
I encountered a rare and subtle race condition that caused a windsock bench to fail.
We have a sanity check that ensures that all `Ec2InstanceWithShotover` arc clones are dropped before issuing a new one: https://github.com/shotover/shotover-proxy/blob/d9a609124851c435efbb134b79055960c1b4a785/shotover-proxy/benches/windsock/aws/mod.rs#L120

This was being accidentally triggered because we did not wait for a background task that had access to such a clone to terminate before ending the bench and proceeding onto the next bench when will request a new `Ec2InstanceWithShotover`
The fix is to tweak our communication with that task to ensure that we wait for the task to end (achieved by `while let Some(event) = self.event_rx.recv().await` which will keep running until the task has terminated.

### DIFF
--- a/shotover-proxy/benches/windsock/aws/mod.rs
+++ b/shotover-proxy/benches/windsock/aws/mod.rs
@@ -279,7 +279,8 @@ impl Ec2InstanceWithShotover {
             .push_file_from_bytes(topology.as_bytes(), Path::new("topology.yaml"))
             .await;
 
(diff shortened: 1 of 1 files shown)


conorbros (approved):

cjrolo (approved):
Race conditions are always nice! (Or maybe not!)

PR #1371, opened 2023-11-20
KafkaSinkSingle: remove dest host config, rely solely on dest port

rukai:
As per the comment I added:
KafkaSinkSingle is designed solely for the use case of running a shotover instance on the same machine as each kafka instance.

I cannot think of a valid configuration where the destination host is set to a value other than the private ip of the machine.
It could technically be set to `localhost` but AFAIK that doesnt offer any advantages over just using the private ip.
Is there a case i'm missing?

So given this, allowing the user to set the full address is just a footgun so we should remove this config instead only allowing the user to set the port.

### DIFF
--- a/shotover/src/transforms/kafka/sink_single.rs
+++ b/shotover/src/transforms/kafka/sink_single.rs
@@ -16,9 +16,10 @@ use tokio::time::timeout;
 
 #[derive(Serialize, Deserialize, Debug)]
 #[serde(deny_unknown_fields)]
+/// KafkaSinkSingle is designed solely for the use case of running a shotover instance on the same machine as each kafka instance.
+/// The kafka instance and shotover instance must run on seperate ports.
 pub struct KafkaSinkSingleConfig {
-    #[serde(rename = "remote_address")]
-    pub address: String,
+    pub destination_port: u16,
     pub connect_timeout_ms: u64,
     pub read_timeout: Option<u64>,
 }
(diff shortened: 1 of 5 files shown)


conorbros (approved):

shotover/src/transforms/kafka/sink_single.rs
```diff
@@ -106,7 +98,11 @@ impl Transform for KafkaSinkSingle {
     async fn transform<'a>(&'a mut self, mut requests_wrapper: Wrapper<'a>) -> Result<Messages> {
         if self.outbound.is_none() {
             let codec = KafkaCodecBuilder::new(Direction::Sink);
-            let tcp_stream = tcp::tcp_stream(self.connect_timeout, &self.address).await?;
+            let tcp_stream = tcp::tcp_stream(
+                self.connect_timeout,
+                (requests_wrapper.local_addr.ip(), self.address_port),
```
cjrolo, line 103: How does this works in case a machine has several IPs and/or interfaces?

cjrolo (approved):
Left a comment regarding machines that might have several NICs and/or IPs

shotover/src/transforms/kafka/sink_single.rs
rukai: In the case where shotover and kafka have seperate NICs/IPs it would not have worked in the first place since we only rewrite the IP.
If you need that kind of setup then I believe you will need to use KafkaSinkCluster instead of KafkaSinkSingle.

Issue #1370, opened 2023-11-20
KafkaSinkSingle: make transform validation fail when kafka port == shotover port


rukai:
It should fail with a message like:
```
KafkaSinkSingle:
  the destination_port is the same as the port used by `KafkaSource`. These ports must be unique as shotover and the destination kafka must be behind the same IP and therefore cannot share a port.
```

rukai:
Since I originally wrote KafkaSinkSingle I've learnt that kafka has a far stronger distinction between advertised address and real address compared to cassandra.
So I think maybe the whole approach of including rewriting in the transform might have been a mistake.
Maybe we should just have kafka advertise shotover's address instead.

PR #1372, opened 2023-11-21
Fix kafka benches to actually bench shotover

rukai:
Before this PR our shotover benches were actually benching directly with kafka completely bypassing shotover!

The KafkaSinkSingle transform is designed to run only on the same machine as the kafka broker.
However the bench currently sets up kafka and shotover on different instances.

One would expect this to result in an error somewhere and the bench to fail.
But what actually happens is that shotover fails to rewrite the metadata query containing the kafka nodes in the cluster and as a result once the bencher receives the metadata query back from shotover it proceeds to send all the benchmark queries directly to kafka.

In more detail:
Shotover will overwrite the port of any brokers returned by the metadata requests.
When shotover and kafka are on different instances they both have ports of 9092.
So we are just overwriting 9092 with 9092, an unintentional noop, and so the client just connects directly to kafka without any error.

This silent failure is really bad and I've raised the following to prevent similar issues from occurring again:
* https://github.com/shotover/shotover-proxy/pull/1371
* https://github.com/shotover/shotover-proxy/issues/1370

This PR fixes the benchmark by moving shotover and kafka onto the same machine for the kafka benchmarks.

kafka windsock results with this PR:
image

### DIFF
--- a/shotover-proxy/benches/windsock/aws/mod.rs
+++ b/shotover-proxy/benches/windsock/aws/mod.rs
@@ -324,11 +324,8 @@ impl RunningShotover {
                         match line {
                             Some(Ok(line)) => {
                                 let event = Event::from_json_str(&line).unwrap();
-                                if let Level::Warn = event.level {
-                                    tracing::error!("shotover warn:\n    {event}");
-                                }
-                                if let Level::Error = event.level {
-                                    tracing::error!("shotover error:\n    {event}");
+                                if let Level::Warn | Level::Error = event.level {
+                                    println!("AWS shotover: {event}");
                                 }
                                 if event_tx.send(event).is_err() {
                                     return
@@ -344,6 +341,7 @@ impl RunningShotover {
                 }
             }
         });
+
         // wait for shotover to startup
         loop {
             let event = event_rx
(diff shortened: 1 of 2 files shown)


conorbros (approved):

shotover-proxy/benches/windsock/aws/mod.rs
```diff
@@ -278,28 +298,34 @@ impl Ec2InstanceWithShotover {
...
+}
 
+impl RunningShotover {
+    async fn new(instance: &Ec2Instance) -> Self {
         let (shutdown_tx, mut shutdown_rx) = tokio::sync::mpsc::unbounded_channel();
         let (event_tx, mut event_rx) = tokio::sync::mpsc::unbounded_channel();
+        let mut receiver = instance
+                    .ssh()
+                    .shell_stdout_lines(r#"
+        killall -w shotover-bin > /dev/null || true
```
cjrolo, line 317: I normally use a -9 to make sure things die fast :D

cjrolo (approved):

shotover-proxy/benches/windsock/aws/mod.rs
rukai: IMO its nice to be able to exercise our clean shutdown process here.

PR #1374, opened 2023-11-23
remove todo

conorbros:
This todo has become irrelevant since the introduction of the TransformBuilder, a new client is not created with each new connection.


rukai (approved):
nice find!

PR #1376, opened 2023-11-27
Protect tests

conorbros:
Added a few more tests and cleaned up the file

### DIFF
--- a/shotover-proxy/tests/cassandra_int_tests/protect.rs
+++ b/shotover-proxy/tests/cassandra_int_tests/protect.rs
@@ -84,6 +84,49 @@ pub async fn test(shotover_session: &CassandraConnection, direct_session: &Cassa
     )
     .await;
 
+    assert_query_result(
+        shotover_session,
+        "SELECT pk, cluster, col1, col2, col3 FROM test_protect_keyspace.test_table",
+        &[
+            &[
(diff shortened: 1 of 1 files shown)


shotover-proxy/tests/cassandra_int_tests/protect.rs
```diff
@@ -13,77 +13,154 @@ pub struct Protected {
...
-
-    let insert_statement = "BEGIN BATCH
-    INSERT INTO test_protect_keyspace.test_table (pk, cluster, col1, col2, col3) VALUES ('pk4', 'cluster', 'encrypted4', 3, true);
-    INSERT INTO test_protect_keyspace.test_table (pk, cluster, col1, col2, col3) VALUES ('pk5', 'cluster', 'encrypted5', 4, false);
-    APPLY BATCH;";
-    run_query(shotover_session, insert_statement).await;
-
+pub async fn select_all(shotover_session: &CassandraConnection) {
     // assert that data is decrypted by shotover
     assert_query_result(
```
rukai, line 18: This large assert seems to be identical to the next one?
Should we run in a loop or move into a function to avoid duplicating it?

rukai (approved):
nice!

cjrolo (approved):

PR #1387, opened 2023-12-06
Update rustls

rukai:
Mostly various reshuffling of types, logic is largely the same.

* dangerous_configuration feature is no longer needed, instead rustls has chosen to convey the same idea via including `danger` in the module name and a `dangerous` method in the `ClientConfigBuilder`
* `RSAKey` was renamed to `Pkcs1Key` here: https://github.com/rustls/pemfile/commit/6f0724bfd5db1a97341218bb22ac32fbfece4a82

### DIFF
--- a/shotover/src/tls.rs
+++ b/shotover/src/tls.rs
@@ -2,10 +2,14 @@
 
 use crate::tcp;
 use anyhow::{anyhow, bail, Context, Error, Result};
-use rustls::client::{InvalidDnsNameError, ServerCertVerified, ServerCertVerifier, WebPkiVerifier};
-use rustls::server::{AllowAnyAuthenticatedClient, NoClientAuth};
-use rustls::{Certificate, CertificateError, PrivateKey, RootCertStore, ServerName};
-use rustls_pemfile::{certs, Item};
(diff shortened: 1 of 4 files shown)


conorbros (approved):

shotover/src/tls.rs
```diff
@@ -174,12 +177,13 @@ impl TlsConnector {
...
             .transpose()?;
 
-        let config_builder = rustls::ClientConfig::builder().with_safe_defaults();
+        let config_builder = ClientConfig::builder();
         let config = match (private_key, certs, tls_config.verify_hostname) {
             (Some(private_key), Some(certs), true) => config_builder
                 .with_root_certificates(root_cert_store)
                 .with_client_auth_cert(certs, private_key)?,
             (Some(private_key), Some(certs), false) => config_builder
+                .dangerous()
```
cjrolo, line 186: I too, like to live dangerously!

cjrolo (approved):

shotover/src/tls.rs
rukai: Tbh I'd rather not 😆 
It's a direct port of what we were doing with openssl which gives no such warnings.
Was convinced to keep the functionality by a coworker.

Issue #1394, opened 2023-12-18
leak chain names


rukai:
Since source + transform names will never change for the execution of shotover, for the purposes of metric reporting we should create the value once and then leak it into a `&'static str` via 
https://doc.rust-lang.org/nightly/alloc/string/struct.String.html#method.leak
This should improve performance by avoiding reallocating for reporting metrics.

PR #1395, opened 2023-12-18
source_to_sink_latency_seconds/sink_to_source_latency_seconds metrics

rukai:
closes https://github.com/shotover/shotover-proxy/issues/1394
    
 This PR introduces two new metrics:
 * source_to_sink_latency_seconds
 * sink_to_source_latency_seconds
 
 These metrics differ from the existing metrics in two important ways:
 * The new metrics time measurement starts and ends when we receive the message off the TCP connection and when we send it off again
     + existing metrics start measuring after we have parsed the message and before we decode it. - this means those parts of shotover are never measured by our metrics.
* The new metrics exclude the time for the database to send back a request. 
    + Existing metrics include that which can be useful to understand the system as a whole but a metric without that is very useful.
 
Abbreviated example output:
image

Some of the histograms are giving really wonky results, percentiles 0.1 through to 0.999 are all the same value.
I tracked this down to the fact that:
* the first message of each connection takes a long time (6-8s)
* regular messages take < 1ms
* the histograms employ some [clever maths to compress the values](https://github.com/mheffner/rust-sketches-ddsketch) for which precision worsens the further apart the values are.

All of these together results in what we are observing.

In production this shouldnt be an issue as the histograms will fix themselves after 1 min once that large value has been cleared.


shotover-proxy/tests/runner/observability_int_tests.rs
```diff
@@ -35,20 +37,32 @@ shotover_chain_messages_per_batch_count{chain="redis source",quantile="0.999"}
 shotover_chain_messages_per_batch_count{chain="redis source",quantile="1"}
 shotover_chain_total_count{chain="redis source"}
 shotover_query_count{name="redis-chain"}
+shotover_sink_to_source_latency_seconds_count{source="redis"}
+shotover_sink_to_source_latency_seconds_sum{source="redis"}
+shotover_sink_to_source_latency_seconds{source="redis",quantile="0"}
+shotover_sink_to_source_latency_seconds{source="redis",quantile="0.1"}
```
cjrolo, line 43: Why 0.1 quantile? Trying to figure out why I would use this vs average (0.5) or the higher ones. Could be useful to try to figure out if I'm on a long tail distribution, but not sure how.

shotover/src/codec/opensearch.rs
```diff
@@ -215,9 +233,10 @@ impl Decoder for OpenSearchDecoder {
                     }
 
                     let body = src.split_to(content_length).freeze();
-                    return Ok(Some(vec![Message::from_frame(Frame::OpenSearch(
-                        OpenSearchFrame::new(http_headers, body),
-                    ))]));
+                    return Ok(Some(vec![Message::from_frame_at_instant(
+                        Frame::OpenSearch(OpenSearchFrame::new(http_headers, body)),
+                        Some(received_at),
+                    )]));
```
cjrolo, line 239: So many brackets! LISP is that you! :)

cjrolo (approved):

shotover-proxy/tests/runner/observability_int_tests.rs
rukai: I added it to help debug strange quantile behavior.
I described it under:
> Some of the histograms are giving really wonky results, percentiles 0.1 through to 0.999 are all the same value.

in the main PR description.

Maybe I should remove it again, but I figured if it was useful now it'll likely be useful later.

conorbros (approved):

PR #1399, opened 2023-12-21
Fix client timeout

rukai:
This code has always been weird.
There have been previous PRs to clean it up:
* https://github.com/shotover/shotover-proxy/pull/140
* https://github.com/shotover/shotover-proxy/pull/644

But they both avoided touching the actual logic to keep the PR scope down.

This PR completely replaces the timeout logic:
* fixes the timeout to actually timeout after specified seconds
    + Currently the logic will timeout after around `log2(seconds)`, I dont think there is any good reason for this old behavior.
* Should improve performance a little:
    + This function showed up in the profiler, which is why I started working on this in the first place.
    + Avoiding using the timeout when we dont need it should help a bit
    + Avoiding breaking out of the await every second should also help a bit.
* I think the logic is easier to follow now
* We convert the config u64 to a Duration earlier, stronger typing is good

### DIFF
--- a/shotover/src/server.rs
+++ b/shotover/src/server.rs
@@ -1,6 +1,6 @@
 use crate::codec::{CodecBuilder, CodecReadError, CodecWriteError};
 use crate::config::chain::TransformChainConfig;
-use crate::message::Messages;
+use crate::message::{Message, Messages};
 use crate::sources::Transport;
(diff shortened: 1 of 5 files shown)


conorbros (approved):

cjrolo (approved):
I love the clarity the code gained with this PR.

PR #1405, opened 2023-12-21
leak chain name for performance

rukai:
closes https://github.com/shotover/shotover-proxy/issues/1394

Putting an Owned string for the chain name into Wrapper is quite expensive.
It means that for each message batch shotover processes it will allocate and deallocate a string just to store it in the wrapper.
Additionally we then clone it in many cases.
Since chain names are set by the config once and then never changed we can simply leak the string and then never have to clone it again as we can hold a &'static reference to it.

This PR does the following:
* Refactors chain_name such that it is set by the chain itself, this has no performance impact just better isolates the chain name logic from the rest of the code base.
* chain_name is always stored as a `&'static str`, it is converted from a `String` via `leak` at the point where the chain is parsed from the yaml.

This PR gives a 25% win in loopback benchmark:
image
In reality it wont be this high as https://github.com/shotover/shotover-proxy/pull/1401 already removes some of those clones. But I expect it to still keep the improvement from non-metric clones.


shotover/src/config/chain.rs
```diff
@@ -19,7 +19,7 @@ impl TransformChainConfig {
         for tc in &self.0 {
             transforms.push(tc.get_builder(name.clone()).await?)
         }
-        Ok(TransformChainBuilder::new(transforms, name))
+        Ok(TransformChainBuilder::new(transforms, name.leak()))
```
cjrolo, line 22: And something new that I learned!

cjrolo (approved):

rukai:
I've rebenched this PR, it gives us a 15% win on top of https://github.com/shotover/shotover-proxy/pull/1401
image

Very nice!

conorbros (approved):

PR #1406, opened 2023-12-22
cassandra optimize murmur hashing

rukai:
This PR improves performance of get_replica_node_in_dc by having `calculate_routing_key` directly convert the query values into a murmur hash instead of returning an intermediate value allocated into a `Vec<u8>`.

I very heavily referenced the scylla driver here as it is much more performant than the cdrs-tokio driver that the previous implementation was based off.

Unfortunately this PR brings a bunch of low level murmur hashing code into shotover, it would be nice if scylla driver exposed it but it feels like too much of a low level implementation detail to ask them to expose it. I think its still worth the performance win though.

I have verified that the changes to calculate_routing_key have reduced the total time spent in that function + dropping the return value.


conorbros (approved):

shotover/src/transforms/cassandra/sink_cluster/murmur.rs
```diff
@@ -0,0 +1,175 @@
...
+use std::num::Wrapping;
+
+pub struct Murmur3PartitionerHasher {
+    total_len: usize,
+    buf: [u8; Self::BUF_CAPACITY],
+    h1: Wrapping<i64>,
+    h2: Wrapping<i64>,
+}
+
+impl Murmur3PartitionerHasher {
```
conorbros, line 15: This is copied directly from Scylla with no logic changes?

rukai: Yeah

conorbros (approved):

cjrolo (approved):

cjrolo (approved):

PR #1408, opened 2024-01-08
Update deps

rukai:
The anyhow update is responsible for the `RUST_LIB_BACKTRACE` changes, see https://github.com/dtolnay/anyhow/issues/342 for more details.

We include anyhow errors in error, warn and info level traces.
We also combine multiple errors into a single trace in some cases.
So including stacktraces there would be way too noisy, and it is reasonable for a user to keep RUST_BACKTRACE=1 enabled for better panic messages, so we dont want to degrade our handling of error/warn/info traces in that case.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -54,9 +54,9 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.8.6"
+version = "0.8.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91429305e9f0a25f6205c5b8e0d2db09e0708a7a6df0f42212bb56c32c8ac97a"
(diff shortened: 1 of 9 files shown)


rukai:
image
I can reproduce the bench regression locally.

Its probably because we arent setting RUST_LIB_BACKTRACE=0 in the benchmarks.

rukai:
I resolved the issue by setting `RUST_LIB_BACKTRACE=0` in our micro benchmarks.
Additionally I set `RUST_LIB_BACKTRACE=1` to make panic performance reproducible, `1` was chosen as that is what CI is configured to use. (And what I recommend developers to set their machine to)

conorbros (approved):

cjrolo (approved):

PR #1411, opened 2024-01-08
Remove chain name from wrapper

rukai:
There is no need to pass in the chain name in this way.
We already give access to the chain name via passing it to the transform builders: https://github.com/shotover/shotover-proxy/blob/df6e283ebf9accb2756252872a6634b77e1de789/shotover/src/transforms/mod.rs#L247 So passing it to the transforms via `CONTEXT_CHAIN_NAME` is just redundant.

Removing the `CONTEXT_CHAIN_NAME.scope(..)` gives us a 7% win to the loopback bench on top of the changes in https://github.com/shotover/shotover-proxy/pull/1405
image

### DIFF
--- a/shotover/src/transforms/mod.rs
+++ b/shotover/src/transforms/mod.rs
@@ -257,7 +257,6 @@ pub struct Wrapper<'a> {
     pub client_details: String,
     /// Contains the shotover source's ip address and port which the message was received on
     pub local_addr: SocketAddr,
-    chain_name: &'static str,
     /// When true transforms must flush any buffered messages into the messages field.
     /// This can occur at any time but will always occur before the transform is destroyed due to either
     /// shotover or the transform's chain shutting down.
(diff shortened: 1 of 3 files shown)


shotover/src/transforms/mod.rs
```diff
@@ -383,50 +375,46 @@ impl<'a> Wrapper<'a> {
...
             local_addr: "127.0.0.1:8000".parse().unwrap(),
-            chain_name: "".to_string(),
             flush: false,
         }
     }
 
     pub fn new_with_chain_name(
         requests: Messages,
-        chain_name: String,
+        _chain_name: String,
```
rukai, line 384: these will be removed by https://github.com/shotover/shotover-proxy/pull/1405

cjrolo (approved):

conorbros (approved):

PR #1424, opened 2024-01-15
split codec benches modules

conorbros:


shotover/benches/benches/codec/kafka.rs
```diff
@@ -0,0 +1,147 @@
...
+    ),
+    (include_bytes!("kafka_requests/fetch.bin"), "request_fetch"),
+    (
+        include_bytes!("kafka_requests/produce.bin"),
+        "request_produce",
+    ),
+];
+
+fn criterion_benchmark(c: &mut Criterion) {
+    let mut group = c.benchmark_group("kafka_codec");
```
rukai, line 25: Since the group is now `kafka_codec` the bench names should have the `kafka_` prefix removed.
Otherwise the bench names have `kafka` twice which is redundant
image


Or alternatively we could just keep as `codec`.

Same thing for cassandra_codec.

rukai (approved):
nice!
Thanks for this

cjrolo (approved):

PR #1430, opened 2024-01-23
refactor Handler::process_messages

rukai:
Makes the logic easier to follow by seperating the two seperate forward/backward paths instead of having them intertwined with a boolean.


conorbros (approved):

cjrolo (approved):
Good one!

PR #1432, opened 2024-01-23
Add v5 benches

conorbros:


shotover/benches/benches/codec/cassandra.rs
```diff
@@ -30,10 +30,93 @@ fn criterion_benchmark(c: &mut Criterion) {
             },
         }))];
 
-        let (_, mut encoder) =
+        let (mut decoder, mut encoder) =
```
rukai, line 33: I think a fresh decoder should be created for each bench.
Otherwise whether or not the encode bench is run could affect the results of the decode bench.
From a quick look at the CassandraCodecBuilder implementation I dont think any of the state that is currently shared between encoder and decoder could cause issues here, but we could very easily introduce more shared state in the future that results in the benches influencing each others results.

shotover/benches/benches/codec/cassandra.rs
```diff
@@ -69,6 +166,183 @@ fn criterion_benchmark(c: &mut Criterion) {
...
+                    let mut bytes = BytesMut::new();
+                    encoder.encode(messages, &mut bytes).unwrap();
+                    black_box(bytes)
+                },
+                BatchSize::SmallInput,
+            )
+        });
+
+        group.bench_function(
+            "decode_cassandra_system.local_result_v5_lz4_compression",
```
rukai, line 333: cassandra was left in the name here

shotover/src/codec/cassandra.rs
```diff
@@ -773,6 +774,16 @@ impl Encoder<Messages> for CassandraEncoder {
 }
 
 impl CassandraEncoder {
+    pub fn set_startup_state_ext(&mut self, compression: String, version: Version) {
```
rukai, line 777: what does `ext` in the function name here mean?

conorbros: external

shotover/benches/benches/codec/cassandra.rs
```diff
@@ -30,10 +30,93 @@ fn criterion_benchmark(c: &mut Criterion) {
...
-        group.bench_function("encode_system.local_query", |b| {
+        encoder.set_startup_state_ext("NONE".to_string(), Version::V4);
+
+        group.bench_function("encode_system.local_query_v4_no_compression", |b| {
+            b.iter_batched(
+                || messages.clone(),
+                |messages| {
+                    let mut bytes = BytesMut::new();
+                    encoder.encode(messages, &mut bytes).unwrap();
+                    black_box(bytes)
```
rukai, line 44: This line could be just `bytes` as criterion will automatically put the return value through a black_box.

shotover/src/codec/cassandra.rs
rukai: I was thinking extension. Maybe we should write it out fully as `set_startup_state_ext`

rukai (approved):
looks good!

cjrolo (approved):

PR #1434, opened 2024-01-25
Remove inferno dependency

rukai:
Im currently searching for ways to fix our iterative build times for release which are current 3 minutes.

I propose that we remove our perf/inferno-based profiler as samply has superseded it for me.
Removing it seems to have cut 5 seconds off our iterative build times although I wasnt very rigorous with my measurements.


conorbros (approved):

cjrolo (approved):
All in for code removal!

PR #1437, opened 2024-01-25
int_tests: avoid including shotover

rukai:
TODO: move the cassandra test into a unit test.

This PR reduces the time to build the integration test binary from 186s to 131s


rukai:
Closing in favor of https://github.com/shotover/shotover-proxy/pull/1441
I think the value of being able to test shotover internals against an actual database is actually quite high and 1441 lets us keep doing that with about 50% of the build time improvement in this PR.

PR #1441, opened 2024-01-29
Add feature flags for each protocol in shotover

rukai:
This improves compile times of windsock without compromising on performance like https://github.com/shotover/shotover-proxy/pull/1438 ended up doing.

Previously `cargo windsock` took 2m 45s to compile after a change to our windsock benches.
Now it takes:
* `cargo windsock-redis` - 1 min 32s
* `cargo windsock-kafka` - 1 min 21s
* `cargo windsock-cassandra` - 2 min 23s
* `cargo windsock` - 2m 45s

We would be getting similar improvements to our integration tests but we rarely need to run them in release so I didnt bother adding cargo aliases for them or measuring them.

Additionally, this PR improves compile times of shotover when developing custom transforms that only need to support a single protocol by allowing developers to disable unneeded dependencies within shotover, this will be quite helpful for our internal kafka project where we are also struggling with build times and have no need for any protocol other than kafka.
As a result of compiling less, the binary size with just kafka enabled is now 13.9MB, while the standard build with everything enabled is 19MB.

Additionally, additionally, this will be quite helpful for investigating the cause of build time issues in the future as we can easily eliminate certain protocols from the build.

### Reduced cfgs

With the help of #1442 #1443 #1444 and #1445 I have reduced the number of cfgs from 283 down to 181

### Improvements for custom transform devs
Testing on our internal kafka-specific project:
* full build times of shotover have been reduced from 8m 53s to 4m 49s
* iterative build times have been reduced from 4m 26s to 1m 57s

As a result with this PR we will cut 4 minutes off the time to run CI as we have no caching setup for our CI on that project.
We will also cut 2 mins 30s from our iterative benchmark workflow which is super important to the project as its very performance sensitive.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -8,6 +8,17 @@ rustflags = [
 linker = "aarch64-linux-gnu-gcc"
 
 [alias]
-windsock = "test --release --bench windsock --features alpha-transforms,rdkafka-driver-tests --"
-windsock-debug = "test --bench windsock --features alpha-transforms,rdkafka-driver-tests --"
-windsock-cloud-docker = "run --package windsock-cloud-docker --"
+# Can run every benchmark
+windsock = "test --release --bench windsock --features kafka,alpha-transforms,rdkafka-driver-tests,cassandra,redis --"
+windsock-debug = "test --bench windsock --features kafka,alpha-transforms,rdkafka-driver-tests,cassandra,redis --"
+
+# Can only run benchmarks specific to the protocol but compiles a lot faster
+windsock-redis = "test --release --bench windsock --no-default-features --features redis,alpha-transforms --"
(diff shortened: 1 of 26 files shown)


cjrolo (approved):
Very good learning to understand how to set feature flags!

rukai:
From a discussion in DMs with @conorbros about this PR:

> conor:
My concern is just a future shotover developer really fighting with the project trying to get it to compile with all these feature flags around.
Feature issues are especially hard to debug, IME
At the moment, I think we should not merge the changes in 1441 and instead figure out how to split out the protocol specific logic, modules and transforms


So I'll make a bunch of prereq PRs to better split out protocol specific logic such that this PR results in far less cfgs.
Additionally I'll measure the build time improvements on my internal kafka project so we have more measurements we can take into consideration.

conorbros (approved):

PR #1448, opened 2024-02-01
Fix ec2-cargo windsock cloud

rukai:
Fixes running `windsock --cloud` within ec2-cargo.
This provides an alternative to `windsock-cloud-docker` with different trade offs.
We have to compile shotover+windsock entirely on EC2 but we get faster communication with EC2 and can be run on a system without docker.

This PR also enables ec2-cargo to use the faster to compile windsock aliases introduced in https://github.com/shotover/shotover-proxy/pull/1441

### DIFF
--- a/ec2-cargo/src/main.rs
+++ b/ec2-cargo/src/main.rs
@@ -73,10 +73,6 @@ sudo /bin/docker1 "$@"
 ' | sudo dd of=/bin/docker
 sudo chmod +x /bin/docker
 
-curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
-unzip awscliv2.zip
-sudo ./aws/install
-
 echo "export RUST_BACKTRACE=1" >> .profile
 echo "export CARGO_TERM_COLOR=always" >> .profile
 echo 'source "$HOME/.cargo/env"' >> .profile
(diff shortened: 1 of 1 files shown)


rukai:
This PR has shown promise: https://github.com/shotover/shotover-proxy/pull/1360

However, I think the next step is to add functionality to windsock to allow reusing EC2 instances.
This will eliminate the noise caused by differences in EC2 instances.
I am thinking an API like this:
```shell
> # Create the resources required to run the benches specified in FILTER and then store the information required to access those instances to disk
> cargo windsock --store-cloud-resources-to-disk FILTER
Creating AWS resources: CloudResourcesRequired {
    shotover_instance_count: 1,
    docker_instance_count: 3,
    include_shotover_in_docker_instance: false,
}
> # Run the benches once, using the instances created in the previous command.
> cargo windsock --use-cloud-resources-from-disk FILTER
Running "kafka,shotover=standard,size=100KB,topology=cluster3"
...
> # Run the benches a second time reusing the same instances
> cargo windsock --use-cloud-resources-from-disk FILTER
Running "kafka,shotover=standard,size=100KB,topology=cluster3"
...
> cargo windsock --cleanup-cloud-resources
```



After that is implemented it should be easier to evaluate #1360

conorbros (approved):

cjrolo (approved):

PR #1457, opened 2024-02-07
windsock-cloud-docker: pass RUST_LOG env var through to windsock

rukai:
Allows us to enable tracing logs for windsock.


conorbros (approved):

cjrolo (approved):
Much more readable!

PR #1460, opened 2024-02-08
remove unused image waiter

conorbros:


shotover-proxy/tests/test-configs/cassandra/cluster-multi-rack-2-per-rack/docker-compose.yaml
```diff
@@ -11,7 +11,7 @@ networks:
 
 services:
   cassandra1_1:
-    image: &image library/cassandra:4.0.6
+    image: &image shotover/cassandra-test:4.0.6-r1
```
rukai, line 14: Converting this to use shotover/cassandra-test is probably worth doing but it will copying the tokens from a `SELECT * FROM system.local; SELECT * FROM system.peers;` of a running `library/cassandra:4.0.6` cluster and then adding them to the CASSANDRA_INITIAL_TOKENS env var.

rukai:
huh, that is less tokens then the other clusters have.
I'm counting 16 * 6 tokens = 96
Whereas the other clusters have 128 * 3 = 384

Maybe cassandra 5 decreased the default token count?
Or maybe its a 6 node cluster thing?

conorbros:
It's not Cassandra 5. It's because there is more nodes to share the tokens.

rukai (approved):
Not sure whats up with the token count but considering this is for an integration test not a benchmark, as long as we have more than one token for each node it should be fine.

rukai:
> It's not Cassandra 5.

Oh true I got it mixed up with the other PR, oops.

cjrolo (approved):

PR #1462, opened 2024-02-08
Update metrics crate

rukai:
The shortcut macros like `counter!` have been removed and the register macros like `register_counter!` have been renamed like `counter!`.
The register macros can be used in place of the shortcut macros since the shortcut macros were just registering and dropping a metric internally anyway.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -116,9 +116,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.5"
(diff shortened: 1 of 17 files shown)


rukai:
I am no longer seeing the performance regression on CI or locally since https://github.com/shotover/shotover-proxy/pull/1463 was merged

conorbros (approved):

cjrolo (approved):

PR #1464, opened 2024-02-08
Update to rust 1.76 🦀🦀🦀

rukai:


rukai:
I cannot reproduce the performance regression in CI or locally, so it must be just noise on the GA VM.

conorbros (approved):

cjrolo (approved):

PR #1465, opened 2024-02-08
Limit max size of main chain batch to 10,000

rukai:
limit buffered incoming messages to 10,000 per connection.
A particular scenario we are concerned about is if it takes longer to send to the server than for the client to send to us, the buffer will grow indefinitely, increasing latency until the buffer triggers an OoM.
To avoid that we have currently hardcoded a limit of 10,000 but if we start hitting that in production we should make this user configurable.


rukai:
Those benchmark improvements are weird and even though they reproduce on CI, I cant reproduce them locally.

I wont stress too much over surprise improvements though!

conorbros (approved):

cjrolo (approved):

PR #1466, opened 2024-02-08
codec send `Message` instead of `Vec<Message>`

rukai:


rukai:
I've reverted the 2nd commit since that was clearly a regression.
For the first commit I will keep this as a draft and evaluate this again after we resolve https://github.com/shotover/shotover-proxy/issues/1471

PR #1472, opened 2024-02-12
windsock workplace deps

conorbros:
Move some of windsock's deps to workspace where possible.


rukai:
ah, I've been keeping these separate to make it easier to split windsock into its own repo.
Sorry :/

Issue #1177, opened 2023-05-18
windsock: implement better abstractions for --flamegraph


rukai:
Currently windsock will instruct benches to take a flamegraph when `--flamegraph` is provided.
However when running multiple benches the flamegraph is overwritten by the last run bench.
So we need to generate flamegraphs with unique names per bench.
Other features we should probably add:

- [x] make room for specifying alternate profiling tools.
- [x] detect benches that dont support flamegraph and skip them.
- [ ] allow displaying some level of arbitrary text in results table from the profiler
- [x] basic CPU usage profiler - shows usage from bencher, shotover and db.

PR #1473, opened 2024-02-12
Direct user where to find samply output

rukai:
closes https://github.com/shotover/shotover-proxy/issues/1177

Adds infrastructure to windsock for benches to make info level notes to the user and then makes use of it to have the  samply profiler output a copy/pasteable command to open the profiler results.
image

I also updated the pinned samply commit since theres been a few bugfixes since we last updated.


conorbros (approved):

windsock/src/report.rs
cjrolo, line 1: Thanks for the comments!

cjrolo (approved):

PR #1476, opened 2024-02-14
Add MessageId (progress towards cleaning up transform invariants)

rukai:
This PR introduces a `MessageId` type and assigns a unique `MessageId` to every message that goes through shotover.
Additionally responses now hold the MessageId of their corresponding request.

The API for this is exposed as methods on the `Message` type:
```rust
impl Message {
    /// Return the shotover assigned MessageId
    pub fn id(&self) -> MessageId {
        self.id
    }

    /// Return the MessageId of the request that resulted in this message
    /// Returns None when:
    /// * The message is a request
    /// * The message is a response but was not created in response to a request. e.g. Cassandra events and redis pubsub
    pub fn request_id(&self) -> Option<MessageId> {
        self.request_id
    }
}
```

This will eventually replace the "requests are in the same order as responses" invariant but for now it is just in addition to it.

I updated the `Transform::transform` docs where we document such invariants and it is rendered below.
The important sections are "Request/Response invariants" and "Deprecated invariants".

image

I have not observed any regressions in benchmark results with this PR and the call to rand doesnt even show up when profiling.

## Benefits

* Improves correctness - Gives us a path forward to resolve some old invariant violations in a performant way https://github.com/shotover/shotover-proxy/issues/499
* Allows future performance improvements:
    + We can return some responses before all of them have been returned.
    + We can start sending a new batch of requests before we have any responses back.
    + Better handle out of order protocols (currently only cassandra)
* Will allow us to simplify handling of pubsub/event messages which are currently special cased via `transform_pushed`
* will allow us to simplify cassandra connection logic, a lot of complexity in there to pretend cassandra is not an out of order protocol
* will allow us to simplify redis pubsub implementation

## Downsides
* The filter transform is now very awkward to implement, we'll have to keep track of the message id of the previous message in order to know where to insert the error message.
    + Maybe in the future we should clean up the filter transform by having relaxed invariants in subchains. Filter is only useful in a subchain anyway and we could make it much more efficient this way by skipping the generation of dummy error messages.
    + Or another possible solution, replace "to be deleted" messages with a dummy frame. That way the codec can just respond with a dummy response and we have a dummy response that we can replace with the error. This would actually be quite efficient as we dont need to insert or remove from the vec of messages.
* for some other transforms a hashmap of message ids is slightly more complicated than a zip with requests
* hashmap look ups are slightly more expensive than a zip with requests - But I believe the other performance wins will largely outweigh this.

## Review
I've included a port of the `protect` and `CassandraSinkCluster` transforms to the message id approach, but I've left other ports out of the PR to keep scope down.
I am fairly confident that this approach will work well for all of our transforms.
But if you have a specific transform you would like me to include in the PR to see how it fares, just let me know.

With "hide whitespace" enabled the protect transform has a very small diff, so for the typical user implemented rewrite transform not much has changed.

### DIFF
--- a/shotover/src/message/mod.rs
+++ b/shotover/src/message/mod.rs
@@ -226,6 +226,17 @@ impl Message {
         self.request_id = Some(request_id);
     }
 
+    pub fn clone_with_new_id(&self) -> Self {
+        Message {
+            inner: self.inner.clone(),
+            meta_timestamp: self.meta_timestamp,
+            received_from_source_or_sink_at: None,
+            codec_state: self.codec_state,
+            id: rand::random(),
+            request_id: self.request_id,
+        }
+    }
+
     pub fn ensure_message_type(&self, expected_message_type: MessageType) -> Result<()> {
         match self.inner.as_ref().unwrap() {
             MessageInner::RawBytes { message_type, .. } => {
--- a/shotover/src/message/mod.rs
+++ b/shotover/src/message/mod.rs
@@ -64,6 +64,9 @@ impl From<&ProtocolType> for CodecState {
 
 pub type Messages = Vec<Message>;
 
+/// Unique identifier for the message assigned by shotover at creation time.
+pub type MessageId = u128;
+
 /// Message holds a single message/query/result going between the client and database.
 /// It is designed to efficiently abstract over the message being in various states of processing.
 ///
@@ -98,6 +101,12 @@ pub struct Message {
     pub(crate) received_from_source_or_sink_at: Option<Instant>,
 
     pub(crate) codec_state: CodecState,
+
+    // TODO: Consider removing the "ignore" down the line, we we need it for now for compatibility with logic using the old style "in order protocol" assumption.
+    #[derivative(PartialEq = "ignore")]
+    pub(crate) id: MessageId,
+    #[derivative(PartialEq = "ignore")]
+    pub(crate) request_id: Option<MessageId>,
 }
 
 // `from_*` methods for `Message`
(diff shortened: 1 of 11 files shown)


shotover/src/message/mod.rs
```diff
@@ -64,6 +64,9 @@ impl From<&ProtocolType> for CodecState {
 
 pub type Messages = Vec<Message>;
 
+/// Unique identifier for the message assigned by shotover at creation time.
+pub type MessageId = u128;
```
cjrolo, line 68: Why 128bits? This means 2 registers, for every single message. Can't this run on 32? 4_294_967_295 Messages before we have to rollover?

cjrolo (approved):

shotover/src/message/mod.rs
rukai: Because it's random, not incrementing. I think you would get collisions on a u32.

I didn't measure any performance issues with u128, so I didn't consider it worth investigating at the moment.
Further down the line I could do some micro benchmarks with a global incrementing u32.

conorbros: Why not incrementing?

conorbros (approved):

shotover/src/message/mod.rs
rukai: Atomic contention.
I'll investigate an incrementing approach in a follow up.

conorbros: It's possible we could just have each message only be unique to it's chain. They'd still be unique across Shotover with the chain name and id number to differentiate.

Issue #1480, opened 2024-02-15
Redesign windsock CLI interface


rukai:
The windsock CLI interface has organically grown over time from 0 to 22 arguments.
Its time to redesign these arguments to form a more cohesive and easier to use API.
Instead of the current CLI, currently accessible via `cargo windsock --help`, I propose the following CLI

```shell
cargo windsock list

# all of these take flags `--bench-length-seconds` `--operations-per-second` `--profilers` and then finally a name or filter.
cargo windsock cloud-setup
cargo windsock cloud-run
cargo windsock cloud-cleanup
cargo windsock cloud-setup-run-cleanup
cargo windsock local-run

cargo windsock baseline-set
cargo windsock baseline-clear
cargo windsock generate-webpage

# all of these take a name or filter
cargo windsock results # additionally includes a `--disable-baseline`
cargo windsock compare
```

Remove `--name` argument, instead either:
* auto-detect that a name is used via the existence of `,` character.
* remove the existence of "names" and just have names output in the form a filter.

PR #1481, opened 2024-02-16
Redesign windsock cli

rukai:
closes https://github.com/shotover/shotover-proxy/issues/1480

This PR cleans up the windsock CLI.
We have accrued a large number of flags at this point and its time we turned them all into proper subcommands to simplify usage of the CLI.

Consider the new `cargo windsock help`:
image

compared to the old:
image

The relevant commands for a task are now a lot clearer, no longer lost among the sea of flags.

To achieve this reduction of the number of flags/subcommand the following was done:
* many flags turned into subcommands
* previously we split flags up as `name`/`by-tags`. But I've now altered names such that they are valid tags, so we no longer need the distinction.
* previously we had a separate flag for comparing against baselines. Now we compare against baselines automatically when they are available, with a `--disable-baseline` flag hidden within the `results` flag for the rare case that its ever needed.

To see the help of a particular subcommand, we can run help on that subcommand, for example here we can see all the flags specific to configuring a run are now within the run subcommands:
image

A particular improvement i'm quite happy with is:
Previously I had to recall do I want to run `cargo windsock --result-by-tags ""` or `cargo windsock --baseline-compare-by-tags ""`.
However now it is correct to just run `cargo windsock results` in both situations.
So much simpler!

The changes to the readme give a good overview of how to use the new CLI.

### DIFF
--- a/.github/workflows/windsock_benches.yaml
+++ b/.github/workflows/windsock_benches.yaml
@@ -37,13 +37,13 @@ jobs:
         echo '1' | sudo tee /proc/sys/kernel/perf_event_paranoid
 
         # run some extra cases that arent handled by nextest
-        cargo windsock --bench-length-seconds 5 --operations-per-second 100 --profilers flamegraph --name cassandra,compression=none,connection_count=1,driver=scylla,operation=read_i64,protocol=v4,shotover=standard,topology=single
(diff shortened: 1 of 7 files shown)


cjrolo (approved):

conorbros (approved):
Really good! Thank you

PR #1483, opened 2024-02-19
Add dummy requests to better support certain transforms

rukai:
This PR ports the `RequestThrottling` and `Filter` transforms to use the new invariants introduced in https://github.com/shotover/shotover-proxy/pull/1476

In order to achieve this I had to extend the semantics of dummy frames and add the implementation of this to our outgoing connection logic.
Dummy frames can now be sent as requests.
When a sink processes a dummy frame a dummy response is generated without touching the database.
This gives us an easy target to replace when rewriting responses in order protocols.
Otherwise we would have to do something strange like record the request just before the request we want to operate on and then insert the generated response directly after the response with the recorded request id.

With this PR the filter transform now conforms to the transform invariants so I removed the `The current filter transform implementation does not obey the current transform invariants. see https://github.com/shotover/shotover-proxy/issues/499` warning.


rukai:
The `redis_filter` benchmark has regressed about 8% on my local machine.
However, this benchmark only has a requests vector of 2 elements.
When running with higher batch sizes of say 50, it would be more expensive to insert and remove elements from vecs, brining the results closer to the new hashmap approach.

Additionally this is a microbenchmark and the integration level benchmarks seem unaffected. Since this change will hopefully lead to improvements to our integration level benchmarks down the line I think we should proceed.

shotover/src/transforms/throttling.rs
```diff
@@ -37,6 +39,7 @@ impl TransformConfig for RequestThrottlingConfig {
 pub struct RequestThrottling {
     limiter: Arc<RateLimiter<NotKeyed, InMemoryState, DefaultClock, NoOpMiddleware>>,
     max_requests_per_second: NonZeroU32,
+    throttled_requests: HashMap<MessageId, Message>,
```
cjrolo, line 42: Aren't MessageId the random 128Bit numbers? 
So, collisions should not be a problem and we could avoid hashing and having a single table ID -> Message?

My view is that, although this looks like a weird optimization step, I think if we are looking at performance, all these trivial things might start to add up under very heavy load. Hashing is a fast operation in today CPUs but if we can skip those, why not?

cjrolo (approved):

shotover/src/transforms/throttling.rs
rukai: I'm not sure what you are suggesting. A `Vec<(MessageId, Message>)` to do a linear look up instead of hashmap?

conorbros (approved):

PR #1486, opened 2024-02-19
Add type alias for common MessageIdMap

rukai:
Allows us to easily swap out our hashmap/hash implementation for the common idiom of `HashMap<MessageId, V>`.
Given how common this will be I think having finer control over our hashmap/hash will be quite valuable.

For now uses the fnv hasher as it is known to perform much better for integer keys than the default siphasher https://crates.io/crates/fnv

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4313,6 +4313,7 @@ dependencies = [
  "csv",
  "dashmap",
  "derivative",
+ "fnv",
  "futures",
  "generic-array",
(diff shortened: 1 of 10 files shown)


conorbros (approved):

cjrolo (approved):
Approved, check my comment on https://github.com/shotover/shotover-proxy/pull/1483 regarding the need for an hashmap.

PR #1490, opened 2024-02-21
Introduce TransformContextConfig

rukai:
This PR introduces the `TransformContextConfig` struct which is passed to each transform at the point where the `TransformConfig` is converted to a `TransformBuilder`. i.e. during shotover initialization

This enables a whole bunch of potential optimizations where we pass in some extra information about the context the transform is in. i.e. the chain its in and what other transforms are in its chain.
A potential future optimization would be detecting if the next transform is going to discard any responses passed to it.
This would allow for a nice optimization to the Filter transform where we skip error generation if its in a Tee subchain configured to discard responses.

However right now, `TransformContextConfig` just contains:
* the name of the chain which we currently use in various diagnostics.
* The protocol that the transform will be processing, currently there is no usages of this information but I plan to use it as part of the port of the Tee transform to use MessageId.

~~Additionally the `CodecBuilder::websocket_subprotocol` is turned into a `CodecBuilder::protocol` so that we can obtain the protocol to insert into the `TransformContextConfig`.~~

### DIFF
--- a/custom-transforms-example/src/redis_get_rewrite.rs
+++ b/custom-transforms-example/src/redis_get_rewrite.rs
@@ -3,7 +3,9 @@ use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
 use shotover::frame::{Frame, RedisFrame};
 use shotover::message::{MessageIdSet, Messages};
-use shotover::transforms::{Transform, TransformBuilder, TransformConfig, Wrapper};
+use shotover::transforms::{
+    Transform, TransformBuilder, TransformConfig, TransformContextConfig, Wrapper,
+};
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(deny_unknown_fields)]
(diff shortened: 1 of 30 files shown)


shotover/src/codec/cassandra.rs
```diff
@@ -139,8 +139,10 @@ impl CodecBuilder for CassandraCodecBuilder {
         )
     }
 
-    fn websocket_subprotocol(&self) -> &'static str {
-        "cql"
+    fn protocol(&self) -> ProtocolType {
```
conorbros, line 142: Could we pull the `ProtocolType` changes out of this?

rukai: Are you thinking to make the ProtocolType a prereq?
It is a requirement of the `TransformContextConfig`.

conorbros: Yep, thanks. I might as a pre req

rukai: sure sounds reasonable,

cjrolo (approved):

conorbros (approved):

PR #1499, opened 2024-02-23
Kafka message parsing

conorbros:


rukai (approved):
Nice!
Thanks for solving this so we dont have to worry about adding anymore in the future!

rukai:
Woah, looks like this PR increases the size of the binary by 4MB probably still worth it though.

rukai:
Oh, this PR adds 3 minutes release build time in CI, maybe we should at least measure how long a release build takes locally before we merge.

rukai (changes requested):
We should measure effect on release build time locally before merging.
Maybe we can alter the kafka-parser crate to improve build times or maybe removing the use of generics here would help

cjrolo (approved):

rukai (approved):
I tested the build time locally.
On my machine, this PR increases the time to run `cargo build --release -p shotover-proxy --no-default-features --features kafka` from 1min 15s to 1min 32s

This is an unfortunate regression, but we do want to be parsing every kafka message, so lets land it and possibly investigate ways to speed it up in the future.

PR #1502, opened 2024-02-26
Replace hyper with axum

rukai:
Simplifies the implementation of our configuration webservers and removes the duplicate hyper dependency.

Currently this PR regresses release build time by 9s on my machine.
I thought we might win back some build time by the removal of hyper 0.14 but in this measurement of 9s I have already removed hyper 0.14

Not sure if its worth proceeding with this cost to build time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -150,9 +150,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.80"
+version = "1.0.81"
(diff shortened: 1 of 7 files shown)


conorbros (approved):

rukai:
https://github.com/hyperium/hyper-util/pull/110 Someone created an example demonstrate using hyper 1.0 as a server, might be worth trying it out before landing this.

cjrolo (approved):

PR #1503, opened 2024-02-26
Replace hyper with reqwest in tests

rukai:
reqwest is higher level, allowing for simpler test logic, and we already use it in other places so we arent paying any cost in compilation time.


conorbros (approved):

cjrolo (approved):
+5/-48 is always nice!

PR #1513, opened 2024-03-05
Kafka cluster sasl

conorbros:
This PR adds the functionality for KafkaSinkCluster SASL authentication.

It works in a similar way to how we implemented authentication for CassandraSinkCluster, intercepting the authentication handshake and using that to initiate new connections.

I had to add a config option because it is difficult to figure out during runtime what configuration the cluster is using. Since SASL has to be enabled in both the client and on each kafka node config, needing to enable within Shotover isn't an issue.

Followups:
- need to make sure that we can authenticate with new nodes that are added to the cluster.
- add a test for SASL with TLS

There might be a slight performance hit to the time for the first connection because we open connections to each available node after receiving the full handshake. Might be possible to do this lazily.


shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -462,6 +499,57 @@ impl KafkaSinkCluster {
                     ..
                 })) => results.push(self.route_to_controller(message).await?),
 
+                Some(Frame::Kafka(KafkaFrame::Request {
+                    body: RequestBody::ApiVersions(_),
+                    ..
+                })) => {
+                    let connection = self
```
rukai, line 506: Should this call route_to_first_node?

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -462,6 +499,57 @@ impl KafkaSinkCluster {
...
+                    ..
+                })) => {
+                    let (tx, rx) = oneshot::channel();
+                    self.route_to_first_node(message.clone(), Some(tx)).await?;
+                    self.connection_factory.add_auth_message(message.clone());
+                    results.push(rx);
+                    self.sasl_status.set_handshake_complete();
+
+                    // now that we have the full handshake, open a connection to all nodes
+                    for node in &mut self.nodes {
```
rukai, line 548: This linearly opens a connection to every node. This will be quite slow in large connections.
But I think we are better off just not doing this at all, so that the cluster sink will lazily open connections as they are needed.

shotover-proxy/tests/test-configs/kafka/cluster-sasl/topology-single.yaml
```diff
@@ -0,0 +1,11 @@
+---
```
rukai, line 1: this is unused

rukai:
I dont see any fundamental reason why we need a config to enable/disable SASL in shotover.
If it proved tricky to implement then I'm happy to land as is to avoid blocking progress, but I'll have a go at removing it after we land the PR and before our next release.

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -484,6 +572,28 @@ impl KafkaSinkCluster {
         Ok(results)
     }
 
+    async fn route_to_first_node(
```
rukai, line 575: Routing to the first node in the list could cause issues in the case where:
* send handshake message 1
* shared_nodes is updated and a new node is first in the list.
* send handshake message 2

In this case the second handshake message goes to a different node.
To avoid this issue we need to a pick a node, store its node id and then continue sending messages to the node with that id.

I'm happy for an issue to be raised and this be addressed in a follow up PR.

rukai (approved):
I've moved my pending concerns to https://github.com/shotover/shotover-proxy/issues/1522
Lets go ahead and land this, so we can start making incremental fixes on top of it.

cjrolo (approved):
This is a great piece of work. Nicely done.

shotover-proxy/tests/kafka_int_tests/mod.rs
```diff
@@ -253,3 +253,63 @@ async fn cluster_2_racks_multi_shotover(#[case] driver: KafkaDriver) {
...
+        Duration::from_secs(10),
+        shotover.shutdown_and_then_consume_events(&[]),
+    )
+    .await
+    .expect("Shotover did not shutdown within 10s");
+}
+
+#[rstest]
+#[cfg_attr(feature = "rdkafka-driver-tests", case::cpp(KafkaDriver::Cpp))]
+#[case::java(KafkaDriver::Java)]
```
rukai, line 283: I believe the failures are due to the recently merged java producer+consumer pr.
Should be able to just disable the java driver for cluster tests added in this pr

shotover-proxy/tests/kafka_int_tests/mod.rs
```diff
@@ -113,10 +113,11 @@ async fn passthrough_sasl(#[case] driver: KafkaDriver) {
...
 }
 
+#[cfg(feature = "rdkafka-driver-tests")]
 #[cfg(feature = "alpha-transforms")]
 #[rstest]
 #[cfg_attr(feature = "rdkafka-driver-tests", case::cpp(KafkaDriver::Cpp))]
-#[case::java(KafkaDriver::Java)]
+// #[case::java(KafkaDriver::Java)]
 #[tokio::test(flavor = "multi_thread")] // multi_thread is needed since java driver will block when consuming, causing shotover logs to not appear
 async fn passthrough_sasl_encode(#[case] driver: KafkaDriver) {
```
rukai, line 122: The passthrough tests are unaffected.

Issue #1522, opened 2024-03-12
Kafka supports SASL authentication


rukai:
These should be addressed in separate PRs.

<s>## Issue 1 - race condition</s>

To move the SASL work forward we've left a race condition in https://github.com/shotover/shotover-proxy/pull/1513#discussion_r1517123818
This needs to be fixed before we start using SASL.

## Issue 2 - simpler config

I dont see any fundamental reason why we need a config to enable/disable SASL in shotover.
Before we make our next release we should change our SASL support to not rely on a config forcing it on/off.

PR #1528, opened 2024-03-14
KafkaSinkCluster race condition fix

conorbros:
fixes the potential race condition raised in https://github.com/shotover/shotover-proxy/issues/1522.


rukai (approved):
Looks good, this fixes the issue I described.

cjrolo (approved):

PR #1531, opened 2024-03-17
Remove pushed_messages mpsc from kafka transforms

rukai:
While kafka sounds like it would make use of pushed messages due to its pubsub model, responses are never sent without a corresponding request.
A fetch request is resent everytime the client wants another fetch response.

So these pushed messages senders can be just removed, I assume they were originally added due to copy/pasting from another transform as a base and it was just assumed that kafka would need push messages.


conorbros (approved):

cjrolo (approved):
Less lines of code are always a good change!

PR #1543, opened 2024-03-21
Dummy message bugfix

rukai:
I noticed this issue while trying to port dummy message handling to the new Connection type.

We plan to delete this code in the future, but we should probably land this fix in case we dont get around to it.

Edit: Oh, this select is also broken because the codec reader is not cancel safe.
This whole thing needs to go, we should just expedite the porting to the new Connection instead.


rukai:
Closing this in favor of just moving everything off this old logic and then deleting it.

PR #1545, opened 2024-03-21
Remove sasl config

conorbros:
closes https://github.com/shotover/shotover-proxy/issues/1522

https://github.com/shotover/shotover-proxy/pull/1513 introduced a config for SASL as a workaround until we arrived at an implementation that dynamically handled both SASL and non-SASL
This PR removes that workaround.


shotover-proxy/config/config.yaml
```diff
@@ -1,5 +1,5 @@
 # configure the first `info` to set the log level for dependencies
 # configure `shotover=info` to set the log level for shotover itself
 # set `shotover::connection_span=info` to `shotover::connection_span=debug` to attach connection info to most log events, this is disabled by default due to a minor performance hit.
-main_log_level: "info, shotover=info, shotover::connection_span=info"
+main_log_level: "debug, shotover=debug, shotover::connection_span=debug"
```
rukai, line 4: left in debugging

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -207,28 +202,22 @@ impl AtomicBrokerId {
 
 #[derive(Debug)]
 struct SaslStatus {
```
rukai, line 204: Should we just replace this with a simple `auth_complete: bool` field in KafkaSinkCluster?

rukai (approved):

cjrolo (approved):

PR #1547, opened 2024-03-22
Port CassandraSinkSingle to Connection + remove transform_pushed impementation

rukai:
This PR includes two changes that needed to be included in a single PR as they are dependent on each other:
* CassandraSinkSingle uses the new Connection type instead of the old Cassandra specific CassandraConnection.
* CassandraSinkSingle and CassandraPeersRewrite no longer use `Transform::transform_pushed` and instead expect cassandra events to come through the regular `Transform::transform` chain.
    + CassandraPeersRewrite only works with CassandraSinkSingle so they had to be done together but I see no issue with leaving CassandraSinkCluster for another PR.

Additionally this PR adds logic to the cassandra codec to assign request_id's to incoming responses.
This could technically be in a prereq, but its fairly straightforward so I'm including it here.
It follows the exact same logic as the kafka and redis codecs.

### DIFF
--- a/shotover/src/codec/cassandra.rs
+++ b/shotover/src/codec/cassandra.rs
@@ -1,8 +1,8 @@
 use super::{CodecBuilder, CodecReadError, CodecWriteError, Direction};
 use crate::codec::CodecState;
-use crate::frame::cassandra::{CassandraMetadata, CassandraOperation, Tracing};
+use crate::frame::cassandra::{CassandraOperation, Tracing};
 use crate::frame::{CassandraFrame, Frame, MessageType};
-use crate::message::{Encodable, Message, Messages, Metadata};
+use crate::message::{Encodable, Message, MessageId, Messages, Metadata};
 use anyhow::{anyhow, Result};
 use atomic_enum::atomic_enum;
 use bytes::{Buf, BufMut, Bytes, BytesMut};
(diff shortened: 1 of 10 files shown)


conorbros (approved):

shotover/src/codec/cassandra.rs
```diff
@@ -785,6 +823,17 @@ impl CassandraEncoder {
             return Ok(());
         }
 
+        if let Some(tx) = &self.stream_id_to_request_id_tx {
+            let Ok(Metadata::Cassandra(meta)) = m.metadata() else {
+                unreachable!("Gauranteed to be cassandra")
```
cjrolo, line 828: Typo! "Guaranteed"

shotover/src/codec/cassandra.rs
```diff
@@ -17,6 +17,7 @@ use lz4_flex::{block::get_maximum_output_size, compress_into, decompress};
 use metrics::{counter, Counter, Histogram};
 use std::collections::HashMap;
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::mpsc;
```
cjrolo, line 20: Good stuff now that this exists!

cjrolo (approved):

shotover/src/codec/cassandra.rs
rukai: This is a common mistake of mine so I fixed it throughout the codebase, thanks.

PR #1553, opened 2024-03-26
CassandraSinkCluster replace recv with try_recv

rukai:
This PR swaps CassandraSinkCluster from recv to try_recv for the performance improvements it brings:
* use try_recv instead of awaiting on recv. - this gives us the benchmark improvement shown below
* when the MessageRewriter invents a new request out of thin air it needs to assign it a unique stream_id that is unique for its destination node. This is a difficult problem. The simplest approach is just to return a BatchMode, allowing the rewriter to dictate whether the current batch of requests is allowed to pipeline the requests for efficiency or whether the current batch must be isolated allowing the rewriter to generate stream_id's as long as they are unique within the current request batch.

Looks like a large win for connection_count=1
image

And when comparing against benches without shotover in the loop we can see that shotover is now very close!
connection_count = 1 has shotover lagging behind by about 20-30%
image

connection_count = 10 has slightly higher throughput for shotover (in reality thats probably just noise)
image

connection_count = 100 is also pretty much identical.
image


cjrolo (approved):

benbromhead (approved):
Nice!

PR #1555, opened 2024-03-27
CassandraSinkCluster index removal

rukai:
https://github.com/shotover/shotover-proxy/pull/1476 ported CassandraSinkCluster to use MessageId.
Well almost, it missed one case, which is the handling of use statements.
This PR fixes that.

I spent a bunch of time thinking about how to handle processing them with MessageId's instead of indexes.
Its definitely possible but it also adds a bunch of extra complexity and I'm not sure if its worth it in the first place.

So @cjrolo I'm interested in your thoughts on how often cassandra clients make use of `use`ing a namespace.
With this PR every time the client uses a `use namespace` statement shotover will close all outgoing connections in use by the transform instance (1 connection per cassandra node in the cluster).
Those connections will then be reopened as shotover routes requests to those nodes. (i.e. there will be delays due to connections being opened in the future)
So if clients regularly change the `use`d namespace then I'd better not go ahead with the approach in this PR.

### DIFF
--- a/shotover/src/transforms/cassandra/sink_cluster/mod.rs
+++ b/shotover/src/transforms/cassandra/sink_cluster/mod.rs
@@ -4,7 +4,7 @@ use crate::frame::cassandra::{CassandraMetadata, Tracing};
 use crate::frame::{CassandraFrame, CassandraOperation, CassandraResult, Frame};
 use crate::message::{Message, MessageIdMap, Messages, Metadata};
 use crate::tls::{TlsConnector, TlsConnectorConfig};
-use crate::transforms::cassandra::connection::{CassandraConnection, Response, ResponseError};
+use crate::transforms::cassandra::connection::{CassandraConnection, ResponseError};
 use crate::transforms::{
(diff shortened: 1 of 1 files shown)


cjrolo:
It will depend by use case. I wasn't using `USE namespace` at all, and I don't remember it being used often at all.
I can ask around in clusters with a lot of keyspaces how the driver looks like. Are they having a specific driver for each keyspace? Is the keyspace set as `keyspace.table`? They prepare a statement bound to the keyspace (How does it translate in different drivers to the query level, I don't know), or they use the `USE` keyword? 

I will try to gather further information for you.

cjrolo:
I do like the current approach, and I think `USE` is a niche thing on the domain of CQL connections via CQLSH. But I'll still do my research and get back to you.

rukai:
Thanks!
I'll go ahead and land this assuming @conorbros is happy with this approach.
I've raised https://github.com/shotover/shotover-proxy/issues/1559 to track the re-adding of keeping the connections open.
@cjrolo if you find anything interesting post it on the issue.

conorbros (approved):

PR #1561, opened 2024-04-03
Port CassandraSinkCluster topology task to new Connection type

rukai:
Pulled out of https://github.com/shotover/shotover-proxy/pull/1553
In this PR:
* We pass a few extra variables around to allow the construction of the SinkConnection. (read_timeout and force_run_chain)
* We alter the logic of the topology task a little to handle the case where we send a request and await a response but get an event instead of a response.
    + This previously did not occur but occurs now because we no longer separate events into `Transform::transform_pushed`

### DIFF
--- a/shotover/src/transforms/cassandra/sink_cluster/topology.rs
+++ b/shotover/src/transforms/cassandra/sink_cluster/topology.rs
@@ -1,13 +1,13 @@
 use super::node::{CassandraNode, ConnectionFactory};
 use super::node_pool::KeyspaceMetadata;
 use super::KeyspaceChanTx;
+use crate::connection::SinkConnection;
 use crate::frame::{
     cassandra::{parse_statement_single, Tracing},
     value::GenericValue,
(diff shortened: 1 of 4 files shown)


cjrolo (approved):

shotover/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -65,133 +65,127 @@ async fn topology_task_process(
...
         connection_info.address
     );
 
+    let mut events = vec![];
     loop {
         // Wait for events to come in from the cassandra node.
         // If all the nodes receivers are closed then immediately stop listening and shutdown the task
-        let pushed_messages = tokio::select! {
-            pushed_messages = pushed_messages_rx.recv() => pushed_messages,
+        tokio::select! {
```
rukai, line 103: We should skip this if events.is_empty otherwise we will await a new message when we already have one  ready to go

conorbros (approved):

PR #1565, opened 2024-04-04
SinkConnection fix hang due to race condition

rukai:
Pulled out of https://github.com/shotover/shotover-proxy/pull/1553

Sink transforms rely on SinkConnection to call `force_run_chain.notify_one()`  to trigger a new chain run to avoid responses getting stuck in the SinkConnection until the client sends another request.

However this approach currently has a race condition allowing messages to get stuck.
Consider this scenario:
1. SinkConnection read_task: `in_tx.send(messsages)`
2. SinkConnection read_task: `force_run_chain.notify_one()`
3. SinkConnection read_task: `in_tx.send(messsages)`
4. SinkConnection read_task: `force_run_chain.notify_one()` - Has no affect since there is already a notification.

At this point we have 2 message batches pending in the read channel.

4. server.rs in main task: `force_run_chain.notified` triggers causing a chain run to occur.
5. SinkConnection try_recv in main task: `in_rx.try_recv()`

At this point, there is still one message batch still in the read channel.
But it will be stuck there until either another response comes back or the client sends another request.

This PR fixes this issue by looping in `SinkConnection::try_recv` to ensure we completely drain the channel after receiving a `force_run_chain.notified()`

### DIFF
--- a/shotover/src/connection.rs
+++ b/shotover/src/connection.rs
@@ -135,14 +135,21 @@ impl SinkConnection {
         if let Some(error) = &self.error {
             Err(error.clone())
         } else {
-            match self.in_rx.try_recv() {
-                Ok(mut messages) => {
-                    self.dummy_response_inserter
-                        .process_responses(&mut messages);
-                    Ok(messages)
+            let mut results = vec![];
+            loop {
+                match self.in_rx.try_recv() {
+                    Ok(mut responses) => {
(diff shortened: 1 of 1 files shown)


conorbros (approved):

cjrolo (approved):
Wonderful world of race conditions.

PR #1569, opened 2024-04-05
SinkConnection improvements

rukai:
This PR improves SinkConnection to be more robust and efficient. https://github.com/shotover/shotover-proxy/pull/1565 started on this work but I didnt realize how deep the issue went.

1. First I realized that the fix in #1565 only fixed try_recv. recv had the same issue and that was unfixed. So this PR adds the fix to `recv` as well.
2. I realized that fixing #1565 introduced a bunch of needless allocations. The fix to this is to have the user pass in the `&mut Vec<Message>` that they will be adding to. This is a fairly ergonomic solution since sinks often directly append the result to another vec anyway.
    + In doing this I renamed `recv` -> `recv_into` and `try_recv` -> `try_recv_into` to better suit the new functionality.
    + Additionally I had to port all the transforms to use this new API.

### DIFF
--- a/shotover/src/connection.rs
+++ b/shotover/src/connection.rs
@@ -71,10 +71,7 @@ impl SinkConnection {
             );
         }
 
-        let dummy_response_inserter = DummyResponseInserter {
-            dummy_requests: vec![],
-            pending_requests_count: 0,
-        };
+        let dummy_response_inserter = DummyResponseInserter::new();
 
         Ok(SinkConnection {
             in_rx,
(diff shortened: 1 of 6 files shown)


shotover/src/connection.rs
```diff
@@ -412,9 +451,10 @@ impl DummyResponseInserter {
         self.pending_requests_count += requests.len();
     }
 
-    pub fn process_responses(&mut self, responses: &mut Vec<Message>) {
+    pub fn process_responses(&mut self, responses: &mut Vec<Message>, start_at: usize) {
```
rukai, line 454: I believe this is correct, I should write some unit tests though, particularly with this extra parameter

shotover/src/connection.rs
```diff
@@ -104,55 +104,94 @@ impl SinkConnection {
...
+                        if responses.capacity() == 0 {
+                            *responses = new;
                         } else {
-                            results.extend(responses)
+                            responses.extend(new)
                         }
                     }
-                    Err(TryRecvError::Disconnected) => return Err(self.set_get_error()),
-                    Err(TryRecvError::Empty) => return Ok(results),
+                    Err(TryRecvError::Disconnected) => {
```
rukai, line 164: Run the dummy inserter here too

shotover/src/connection.rs
```diff
@@ -104,55 +104,94 @@ impl SinkConnection {
...
-            }
-
-            match self.in_rx.recv().await {
-                Some(mut messages) => {
-                    self.dummy_response_inserter
-                        .process_responses(&mut messages);
-                    Ok(messages)
+            let initial_count = responses.len();
+            // exhaust channel and ensure dummy messages are inserted by running try_recv_into
+            self.try_recv_into(responses)?;
```
rukai, line 113: We can remove the self.error check since this does it for us

shotover/src/connection.rs
```diff
@@ -104,55 +104,94 @@ impl SinkConnection {
...
-                    Err(TryRecvError::Disconnected) => return Err(self.set_get_error()),
-                    Err(TryRecvError::Empty) => return Ok(results),
+                    Err(TryRecvError::Disconnected) => {
+                        return self.handle_disconnect_on_recv(responses, initial_count)
+                    }
+                    Err(TryRecvError::Empty) => {
+                        // We need to ensure this runs even when we receive nothing from the channel.
+                        // this will ensure that the dummy message at index 0 is inserted if it exists
+                        self.dummy_response_inserter
+                            .process_responses(responses, initial_count);
```
rukai, line 171: I think I can ditch the initial_count Arg if I move this call directly onto new

rukai: I think inserting to responses only results in less allocations though

cjrolo (approved):

PR #1573, opened 2024-04-15
KafkaSinkCluster accept shotover addresses as domain names

rukai:
We should be able to specify the addresses of shotover nodes as a domain name.
This PR makes that possible by allowing the host name to be an arbitrary string.


conorbros:
Worth testing by setting up DNS with Docker?

rukai:
that sounds like a good idea, I'll give it a go

rukai:
I dont think I can use docker to setup DNS here since shotover is not run within docker.
However I can just use `localhost` instead of `127.0.0.1`.
So thats what I did.

To make that work however I have to set an ipv4 config for the CPP driver otherwise it tries to connect with ipv6, which fails and then it gives up.

cjrolo (approved):

PR #1576, opened 2024-04-16
Remove dummy message insertion

rukai:
Some protocols, so far only kafka, have some requests which do not trigger a response.
This is forbidden by our transform invariants, so to handle this we have logic in SinkConnection to insert dummy responses that correspond to those requests.
This works but I've done some thinking and I am proposing that we throw this away and instead have the user just call the new `respondable_id()` method when they need to wait for a response on a protocol which may not send a response:
* The dummy insertion logic is very complicated, I'd rather not have to maintain it. For example I think its causing mysterious problems with https://github.com/shotover/shotover-proxy/pull/1569
* It should be a performance improvement for cases with many responseless requests, for example a kafka client configured with `acks = 0`:
   + In SinkConnection, avoids bookkeeping vec allocations, message vec insertions.
   + In certain transforms, responseless requests do not need to have their ID's inserted and removed from hashmaps to correlate to a response that is in reality only a dummy response to be ignored.
* We are swapping around complexity, hard to say which is better.
   + Previously, transforms for certain protocols need to handle the possibility of some responses containing Dummy frames instead of the frame of the protocol.
   + With this PR, transforms for certain protocols need to call `respondable_id()` instead of `id()` when correlating responses to requests, currently none of our transforms actually need to do this.

### DIFF
--- a/shotover/src/connection.rs
+++ b/shotover/src/connection.rs
@@ -1,8 +1,7 @@
 //! All Sink transforms use SinkConnection for their outgoing connections.
 
 use crate::codec::{CodecBuilder, CodecReadError, CodecWriteError};
-use crate::frame::Frame;
-use crate::message::{Message, MessageId, Messages};
+use crate::message::{Message, Messages};
 use crate::tcp;
 use crate::tls::{TlsConnector, ToHostname};
 use futures::{SinkExt, StreamExt};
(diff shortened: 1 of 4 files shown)


rukai:
Whoops, this doesnt account for dummy messages needing a dummy response.

PR #1577, opened 2024-04-17
Remove TuneableConsistencyScatter and RedisTimestampTagger transforms

rukai:
This PR removes TuneableConsistencyScatter and RedisTimestampTagger.
For consideration of removal I am considering these transforms to be a single transform since they both rely on the other to do anything useful.

Reasons for removal:
* To my knowledge these have never been used by us or anyone else
* There are currently no plans for these transforms to be used in the future
* They add significant maintenance burden.
* They rely on the existence of a [special field in `Message`](https://github.com/shotover/shotover-proxy/blob/61471324552614c8473152826633dbf2a1388e49/shotover/src/message/mod.rs#L58) only used by these transforms, the existence of this field makes message creation/deletion/moving more expensive for all shotover use cases.
* Current implementation has design issues:
   * Should be a single transform instead of splitting into 2 transforms that always rely on each other.
   * Doesnt handle redis transactions
   * Relies on determining the size of the message but this is ill defined and currently stubbed out

Reasons for keeping:
* In the future we may want to support this exact use case.
   + The current implementation has design issues and even if we wanted to support it in the future I think we would be better off starting from scratch anyway. using redis's lua support to add a timestamp is neat, but we can always check the git history to retrieve that part of the implementation if needed.
* In the future we may want to support similar use cases, keeping the transforms around ensuresthat shotover's underlying abstractions continue to meet the needs for this kind of use case.
    + We still have the ParallelMap transform which I consider to be close enough to demonstrate the functionality required for this use case.


conorbros (approved):
The TuneableConsistencyScatter usecase could certainly come up again but if there is issues with the design I am not opposed to removing it and re implementing later if required.

cjrolo (approved):

benbromhead:
Just to round out my thoughts for why they need to get removed:
- Both these implementations were very exploratory and had a whole bunch of edge case issues
- Shoehorning consistency mechanisms on protocols that don't implicitly support it is always going to be really really hard at best and the average case is just one giant footgun
- Both required external consistency recovery mechanisms that never got implemented.

PR #1581, opened 2024-04-18
Add bug investigation docs

rukai:
Introduces some documentation on tracking down the cause of bugs aimed at absolute beginner shotover developers.
Includes shotover specific guidance as well as general rust guidance.

While writing the docs I noticed that shotover defaults to log level `INFO` while the integration tests default to log level `WARN`, rather than documenting the mismatch I just fixed it so that the integration tests log at INFO level.

### DIFF
--- a/docs/src/dev-docs/debugging.md
+++ b/docs/src/dev-docs/debugging.md
@@ -26,6 +26,13 @@ sources:
             connect_timeout_ms: 3000
 ```
 
+For a simple redis request/response, the logs will look like:
+
+```plain
+shotover   06:37:14.712042Z  INFO connection{id=2 source="redis"}: shotover::transforms::debug::printer: Request: Redis Array([BulkString(b"GET"), BulkString(b"bar")])
+shotover   06:37:14.712212Z  INFO connection{id=2 source="redis"}: shotover::transforms::debug::printer: Response: Redis BulkString(b"foo")
+```
+
 ## Run the test
 
 Run the test by:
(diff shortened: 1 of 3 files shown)


docs/src/dev-docs/debugging.md
```diff
@@ -0,0 +1,86 @@
...
+```shell
+cargo nextest run --no-default-features --features kafka kafka_int_tests::passthrough_standard --nocapture
+```
+
+Broken down we have:
+
+* `--no-default-features --features kafka` will compile only the parts of shotover and the integration tests required for the specified protocol.
+  * This will drastically improve time to compile shotover and the tests, currently it drops from 15s to 5s. The difference will grow as shotover grows.
+  * For a full list of features refer to the `[features]` section in [shotover-proxy/Cargo.toml](https://github.com/shotover/shotover-proxy/blob/main/shotover-proxy/Cargo.toml)
+* `kafka_int_tests::passthrough_standard::case_1` the exact name of the test to run.
```
cjrolo, line 49: Doesn't match the shell command above. While you explain in the line below how to filter, I think this explanation should match the shell command.

docs/src/dev-docs/debugging.md
```diff
@@ -0,0 +1,86 @@
...
+* Responses are coming from the sink and into the sink from the sink's codec.
+
+So determine whether to first investigate the source codec or sink by determining if the problem is occurring in the debug printer's requests or responses.
+
+Sprinkle `tracing::info!("some log {}", some.variable)` around suspect areas in the codebase to narrow down whats going wrong.
+Rerun the test each time to learn more about the behavior of the system and narrow down your search.
+
+Delete these extra logs when you are finished your investigation. Some of them could be downgraded to a `tracing::debug!()` and kept if they are found to be generally valuable.
+
+## Ensure the project has no warnings
```
cjrolo, line 78: I would put this higher in the list. Maybe even make it the first topic.

cjrolo (approved):
Addressing the comments is more of a personal take. Very good commit, very useful documentation.

conorbros (approved):

docs/src/dev-docs/debugging.md
rukai: good catch!

rukai:
Thanks for the feedback!
Merging.

PR #1588, opened 2024-04-23
Fix Kafka java driver multi shotover tests

rukai:
This PR brings us closer to full support for the java driver.
The java driver integration tests that are currently disabled are now enabled but some of the test cases within those integration tests are still disabled.
This is done by calling a `minimal_test_suite` function and calling it when running the java driver.

The fixes to shotover was to handle prefetching topic metadata for fetches using the newer protocol version used by the java driver.
The new protocol specifies topics by their id instead of their name.
Previously we were only prefetching topic metadata by name but now we also do it by topic id.
This involved:
1. finding a list of all topic id's in the current message batch. (route_requests method)
2. Including these topic id's in the metadata request (get_metadata_of_topics method)
   + I chose to revert to a lower api_version when not dealing with topic id's, this isnt required but I figured we may as well support the older kafka versions since it doesn't add much complexity.

### DIFF
--- a/shotover-proxy/tests/kafka_int_tests/mod.rs
+++ b/shotover-proxy/tests/kafka_int_tests/mod.rs
@@ -5,6 +5,8 @@ use rstest::rstest;
 use std::time::Duration;
 use test_helpers::connection::kafka::{KafkaConnectionBuilder, KafkaDriver};
 use test_helpers::docker_compose::docker_compose;
+use test_helpers::shotover_process::{Count, EventMatcher};
+use tokio_bin_process::event::Level;
 
 #[rstest]
 #[cfg_attr(feature = "kafka-cpp-driver-tests", case::cpp(KafkaDriver::Cpp))]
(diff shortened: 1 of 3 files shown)


shotover-proxy/tests/kafka_int_tests/mod.rs
```diff
@@ -263,22 +264,25 @@ async fn cluster_1_rack_multi_shotover(#[case] driver: KafkaDriver) {
     }
 
     let connection_builder = KafkaConnectionBuilder::new(driver, "127.0.0.1:9192");
-    test_cases::standard_test_suite(connection_builder).await;
+    match driver {
+        #[cfg(feature = "kafka-cpp-driver-tests")]
+        KafkaDriver::Cpp => test_cases::standard_test_suite(connection_builder).await,
+        KafkaDriver::Java => test_cases::minimal_test_suite(connection_builder).await,
```
conorbros, line 270: This is intended to be temporary?

conorbros: Ah I see the todo is in the minimal_test_suite function.

conorbros (approved):

cjrolo (approved):

PR #1590, opened 2024-04-26
KafkaSinkCluster sasl failure tests

rukai:
This PR introduces tests to handle sasl failures:
* cluster_sasl_plain_multi_shotover is extended to also test plain sasl with an incorrect password
  * It ensures that requests fail when the incorrect password is used.
* cluster_sasl_scram_single_shotover tests scram sasl with a correct password.
   * Currently shotover fails when using SCRAM with KafkaSinkCluster, this is because we cannot replay SCRAM sasl requests like we can with plain sasl.
   * This test demonstrates the issue by asserting that the request fails when it should actually pass.
   * Additionally we alter how shotover performs auth replays so that it catches auth failures and immediately fails the connection. This considerably improves our error reporting and the test asserts that such error reporting occurs.
   
test-helpers was adjusted to allow asserting that `new_topics` failed, there is nothing particularly interesting about new_topics itself, any request would have worked, I just picked it at random.


shotover-proxy/tests/kafka_int_tests/mod.rs
```diff
@@ -316,6 +334,20 @@ async fn cluster_sasl_plain_multi_shotover(#[case] driver: KafkaDriver) {
...
 
+    // Test invalid credentials
+    // We perform the regular test suite first in an attempt to catch a scenario
+    // where valid credentials are leaked into subsequent connections.
+    let connection_builder = KafkaConnectionBuilder::new(driver, "127.0.0.1:9192")
+        .use_sasl_plain("user", "not_the_password");
+    assert_eq!(
+        connection_builder.assert_admin_error().await.to_string(),
+        match driver {
+            #[cfg(feature = "rdkafka-driver-tests")]
```
conorbros, line 345: Seems every test in this module has this feature guard? Might be able to move it up a level - not for this PR though.

conorbros (approved):

shotover-proxy/tests/kafka_int_tests/mod.rs
rukai: This feature guard is for the cpp driver, the test cases that use java driver can still be run when its disabled.
Now that you've brought this up, I'm thinking rdkafka-driver-tests should be renamed to `kafka-cpp-driver-tests`, since its not obvious that rdkafka is the name of the cpp driver for kafka. I'll do this in another PR.

cjrolo (approved):

PR #1591, opened 2024-04-26
Redo internal error response generation

rukai:
This PR addresses a few issues:
* internal error response generation has been broken since the recent transform invariant refactorings.
   + To resolve this, internal error response generation is completely redone. A `PendingRequests` type is introduced to keep track of pending requests. It uses a different method for tracking depending on the protocol.
* flushing was triggering after internal error generation, this is bad since transforms do not attempt to maintain their internal invariants when returning an error.
   + To resolve this, I changed shotover to skip flushing if the chain was closed due to a transform error and additionally documented shotover's behavior here as a transform invariant.
* We were cloning entire messages which is a performance concern since, while its true that newly created messages should have no allocations:
    + They still take up a lot of stack space, its currently like ~400bytes per message.
    + I think some codecs DO parse messages if they really need to in order to handle weird edge cases. .e.g cassandra batch messages.
    + This PR resolves the issue by storing the request as a `Metadata` instead of a `Message`, just 4 bytes of stack space.
* to_error_responses was leaving the responses without the appropriate request_id set. In order to correctly set the id its required to know if the message is a request or a response so I split the method into `from_request_to_error_response` and `from_response_to_error_response`

### DIFF
--- a/shotover/src/server.rs
+++ b/shotover/src/server.rs
@@ -1,6 +1,7 @@
 use crate::codec::{CodecBuilder, CodecReadError, CodecWriteError};
 use crate::config::chain::TransformChainConfig;
-use crate::message::{Message, Messages};
+use crate::frame::MessageType;
+use crate::message::{Message, MessageIdMap, Messages, Metadata};
 use crate::sources::Transport;
 use crate::tls::{AcceptError, TlsAcceptor};
 use crate::transforms::chain::{TransformChain, TransformChainBuilder};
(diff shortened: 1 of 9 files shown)


shotover/src/server.rs
```diff
@@ -814,3 +805,130 @@ impl Shutdown {
         self.shutdown = true;
     }
 }
+
+/// Keeps track of all currently pending requests.
+/// This allows error responses to be generated if the connection needs to be terminated before the response comes back.
+enum PendingRequests {
```
cjrolo, line 811: Would this be prone to a massive memory usage by a badly coded client?
One could even code a client in a way to DOS shotover this way. I would put a hard cap somewhere (might even be configurable).

cjrolo (approved):
Approving,

Left a comment that might be addressed in a different PR.

conorbros (approved):

PR #1592, opened 2024-04-29
Cassandra collection tests refactor

conorbros:
This PR refactors all Cassandra collection tests to have multiple elements in each collection for testing.

Removed boolean because sets require unique values and it's adequately covered by the other types. Not worth handling the edge case.


shotover-proxy/tests/cassandra_int_tests/collections.rs
```diff
@@ -350,42 +546,130 @@ mod map {
...
+        })
+        .await;
+        insert_table(connection, "map_list", |elements| {
+            let lists = format!("[{}]", elements.join(","));
+            let mut map = String::from("{");
+
+            for i in 0..2 {
+                map.push_str(&format!("{}: {}, ", i, lists));
+            }
+            map.push_str(&format!("{}: {}}}", 2, lists));
```
rukai, line 579: would this do the same thing?
```rust
            for i in 0..3 {
                map.push_str(&format!("{}: {}, ", i, lists));
            }
            map.push_str("}");
```

conorbros: That would leave a trailing comma, assumed invalid syntax but I can double check.

rukai: oh its that trailing comma, isnt it?

rukai:
Now we dont test boolean types at all, that seems unfortunate.
It should be straightforward to change `insert` and friends to work on an arbitrarily long `elements` instead of hardcoding to 3.
Then we can just have boolean contain 2 values (true and false).

shotover-proxy/tests/cassandra_int_tests/collections.rs
conorbros: How does it work for maps?

rukai: I'll give it a go and report back.

rukai: Here it is for map_list:

```rust
        insert_table(connection, "map_list", |elements| {
            let lists = format!("[{}]", elements.join(","));
            format!("{{{}}}", (0..3).map(|i| format!("{i}: {lists}")).join(","))
        })
        .await;
```

I also edited the example for set_set that I gave, there is hardcoded number of 3 outer sets that has no relation to the len of elements.

shotover-proxy/tests/cassandra_int_tests/collections.rs
```diff
@@ -350,42 +504,125 @@ mod map {
...
+                elements
+                    .iter()
+                    .enumerate()
+                    .map(|(i, element)| { format!("{i}: {element}") })
+                    .join(",")
+            );
+
+            format!("{{{}}}", (0..3).map(|i| format!("{i}: {maps}")).join(","))
+
+            // let mut maps = String::from("{");
```
rukai, line 540: dead code

rukai:
I pushed a commit to fix the boolean issue.
Should be good to go now.

rukai (approved):

shotover-proxy/tests/cassandra_int_tests/collections.rs
```diff
@@ -57,57 +57,148 @@ fn get_type_str(col_type: ColType) -> &'static str {
...
+            ),
+            ResultValue::TimeUuid(
+                uuid::Uuid::parse_str("84196262-53de-11ec-bf63-0242ac130002").unwrap(),
+            ),
+            ResultValue::TimeUuid(
+                uuid::Uuid::parse_str("84196262-53de-11ec-bf63-0242ac130003").unwrap(),
+            ),
+        ],
+        ColType::Inet => vec![
+            ResultValue::Inet("127.0.0.1".parse().unwrap()),
```
cjrolo, line 178: Inet is only IPv4?

rukai: https://docs.datastax.com/en/cql-oss/3.x/cql/cql_reference/cql_data_types_c.html 
image

IPV4 or IPV6

cjrolo (approved):

PR #1594, opened 2024-04-29
Remove unneeded derives on transforms

rukai:
The Clones are left over from when we used to Clone transforms to get a new one instead of using `TransformBuilder`s.

I'm not sure where the Debug derives came from?
They could be useful for debug printing but I've never wanted to debug print an entire transform, printing just a few relevant fields has always been more useful for me.

edit:
I also added in some extra changelog entries that were missed in other recent work.


conorbros:
Technically breaks our public API? I doubt it will matter in this instance, just something to keep in mind when changing the transform structs.

rukai:
True, I'll add a note to the changelog.

cjrolo (approved):

PR #1596, opened 2024-04-30
Transforms declare supported protocols

rukai:
Progress towards https://github.com/shotover/shotover-proxy/issues/520

This PR improves topology validation and allows for new kinds of topologies:
* We now catch and report to the user when two protocol incompatible transforms are connected to each other. Previously this would just blow up at runtime with a confusing error.
* We now allow topologies like: `CassandraToRedisTransform` -> `Tee`
   + Previously this would have failed since Tee relies on the protocol type being correctly set in `TransformContextConfig` but the protocol was only set to the protocol of the source and any changes to the protocol via previous transforms was ignored.

Not sure if up chain and down chain are the clearest way to name these but I also cant think of anything better.
I considered incoming and outgoing but I feel that is confusing since that its not clear what perspective its coming from.
Incoming could mean coming from the client but that would only be correct from the perspective of requests, since responses are actually outgoing to the client.

Work to do in follow up PRs:
* Correctly declare protocols for all existing transforms and make `up_chain_protocol` and `down_chain_protocol` mandatory
* Improve diagnostic formatting, when compared to the existing diagnostics for `TransformBuilder::validate` and `TransformBuilder::is_terminate` the diagnostics formatting is not as good. We will need to build up similar abstractions for diagnostics on config as we have for builders.
* Remove `TransformBuilder::is_terminating`, its redundant now since `DownChainProtocol` is a superset of that information.

### DIFF
--- a/shotover-proxy/tests/runner/runner_int_tests.rs
+++ b/shotover-proxy/tests/runner/runner_int_tests.rs
@@ -1,5 +1,5 @@
 use crate::shotover_process;
-use test_helpers::shotover_process::{Count, EventMatcher, Level};
+use test_helpers::shotover_process::{EventMatcher, Level};
 
 #[tokio::test]
 async fn test_early_shutdown_cassandra_source() {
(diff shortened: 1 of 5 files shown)


conorbros:
Request/Response protocol might be clearer but I think up/down protocol is fine too.

rukai:
Request could refer to both sides.

conorbros (approved):

cjrolo (approved):

PR #1598, opened 2024-05-01
valkey passthrough test

conorbros:
Valkey is looking to be the dominant open source fork for Redis, makes sense to run this smoke against it. There hasn't been an official release so I've pinned to latest. This would work to our advantage because it will notify us immediately if anything changes that breaks Shotover compatibility.

### DIFF
--- a/shotover-proxy/tests/test-configs/redis/passthrough-valkey/docker-compose.yaml
+++ b/shotover-proxy/tests/test-configs/redis/passthrough-valkey/docker-compose.yaml
@@ -1,0 +1,6 @@
+services:
+  valkey-one:
+    image: chainguard/valkey:latest
+    ports:
+      - "1111:6379"
+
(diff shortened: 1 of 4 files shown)


shotover-proxy/tests/test-configs/redis/passthrough-valkey/docker-compose.yaml
```diff
@@ -0,0 +1,6 @@
+services:
+  valkey-one:
+    image: chainguard/valkey:latest
```
rukai, line 3: Need to pin a specific version to avoid test breakage

conorbros: Updated description

rukai:
I think we should give this some more time before we start supporting a redis fork.
Googling for valkey doesnt even bring up a repo.
image

Lets reconsider this once valkey is large enough and it has a version we can pin to.
Happy for this to remain up as a draft until then.

conorbros:
Can close now for https://github.com/shotover/shotover-proxy/pull/1809/files

PR #1602, opened 2024-05-01
Cassandra test helpers module 

conorbros:


rukai (approved):
LGTM!

Splitting ResultValue and CassandraConnection into their own modules is nice.

I checked with a diff tool: `cassandra/mod.rs`, `cassandra/connection.rs`, `cassandra/result_value.rs` are all directly split from the original `cassandra.rs` file, without any other changes sneaking in.

cjrolo (approved):

PR #1603, opened 2024-05-01
Move cql ws to shotover

conorbros:
We had some maintenance issues keeping this dependency in a repo. https://github.com/shotover/cql-ws
It's small enough that it can just be a module within our `test-helpers` crate. 

~~https://github.com/shotover/shotover-proxy/pull/1602 is a pre-req~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1211,20 +1211,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "cql-ws"
-version = "0.1.0"
-source = "git+https://github.com/shotover/cql-ws#11ec9db0acc0c1e605eae214a81a327571f4cd97"
-dependencies = [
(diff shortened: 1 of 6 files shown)


rukai (approved):
The websocket project has been in limbo in a while due to difficulty upstreaming required changes to dependencies and drivers.
This has left us with a large maintenance burden of keeping this chain of git dependencies up to date. The chain looks like `cql-ws` -> `tokio-tungstenite` -> `tungstenite`.

I think that moving cql-ws into the shotover repo will help reduce that chain while we are in limbo.
In the future, if this project is worth putting more time into, it would be good to see cql-ws replaced with a more complete external driver.
For now it is sufficient for integration testing to ensure what we have doesnt break.

cjrolo (approved):

PR #1604, opened 2024-05-01
Fix FindCoordinatorResponse rewriting for java driver

rukai:
Previously we were creating multiple coordinators per key.
The java driver does not like this and raises an exception, so this PR avoids doing that.
The comments in the code change do a pretty good job of explaining more details.

This issue showed up in the currently disabled `cluster_1_rack_multi_shotover::case_1_java` but I cant enable the test yet since there are still other deeper issues causing the test to fail.

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/mod.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/mod.rs
@@ -9,7 +9,6 @@ use anyhow::{anyhow, Context, Result};
 use async_trait::async_trait;
 use dashmap::DashMap;
 use kafka_protocol::messages::fetch_request::FetchTopic;
-use kafka_protocol::messages::find_coordinator_response::Coordinator;
 use kafka_protocol::messages::metadata_request::MetadataRequestTopic;
 use kafka_protocol::messages::metadata_response::MetadataResponseBroker;
 use kafka_protocol::messages::{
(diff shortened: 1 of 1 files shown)


conorbros (approved):

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1221,15 +1220,22 @@ impl KafkaSinkCluster {
...
         } else {
-            // for version > 3 we can include as many coordinators as we want.
-            // so we include all shotover nodes in the rack of the coordinator.
-            let mut shotover_coordinators: Vec<Coordinator> = vec![];
-            for coordinator in find_coordinator.coordinators.drain(..) {
+            // For version > 3 we have to replace multiple coordinators.
+            // It may be tempting to include all shotover nodes in the rack of the coordinator, assuming the client will load balance between them.
+            // However it doesnt work like that.
+            // AFAIK there can only be one coordinator per unique `FindCoordinatorResponse::key`.
+            // In fact the java driver doesnt even support multiple coordinators of different types yet:
```
benbromhead, line 1233: Is this something that needs a test case, todo or open issue to catch once the driver does support multiple coordinators of a different type?

Is there a win in supporting this when the java driver does or should we just fix and move on?

benbromhead (approved):
Above comment can be addressed outside of this PR - I'm happy with the approach otherwise.

shotover/src/transforms/kafka/sink_cluster/mod.rs
rukai: I believe the code I've written should handle multiple coordinators of different types.
If the java driver starts making use of that and my implementation is broken somehow, we'll know about it because it'll be broken.

PR #1616, opened 2024-05-08
Update aws_config

conorbros:
Caught this in another PR

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -151,9 +151,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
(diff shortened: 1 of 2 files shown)


rukai:
https://docs.rs/aws-config/latest/aws_config/struct.BehaviorVersion.html#method.latest maybe we should just use `latest`?
From the description in the docs, all versions are compatible, it just changes some performance characteristics. And we certainly arent "tuning to rust SDK behaviors" as they say.

rukai (approved):
Lgtm

PR #1624, opened 2024-05-14
pretty assert

conorbros:
Seeing if there is interest in using this library. An example can be seen in the logs.

### DIFF
--- a/test-helpers/Cargo.toml
+++ b/test-helpers/Cargo.toml
@@ -39,3 +39,4 @@ http = "1.1.0"
 rustls = { version = "0.21.2" ,features = ["dangerous_configuration"] }
 rustls-pemfile = "1.0.2"
 tokio-tungstenite = { git = "https://github.com/shotover/tokio-tungstenite", features = ["rustls-tls-native-roots"], rev = "2f1cc11e491c2d0401d01a83f623623182784b3a" }
+pretty_assertions.workspace = true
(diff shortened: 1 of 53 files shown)


rukai:
Woah, I love this!
Lets set it to a dev-dependency instead of regular dependency and land this!

rukai:
Is the test failure intentional?

rukai (approved):
nice!

test-helpers/Cargo.toml
```diff
@@ -39,3 +39,4 @@ http = "1.1.0"
 rustls = { version = "0.21.2" ,features = ["dangerous_configuration"] }
 rustls-pemfile = "1.0.2"
 tokio-tungstenite = { git = "https://github.com/shotover/tokio-tungstenite", features = ["rustls-tls-native-roots"], rev = "2f1cc11e491c2d0401d01a83f623623182784b3a" }
+pretty_assertions = "1.4.0"
```
rukai, line 42: may as well make this a dev-dependency

rukai (changes requested):
noticed some issues

test-helpers/Cargo.toml
conorbros: It's in the "test-helpers" module which means it's already only included as dev-dependency.

rukai (approved):

ronycsdu (approved):

Issue #1618, opened 2024-05-09
KafkaSinkCluster: authorize_scram_over_mtls phase 2


rukai:
https://github.com/shotover/shotover-proxy/pull/1605 introduced basic functionality for authorize_scram_over_mtls.
However its not as efficient or as robust as we would like.
* We recreate a delegation token for every new incoming connection.
* We rely on a hardcoded 4 second wait to ensure the delegation token has propagated across the cluster.

To solve both of these I propose a background task to manage tokens similar to the topology task shared between all `CassandraSinkCluster` instances:

`KafkaSinkClusterConfig::get_builder` spins up a background task that receives requests for new tokens and then sends the tokens back across a oneshot.
The task will be responsible for requesting tokens over mTLS, waiting until the token is usable, and then finally sending the token details along the oneshot.
From the perspective of the transform the API would look like:

```rust
KafkaSinkCluster {
  token_request_tx: mpsc::Sender<TokenRequest>
  ..
}

struct TokenRequest {
   username: String,
   response: oneshot::Receiver<DelegationToken>,
}

pub struct DelegationToken {
    pub token_id: String,
    pub hmac: Vec<u8>,
}
```

PR #1626, opened 2024-05-14
scram_over_mtls - phase2 - caching task

rukai:
progress towards: https://github.com/shotover/shotover-proxy/issues/1618

This PR is the first step towards a more robust scram_over_mtls implementation:
* It introduces a task that runs in the background and manages delegation tokens for all KafkaSinkCluster instances.
* The task currently just implements caching of tokens across all connections.

In the future the task will also be used for:
* detecting when tokens are ready to be used without a hardcoded sleep.
* recreating tokens before they expire
* prefetching all known tokens at startup

Since this task going down would leave the shotover instance in a very broken state:
* unwrap/panic is never used, instead an error is returned for all error cases.
* The task is wrapped in a loop that will restart the task from scratch if it encounters an error.

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/scram_over_mtls.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/scram_over_mtls.rs
@@ -1,5 +1,6 @@
 use super::node::{ConnectionFactory, KafkaAddress};
 use crate::{
+    connection::SinkConnection,
     frame::{
         kafka::{KafkaFrame, RequestBody, ResponseBody},
         Frame,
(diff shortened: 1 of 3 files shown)


shotover/src/transforms/kafka/sink_cluster/scram_over_mtls.rs
```diff
@@ -12,11 +13,92 @@ use kafka_protocol::{
...
+    tx: mpsc::Sender<TokenRequest>,
+}
+
+impl TokenTask {
+    #[allow(clippy::new_without_default)]
+    pub fn new(
+        mtls_connection_factory: ConnectionFactory,
+        mtls_port_contact_points: Vec<String>,
+    ) -> TokenTask {
+        let (tx, mut rx) = mpsc::channel::<TokenRequest>(100);
```
benbromhead, line 39: Why 100?

shotover/src/transforms/kafka/sink_cluster/scram_over_mtls.rs
```diff
@@ -12,11 +13,92 @@ use kafka_protocol::{
...
+                // Must be unauthed since mTLS is its own auth.
+                .create_connection_unauthed(&address)
+                .await?;
+
+            let token =
+                create_delegation_token_for_user(&mut connection, request.username.clone()).await?;
+
+            // TODO: This sleep is currently load bearing...
+            //       Need to delay progression until token has propagated.
+            tokio::time::sleep(std::time::Duration::from_secs(4)).await;
```
benbromhead, line 91: Don't need to address in this PR, but is there a better way to detect token propagation

rukai: yep, figuring that out is my next priority

shotover/src/transforms/kafka/sink_cluster/scram_over_mtls.rs
rukai: I just picked the value at random.
An alternative would be to make it unlimited but I feel having some kind of limit in place would be better off.
This code is only called on connection creation so the only scenario I can see it coming into place is if a client startsup and creates a whole bunch of connections at once.

I think I'll bump the value to 1000, since processing only 100 incoming connections at once does feel possibly a bit low.

benbromhead (approved):
Happy with this as it stands... I think the interesting code will be later and the approach makes sense

cjrolo (approved):

Issue #1585, opened 2024-04-22
KafkaSinkCluster IncrementalFetchRequests/session_id with multiple partitions


rukai:
important context: https://cwiki.apache.org/confluence/display/KAFKA/KIP-227%3A+Introduce+Incremental+FetchRequests+to+Increase+Partition+Scalability

shotover routes a fetch request.
the client sends another fetch request with the same session_id but a different partition.
shotover reused the old routing (as it should) by looking up the session_id but the route is actually invalid since the client requested a partition not on the broker who created that session_id.

The problem is that to the client multiple kafka brokers are hidden by a single shotover instance.
The client believes it can reuse a single session_id forever since it believes there is only one broker. (the shotover instance)
In reality however it needs to use multiple session_id's one for each backing broker.
We cant make the client do that, so instead our options are:
* Instruct users of shotover to disable kafka from creating session_id's (undesirable)
* Translate session_id's between shotover session_id's and kafka session_id's (complex)

To achieve translation we need to reimplement a large chunk of kafka's session_id logic.
We should be able to split up the implementation into 2 phases.

## Phase 0 - shotover <-> kafka (required)
make requests routable

### Request translation
Split requests such that they can be routed to the correct partitions.

### Response translation
Recombine requests into a single response
set session_id to 0

## Phase 1 - shotover <-> kafka (optional, reduces network traffic + broker cpu usage)
make requests smaller and prevent broker from recreating session_id's by utilizing knowledge of session_id's

### Request translation
remove state from a fetch request that is implicit according to stored session_id state

## Phase 2 - client <-> shotover (optional, reduces network traffic)
Implement session_id emulation in shotover.

### Request translation
reintroduce implicit state from session_id into the fetch request

### Response translation
set session_id to shotover value

rukai:
Me and conor had a discussion about this.
We decided:
* phase 1 should be altered
    + its too difficult for shotover to make good use of session_id's.
        + the client has context around what each session id will be used for. Once shotover receives a request that context is lost and its very expensive for shotover to determine which session id to make use of without that context.
        + so its not clear that implementing the current phase 1 design would actually bring any value.
        + Instead we will force shotover to disable session_id by setting session_id to -1 in the fetch request.
        + Alternatively we could just request that kafka be configured with session_id's disabled, but we will evaluate that after seeing the performance of the session_id = -1 approach.
* phase 2 is still worth attempting as described.

PR #1636, opened 2024-05-23
KafkaSinkCluster disable session creation

conorbros:
Completes Phase 1 in https://github.com/shotover/shotover-proxy/issues/1585

To improve the performance of `FetchRequests`, Kafka allows clients to create a "session" so the broker doesn't send unnecessary information over the network, i.e. topics or partitions that have unchanged since the last time the client received information about them.

This is done with a `session_id` that is sent with an `IncrementalFetchRequest` by the client to let the broker know it is part of a particular session.

`KafkaSinkCluster` currently sets the `session_id` to 0 before sending to the routed node because it's possible for Shotover to send a request with a `session_id` to a different broker to the one the session was created on, causing a `SessionIdNotFound` error.

Setting the `session_id` to 0 has the side effect of creating a new session for each fetch request to a broker. This is expensive for CPU on the broker. 

As seen on the table [here](https://cwiki.apache.org/confluence/display/KAFKA/KIP-227%3A+Introduce+Incremental+FetchRequests+to+Increase+Partition+Scalability#KIP227:IntroduceIncrementalFetchRequeststoIncreasePartitionScalability-FetchRequestMetadatameaning), the `session_epoch` can be set to `-1` to inform the broker not to create a session.

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/mod.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/mod.rs
@@ -502,6 +502,9 @@ impl KafkaSinkCluster {
                         self.store_topic_names(&mut topic_names, topic.topic.clone());
                         self.store_topic_ids(&mut topic_ids, topic.topic_id);
                     }
+                    fetch.session_id = 0;
+                    fetch.session_epoch = -1;
+                    request.invalidate_cache();
                 }
                 Some(Frame::Kafka(KafkaFrame::Request {
                     body:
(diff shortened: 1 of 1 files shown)


rukai (approved):
Nice!
This should be a good performance improvement and is also simpler than what we had before.

I dont think any of our existing integration benchmarks would really push this scenario hard enough to notice the benefit.
But its pretty clear that this will reduce load on the kafka instance itself due to making it do less work.
So I'm happy to land this PR without introducing such benchmarks.

For the next reviewer, make sure to set "hide whitespace".

justinweng-instaclustr (approved):

Issue #1526, opened 2024-03-13
KafkaSinkCluster is rack aware


rukai:
KafkaSinkCluster needs to follow the same rack aware model as CassandraSinkCluster as we want to use it for the same use case.

## Extended shotover nodes config

The config needs to be extended to:
```yaml
        - KafkaSinkCluster:
            shotover_nodes:
              - address: "172.16.1.100:9042"
                rack: "rack0"
                broker_id: 0
              - address: "172.16.1.101:9042"
                rack: "rack1"
                broker_id: 1
              - address: "172.16.1.102:9042"
                rack: "rack2"
                broker_id: 2
            local_shotover_broker_id: 0
            first_contact_points: ["172.16.1.2:9092"]
            connect_timeout_ms: 3000
```

## Rack aware routing 

We already have fairly good routing in place, requests are sent to the nodes that can handle them.
But the routing needs to be adjusted to favor nodes within the rack that the shotover node is representing.

The rack of each broker can be obtained from the Metadata queries that we are already processing.

## Message rewriting

And the message rewriting needs to be completely redone.
* the Metadata message needs to list all of the shotover nodes.
    + currently it just rewrites addresses to known shotover nodes and deletes duplicates - this turned out to be very fragile and I'll be glad to be rid of it.
* The DescribeCluster message needs to list all of the shotover nodes
* the FindCoordinator message needs to find a shotover node to report the coordinator as by fetching the rack of the coordinator and then mapping that to a shotover node.

PR #1639, opened 2024-05-29
KafkaSinkCluster: increment out_of_rack_requests metric for all out of rack requests

rukai:
closes https://github.com/shotover/shotover-proxy/issues/1526

I thought the `out_of_rack_requests` metric would need to be reported at the routing stage.
Which is why I left it for a follow up PR, I thought it was going to require modifications to all of our `route_to_*` methods.
But it turns out its far simpler to just report it at the sending stage, which is what this PR does.
We just compare the rack of the node we are about to send against shotover's rack and record the metric if there is a mismatch.

I left a `tracing::debug` at the previous location of `self.out_of_rack_requests.increment(1);` to help disambiguate why we are getting out of rack requests.

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/mod.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/mod.rs
@@ -753,7 +753,9 @@ impl KafkaSinkCluster {
                         {
                             node.broker_id
                         } else {
-                            self.out_of_rack_requests.increment(1);
+                            tracing::debug!(
+                                "Routing fetch request to replica outside of shotover's rack"
+                            );
                             self.nodes
                                 .iter_mut()
                                 .filter(|node| {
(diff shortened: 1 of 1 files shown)


conorbros (approved):

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1008,17 +1010,28 @@ impl KafkaSinkCluster {
...
-                    &self.sasl_mechanism,
-                )
-                .await?
-                .send(requests.requests)?;
+                .unwrap();
+
+            if node
+                .rack
+                .as_ref()
+                .map(|rack| rack == &self.rack)
```
rukai, line 1022: !=

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1008,17 +1010,28 @@ impl KafkaSinkCluster {
...
+                .unwrap_or(true)
```
rukai, line 1023: change to false

PR #1642, opened 2024-05-30
Make TLS connection builders take references to configs instead of cloning

### LINKED ISSUE #1619: Make TlsConnector::new + TlsAcceptor::new take references to the config
`TlsConnector::new` and `TlsAcceptor::new`  both take a `TlsConnectorConfig`/`TlsAcceptorConfig` but they could easily take a `&TlsConnectorConfig`/`&TlsAcceptorConfig` instead.
So lets do that.

Once changed we will need to find all the locations that call `new` and swap the [Clone::clone()](https://doc.rust-lang.org/std/clone/trait.Clone.html#tymethod.clone) with an [Option::as_ref](https://doc.rust-lang.org/std/option/enum.Option.html#method.as_ref):
```diff
-        let tls = self.tls.clone().map(TlsConnector::new).transpose()?;
+        let tls = self.tls.as_ref().map(TlsConnector::new).transpose()?;
```
Here, `Option::as_ref` takes an `&Option<TlsConnectorConfig>` and returns it as `Option<&TlsConnector>`.
This is important because we need the value that the function in `.map(..)` is given to be `&TlsConnector` instead of `TlsConnector`.

If this were a code path that every message took this would be a nice performance improvement.
In reality this wont make shotover any faster since the code path only occurs during shotover's initialization, so it will only make initialization a tiny bit faster.
But its a nice isolated introduction to working with borrows vs clones and makes the code follow patterns used in the rest of the codebase.

justinweng-instaclustr:
This PR changes the TLS connection builders (i.e., `TlsConnector::new` and `TlsAcceptor::new`) to use references to the TLS configs (i.e., `TlsConnectorConfig` and `TlsAcceptorConfig`) rather than clone the TLS config objects, which in theory should improve shotover performance to a slight extent.

Closes #1619.

### DIFF
--- a/shotover-proxy/tests/cassandra_int_tests/cluster/mod.rs
+++ b/shotover-proxy/tests/cassandra_int_tests/cluster/mod.rs
@@ -22,7 +22,7 @@ pub async fn run_topology_task(ca_path: Option<&str>, port: Option<u32>) -> Vec<
     let (keyspaces_tx, _keyspaces_rx) = watch::channel(HashMap::new());
     let (task_handshake_tx, task_handshake_rx) = mpsc::channel(1);
     let tls = ca_path.map(|ca_path| {
-        TlsConnector::new(TlsConnectorConfig {
+        TlsConnector::new(&TlsConnectorConfig {
             certificate_authority_path: ca_path.into(),
             certificate_path: None,
             private_key_path: None,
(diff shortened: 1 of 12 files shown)


rukai (approved):
LGTM! Well done on your first issue

ronycsdu (approved):

PR #1644, opened 2024-05-31
#1643 Fix cassandra_int_tests::passthrough_cassandra_down on macOS

### LINKED ISSUE #1643: cassandra_int_tests::passthrough_cassandra_down fails on macos
failure:
image

This is the same issue as was resolved for `passthrough_redis_down` in https://github.com/shotover/shotover-proxy/pull/1617
You should be able to use the same solution as done in that PR, in fact you should be able to reuse the CONNECTION_REFUSED_OS_ERROR constant that was introduce in that PR.

justinweng-instaclustr:
Similar to https://github.com/shotover/shotover-proxy/pull/1617, this PR fixes the OS error code in `cassandra_int_tests::passthrough_cassandra_down` for macOS.

All the error messages of `Connection refused (os error ...)` should be using the OS-dependent `CONNECTION_REFUSED_OS_ERROR` now.

All the passthrough tests now pass on macOS.

image

Closes #1643


rukai (approved):
nice!

conorbros (approved):

PR #1646, opened 2024-06-03
#1622 Upgrade cached and scylla

### LINKED ISSUE #1622: Update `scylla` and `cached` dependencies
We can keep track of outdated dependencies in the shotover repo by looking at [deps.rs](https://deps.rs/repo/github/shotover/shotover-proxy) or setting up and using the [cargo upgrades](https://crates.io/crates/cargo-upgrades) tool.
Of all the currently out of date dependencies we have the following which are trivially upgradable:
* `scylla` - One of the cassandra drivers used in our integration tests and benchmarks.
* `cached` - a macro for caching the results of function calls.

These dependencies can be trivially updated without changing the code, we just need to change the version used in the `Cargo.toml` files.

Cargo follows [semantic versioning](https://semver.org/).

In general, when updating dependencies, we need to change our `Cargo.toml` when a dependency has a new major version released.
The minor and patch versions will be managed by cargo automatically, which it keeps track of in the `Cargo.lock` file, we can instruct cargo to update the minor and patch versions by running `cargo update`. (You should do so after upgrading `scylla` and `cached`)

In the case of both `scylla` and `cached` they both have a major version of 0. i.e. the version looks like `0.x.y`
In rust, a major version of 0, means that the version is effectively shifted left by one.
The minor version takes on the role of the major version.
And the patch version takes on the role of the minor version.

Normally I would just include such straightforward upgrades in with the more involved upgrades, but I've made this a separate issue to serve as an introduction to dependency management with rust.

justinweng-instaclustr:
This PR upgrades `cached` from 0.50 to 0.51, and `scylla` from 0.12 to 0.13.

`cargo upgrades` don't show `cached` and `scylla` anymore.
image

Closes #1622

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,9 +18,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
+checksum = "6e4503c46a5c0c7844e948c9a4d6acd9f50cccb4de1c48eb9e291ea17470c678"
 dependencies = [
  "gimli",
 ]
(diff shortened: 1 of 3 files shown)


rukai (approved):
Thanks!

ronycsdu (approved):

PR #1655, opened 2024-06-05
#1611 Allow disabling Shotover metrics

### LINKED ISSUE #1611: Allow disabling metrics reporting
The main benefit is allowing us to determine the performance cost of collecting metrics is costing us.
Not to evaluate removing it, but to see if optimizing it could be useful.
But it also makes sense to allow disabling for potential users who dont make use of metrics.

Shotover has two config files:
* [topology.yaml](https://github.com/shotover/shotover-proxy/blob/79b4d1fca73a086d687f39137680b365ac76c7fd/shotover-proxy/config/topology.yaml) - the main config file, determines where to listen, where to connect and what rewriting our routing of messages needs to occur in between.
* [config.yaml](https://github.com/shotover/shotover-proxy/blob/79b4d1fca73a086d687f39137680b365ac76c7fd/shotover-proxy/config/config.yaml) - a secondary config file, defines side concerns such as metrics and logging.

The default `config.yaml` is quite simple and only has two fields:
```yaml
main_log_level: "info, shotover=info, shotover::connection_span=info"
observability_interface: "0.0.0.0:9001"
```

This issue proposes that we make the observability_interface field optional.
When its not provided, the observability interface aka metrics is disabled.
To ensure that users get observability enabled by default, the example `config.yaml` shipped with shotover will retain the field.

To make the field optional the type of the config field should be changed from `String` to `Option<String>`.
The function that enables metrics is `start_observability_interface`.

## Testing 
An integration test should be added to `observability_int_tests.rs` to test that shotover still works when metrics is disabled.
Use `passthrough_standard` from `redis_int_tests` as a base.
But instead of running the whole test suite, just check that a single message can pass through as a smoke test.
This smoke test will be similar to `redis::cmd`s in `test_metrics` but we should assert that they succeed instead of fail since we will have an actual redis instance running and we only need to send one message.

justinweng-instaclustr:
This PR makes the field `observability_interface` optional in `configuration.yaml`. The Shotover metrics reporting will be disabled when this field is not specified.

A test function has been added to verify Shotover can still process Redis requests with metrics disabled. As discussed with @rukai, we don't need to check that querying the prometheus endpoint returns the error of connection refused because the endpoint address is not specified at all.

Closes #1611

### DIFF
--- a/docs/src/user-guide/configuration.md
+++ b/docs/src/user-guide/configuration.md
@@ -8,7 +8,7 @@ and a topology file specified by `--topology-file`
 The configuration file is used to change general behavior of Shotover. Currently it supports two values:
 
 * `main_log_level`
-* `observability_interface`
+* `observability_interface` (optional)
 
 ### main_log_level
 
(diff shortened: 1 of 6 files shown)


rukai (approved):
LGTM!
Good thinking to update the docs.

conorbros (approved):

ronycsdu (approved):

PR #1658, opened 2024-06-11
#1650 RedisSinkCluster - Avoid recreating Counter

### LINKED ISSUE #1650: RedisSinkCluster - Avoid recreating `Counter`
The [counter!()](https://docs.rs/metrics/latest/metrics/macro.counter.html) macro is used to create a [Counter](https://docs.rs/metrics/latest/metrics/struct.Counter.html) instance which we can then call [increment(1)](https://docs.rs/metrics/latest/metrics/struct.Counter.html#method.increment) on to increment a metric reported in shotover's prometheus metrics.

Counter instances can be either:
* stored to be reused many times
* created, used, discarded on each use.

The end result of both approaches is identical.

However, the advantages of reusing Counters is:
* Better performance - creating a Counter can be expensive
* Easier maintenance - the metrics labels are defined in a single place

RedisSinkCluster contains a metric named `shotover_failed_requests_count`, we should create that metric in `RedisSinkCluster::new()`.
We already register the `shotover_failed_requests_count` counter metric in `RedisSinkCluster::new`, we just dont hold onto the `Counter` for later use.
So we should take the registration, and store the returned value in RedisSinkCluster.
To do this we will need to replace the call to `sink_cluster.get_name()` with just the `NAME` constant, which enables further cleanup in removing the get_name method. (just follow what the warnings tell you)

When implementing this PR, you can refer to how other transforms store and reuse Counter's: https://github.com/shotover/shotover-proxy/blob/a9ff5186d9348d25ed187846118d7dce1e36748a/shotover/src/transforms/kafka/sink_cluster/mod.rs#L190

justinweng-instaclustr:
This PR registers and stores the `Counter` instance in `RedisSinkCluster` instead of recreating the instance every time the method `send_error_response` is called.

Closes #1650

### DIFF
--- a/shotover-proxy/tests/redis_int_tests/mod.rs
+++ b/shotover-proxy/tests/redis_int_tests/mod.rs
@@ -45,6 +45,8 @@ async fn passthrough_standard() {
     shotover
         .shutdown_and_then_consume_events(&[invalid_frame_event()])
         .await;
+
+    assert_failed_requests_metric_is_incremented_on_error_response().await;
 }
 
 #[tokio::test(flavor = "multi_thread")]
@@ -323,9 +325,7 @@ async fn cluster_dr() {
     shotover.shutdown_and_then_consume_events(&[]).await;
 }
 
-#[tokio::test(flavor = "multi_thread")]
-async fn test_failed_requests_metric_sink_single() {
-    let _compose = docker_compose("tests/test-configs/redis/passthrough/docker-compose.yaml");
+pub async fn assert_failed_requests_metric_is_incremented_on_error_response() {
     let shotover = shotover_process("tests/test-configs/redis/passthrough/topology.yaml")
         .start()
         .await;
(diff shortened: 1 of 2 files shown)


shotover-proxy/tests/redis_int_tests/mod.rs
```diff
@@ -321,3 +322,54 @@ async fn cluster_dr() {
...
+
+    redis::cmd("INVALID_COMMAND")
+        .arg("foo")
+        .query_async::<_, ()>(&mut connection)
+        .await
+        .unwrap_err();
+
+    // Redis client driver initialization sends 2 CLIENT SETINFO commands which trigger 2 errors
+    // because those commands are not available in the currently used redis version.
+    assert_metrics_key_value(
```
rukai, line 342: I think we should move these assertions into assertion methods and then call them from redis_passthrough and redis_cluster_handling.
e.g. define a method named `assert_failed_requests_metric_is_incremented_on_error_response` that contains lines 329 to 348. We should include creation of shotover in the assertion since we need a fresh shotover to be able to make assertions on the state of the counter.
Alternatively we could avoid creating shotover in the assertion method and create a 2nd shotover in redis_passthrough/redis_cluster_handling.
I can see arguments for both approaches so I dont really mind which one we take.

The main reason is to reduce the time it takes to run the test suite.
On my machine, it takes 1 second to spin up `redis/passthrough/docker-compose.yaml` and 5 seconds to spin up `redis/cluster-handling/docker-compose.yaml`.
So by combining these cases into `redis_passthrough` and `redis_cluster_handling` we cut off 6 seconds from the test suite run time.
In isolation 6 seconds isnt much, but if every PR added 6 seconds we would have a problem.

The implementation within shotover looks good btw!

justinweng-instaclustr: Cool will do

justinweng-instaclustr: Just moved the test code into a new function `assert_failed_requests_metric_is_incremented_on_error_response()` to avoid spinning up a new redis container

rukai (approved):
nice!

ronycsdu (approved):

PR #1659, opened 2024-06-12
KafkaSinkCluster - routing error handling

rukai:
The kafka cluster has a lot of state that shotover must keep track of for routing purposes.
This includes things like which brokers are holding which partitions.
Currently shotover populates its internal records of this state but it has no way to invalidate these records.
This PR implements this missing invalidation.

To handle changes to the cluster we need to handle routing errors as these indicate that the cluster has changed. 

This PR handles the 3 kinds of routing errors as follows: 
* NOT_CONTROLLER set controller_broker to BrokerId(-1) 
* NOT_COORDINATOR remove group from group_to_coordinator_broker 
* NOT_LEADER_OR_FOLLOWER 
  * remove topic from topic_by_name and topic_by_id 
  * alternatively, if supported on this api version, immediately update topic_by_name/topic_by_id as per [KIP-951](https://cwiki.apache.org/confluence/display/KAFKA/KIP-951%3A+Leader+discovery+optimisations+for+the+client) 
    * if the produce response is NOT_LEADER_OR_FOLLOWER and includes a newer leader epoch then we can update the topic entry with the provided broker ids 

This PR ensures that every request type that we perform routing also has a response handler that invalidates the routing state when we get a routing error.

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/mod.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/mod.rs
@@ -216,7 +216,11 @@ impl TransformBuilder for KafkaSinkClusterBuilder {
             first_contact_node: None,
             control_connection: None,
             pending_requests: Default::default(),
+            // TODO: this approach with `find_coordinator_requests` and `routed_to_coordinator_for_group`
+            //       is prone to memory leaks and logic errors.
+            //       We should replace these fields with extra state within `PendingRequestTy::Received/Sent`.
             find_coordinator_requests: Default::default(),
+            routed_to_coordinator_for_group: Default::default(),
             temp_responses_buffer: Default::default(),
             sasl_mechanism: None,
             authorize_scram_over_mtls: self.authorize_scram_over_mtls.as_ref().map(|x| x.build()),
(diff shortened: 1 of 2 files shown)


shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1234,6 +1243,108 @@ routing message to a random node so that:
...
+                    ..
+                })) => {
+                    for (topic_name, topic) in &produce.responses {
+                        for response_partition in &topic.partition_responses {
+                            if let Some(ResponseError::NotLeaderOrFollower) =
+                                ResponseError::try_from_code(response_partition.error_code)
+                            {
+                                if response_partition.current_leader.leader_id != -1 {
+                                    // The broker has informed us who the new leader is, we can just directly update the leader
+                                    if let Some(mut topic) = self.topic_by_name.get_mut(topic_name)
```
rukai, line 1257: rename topic to avoid shadowing.

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1234,6 +1243,108 @@ routing message to a random node so that:
                 })) => {
                     self.process_sasl_authenticate(authenticate).await?;
                 }
+                Some(Frame::Kafka(KafkaFrame::Response {
```
rukai, line 1246: Add logging for all cluster state changes

justinweng-instaclustr (approved):

ronycsdu (approved):

PR #1660, opened 2024-06-12
KafkaSinkCluster: use leader_epoch

rukai:
We need to change our metadata response processing to store each partitions leader_epoch and make use of it when updating our partitions metadata state. 
We should only update a partition when the leader epoch in the response is newer (a smaller integer) than the existing epoch. This ensures that if we get the metadata from a node that is a bit out of date we dont regress the state of the metadata.
However in order to enable cleaning up of deleted partitions we should still delete partitions if missing in the response, since otherwise there would be no way to clean these up.

This PR implements this.

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/mod.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/mod.rs
@@ -1371,7 +1371,7 @@ routing message to a random node so that:
             if ResponseError::try_from_code(topic.error_code).is_none() {
                 // We use the response's partitions list as a base
                 // since if it has deleted an entry then we also want to delete that entry.
-                let mut base_partitions: Vec<_> = topic
+                let mut new_partitions: Vec<_> = topic
                     .partitions
                     .iter()
                     .map(|partition| Partition {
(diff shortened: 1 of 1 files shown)


shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1389,18 +1392,45 @@ routing message to a random node so that:
...
+                // If topic_by_name contains any partitions with a more recent leader_epoch use that instead.
+                // The out of date epoch is probably caused by requesting metadata from a broker that is slightly out of date.
+                // We use topic_by_name instead of topic_by_id since its always used regardless of protocol version.
+                if let Some(topic) = self.topic_by_name.get(topic_name) {
+                    for partition in &topic.partitions {
+                        if let Some(base_partition) = base_partitions
+                            .iter_mut()
+                            .find(|p| p.index == partition.index)
+                        {
+                            if partition.leader_epoch > base_partition.leader_epoch {
```
rukai, line 1406: partition -> old_partition

ronycsdu (approved):

justinweng-instaclustr (approved):

Issue #1649, opened 2024-06-03
Update metrics + metrics-exporter-prometheus


rukai:
These two crates must be upgraded together since metrics-exporter-prometheus exposes types from metrics in its public API that we use.

However, the new version of metrics has improved its diagnostics such that unused diagnostics are shown as compiler warnings. 
We have 2 such instances of these warnings.
![image](https://github.com/shotover/shotover-proxy/assets/5120858/f09dc6c2-d64d-4d26-838e-48893a6a7cc4)

Since CI will prevent merging with warnings, we need to either ignore the warnings or fix them.
I can see that our usages that are getting warned do have an associated performance cost that we should try to avoid.
* For RedisSinkCluster
  + we can easily store the Counter in the transform instance to avoid recreating it. https://github.com/shotover/shotover-proxy/issues/1650
* For QueryCounter we can:
  + leak the counter_name so we store it as `&'static str` - this will definitely be faster.
  + store every `Counter` we create in a `HashMap<String, Counter>` this should be faster, we should measure it with a microbenchmark though.

But also I think there are legitimate uses of creating a metric without using it so I've made a comment to see what upstream thinks: https://github.com/metrics-rs/metrics/pull/475#issuecomment-2144361030

TODO, all in separate PRs:
- [x] https://github.com/shotover/shotover-proxy/issues/1650
- [x] Write a microbenchmark for QueryCounter: https://github.com/shotover/shotover-proxy/pull/1663
   * Add the bench to `shotover/benches/benches/chain.rs`
   * Copy `redis_filter` or `loopback` as a base for the new bench.
   * The bench should have a chain consisting of `QueryCounter` followed by `Loopback`. This is because QueryCounter ignores the responses and Loopback is the fastest transform for when we dont care about the value of responses.
   * The bench can send in redis requests as they are simplest and fastest to work with.
- [x] QueryCounter - `String::leak()` the counter_name so we store it as `&'static str` - this will definitely be faster, run the benchmark to double check. https://github.com/shotover/shotover-proxy/pull/1669
   * leaking memory sounds bad but its actually fine since we only create this value once for the lifetime of shotover. And leaking the memory gives us 2 advantages:
      + we never have to pay the cost of freeing its allocation when dropping it
      + we can pass it around as a reference without having to worry about lifetimes, this reduces cloning.
[...]

PR #1663, opened 2024-06-14
#1649 Add a microbenchmark for QueryCounter

justinweng-instaclustr:
This PR adds a benchmark function for QueryCounter which includes a chain of QueryCounter followed by Loopback. It sends several redis commands to trigger the increment of the QueryCounter metric.

Progress towards #1649


rukai:
for the regression on the `encode_system.local_result_v5_no_compression` benchmark, I think that changing the benchmarks has just added some noise by making the benchmark executable compile slightly differently or leave the heap in a slightly different state.

So dont worry about it.

rukai (approved):
looks good!

ronycsdu (approved):

PR #1669, opened 2024-06-17
#1649 Leak query counter name to improve QueryCounter performance

justinweng-instaclustr:
Stores QueryCounter's `counter_name` as `&'static str` (i.e., leaks the `counter_name` string so that its memory allocation is not freed for the lifetime of the shotover process. This should improve shotover performance as it reduces cloning.

We can observe minor improvement in query_counter benchmark regarding allocating and freeing memory.
image

Progress towards #1649.

### DIFF
--- a/shotover/src/transforms/query_counter.rs
+++ b/shotover/src/transforms/query_counter.rs
@@ -29,6 +29,7 @@ impl QueryCounter {
         counter!("shotover_query_count", "name" => counter_name.clone());
 
         QueryCounter {
+            // Leaking here is fine since the builder is created only once during shotover startup.
             counter_name: counter_name.leak(),
         }
     }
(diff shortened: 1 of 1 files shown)


shotover/src/transforms/query_counter.rs
```diff
@@ -28,7 +28,9 @@ impl QueryCounter {
     pub fn new(counter_name: String) -> Self {
         counter!("shotover_query_count", "name" => counter_name.clone());
 
-        QueryCounter { counter_name }
+        QueryCounter {
+            counter_name: counter_name.leak(),
```
rukai, line 32: Lets add a comment to reassure readers that leaking is fine in this case. Something like:
```rust
// Leaking here is fine since the builder is created only once during shotover startup.
```

rukai (approved):
Nice!

ronycsdu (approved):

PR #1671, opened 2024-06-19
#1649 Reuse query counters

justinweng-instaclustr:
This PR stores the query counters in a HashMap and reuses them when processing messages instead of registering query counters in every transform. According to the benchmark report, `query_counter_pre_used` gains 14% improvement thanks to the faster `transform`.
image

`query_counter_fresh` becomes a bit slower but it's expected because the first transform runs create the counters and insert them into the hashmap.
image

Progress towards #1649

### DIFF
--- a/shotover/src/transforms/query_counter.rs
+++ b/shotover/src/transforms/query_counter.rs
@@ -9,6 +9,7 @@ use metrics::counter;
 use metrics::Counter;
 use serde::Deserialize;
 use serde::Serialize;
+use std::borrow::Cow;
 use std::collections::HashMap;
 
 use super::DownChainProtocol;
@@ -18,7 +19,7 @@ use super::UpChainProtocol;
 #[derive(Clone)]
 pub struct QueryCounter {
     counter_name: &'static str,
-    counters: HashMap<String, Counter>,
+    query_to_counter: HashMap<String, Counter>,
 }
 
 #[derive(Serialize, Deserialize, Debug)]
@@ -29,26 +30,22 @@ pub struct QueryCounterConfig {
 
 impl QueryCounter {
     pub fn new(counter_name: String) -> Self {
+        // Leaking here is fine since the builder is created only once during shotover startup.
         let counter_name_ref: &'static str = counter_name.leak();
-        let mut counters = HashMap::new();
-        let query_counter = counter!("shotover_query_count", "name" => counter_name_ref);
-        counters.insert("shotover_query_count".to_string(), query_counter);
+
+        // Although not incremented, this counter needs to be created to ensure shotover_query_count is 0 on shotover startup.
+        counter!("shotover_query_count", "name" => counter_name_ref);
 
         QueryCounter {
-            // Leaking here is fine since the builder is created only once during shotover startup.
             counter_name: counter_name_ref,
-            counters,
+            query_to_counter: HashMap::new(),
         }
     }
 
-    fn increment_counter(&mut self, query: String, query_type: &'static str) {
-        if self.counters.contains_key(query.as_str()) {
-            self.counters.get(query.as_str()).unwrap().increment(1);
-        } else {
-            let query_counter = counter!("shotover_query_count", "name" => self.counter_name, "query" => query.clone(), "type" => query_type);
-            query_counter.increment(1);
-            self.counters.insert(query, query_counter);
-        }
+    fn increment_counter(&mut self, query: Cow<str>, query_type: &'static str) {
+        self.query_to_counter.entry(query.to_string())
+            .or_insert_with(|| counter!("shotover_query_count", "name" => self.counter_name, "query" => query.to_string(), "type" => query_type))
+            .increment(1);
     }
 }
 
@@ -74,20 +71,20 @@ impl Transform for QueryCounter {
                 #[cfg(feature = "cassandra")]
                 Some(Frame::Cassandra(frame)) => {
                     for statement in frame.operation.queries() {
-                        self.increment_counter(statement.short_name().to_string(), "cassandra");
+                        self.increment_counter(Cow::from(statement.short_name()), "cassandra");
                     }
                 }
                 #[cfg(feature = "redis")]
                 Some(Frame::Redis(frame)) => {
                     if let Some(query_type) = crate::frame::redis::redis_query_name(frame) {
-                        self.increment_counter(query_type, "redis");
+                        self.increment_counter(Cow::from(query_type), "redis");
                     } else {
-                        self.increment_counter("unknown".to_string(), "redis");
+                        self.increment_counter(Cow::from("unknown"), "redis");
                     }
                 }
                 #[cfg(feature = "kafka")]
                 Some(Frame::Kafka(_)) => {
-                    self.increment_counter("unknown".to_string(), "kafka");
+                    self.increment_counter(Cow::from("unknown"), "kafka");
                 }
                 Some(Frame::Dummy) => {
                     // Dummy does not count as a message
@@ -97,7 +94,7 @@ impl Transform for QueryCounter {
                     todo!();
                 }
                 None => {
-                    self.increment_counter("unknown".to_string(), "none");
+                    self.increment_counter(Cow::from("unknown"), "none");
                 }
             }
         }
--- a/shotover/src/transforms/query_counter.rs
+++ b/shotover/src/transforms/query_counter.rs
@@ -6,8 +6,10 @@ use crate::transforms::{Transform, TransformBuilder, Wrapper};
 use anyhow::Result;
 use async_trait::async_trait;
 use metrics::counter;
+use metrics::Counter;
 use serde::Deserialize;
 use serde::Serialize;
+use std::collections::HashMap;
 
 use super::DownChainProtocol;
 use super::TransformContextConfig;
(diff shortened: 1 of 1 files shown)


shotover/src/transforms/query_counter.rs
```diff
@@ -26,11 +29,36 @@ pub struct QueryCounterConfig {
...
         QueryCounter {
             // Leaking here is fine since the builder is created only once during shotover startup.
-            counter_name: counter_name.leak(),
+            counter_name: counter_name_ref,
+            counters,
+        }
+    }
+
+    fn increment_counter(&mut self, query: &'static str, query_type: &'static str) {
+        let query_counter_key = format!(
```
rukai, line 48: We should be able to use `(&'static str, &'static str)` as a key instead of constructing a new string

justinweng-instaclustr: In that case, how can we differentiate between the counters with `query=unknown`?

rukai: `query=unknown` is using the query field which is the field I am proposing we keep, so it should work fine.

shotover/src/transforms/query_counter.rs
```diff
@@ -16,6 +18,7 @@ use super::UpChainProtocol;
 #[derive(Clone)]
 pub struct QueryCounter {
     counter_name: &'static str,
+    counters: HashMap<&'static str, Counter>,
```
rukai, line 21: HashMap should be a `HashMap<String, Counter>`.
I previously said it should be `HashMap<&'static str, Counter>` but that was incorrect.
I saw that `.short_name()` returned an &'static str for the cassandra logic.
But I missed that `redis_query_name` returned a String for the redis logic.

Using `String` instead of &'static str will increase the cost of inserting an element, but it will not affect the cost of retrieving an element which is where we are most interested.

shotover/src/transforms/query_counter.rs
rukai: We still want to include the "type" field when we create the counter.
However we do not need to worry about different type values colliding in the hashmap since a single transform can only process a single protocol. e.g. it will process only redis message or only kafka messages.

The only exception to this is `Frame::Dummy` which can be sent by any protocol but does not record to a counter since its a shotover implementation detail.

shotover/src/transforms/query_counter.rs
```diff
@@ -26,11 +29,25 @@ pub struct QueryCounterConfig {
...
 
         QueryCounter {
             // Leaking here is fine since the builder is created only once during shotover startup.
-            counter_name: counter_name.leak(),
+            counter_name: counter_name_ref,
+            counters,
+        }
+    }
+
+    fn increment_counter(&mut self, query: String, query_type: &'static str) {
```
rukai, line 44: To efficiently handle both owned and reference cases we can split this into:
```rust
fn increment_counter(&mut self, query: String, query_type: &'static str)
fn increment_counter_ref(&mut self, query: &str, query_type: &'static str)
```
This way if we are using the cassandra protocol we can completely avoid cloning the string when fetching from the hashmap since query is an `&str`.
While at the same time we can avoid an unncessesary clone for redis where query is a `String`.

## Alternatively
Or we could use [Cow](https://doc.rust-lang.org/std/borrow/enum.Cow.html) type to handle both cases from the one function:
```rust
fn increment_counter(&mut self, query: Cow<str>, query_type: &'static str) { .. }

self.increment_counter(Cow::borrowed("foo"), "kafka")
self.increment_counter(Cow::owned("foo".to_string()), "kafka")
```

The cow solution is probably nicer, but duplicating the function is probably easier.
I would happily accept either solution.

shotover/src/transforms/query_counter.rs
```diff
@@ -26,11 +29,25 @@ pub struct QueryCounterConfig {
...
+        if self.counters.contains_key(query.as_str()) {
```
rukai, line 45: The hashmap provides an API to:
*  fetch the value if it exists
* otherwise reuse create, insert and return a value

It used by calling [entry](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry) and then calling [https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_insert_with](Entry::or_insert_with)
Which will look like:
```rust
self.counters
  .entry(query.as_str())
  .or_insert_with(|| counter!("shotover_query_count", "name" => self.counter_name, "query" => query.clone(), "type" => query_type))
  .increment(1);
```

This should be:
* simpler - its now a one liner
* faster - we only have to perform the hashmap lookup once

shotover/src/transforms/query_counter.rs
```diff
@@ -26,11 +29,25 @@ pub struct QueryCounterConfig {
 
 impl QueryCounter {
     pub fn new(counter_name: String) -> Self {
-        counter!("shotover_query_count", "name" => counter_name.clone());
+        let counter_name_ref: &'static str = counter_name.leak();
+        let mut counters = HashMap::new();
+        let query_counter = counter!("shotover_query_count", "name" => counter_name_ref);
+        counters.insert("shotover_query_count".to_string(), query_counter);
```
rukai, line 35: We dont actually need to insert this into counters since we never increment this exact counter.
We do want to continue creating it though because it ensures that shotover_query_count shows up with 0 on shotover startup, before any messages have passed through.

shotover/src/transforms/query_counter.rs
```diff
@@ -16,6 +18,7 @@ use super::UpChainProtocol;
 #[derive(Clone)]
 pub struct QueryCounter {
     counter_name: &'static str,
+    counters: HashMap<String, Counter>,
```
rukai, line 21: Can we name this something like `query_to_counter` to indicate the relation between the key and value?
I find this especially important when we have a string or integer type as the key, since in that case its not clear what kind of value the key actually is.

shotover/src/transforms/query_counter.rs
justinweng-instaclustr: This is handy!

shotover/src/transforms/query_counter.rs
justinweng-instaclustr: I'll try to use Cow

shotover/src/transforms/query_counter.rs
```diff
@@ -26,13 +30,23 @@ pub struct QueryCounterConfig {
...
         QueryCounter {
-            // Leaking here is fine since the builder is created only once during shotover startup.
-            counter_name: counter_name.leak(),
+            counter_name: counter_name_ref,
+            query_to_counter: HashMap::new(),
         }
     }
+
+    fn increment_counter(&mut self, query: Cow<str>, query_type: &'static str) {
+        self.query_to_counter.entry(query.to_string())
```
justinweng-instaclustr, line 46: It's interesting that `HashMap.entry()` doesn't support lookup by reference. I didn't go through the entire discussion in https://github.com/rust-lang/rfcs/pull/1769 but it kind of makes sense as we may need to insert key so that it needs to be owned?

rukai: ah yeah, thats unfortunate.

Given that we need to convert the string to owned to use entry, the cow isnt actually giving us any advantage, we should revert the cow changes and go back to:
```rust
fn increment_counter(&mut self, query: String, query_type: &'static str) {
```

To do that, we also need to swap `or_insert_with` to `or_insert_with_key` so that we access the key while creating:
```rust
.or_insert_with_key(|query| counter!("shotover_query_count", "name" => self.counter_name, "query" => query.to_string(), "type" => query_type))
```

## Alternatively

We could alternatively keep the cow and revert the move to the `entry` API.
This should be faster since I expect allocating to be slower than hashing.
However its impossible to tell without implementing both approaches and benchmarking.

## But ultimately

But ultimately we are working on this to resolve a warning, so we shouldn't get too bogged down into these optimizations.
So I think just pick one of these solutions and go with it.
I'm happy with whichever one you choose.

rukai (approved):
nice work!
lets land this.

I think that if we ever want to further improve performance here, the best way would be to make redis_query_name return a new enum named `RedisCommandType` and then have a `to_str()` method on that enum.
That way we can make this work completely with `&'static str` and other usages of `redis_query_name` across the codebase would also be able to avoid the allocation.
But again, I think we've hit the sweetspot of how much time to spend on this problem since this will resolve the performance impacts that are being warned about in the new `metrics` release.

justinweng-instaclustr:
> nice work! lets land this.
> 
> I think that if we ever want to further improve performance here, the best way would be to make redis_query_name return a new enum named `RedisCommandType` and then have a `to_str()` method on that enum. That way we can make this work completely with `&'static str` and other usages of `redis_query_name` across the codebase would also be able to avoid the allocation. But again, I think we've hit the sweetspot of how much time to spend on this problem since this will resolve the performance impacts that are being warned about in the new `metrics` release.

Yeah sounds good. We can probably change redis_query_name to improve performance opportunistically in the future. Do we want to raise an issue for it just for tracking purpose?

rukai:
I think just leave it, better in general to let performance work be driven by observed issues in the system rather then guessing which parts of the system are going to end up as the bottleneck.
I consider this work an exception since we had a warning pointing out a potential issue, better to resolve it then just ignore it.

ronycsdu (approved):

PR #1676, opened 2024-06-25
Cassandra 5 vector support

rukai:
This PR fixes a bug causing shotover to error out when processing cassandra custom types.
In doing so we enable support for cassandra 5's new vector types.

The problem is that the cassandra-protocol crate will return an Error when it tries to parse a Custom cassandra value, this makes sense from the perspective of the crate since it does not know how to handle this type. If the user wants to use the custom type they should use a driver that knows about this type.
However from the shotover perspective its reasonable for the client and cassandra to use a custom type that shotover knows nothing about, in this case we should just pass on the bytes as is.

To achieve this we add an enum `CustomOrStandardType` which allows us to store either a Custom type (not supported by cassandra-protocol) or any other standard type supported by cassandra-protocol.
Long term I consider this a hack, but unless we have a push to improve cassandra performance I think we'll be living with this hack forever.

As demonstrated [here](https://github.com/rukai/cassandra_vector_bug/blob/main/src/main/java/com/mycompany/app/App.java) the java driver only supports 4 data types for use with vectors. So even though cassandra returns results for the other data types that could in theory be interpreted by a driver, we dont test those since there's no way to do that through the java driver. Shotover doesn't parse these anyway so in reality the coverage isnt important.

### DIFF
--- a/shotover-proxy/tests/cassandra_int_tests/collections/vector.rs
+++ b/shotover-proxy/tests/cassandra_int_tests/collections/vector.rs
@@ -1,0 +1,47 @@
+use super::*;
+
+// We dont use the collection abstractions used by list/map/set since vectors only support a small subset of data types.
+
+async fn create(connection: &CassandraConnection) {
+    run_query(
+        connection,
+        "CREATE TABLE collections.vector (id int PRIMARY KEY, col0 vector<int, 1>, col1 vector<bigint, 2>, col2 vector<float, 2>, col3 vector<double, 2>);",
+    )
+    .await;
+}
+
+async fn insert(connection: &CassandraConnection) {
+    run_query(
+        connection,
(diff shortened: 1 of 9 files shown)


shotover-proxy/tests/cassandra_int_tests/collections/vector.rs
```diff
@@ -0,0 +1,47 @@
...
+    )
+    .await;
+}
+
+async fn select(connection: &CassandraConnection) {
+    let results = vec![
+        ResultValue::Vector(vec![ResultValue::Int(1)]),
+        ResultValue::Vector(vec![ResultValue::BigInt(2), ResultValue::BigInt(3)]),
+        ResultValue::Vector(vec![
+            ResultValue::Float(4.0.into()),
```
rukai, line 26: use values that test the floatness

ronycsdu (approved):

justinweng-instaclustr (approved):

PR #1677, opened 2024-06-27
#1665 Clear optional fetch response fields to avoid making clients try to bypass shotover

### LINKED ISSUE #1665: KafkaSinkCluster - rewrite fetch responses to not refer to real brokers
rewrite `FetchResponse::node_endpoints`, `PartitionData::preferred_read_replica` and `PartitionData::current_leader` to avoid pointing at the real kafka broker.
The logic can go in `KafkaSinkCluster::process_responses`.
We should be able to just clear these values from the produce response and then call `Message::invalidate_cache()`

To learn more about these kafka fields you can take a look at: https://cwiki.apache.org/confluence/display/KAFKA/KIP-951%3A+Leader+discovery+optimisations+for+the+client
But the general idea is that they are optional new fields that enable faster rerouting in the case that the client attempted to route to a partition that was recently moved to another broker.
The problem with these fields is that they refer to internal kafka state such as broker IP addresses and broker id's that are incorrect from the clients perspective.
If the client attempted to connect to the IP address they would bypass shotover which would either fail (due to the IP being inaccessible to the client) or bypass shotovers transform logic. Both of which are bad outcomes.
Since these new fields are optional the easiest fix here is to rewrite the fields to be empty so that the optional functionality remains unused.

Normally when working with the kafka protocol I like to refer to https://kafka.apache.org/protocol.html#protocol_messages
However in this case `node_endpoints` is not included there since its implemented as one of the [_tagged_fields](https://cwiki.apache.org/confluence/display/KAFKA/KIP-482%3A+The+Kafka+Protocol+should+Support+Optional+Tagged+Fields).
In this case you are better off looking at the rust protocol crate: https://docs.rs/kafka-protocol/latest/kafka_protocol/messages/fetch_response/struct.FetchResponse.html
Or at the json files from which the java driver and rust protocol crate are generated from: 
https://github.com/apache/kafka/blob/a6718dbbdbcf3c71a8a3ab9844107cd994ceec25/clients/src/main/resources/common/message/FetchResponse.json#L108-L115

justinweng-instaclustr:
This PR clears the optional fields `FetchResponse::node_endpoints`, `PartitionData::preferred_read_replica` and `PartitionData::current_leader` in the Fetch response processed by the KafkaSinkCluster `process_responses()` function . This is to avoid passing these values to the client which will make the client try to bypass shotover and connect to Kafka brokers directly.

Closes #1665

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/mod.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/mod.rs
@@ -1294,7 +1294,7 @@ routing message to a random node so that:
                     body: ResponseBody::Fetch(fetch),
                     ..
                 })) => {
-                    // Clear this optional field to avoid making clients trying to bypass shotover
+                    // Clear this optional field to avoid making clients try to bypass shotover
                     // partition.current_leader and partition.preferred_read_replica are cleared due to the same reason
                     fetch.node_endpoints.clear();
                     for fetch_response in &mut fetch.responses {
--- a/shotover/src/transforms/kafka/sink_cluster/mod.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/mod.rs
@@ -1297,8 +1297,8 @@ routing message to a random node so that:
                     // Clear this optional field to avoid making clients trying to bypass shotover
                     // partition.current_leader and partition.preferred_read_replica are cleared due to the same reason
                     fetch.node_endpoints.clear();
-                    for fetch_response in &fetch.responses {
-                        for mut partition in &fetch_response.partitions {
+                    for fetch_response in &mut fetch.responses {
+                        for partition in &mut fetch_response.partitions {
                             partition.current_leader = LeaderIdAndEpoch::default();
                             partition.preferred_read_replica = BrokerId(-1);
                             if let Some(ResponseError::NotLeaderOrFollower) =
--- a/shotover/src/transforms/kafka/sink_cluster/mod.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/mod.rs
@@ -12,6 +12,7 @@ use anyhow::{anyhow, Context, Result};
 use async_trait::async_trait;
 use dashmap::DashMap;
 use kafka_protocol::messages::fetch_request::FetchTopic;
+use kafka_protocol::messages::fetch_response::LeaderIdAndEpoch;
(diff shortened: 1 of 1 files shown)


shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1293,24 +1294,30 @@ routing message to a random node so that:
...
                     ..
                 })) => {
-                    for response in &fetch.responses {
-                        for partition in &response.partitions {
+                    // Clear this optional field to avoid making clients try to bypass shotover
+                    // partition.current_leader and partition.preferred_read_replica are cleared due to the same reason
+                    fetch.node_endpoints.clear();
+                    for fetch_response in &mut fetch.responses {
+                        for partition in &mut fetch_response.partitions {
+                            partition.current_leader = LeaderIdAndEpoch::default();
```
rukai, line 1302: While setting current_leader.leader_id to -1 is good.
I'm not sure if current_leader.leader_epoch should be set to -1 or left as is.
I think that drivers encountering a -1 value here would just ignore it, but if the DB always populates this on routing error then the driver might assume its populated.

In conclusion, I think that what you have is fine.

justinweng-instaclustr: I'm happy to change this line to rewrite current_leader.leader_id to -1 while keeping current_leader.leader_epoch unchanged, if that's a safer approach.

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1293,24 +1294,30 @@ routing message to a random node so that:
...
+                            partition.preferred_read_replica = BrokerId(-1);
```
rukai, line 1303: I checked to see if this field was used in non error cases.
The DB is sending protocol version 15 fetch responses and they all have BrokerId(-1) so I think its safe to assume this is only used in error cases.

So, looks good :+1:

shotover/src/transforms/kafka/sink_cluster/mod.rs
rukai: I think what you have is the better approach, just noting the possibility here in case we do hit issues down the road.

rukai (approved):
Looks good! nice work.

ronycsdu (approved):

Issue #1664, opened 2024-06-14
KafkaSinkCluster - rewrite produce responses to not refer to real brokers


rukai:
rewrite `ProduceResponse::node_endpoints` and `PartitionProduceResponse::current_leader` to avoid pointing at the real kafka broker.
The logic can go in `KafkaSinkCluster::process_responses`.
We should be able to just clear these values from the produce response and then call `Message::invalidate_cache()`

There are some extra details in: https://github.com/shotover/shotover-proxy/issues/1665

PR #1678, opened 2024-06-28
#1664 Clear optional produce response fields to avoid making clients try to bypass shotover

justinweng-instaclustr:
This PR clears the optional fields `ProduceResponse::node_endpoints` and `PartitionProduceResponse::current_leader` in the Produce response processed by the KafkaSinkCluster `process_responses()` function . This is to avoid passing these values to the client which will make the client try to bypass shotover and connect to Kafka brokers directly.

Closes https://github.com/shotover/shotover-proxy/issues/1664

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/mod.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/mod.rs
@@ -1250,10 +1250,9 @@ routing message to a random node so that:
                     ..
                 })) => {
                     // Clear this optional field to avoid making clients try to bypass shotover
-                    // response_partition.current_leader is cleared due to the same reason
                     produce.node_endpoints.clear();
                     for (topic_name, response_topic) in &mut produce.responses {
-                        for response_partition in &mut response_topic.partition_responses {
+                        for response_partition in &response_topic.partition_responses {
                             if let Some(ResponseError::NotLeaderOrFollower) =
                                 ResponseError::try_from_code(response_partition.error_code)
                             {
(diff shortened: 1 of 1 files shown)


shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1284,11 +1288,16 @@ routing message to a random node so that:
                                         "Produce response included error NOT_LEADER_OR_FOLLOWER and so cleared topic {:?}",
                                         topic_name,
                                     );
+                                    response_partition.current_leader =
```
rukai, line 1291: why do we set this twice?

justinweng-instaclustr: If I'm reading the code correctly, without this line `response_partition.current_leader` won't get cleared if `break;` happens?

rukai: ah true.
I think we should pull the clearing logic added by this PR into a separate for loop.
This will make the logic easier to follow, as it will avoid entangling the separate processes of "modifying the message to hide the real cluster" and "reading the message to update shotover's internal metadata".

justinweng-instaclustr: Ok I can do it

rukai (approved):
looks good!

ronycsdu (approved):

Issue #1667, opened 2024-06-16
KafkaSinkCluster scram_over_mtls - improve token handling


rukai:
prereq: https://github.com/shotover/shotover-proxy/issues/1656

In two separate PRs implement the following:
1. Dont cache the delegation token within the KafkaSinkCluster instance. Instead refetch from the token task every time its needed. https://github.com/shotover/shotover-proxy/pull/1683
    * This ensures we dont use tokens that the task has already replaced, and is important for the next step.
    * This will look like removing the `delegation_token` field from `AuthorizeScramOverMtls` and then resolving the compiler issues by fetching the token from the task when its needed.
2. Invalidate expired tokens, we should immediately expire a token that has been replaced. (This is not possible at the moment as Kafka doesn't allow users to expire delegation tokens for other users. See more details in https://github.com/shotover/shotover-proxy/pull/1685. Instead we added a comment into the code regarding this in https://github.com/shotover/shotover-proxy/pull/1686)
    * This will improve security and keep the list of tokens on the brokers clean which is useful if we ever need to manually revoke tokens.

PR #1683, opened 2024-07-05
#1667 Remove cached delegation token from AuthorizeScramOverMtls

justinweng-instaclustr:
This PR removes the field `delegation_token` from the struct `AuthorizeScramOverMtls` to avoid caching the token in `AuthorizeScramOverMtls` instances. Instead, we store the username used to generate delegation tokens in this struct and fetch the delegation token from the `TokenTask` every time it's needed.

Progress towards #1667

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/node.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/node.rs
@@ -148,7 +148,7 @@ impl ConnectionFactory {
 
         // SCRAM client-first
         let mut scram = Scram::<Sha256>::new(
-            delegation_token.token_id.clone(),
+            delegation_token.token_id,
             delegation_token.hmac.to_string(),
             ChannelBinding::None,
             "tokenauth=true".to_owned(),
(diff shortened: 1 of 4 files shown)


shotover/src/transforms/kafka/sink_cluster/node.rs
```diff
@@ -141,10 +141,15 @@ impl ConnectionFactory {
...
+        let delegation_token = scram_over_mtls
+            .token_task
+            .get_token_for_user(scram_over_mtls.username.clone())
+            .await?;
+
         // SCRAM client-first
         let mut scram = Scram::<Sha256>::new(
-            scram_over_mtls.delegation_token.token_id.clone(),
-            scram_over_mtls.delegation_token.hmac.to_string(),
+            delegation_token.token_id.clone(),
```
rukai, line 151:
```suggestion
            delegation_token.token_id,
```
We can avoid the clone here since we own `delegation_token` and dont use it again after this line, the `token_id` will just be moved out of `delegation_token`.

justinweng-instaclustr: Ah yes, I'll remove the clone

rukai (approved):
Nice work!

ronycsdu (approved):

PR #1685, opened 2024-07-10
#1667 Expire delegation tokens immediately after they are replaced

### LINKED ISSUE #1667: KafkaSinkCluster scram_over_mtls - improve token handling

justinweng-instaclustr:
This PR creates the module `expire_token` which contains functions to send the ExpireDelegationTokenRequest to the Kafka cluster. The old delegation token is expired immediately after the new delegation token is created in token task in `scram_over_mtls`.

Closes #1667

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/scram_over_mtls.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/scram_over_mtls.rs
@@ -23,6 +23,7 @@ use crate::{
 use super::node::{ConnectionFactory, KafkaAddress};
 
 mod create_token;
+mod expire_token;
 mod recreate_token_queue;
 
 pub struct TokenRequest {
@@ -99,7 +100,7 @@ async fn task(
     token_creation_time_metric: &Histogram,
 ) -> Result<()> {
     let mut rng = SmallRng::from_rng(rand::thread_rng())?;
-    let mut username_to_token = HashMap::new();
+    let mut username_to_token: HashMap<String, DelegationToken> = HashMap::new();
     let mut recreate_queue =
         recreate_token_queue::RecreateTokenQueue::new(delegation_token_lifetime);
     let mut nodes = vec![];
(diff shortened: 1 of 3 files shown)


shotover/src/transforms/kafka/sink_cluster/scram_over_mtls.rs
```diff
@@ -108,21 +109,34 @@ async fn task(
...
                     &mut nodes,
                     &mut rng,
                     mtls_connection_factory,
                     &username,
                     delegation_token_lifetime
                 ).await
                 .with_context(|| format!("Failed to recreate delegation token for {:?}", username))?;
-                username_to_token.insert(username.clone(), token);
+
+                username_to_token.insert(username.clone(), new_token);
```
rukai, line 123: `.insert` will return the old token, so lets do `let old_token = username_to_token.insert(...`

shotover/src/transforms/kafka/sink_cluster/scram_over_mtls/expire_token.rs
```diff
@@ -0,0 +1,74 @@
...
+    hmac: StrBytes,
+    rng: &mut SmallRng,
+) -> anyhow::Result<()> {
+    let connection = get_node_connection(nodes, rng).await?;
+
+    connection.send(vec![Message::from_frame(Frame::Kafka(
+        KafkaFrame::Request {
+            header: RequestHeader::builder()
+                .request_api_key(ApiKey::ExpireDelegationTokenKey as i16)
+                .request_api_version(3)
```
rukai, line 43: There is no version 3 of this message.
The highest it goes to is 2.

PR #1695, opened 2024-07-22
Set aws-throwaway AZ

rukai:
This allows us to use ec2-cargo and windsock on our internal aws accounts.

For now I'm happy to hardcode `us-east-1b` since we are an unusual case where `us-east-1c` is not suitable.

I also added a readme.md to document how to use ec2-cargo.

@justinweng-instaclustr you should give this a go to make sure it works for you.
Note that cassandra tests currently fail on arm due to lack of support from the docker image we use.
So if using the default arm instance, I would just run the kafka tests.

### DIFF
--- a/ec2-cargo/src/main.rs
+++ b/ec2-cargo/src/main.rs
@@ -45,7 +45,10 @@ async fn main() {
         return;
     }
 
-    let aws = Aws::builder(CleanupResources::AllResources).build().await;
+    let aws = Aws::builder(CleanupResources::AllResources)
+        .use_az(Some("us-east-1b".into()))
+        .build()
+        .await;
     let instance_type = InstanceType::from(args.instance_type.as_str());
     let instance = aws
         .create_ec2_instance(Ec2InstanceDefinition::new(instance_type).volume_size_gigabytes(40))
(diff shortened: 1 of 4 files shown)


justinweng-instaclustr (approved):

ec2-cargo/src/main.rs
```diff
@@ -45,7 +45,10 @@ async fn main() {
         return;
     }
 
-    let aws = Aws::builder(CleanupResources::AllResources).build().await;
+    let aws = Aws::builder(CleanupResources::AllResources)
+        .use_az(Some("us-east-1b".into()))
```
ronycsdu, line 49: why us-east-1b? I thought us-east-1c was the default.

rukai: us-east-1c lacks a default subnet in our internal environment, therefore we are using us-east-1b for shotover infra.

ronycsdu: fair enough. In that case, should the default in https://github.com/shotover/aws-throwaway/blob/main/aws-throwaway/src/backend/sdk/aws.rs#L59 be us-east-1b as well?

rukai: I kept it the same there for backwards compatibility in case someone else is using the library.

ronycsdu (approved):

ronycsdu (approved):

PR #1705, opened 2024-07-25
Update to rust 1.80 🦀🦀🦀

rukai:
All code changes are due to new clippy lints added in 1.80.
The documentation changes are due to a clippy lint on invalid markdown. (every list needs to be followed by a newline)


shotover-proxy/benches/windsock/cloud/mod.rs
```diff
@@ -174,7 +174,7 @@ impl Cloud for AwsCloud {
         {
             let mut futures =
                 FuturesUnordered::<Pin<Box<dyn Future<Output = Result<()>> + Send>>>::new();
-            for instance in &mut resources.bencher {
+            if let Some(instance) = &mut resources.bencher {
```
ronycsdu, line 177: Has the possibility changed from multiple to single?

rukai: No, the logic is unchanged.
Since `Option` implements [IntoIterator](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html) you can use an `Option` in a for loop.
In that case the code block will run once if the Option is `Some` and not at all if it is `None`

However it is considered confusing to do so, hence why clippy has added a lint warning against this in rust 1.80.

ronycsdu: My understanding from the previous deleted code is that the for loop could have added more than 1 entry in the futures data structure whereas this new change will only add 1 entry if needed. Am I reading correctly?

rukai: Both the old and new implementation can only add 1 or 0 entries.

The value in `resources.bencher` is an `Option<Ec2InstanceWithBencher>`
This can either be `Some` containing exactly 1 `Ec2InstanceWithBencher`.
Or it can be `None` in which case it contains nothing.
Refer to my previous comment about why an Option can be used in a for loop.

Inspecting the types in an IDE might help? Or I'm happy to jump on a huddle to walk through it.

ronycsdu: Got it now.

ronycsdu (approved):

justinweng-instaclustr (approved):

PR #1712, opened 2024-07-30
Update tokio-tungstenite and rustls in test-helpers

rukai:
rustls needs to be updated at the same time since its part of tokio-tungstenite's public API

Changes might look complicated but they are entirely within test logic and is the equivalent to the upgrade we did within shotover itself in https://github.com/shotover/shotover-proxy/pull/1387

### DIFF
--- a/test-helpers/src/connection/cassandra/cql_ws.rs
+++ b/test-helpers/src/connection/cassandra/cql_ws.rs
@@ -1,3 +1,4 @@
+use anyhow::{Context, Result};
 use cassandra_protocol::compression::Compression;
 use cassandra_protocol::frame::message_query::BodyReqQuery;
 use cassandra_protocol::frame::message_response::ResponseBody;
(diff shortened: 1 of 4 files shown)


test-helpers/src/connection/cassandra/cql_ws.rs
```diff
@@ -280,4 +284,26 @@ impl ServerCertVerifier for SkipVerifyHostName {
             Err(err) => Err(err),
         }
     }
+
```
justinweng-instaclustr, line 287: Why do we need to add the following new functions here?

test-helpers/src/connection/cassandra/cql_ws.rs
```diff
@@ -228,25 +231,28 @@ fn spawn_read_write_tasks<S: AsyncRead + AsyncWrite + Unpin + Send + 'static>(
...
+            .context("Failed to add cert to cert store")?;
     }
-    root_cert_store
+    Ok(root_cert_store)
 }
 
+#[derive(Debug)]
 pub struct SkipVerifyHostName {
-    verifier: WebPkiVerifier,
+    verifier: Arc<WebPkiServerVerifier>,
```
justinweng-instaclustr, line 247: Why do we need Arc here?

test-helpers/src/connection/cassandra/cql_ws.rs
rukai: Its just following the new rustls API.
Refer to https://github.com/shotover/shotover-proxy/pull/1387/files#diff-b37d1d20c43783d212ac312f79ebbba68058679501116dcdf45beda95fe4ea5bR271-R291 where we made the same change in shotover itself.

test-helpers/src/connection/cassandra/cql_ws.rs
rukai: Its because `WebPkiServerVerifier::builder(Arc::new(roots)).build()` returns an Arc which is part of rustls API, and we need to store the value it returns here.

justinweng-instaclustr (approved):

ronycsdu (approved):

PR #1722, opened 2024-08-05
Add Wrapper::close_client_connection

rukai:
In order for shotover to properly recreate the behavior of certain databases it needs to be able to close incoming connections from the client.
Databases will close the connection on the client if it misbehaves, this can include things like failed auth and sending requests that cannot be parsed according to the databases wire protocol.
We need to be able to recreate this behavior to avoid subtle differences in behavior between shotover and the database.

Shotover currently lacks a way for a transform to close the incoming connection to the client.
This is a problem since cluster sink transforms need to be able to close the connection since they need to emulate a chunk of the servers behavior.

This PR fills this missing hole in the API shotover exposes to transforms and makes use of this new API in KafkaSinkCluster in order to close the client connection when authentication fails.

This is achieved by adding a `close_client_connection` field to `Wrapper`.
By default the field is false, but any transform can set it to true, allowing any transform to close the connection to the client.

Additionally we use the CloseReason enum to avoid flushing the chain when a transform requests the connection to be closed.
Chain flushing is needed to ensure we flush all pending responses back to the client, since shotover will not be sending anymore responses to the client it does not make sense to flush the chain.

This enables use cases such as:
* cluster sink transforms to close the connection for the purpose of better emulating a single server instance.
* single sink transforms to close the connection if they detect the server has closed its connection.
* custom auth implementations, implemented as a separate transform, to close the connection on auth failure.

In this PR, we only use `close_client_connection` for `KafkaSinkCluster` auth, as this is all we need for the "handling of down kafka nodes" project.

### DIFF
--- a/shotover/src/server.rs
+++ b/shotover/src/server.rs
@@ -634,9 +634,9 @@ impl<C: CodecBuilder + 'static> Handler<C> {
             .run_loop(&client_details, local_addr, in_rx, out_tx, force_run_chain)
             .await;
 
-        // Only flush messages if we are shutting down due to application shutdown
+        // Only flush messages if we are shutting down due to shotover shutdown or client disconnect
         // If a Transform::transform returns an Err the transform is no longer in a usable state and needs to be destroyed without reusing.
-        if result.is_ok() {
+        if let Ok(CloseReason::ShotoverShutdown | CloseReason::ClientClosed) = result {
             match self.chain.process_request(&mut Wrapper::flush()).await {
                 Ok(_) => {}
                 Err(e) => error!(
@@ -649,7 +649,7 @@ impl<C: CodecBuilder + 'static> Handler<C> {
             }
         }
 
-        result
+        result.map(|_| ())
     }
 
     async fn receive_with_timeout(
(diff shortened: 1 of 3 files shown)


shotover/src/server.rs
```diff
@@ -707,30 +707,29 @@ impl<C: CodecBuilder + 'static> Handler<C> {
...
                 },
             };
         }
 
-        Ok(())
+        Ok(CloseReason::ShotoverShutdown)
     }
 
-    async fn send_receive_chain(
+    async fn process_requests(
```
justinweng-instaclustr, line 724: A minor question, why do we rename this function again after it was renamed in the previous PR https://github.com/shotover/shotover-proxy/pull/1727/files#diff-a583948a329ee7447a9001d675ee60ab5789970c2b49cf275173357073199260R724 ?

rukai: good catch, this was a bad merge, I'll fix it

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1209,6 +1216,12 @@ routing message to a random node so that:
                     body: ResponseBody::SaslAuthenticate(authenticate),
                     ..
                 })) => {
+                    // The broker always closes the connection after an auth failure response,
+                    // so we should do the same.
+                    if authenticate.error_code != 0 {
```
justinweng-instaclustr, line 1221: should we do this check here or in `self.process_sasl_authenticate(authenticate)?;`?

rukai: The only downside is having to pass in close_client_connection, but you are right, it should really be in process_sasl_authenticate.

justinweng-instaclustr (approved):

ronycsdu (approved):

PR #1728, opened 2024-08-20
KafkaSinkCluster - handle receive errors

rukai:
Currently KafkaSinkCluster ignores failures to receive messages:
https://github.com/shotover/shotover-proxy/blob/17ccad5e257bf06836283ad8c49970b3dd63e749/shotover/src/transforms/kafka/sink_cluster/mod.rs#L1041
This is quite bad since many kinds of issues are silently ignored, none of our other transforms behave this way.

This PR fixes this by bubbling up errors when receiving from a connection.
However, there are a few latent issues hidden by the lack of error handling here:
* https://github.com/shotover/shotover-proxy/pull/1722
* recreate connections before the delegation token expires - done in this PR.

These must be resolved before we can enable handling of receive errors.

## recreate connections before the delegation token expires

When scram_over_mtls is enabled, connections will be closed when the token it auth'd with expires.
This PR introduces a system for recreating token authed connections before the token expires.
This new system is implemented by:
* Wrapping over SinkConnection with a KafkaConnection enum. It can be either:
   + `KafkaConnection::General` - the existing behaviour, just routes directly to SinkConnection methods
   + `KafkaConnection::ScramOverMtls` - the new behaviour for scram over mtls
       - it returns `ConnectionState::AtRiskOfAuthTokenExpiry` before the scram connection expires.
       - it maintains an old_connection to gracefully transition to a new connection when the connection is recreated.
       
## Alternate approach

As an alternative to the timeout approach we could rewrite the token task to send an `ArcSwap<DelegationToken>` which is then live updated, we could then check this token everytime we send to determine if we need to recreate the connection.

We still need the old_connection fallback logic though, so largely this PR would remain the same.
So I propose we land this and we could possibly swap to the ArcSwap approach in the future if it was valuable.

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/connections.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/connections.rs
@@ -1,14 +1,21 @@
-use crate::connection::SinkConnection;
-use anyhow::{Context, Result};
+use crate::{
+    connection::{ConnectionError, SinkConnection},
+    message::Message,
+};
+use anyhow::{anyhow, Context, Result};
 use fnv::FnvBuildHasher;
 use kafka_protocol::{messages::BrokerId, protocol::StrBytes};
 use metrics::Counter;
 use rand::{rngs::SmallRng, seq::SliceRandom};
-use std::collections::HashMap;
+use std::{
+    collections::HashMap,
+    time::{Duration, Instant},
+};
 
 use super::{
     node::{ConnectionFactory, KafkaAddress, KafkaNode},
     scram_over_mtls::AuthorizeScramOverMtls,
+    SASL_SCRAM_MECHANISMS,
 };
 
 #[derive(Debug, PartialEq, Eq, Hash, Clone, Copy)]
@@ -28,7 +35,7 @@ pub enum Destination {
 }
 
 pub struct Connections {
-    pub connections: HashMap<Destination, SinkConnection, FnvBuildHasher>,
+    pub connections: HashMap<Destination, KafkaConnection, FnvBuildHasher>,
     out_of_rack_requests: Counter,
 }
 
@@ -50,8 +57,9 @@ impl Connections {
         nodes: &[KafkaNode],
         contact_points: &[KafkaAddress],
         local_rack: &StrBytes,
+        recent_instant: Instant,
         destination: Destination,
-    ) -> Result<&mut SinkConnection> {
+    ) -> Result<&mut KafkaConnection> {
         let node = match destination {
             Destination::Id(id) => Some(nodes.iter().find(|x| x.broker_id == id).unwrap()),
             Destination::ControlConnection => None,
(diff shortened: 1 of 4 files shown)


shotover/src/transforms/kafka/sink_cluster/connections.rs
```diff
@@ -67,22 +75,230 @@ impl Connections {
...
+                    None,
+                    destination,
+                )
+                .await
+                .context("Failed to create a new connection")?;
+            }
+            ConnectionState::AtRiskOfTimeout => {
+                let old_connection = self.connections.remove(&destination).unwrap();
+                if old_connection.old_connection.is_some() {
+                    return Err(anyhow!("Old connection had an old connection"));
```
rukai, line 99: add more details as to why thats bad

shotover/src/transforms/kafka/sink_cluster/connections.rs
```diff
@@ -67,22 +75,230 @@ impl Connections {
...
+                    node,
+                    contact_points,
+                    None,
+                    destination,
+                )
+                .await
+                .context("Failed to create a new connection")?;
+            }
+            ConnectionState::AtRiskOfTimeout => {
+                let old_connection = self.connections.remove(&destination).unwrap();
```
rukai, line 97: old_connection -> existing_connection

rukai:
rewrite PR to focus only on scram_over_mtls handling, this should make things a little simpler.

shotover/src/transforms/kafka/sink_cluster/scram_over_mtls/connection.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    pub fn state(
+        &self,
+        recent_instant: Instant,
+        authorize_scram_over_mtls: &Option<AuthorizeScramOverMtls>,
+    ) -> ConnectionState {
+        // The delegation token is recreated after `0.5 * delegation_token_lifetime`
+        // Consider what happens when we match that timing for our connection timeout here:
+        //
```
rukai, line 102: mention this being a timeline

shotover/src/transforms/kafka/sink_cluster/scram_over_mtls/connection.rs
```diff
@@ -0,0 +1,159 @@
...
+        // |                    |--------------------|
+        // This case is a race condition.
+        // We could start with either token t2 or t1.
+        // If we start with t1 we could go past the end of t1's lifetime.
+        // To avoid this issue we reduce the size of the connection lifetime by a further 25%
+        //
+        // At low values of delegation_token_lifetime all of this falls apart since something
+        // like a VM migration could delay shotover execution for many seconds.
+        // However for sufficently large delegation_token_lifetime values (> 1 hour) this should be fine.
+        let timeout = authorize_scram_over_mtls
```
justinweng-instaclustr, line 143: A minor comment. Since we do this calculation every time the connection state is checked, would it affect the performance?

justinweng-instaclustr: since the timeout is a constant based on delegation_token_lifetime, would it be worth storing it in authorize_scram_over_mtls as well?

rukai: I had assumed it would compile down to a single instruction. (or at least very few)
But I looked into it just now and found some interesting results.

On the right we have the ASM generated for the current code.
On the left we have the ASM generated when I combine the `mul_f32` into a single `mul_f32(0.5 * 0.75)`
image
As you can see reducing the number of `mul_f32` calls greatly reduces the number of instructions.

Inspecting the implementation of mul_f32 I can see that multiplying a Duration is actually fairly complicated, so the results do make some sense.
I'll look into moving the multiply into ScramOverMtlsConnection::new and storing the result.

justinweng-instaclustr: Interesting find!

rukai: After moving the calculation out of the function entirely:
image

justinweng-instaclustr (approved):

ronycsdu (approved):

PR #1729, opened 2024-08-20
test codspeed benches on ubuntu 24.04

rukai:


rukai:
codspeed doesnt support ubuntu 24.04.
For context, I thought its weird that gettime shows up as a syscall in codspeed flamegraphs since it should go through vdso and avoid a syscall. I thought maybe the newer linux kernel in 24.04 might avoid this issue, so I attempted to test my hypothesis with this PR.

I'll retry this test in the future once codspeed supports ubuntu 24.04

PR #1734, opened 2024-08-26
KafkaSinkCluster route by NodeState

rukai:
blocked on https://github.com/shotover/shotover-proxy/pull/1715

## Tests
A new test case is added to certain tests to kill a kafka node while accessing an RF3 partition.
An extra node is added to these test's docker-compose.yaml (making a total of 4 nodes) to give the cluster room to add another replica and maintain RF3.

To support clusters of more than 3 nodes with TLS (needed for the `cluster_sasl_scram_over_mtls_single_shotover` test), extra DNS names (kafka3, kafka4, kafka5) are added to the TLS cert we generate for testing.
Technically we only need kafka3 added, but I added more since these kinds of TLS failures are tricky to debug.

### Routing
The methods `store_topic_names`/`store_topic_ids`/`store_group` contain logic for checking if we are missing any metadata needed for routing incoming requests.
This PR extends that logic by making them also check if the metadata points at a node that is now down as this would indicate that the metadata may be out of date, so it should be refetched.

## Dont explode when nodes disappear

Some of the logic in KafkaSinkCluster is naive and assumes that once a node has existed in the metadata of kafka nodes it will always exist.
This is however not true, a "down" node will be eventually removed from the clusters advertised list of nodes.
So to fix this, in the `rewrite_metadata_response` method, this PR replaces some `unwrap`s with `if let`s to gracefully handle the absence of nodes that we expect to exist.


shotover/src/transforms/kafka/sink_cluster/connections.rs
```diff
@@ -227,23 +226,25 @@ impl Connections {
...
         if old_connection
             .map(|old| old.pending_requests_count())
             .unwrap_or(0)
             > 0
         {
-            Err(error.context("Succesfully reopened outgoing connection but previous outgoing connection had pending requests."))
-        } else {
-            self.connections.insert(destination, connection);
-            Ok(())
+            return Err(error.context("Outgoing connection had pending requests, those requests/responses are lost so connection recovery cannot be attempted."));
```
rukai, line 235: this is moved before the `connection.context(..)?` to ensure that we return this error regardless of if the new connection is succesfully created or not.

ronycsdu (approved):

shotover-proxy/tests/kafka_int_tests/mod.rs
```diff
@@ -247,21 +247,55 @@ async fn single_sasl_scram_plaintext_source_tls_sink(#[case] driver: KafkaDriver
...
+
+        test_cases::produce_consume_partitions1_kafka_node_goes_down(
+            driver,
+            &docker_compose,
+            &connection_builder,
+            "kafka_node_goes_down_test",
+        )
+        .await;
+
+        // Shotover can reasonably hit many kinds of errors due to a kafka node down so ignore all of them.
```
justinweng-instaclustr, line 284: Is it possible to specify the expected error types here so that we can still catch errors that shouldn't happen during a kafka node down?

shotover-proxy/tests/kafka_int_tests/test_cases.rs
```diff
@@ -169,6 +172,138 @@ pub async fn produce_consume_partitions1(
     }
 }
 
+pub async fn produce_consume_partitions1_kafka_node_goes_down(
+    driver: KafkaDriver,
+    docker_compose: &DockerCompose,
+    connection_builder: &KafkaConnectionBuilder,
+    topic_name: &str,
+) {
+    if driver.is_cpp() {
```
justinweng-instaclustr, line 181: I recall we can use the feature flag of driver types to control if tests cases are run. Can we use that approach here?

rukai: Say if we used feature flags here and set this test case to `#[cfg(not(feature = "kafka-cpp-driver-tests")]`, then if we enable all feature flags (which is how we run our tests in CI) then this test case would not run at all.

We dont have a feature flag for the java driver since its always available.

So the check really needs to be done at runtime rather than compile time.

In general, the purpose of feature flags are to enable compiling on platforms that dont support certain drivers or to reduce compile times, not to exclude test cases in certain conditions.

shotover-proxy/tests/kafka_int_tests/mod.rs
rukai: There are a lot of errors that can occur during a kafka node being down so I dont think its really feasible at the moment.
The test logic itself still guarantees that all the kafka functionality is upheld, so I'm not too worried.

I think there would be value in extending our [test infrastructure](https://github.com/shotover/tokio-bin-process) to support matching errors on a regex, but that is out of scope of this work.

justinweng-instaclustr (approved):

PR #1738, opened 2024-08-28
Test kafka with replication_factor 3

rukai:
This was a missing test case that upon implementation revealed a bug in shotover.
Since we cannot create a topic with replication_factor 3 on a kafka cluster with only 1 node, `cluster_test_suite` is introduced for the cluster tests to use without regressing the single/passthrough tests.

To get the tests to pass we required:
* one change to the test logic - we need to set produce [acks](https://docs.confluent.io/platform/current/installation/configuration/producer-configs.html#acks) to "all"
  + It kind of makes sense in that, adding replicas means that there is now a difference in behavior between `acks=1` and `acks=all`, acks=1 will not wait for the replicas to acknowledge the produced record.
  + On the other hand I'm not entirely sure why we need it since an ack from the leader should be sufficient as the consumer and producer only routes to the leader.
* one change to shotover routing logic - we need to route fetch requests to the leader instead of the replicas.
  + The comment I added to the code justifies the change pretty well. But basically implementing shotover to route fetch requests to the replicas was a mistake that sounded correct at the time. kafka does not allow routing to replicas in the same way that cassandra and redis do. In kafka, replicas are purely for "high availability" with some extra request handling tacked on.

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/mod.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/mod.rs
@@ -825,32 +825,18 @@ routing message to a random node so that:
                     let destination = if let Some(partition) =
                         topic_meta.partitions.get(partition_index)
                     {
-                        if let Some(node) = self
-                            .nodes
-                            .iter_mut()
-                            .filter(|node| {
-                                partition
-                                    .shotover_rack_replica_nodes
-                                    .contains(&node.broker_id)
-                            })
-                            .choose(&mut self.rng)
-                        {
-                            node.broker_id
-                        } else {
-                            tracing::debug!(
-                                "Routing fetch request to replica outside of shotover's rack"
+                        // While technically kafka has some support for fetching from replicas, its quite weird.
+                        // See https://cwiki.apache.org/confluence/display/KAFKA/KIP-392%3A+Allow+consumers+to+fetch+from+closest+replica
+                        // We should never route to replica_nodes from the metadata response ourselves.
+                        // Instead, when its available, we can make use of preferred_read_replica field in the fetch response as an optimization.
+                        // However its always correct to route to the partition.leader_id which is what we do here.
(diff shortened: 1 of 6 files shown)


shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -825,32 +825,18 @@ routing message to a random node so that:
...
-                                .filter(|node| {
-                                    partition
-                                        .external_rack_replica_nodes
-                                        .contains(&node.broker_id)
-                                })
-                                .choose(&mut self.rng)
-                                .unwrap()
-                                .broker_id
                         }
+                        partition.leader_id
```
justinweng-instaclustr, line 839: I'm not too familiar with kafka fetch requests, but should shotover route it to replica nodes if `partition.leader_id == -1` ?

rukai: unfortunately not, from my understanding of https://cwiki.apache.org/confluence/display/KAFKA/KIP-392%3A+Allow+consumers+to+fetch+from+closest+replica only the leader should be used, in some circumstances however we can refer to `preferred_read_replica` field which may point at replicas but I've not attempted that in this PR.

justinweng-instaclustr: probably a silly question but what happens to the fetch requests when `partition.leader_id == -1`?

rukai: The only time I've seen a leader_id of -1 is when the leader broker was killed and a new leader has not yet been found.

Any time shotover routes a request to broker -1 it routes the request to a random node.
Kafka will then give an appropriate error message for being unable to handle the request which we then return the client.
The client will then retry the request.

justinweng-instaclustr (approved):

ronycsdu (approved):

PR #1740, opened 2024-08-29
KafkaSinkCluster - improve metadata response rewriting

rukai:
Previously we were exposing internal cluster BrokerId's to the client.
However these need to be rewritten to the BrokerId's used by the shotover instances.
This PR introduces the required message rewriting logic.

The `isr_nodes` (in-sync-replica nodes) can follow the same logic as `replica_nodes`, so the logic is just copied over from there.

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/mod.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/mod.rs
@@ -1821,6 +1821,36 @@ routing message to a random node so that:
                     }
                 }
                 partition.replica_nodes = shotover_replica_nodes;
+                // Every isr (in-sync-replica) node has its entire corresponding shotover rack included.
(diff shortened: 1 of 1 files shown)


ronycsdu (approved):

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1821,6 +1821,36 @@ routing message to a random node so that:
                     }
                 }
                 partition.replica_nodes = shotover_replica_nodes;
+                // Every isr node has its entire corresponding shotover rack included.
```
justinweng-instaclustr, line 1824: Minor: should we explain isr here?

rukai: yeah, easy enough

justinweng-instaclustr (approved):

PR #1743, opened 2024-09-04
kafka docker-compose.yaml fixes

rukai:
This PR makes 2 config changes to the kafka docker-compose.yaml files.
* all cluster configs are changed to set the RF of system topics to 3. For example the `__consumer_offset` topic.
   + Previously these were set to an RF1 of one due to some bitnami default, kafka's default for these is RF3 which is much more sane.
   + Without this change we were getting weird intermittent failures in https://github.com/shotover/shotover-proxy/pull/1734
   + This change more accurately reflects actual production usage.
  + Not every test cluster config NEEDS these new configs, but I've applied it to every cluster config in case we run high availability tests on those clusters in the future. It would be impossible to apply to single kafka node setups, since we can only achieve an RF of 1 there.
* all inter_broker users are set to the super user
   + This was a latent problem that was only revealed when making the system topics RF change described above.
   + When coupled with the RF3 change this issue caused basic kafka functionality to fail since the `__consumer_offset` topic became unwritable, not sure why this triggered it exactly, but its definitely a misconfiguration so not worth worrying about.
   + For many yamls this means removing the `KAFKA_INTER_BROKER_USER` so it will fallback to the default value which is `user`, while specifying the password of `user` in `KAFKA_INTER_BROKER_PASSWORD`.
   + Sometimes the super user is set to a name like `super_user` instead of just `user`. In those cases we set `KAFKA_INTER_BROKER_USER` to `super_user`.
   + `KAFKA_CONTROLLER_USER` and `KAFKA_CONTROLLER_PASSWORD` can just be deleted entirely since the default values are fine and manually setting them is just confusing.
   
This readme describes how to [configure bitnami kafka images](https://github.com/bitnami/containers/blob/main/bitnami/kafka/README.md)


ronycsdu (approved):

rukai:
rony question what `KAFKA_CFG_OFFSETS_TOPIC_REPLICATION_FACTOR=2` is for, and its a good question.
I saw it here https://github.com/bitnami/containers/blob/main/bitnami/kafka/README.md#step-1-create-the-first-node-for-apache-kafka when I was looking into how to set KAFKA_CFG_OFFSETS_TOPIC_REPLICATION_FACTOR, figured it was important in the same way if they included it here as well.

Looking into it further we have:
https://docs.confluent.io/platform/current/installation/configuration/broker-configs.html#transaction-state-log-min-isr
which references:
https://docs.confluent.io/platform/current/installation/configuration/broker-configs.html#min-insync-replicas

So it looks like it enforces that the internal transactions topic must have at least 2 in sync replicas before it can send an ack to the broker who wrote to it.
We are setting it to 2 and the default kafka value is 2 as well.
Likely bitnami has changed the default of this field as well, so we should set it back to 2 as we do in this PR.

justinweng-instaclustr (approved):

PR #1744, opened 2024-09-04
Fix setting node to down during send_requests

rukai:
pulled out of https://github.com/shotover/shotover-proxy/pull/1734

In `send_requests` we currently call `handle_connection_error` and therefore set the node as down when a failure to send occurs.
However we do not do that when a failure to create a connection occurs.
So this PR adds error handling for failure to create connection case:
* It sets the node as down.
* To improve debuggability it includes the requests that were routed to a node that could not be connected to.
   This is useful since it could help diagnose routing issues where we arent routing away from down nodes.

This PR also makes Node::state field private by adding a setter method, this simplifies access of the field since it hides the need to think about `Ordering`, relaxed ordering will suffice for all possible use cases here.


shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1080,22 +1078,62 @@ routing message to a random node so that:
...
+                        .set_state(NodeState::Down);
+
+                    // bubble up error
+                    let request_types: Vec<String> = requests
+                        .requests
+                        .iter_mut()
+                        .map(|x| match x.frame() {
+                            Some(Frame::Kafka(KafkaFrame::Request { header, .. })) => {
+                                format!("{:?}", ApiKey::try_from(header.request_api_key).unwrap())
+                            }
```
justinweng-instaclustr, line 1129: Why do we log the api key here?

rukai: API key is the message type. E.g. fetch, produce, metadata
I included it since I found it useful for debugging routing issues that were causing connections to be opened against down nodes.

justinweng-instaclustr: Right, just read https://kafka.apache.org/0101/protocol.html#protocol_api_keys and found out Kafka defines the request type using "API key" which is a bit confusing because I thought it's a credential.

justinweng-instaclustr (approved):

shotover/src/transforms/kafka/sink_cluster/mod.rs
rukai: oh yeah, that would be very concerning :laughing:

ronycsdu (approved):

PR #1752, opened 2024-09-17
KafkaSinkCluster: handle broker closing connection due to idle timeout

rukai:
Kafka brokers have a config field `connections.max.idle.ms`, when a connection has been idle for longer than this timeout.
Normally a client can expect idle connections to timeout after some time of being unused.
However when KafkaSinkCluster is sitting in between the client and the broker there can be some unexpected behaviour.

For example:
1. We have a kafka cluster configured with `connections.max.idle.ms=5s`.
2. A client sends a request every 1s.
3. Shotover routes the first request to node A
4. Shotover routes the next 10 requests to node B
5. Shotover routes another request to node A
    + this request will fail due to the connection hitting idle time out even though the client was sending one request per second, well within the timeout!

With more reasonable timeout values this issue is harder to hit, but we are still still hitting this in our preprod environment with the default idle timeout of 10 minutes.

## Integration test

This PR introduces an integration test that runs kafka with a very low connections.max.idle.ms (30s), this reproduces the issue described above.
The new integration test fails when running the shotover main branch.

## Fix implementation

This PR then fixes the failing test by adding a check into our connection getting logic to recreate the connection if it has hit an error.
This does not attempt to fix all possible failures due to closed connection, as that would be impossible due to race conditions.
Instead we make a best effort handling which will catch the regular failures we are seeing.

### DIFF
--- a/shotover-proxy/tests/kafka_int_tests/mod.rs
+++ b/shotover-proxy/tests/kafka_int_tests/mod.rs
@@ -302,6 +302,37 @@ async fn cluster_1_rack_single_shotover(#[case] driver: KafkaDriver) {
 #[cfg_attr(feature = "kafka-cpp-driver-tests", case::cpp(KafkaDriver::Cpp))]
 #[case::java(KafkaDriver::Java)]
 #[tokio::test(flavor = "multi_thread")] // multi_thread is needed since java driver will block when consuming, causing shotover logs to not appear
+async fn cluster_1_rack_single_shotover_broker_idle_timeout(#[case] driver: KafkaDriver) {
+    let _docker_compose = docker_compose(
+        "tests/test-configs/kafka/cluster-1-rack/docker-compose-short-idle-timeout.yaml",
+    );
+
+    let shotover = shotover_process("tests/test-configs/kafka/cluster-1-rack/topology-single.yaml")
+        .start()
+        .await;
+
+    let connection_builder = KafkaConnectionBuilder::new(driver, "127.0.0.1:9192");
+
+    // We do not run the regular test suite since there is a race condition where the timeout
+    // could occur between checking if the connection is live and sending a request.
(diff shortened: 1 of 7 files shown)


test-helpers/src/connection/kafka/mod.rs
```diff
@@ -162,10 +162,13 @@ impl KafkaConsumer {
             expected_response.key, response.key,
             "Unexpected key for topic {topic}"
         );
-        assert_eq!(
-            expected_response.offset, response.offset,
-            "Unexpected offset for topic {topic}"
-        );
+
+        if expected_response.offset.is_some() {
```
rukai, line 166: The documentation and PartialEq implementation of ExpectedResponse already match this logic, but we missed it here somehow.

shotover-proxy/tests/kafka_int_tests/mod.rs
```diff
@@ -656,7 +687,7 @@ fn multi_shotover_events(driver: KafkaDriver) -> Vec<EventMatcher> {
             .with_level(Level::Warn)
             .with_target("shotover::transforms::kafka::sink_cluster")
             .with_message(
-                r#"no known coordinator for GroupId("some_group"), routing message to a random node so that a NOT_COORDINATOR or similar error is returned to the client"#,
+                r#"no known coordinator for GroupId("some_group"), routing message to a random broker so that a NOT_COORDINATOR or similar error is returned to the client"#,
```
rukai, line 690: Fixes an unrelated intermittent failure that slip in through: https://github.com/shotover/shotover-proxy/pull/1750/files

ronycsdu (approved):

justinweng-instaclustr:
Hey Lucas, the `connections.max.idle.ms` in description should be milliseconds, right? But it's referred to as seconds

justinweng-instaclustr (approved):

PR #1758, opened 2024-09-30
Add NodeState to ShotoverNode + some renaming

justinweng-instaclustr:
- Add a new `ShotoverNodeState` field to `ShotoverNode`. We are not reusing the existing `NodeState` for Kafka because it's possible that we will have different node states for Shotover in the future.
- Add a new configuration `check_shotover_peers_delay_ms` to `KafkaSinkCluster` which will be used when checking connection to nodes and setting node states.
- Move `ShotoverNode` code into a new file `kafka/sink_cluster/shotover_node.rs`
- Rename `kafka/sink_cluster/node.rs` to `kafka/sink_cluster/kafka_node.rs` and `NodeState` to `KafkaNodeState`

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/shotover_node.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/shotover_node.rs
@@ -14,7 +14,7 @@ pub struct ShotoverNodeConfig {
 }
 
 impl ShotoverNodeConfig {
-    pub fn build(self) -> anyhow::Result<ShotoverNode> {
+    pub(crate) fn build(self) -> anyhow::Result<ShotoverNode> {
         Ok(ShotoverNode {
             address: KafkaAddress::from_str(&self.address)?,
             rack: StrBytes::from_string(self.rack),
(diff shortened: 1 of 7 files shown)


shotover/src/transforms/kafka/sink_cluster/shotover_node.rs
```diff
@@ -0,0 +1,41 @@
...
+            address: KafkaAddress::from_str(&self.address)?,
+            rack: StrBytes::from_string(self.rack),
+            broker_id: BrokerId(self.broker_id),
+            state: Arc::new(AtomicShotoverNodeState::new(ShotoverNodeState::Up)),
+        })
+    }
+}
+
+#[derive(Clone)]
+pub struct ShotoverNode {
```
rukai, line 28: we should be able to keep ShotoverNode and ShotoverNodeState private?

justinweng-instaclustr: Done!

rukai (approved):
Looks good!

conorbros (approved):

PR #1759, opened 2024-09-30
Update kafka-protocol crate

rukai:
Update to the [recently released](https://github.com/tychedelia/kafka-protocol-rs/pull/95) kafka-protocol 0.13.0 

The largest change was https://github.com/tychedelia/kafka-protocol-rs/pull/90 which converted all IndexMap fields to use Vec instead.
This change was required due to the protocol actually supporting duplicate keys at the protocol level. (Its up to the broker to reject responses that are invalid due to duplicate keys)
The easiest way to fix this is to change those fields to Vecs and store the key in the struct.
This approach also comes with some performance benefits at the decode stage as can be seen by the improvement to the `decode_request_produce_create` benchmark.

Due to the replacement of IndexMap with Vec there was only 1 lookup that would have to be turned into an O(N) lookup.
To avoid it becoming an O(N) lookup I've added an intermediate hashmap which is converted to a vec at the end.
I'm not entirely sure its needed but its the safer option (better worst case) so I took it.

### DIFF
--- a/shotover/Cargo.toml
+++ b/shotover/Cargo.toml
@@ -115,7 +115,8 @@ aws-config = { version = "1.0.0", optional = true }
 aws-sdk-kms = { version = "1.1.0", optional = true }
 chacha20poly1305 = { version = "0.10.0", features = ["std"], optional = true }
 generic-array = { version = "0.14", features = ["serde"], optional = true }
-kafka-protocol = { version = "0.12.0", optional = true, default-features = false, features = ["messages_enums", "broker", "client"] }
+#kafka-protocol = { version = "0.12.0", optional = true, default-features = false, features = ["messages_enums", "broker", "client"] }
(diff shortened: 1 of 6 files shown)


shotover/Cargo.toml
```diff
@@ -126,6 +126,11 @@ atoi = { version = "2.0.0", optional = true }
 fnv = "1.0.7"
 sasl = { version = "0.5.1", optional = true, default-features = false, features = ["scram"] }
 
+# Force C dependencies to be built in parallel e.g. ring has some C code it compiles with cc
+# Remove this if we no longer have cc in our dep tree.
+[build-dependencies]
+cc = { version = "1.0", features = ["parallel"] }
```
rukai, line 132: This feature was previously enabled by a dependency of kafka-protocol, to retain this feature I've manually added it here.

justinweng-instaclustr (approved):

ronycsdu (approved):

PR #1762, opened 2024-10-03
Detect offline shotover nodes for KafkaSinkCluster

justinweng-instaclustr:
After introducing `ShotoverNodeState` to `ShotoverNode` in https://github.com/shotover/shotover-proxy/pull/1758, we should add a task to detect down shotover nodes and set `ShotoverNodeState` accordingly. 

This PR adds a background task `check_shotover_peers` looping over peer shotover nodes and trying to open a TCP connection to each peer shotover node. If the connection cannot be established within `connect_timeout_ms`, the peer node is marked as down. 
- `connect_timeout_ms` is the same configuration used when creating a connection to a destination kafka broker. 
- Each check is delayed for (`check_shotover_peers_delay_ms` + random(-`check_shotover_peers_delay_ms`/10, `check_shotover_peers_delay_ms`/10)) before moving to the next peer shotover node. 
- `start_shotover_peers_check` is called when the instance of `KafkaSinkClusterBuilder` is being created and hence is called exactly once.
- `check_shotover_peers` is be invoked at all if there's no peer shotover node (i.e., there's only 1 shotover node in the cluster)
- `check_shotover_peers` is restarted if the creation of random number generator fails.

The next PR will change metadata rewrites to exclude down shotover nodes.


shotover/src/transforms/kafka/sink_cluster/shotover_node.rs
```diff
@@ -1,9 +1,13 @@
-use crate::transforms::kafka::sink_cluster::kafka_node::KafkaAddress;
 use atomic_enum::atomic_enum;
 use kafka_protocol::messages::BrokerId;
 use kafka_protocol::protocol::StrBytes;
 use serde::{Deserialize, Serialize};
+use std::net::SocketAddr;
+use std::net::TcpStream;
```
rukai, line 6: we need to use https://docs.rs/tokio/latest/tokio/net/struct.TcpStream.html instead of `std::net::TcpStream`.
The std TcpStream will block the current thread when sending, receiving or opening a connection. In this case the blocking would last up to connect_timeout_ms and it will completely prevent execution of other tokio tasks on the current thread until the operation completes.

justinweng-instaclustr: ok will do.

shotover-proxy/tests/kafka_int_tests/test_cases.rs
```diff
@@ -539,6 +539,102 @@ pub async fn produce_consume_partitions1_kafka_node_goes_down(
...
+        consumer
+            .assert_consume(ExpectedResponse {
+                message: "initial".to_owned(),
+                key: Some("Key".to_owned()),
+                topic_name: topic_name.to_owned(),
+                offset: Some(0),
+            })
+            .await;
+
+        docker_compose.kill_service(shotover_node_to_kill);
```
rukai, line 592: shotover is not running on docker, so this is incorrect.
You will need to call `shotover.shutdown_and_then_consume_events` instead.

shotover/src/transforms/kafka/sink_cluster/shotover_node.rs
```diff
@@ -26,16 +32,79 @@ impl ShotoverNodeConfig {
...
+
+pub(crate) fn start_shotover_peers_check(
+    shotover_peers: Vec<ShotoverNode>,
+    check_shotover_peers_delay_ms: u64,
+    connect_timeout_ms: u64,
+) {
+    tokio::spawn(async move {
+        // Wait for all shotover nodes to start
+        sleep(Duration::from_secs(10)).await;
+        loop {
```
rukai, line 68: Do we need this outer loop?

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -2125,8 +2141,9 @@ impl KafkaSinkCluster {
                     })
                     .unwrap();
 
-                find_coordinator.host = shotover_node.address.host.clone();
-                find_coordinator.port = shotover_node.address.port;
+                find_coordinator.host =
+                    StrBytes::from_string(shotover_node.address.ip().to_string());
```
rukai, line 2145: Converting from SocketAddr (entirely on the stack) to StrBytes (reference counted string stored on the heap) results in an allocation, which would be good to avoid.

How about we have 2 address fields in ShotoverNode.
```rust
    /// Use this field when replacing fields in kafka messages.
    pub address_as_string KafkaAddress,
    /// Use this field when connecting to the shotover node
    pub address_as_socket_addr: SocketAddr,
```
Since ShotoverNode is treated as immutable we dont have to worry about "source of truth" issues with this duplication.

shotover/src/transforms/kafka/sink_cluster/shotover_node.rs
justinweng-instaclustr: I was thinking about adding error handling for check_shotover_peers like in https://github.com/shotover/shotover-proxy/blob/0db4f69e39eec2944c6a9141bbad7b2919613aee/shotover/src/transforms/kafka/sink_cluster/scram_over_mtls.rs#L47 so that it can be restarted if somehow fails. But since I was not sure how check_shotover_peers should return a result, the error handling was not added and this loop is not useful in this case. Do you think we should handle its error at all?

rukai: Thats a good point.
For something like `StdRng::from_rng(rand::thread_rng()).unwrap();` we should return that error, log it, and restart the task from the begginning, like we do for scram_over_mtls

justinweng-instaclustr:
The regression benchmark `encode_system.local_result_v5_no_compression` is for Cassandra and hence a noise.

shotover/src/transforms/kafka/sink_cluster/shotover_node.rs
```diff
@@ -33,9 +40,85 @@ pub(crate) struct ShotoverNode {
...
+    shotover_peers: &[ShotoverNode],
+    check_shotover_peers_delay_ms: u64,
+    connect_timeout_ms: u64,
+) -> Result<(), Error> {
+    let mut shotover_peers_cycle = shotover_peers.iter().cycle();
+    let mut rng = StdRng::from_rng(rand::thread_rng())?;
+    let check_shotover_peers_delay_ms = check_shotover_peers_delay_ms as i64;
+    loop {
+        if let Some(shotover_peer) = shotover_peers_cycle.next() {
+            let tcp_stream = tcp_stream(
```
rukai, line 100: This can be written as:
```rust
            let tcp_stream = tcp_stream(
                Duration::from_millis(connect_timeout_ms),
                (
                    shotover_peer.address.host.as_str(),
                    shotover_peer.address.port as u16,
                ),
            )
            .await;
```
Which will avoid the cost of allocation of the string here.

shotover/src/transforms/kafka/sink_cluster/shotover_node.rs
```diff
@@ -33,9 +40,85 @@ pub(crate) struct ShotoverNode {
...
+                };
+            }
+        });
+    }
+}
+
+async fn check_shotover_peers(
+    shotover_peers: &[ShotoverNode],
+    check_shotover_peers_delay_ms: u64,
+    connect_timeout_ms: u64,
```
rukai, line 93: I agree that check_shotover_peers_delay_ms should stay as is so we can easily perform the +- random jitter.
However, `connect_timeout_ms: u64` should be converted to a `connect_timeout: Duration` earlier, in `KafkaSinkClusterBuilder::new` would be reasonable.
When there is a stronger typed representation available its better to make use of that as much as we can.

shotover/src/transforms/kafka/sink_cluster/shotover_node.rs
```diff
@@ -33,9 +40,85 @@ pub(crate) struct ShotoverNode {
...
+            )
+            .await;
+            match tcp_stream {
+                Ok(_) => {
+                    shotover_peer.set_state(ShotoverNodeState::Up);
+                }
+                Err(_) => {
+                    tracing::warn!(
+                        "Shotover peer {} is down",
+                        shotover_peer.address.to_string()
```
rukai, line 112: Since `KafkaAddress` implements `Display`, we can do:

```suggestion
                        shotover_peer.address
```

This will prevent an allocation and also include the port in the warning which might be required to uniquely identify the address.

shotover/src/transforms/kafka/sink_cluster/shotover_node.rs
```diff
@@ -1,9 +1,16 @@
+use crate::tcp::tcp_stream;
 use crate::transforms::kafka::sink_cluster::kafka_node::KafkaAddress;
 use atomic_enum::atomic_enum;
 use kafka_protocol::messages::BrokerId;
 use kafka_protocol::protocol::StrBytes;
+use rand::rngs::StdRng;
+use rand::{Rng, SeedableRng};
+use rand_core::Error;
```
rukai, line 8: Anytime we dont care about the type of the error but just need to bubble something up so that a process can be aborted/restarted and an error logged we should use the anyhow error.
That is because it is very easy to work with and most errors will automatically convert into this error type.
If we needed a concrete error type that we could match on and handle in different ways then we should use the original error type instead.

In this case we should use `anyhow::Error` instead, since the error is only used for logging and triggering a restart of the task. This will allow us to return other error types in the future if we find that we add more logic that can error.

shotover-proxy/tests/kafka_int_tests/mod.rs
```diff
@@ -400,6 +400,64 @@ async fn cluster_2_racks_multi_shotover(#[case] driver: KafkaDriver) {
...
+            shotover.shutdown_and_then_consume_events(&[EventMatcher::new()
+                .with_level(Level::Warn)
+                .with_target("shotover::transforms::kafka::sink_cluster::shotover_node")
+                .with_message(r#"Shotover peer localhost:9191 is down"#)
+                .with_count(Count::Any)]), // with count > 0 is not supported
+        )
+        .await
+        .expect("Shotover did not shutdown within 10s");
+
+        events.assert_contains(
```
rukai, line 453: lets add a comment explaining why we need to add this assert twice, I was a bit confused at first.

We should add a `Count::Some` (count > 1) to https://github.com/shotover/tokio-bin-process to make this case easier to test but lets not block this PR on it.

docs/src/transforms.md
```diff
@@ -293,6 +293,11 @@ If SCRAM authentication against the first kafka broker fails, shotover will term
     # When a timeout occurs the connection to the client is immediately closed.
     # read_timeout: 60
 
+    # Delay in milliseconds before the local shotover node checks its next peer's node state.
+    # The local shotover node will try to establish a TCP connection with the next peer in the list.
+    # If the connection cannot be established within connect_timeout_ms, then the peer is considered down.
```
rukai, line 298:
```
Delay in milliseconds before the local shotover node checks its next peer's node state.
```
Reading this sentence before the other sentences is confusing because we dont have enough context yet. I think it should make more sense if we swap up the order.

```
Shotover will regularly open a TCP connection to each of its peers to check if they are up or down.
If shotover detects that a peer is down shotover will exclude the down peer from its metadata reports to the client.
Each peer is checked in a round robin fashion and this `check_shotover_peers_delay_ms` field defines the milliseconds delay taken before moving onto the next peer to check.
If the connection cannot be established within connect_timeout_ms, then the peer is considered down.
```

rukai (changes requested):
Looking good, I've left some minor feedback.

shotover-proxy/tests/kafka_int_tests/mod.rs
justinweng-instaclustr: Yeah I'll open a PR to add Count::Some to https://github.com/shotover/tokio-bin-process, and update this test in the 3rd PR for handling down shotover nodes

shotover/src/transforms/kafka/sink_cluster/shotover_node.rs
justinweng-instaclustr: Thanks for the tip!

shotover/src/transforms/kafka/sink_cluster/shotover_node.rs
justinweng-instaclustr: Nice

docs/src/transforms.md
justinweng-instaclustr: updated

rukai: Was the first line: `Shotover will regularly open a TCP connection to each of its peers to check if they are up or down.` intentionally excluded?

shotover-proxy/tests/kafka_int_tests/mod.rs
```diff
@@ -400,6 +400,65 @@ async fn cluster_2_racks_multi_shotover(#[case] driver: KafkaDriver) {
...
+    .await
+    .expect("Shotover did not shutdown within 10s");
+
+    // Wait for the other shotover node to detect the down node
+    tokio::time::sleep(Duration::from_secs(5)).await;
+
+    for shotover in shotovers {
+        let events = tokio::time::timeout(
+            Duration::from_secs(10),
+            shotover.shutdown_and_then_consume_events(&[EventMatcher::new()
```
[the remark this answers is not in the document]
rukai: in fact probably best to just concatenate the two lists in this case.

docs/src/transforms.md
justinweng-instaclustr: my bad, it was missed

rukai (approved):
LGTM, happy for this to land as is, we can make use of https://github.com/shotover/tokio-bin-process/pull/41 and https://github.com/shotover/tokio-bin-process/pull/42 in a follow up cleanup if we want.

ronycsdu (approved):

PR #1765, opened 2024-10-03
Fix local_shotover_broker_id in test configs

justinweng-instaclustr:
Copied from Lucas's comment below:
The `local_shotover_broker_id` field is currently only used for rack aware routing.
We weren't correctly setting it in our integration tests, the tests still passed since rack aware routing is purely an optimization and not needed for correctness.
We really should fix these configs so that we are running our integration tests with a configuration closer to production.

The way the field works is it configures the broker id that shotover should become.
Shotover looks up the entry in `shotover_nodes` that has that broker id and acts as that node.

### DIFF
--- a/shotover-proxy/tests/test-configs/kafka/cluster-1-rack/topology2.yaml
+++ b/shotover-proxy/tests/test-configs/kafka/cluster-1-rack/topology2.yaml
@@ -15,6 +15,6 @@ sources:
               - address: "127.0.0.1:9193"
                 rack: "rack0"
                 broker_id: 2
-            local_shotover_broker_id: 0
+            local_shotover_broker_id: 1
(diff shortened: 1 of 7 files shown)


rukai (approved):
Nice catch!

Context for the next reviewer:
The `local_shotover_broker_id` field is currently only used for rack aware routing.
We weren't correctly setting it in our integration tests, the tests still passed since rack aware routing is purely an optimization and not needed for correctness.
We really should fix these configs so that we are running our integration tests with a configuration closer to production.

The way the field works is it configures the broker id that shotover should become.
Shotover looks up the entry in `shotover_nodes` that has that broker id and acts as that node.

conorbros (approved):

PR #1767, opened 2024-10-04
Implement routing for ListOffsets

rukai:
From inspecting the messages passing through shotover I could see that ListOffsets is hitting NOT_LEADER_OR_FOLLOWER errors. (error code 6)

For example:
```
shotover   02:07:50.445929Z  INFO connection{id=41 source="kafka"}: shotover::transforms::debug::printer: Response: Kafka version:7 correlation_id:54 ListOffsets(ListOffsetsResponse { throttle_time_ms: 0, topics: [ListOffsetsTopicResponse { name: TopicName("partitions3_case4"), partitions: [ListOffsetsPartitionResponse { partition_index: 0, error_code: 6, old_style_offsets: [], timestamp: -1, offset: -1, leader_epoch: -1, unknown_tagged_fields: {} }], unknown_tagged_fields: {} }], unknown_tagged_fields: {} })
```

After some retries it will eventually hit the correct node and succeed:
```
shotover   02:07:51.047014Z  INFO connection{id=41 source="kafka"}: shotover::transforms::debug::printer: Response: Kafka version:7 correlation_id:60 ListOffsets(ListOffsetsResponse { throttle_time_ms: 0, topics: [ListOffsetsTopicResponse { name: TopicName("partitions3_case4"), partitions: [ListOffsetsPartitionResponse { partition_index: 0, error_code: 0, old_style_offsets: [], timestamp: -1, offset: 0, leader_epoch: 0, unknown_tagged_fields: {} }], unknown_tagged_fields: {} }], unknown_tagged_fields: {} })
```
It seems that the java driver is robust enough to retry only the parts of the request that failed, so eventually it will succesfully complete and move on. However, it is still quite slow and wasteful to error like this.
To fix these errors, we need to split and combine the request like we do for fetch and produce messages.

So this PR implements that fetch/combine logic for ListOffsets.

I added an integration test to call the listOffsets method from the admin API.
This test does not fail due to java's robust retry mechanism but I've manually verified that the errors are gone with the new routing logic and the new test adds more coverage of the driver.

Another nice outcome of this fix is that `cluster_1_rack_single_shotover::case_2_java` now completes in 100s down from 110s

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/mod.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/mod.rs
@@ -16,15 +16,17 @@ use kafka_node::{ConnectionFactory, KafkaAddress, KafkaNode, KafkaNodeState};
 use kafka_protocol::indexmap::IndexMap;
 use kafka_protocol::messages::fetch_request::FetchTopic;
 use kafka_protocol::messages::fetch_response::LeaderIdAndEpoch as FetchResponseLeaderIdAndEpoch;
+use kafka_protocol::messages::list_offsets_request::ListOffsetsTopic;
 use kafka_protocol::messages::metadata_request::MetadataRequestTopic;
 use kafka_protocol::messages::metadata_response::MetadataResponseBroker;
 use kafka_protocol::messages::produce_request::TopicProduceData;
 use kafka_protocol::messages::produce_response::LeaderIdAndEpoch as ProduceResponseLeaderIdAndEpoch;
 use kafka_protocol::messages::{
     ApiKey, BrokerId, FetchRequest, FetchResponse, FindCoordinatorRequest, FindCoordinatorResponse,
-    GroupId, HeartbeatRequest, JoinGroupRequest, LeaveGroupRequest, MetadataRequest,
-    MetadataResponse, ProduceRequest, ProduceResponse, RequestHeader, SaslAuthenticateRequest,
-    SaslAuthenticateResponse, SaslHandshakeRequest, SyncGroupRequest, TopicName,
+    GroupId, HeartbeatRequest, JoinGroupRequest, LeaveGroupRequest, ListOffsetsRequest,
+    ListOffsetsResponse, MetadataRequest, MetadataResponse, ProduceRequest, ProduceResponse,
+    RequestHeader, SaslAuthenticateRequest, SaslAuthenticateResponse, SaslHandshakeRequest,
+    SyncGroupRequest, TopicName,
 };
 use kafka_protocol::protocol::StrBytes;
 use kafka_protocol::ResponseError;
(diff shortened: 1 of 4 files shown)


shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1101,6 +1105,140 @@ impl KafkaSinkCluster {
...
+    fn route_list_offsets(&mut self, mut request: Message) -> Result<()> {
+        if let Some(Frame::Kafka(KafkaFrame::Request {
+            body: RequestBody::ListOffsets(list_offsets),
+            ..
+        })) = request.frame()
+        {
+            let routing = self.split_list_offsets_request_by_destination(list_offsets);
+
+            if routing.is_empty() {
+                // Fetch contains no topics, so we can just pick a random destination.
```
rukai, line 1169: ListOffsets

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1576,6 +1718,46 @@ impl KafkaSinkCluster {
...
+        drain: impl Iterator<Item = Message>,
+    ) -> Result<()> {
+        for mut next in drain {
+            if let Some(Frame::Kafka(KafkaFrame::Response {
+                body: ResponseBody::ListOffsets(next_list_offsets),
+                ..
+            })) = next.frame()
+            {
+                for next_topic in std::mem::take(&mut next_list_offsets.topics) {
+                    if let Some(base_response) = base_fetch
```
rukai, line 1732: base_list_offsets

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1576,6 +1718,46 @@ impl KafkaSinkCluster {
...
+                        .topics
+                        .iter_mut()
+                        .find(|topic| topic.name == next_topic.name)
+                    {
+                        for next_partition in &next_topic.partitions {
+                            for base_partition in &base_response.partitions {
```
rukai, line 1738: base_topic

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1548,6 +1698,7 @@ impl KafkaSinkCluster {
                 ))
             }
         }
+        base.invalidate_cache();
```
justinweng-instaclustr, line 1701: why do we need 2 `base.invalidate_cache();` here?

ronycsdu (approved):

shotover/src/transforms/kafka/sink_cluster/mod.rs
rukai: good catch, its a bad merge, fixed

justinweng-instaclustr (approved):

PR #1770, opened 2024-10-10
Exclude offline shotover nodes when rewriting responses

justinweng-instaclustr:
- This PR updates `rewrite_find_coordinator_response` and `rewrite_metadata_response` to exclude down shotover nodes.
- For responses where a single shotover's info is to be returned, we try to deterministically select an UP shotover node in the rack if available, and fall back to UP nodes out of the rack otherwise. Because the local shotover node is always UP, at least one UP shotover node should be available.
- For responses where many shotover nodes' info is to be returned, we try to select all UP shotover node in the rack if available, and fall back to UP nodes out of the rack otherwise.


shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -2374,169 +2376,191 @@ impl KafkaSinkCluster {
...
+                            .collect();
+
+                        if !shotover_nodes_in_rack.is_empty() {
+                            let hash = hash_partition(topic.topic_id, partition.partition_index);
+                            let shotover_node =
+                                &shotover_nodes_in_rack[hash % shotover_nodes_in_rack.len()];
+                            partition.leader_id = shotover_node.broker_id;
+                        } else {
+                            // If there is no UP shotover node in the rack, we should mark the leader as not existing by setting it to -1.
+                            partition.leader_id = BrokerId(-1);
```
rukai, line 2453: we should also fall back to out of rack shotover nodes here.

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -2716,19 +2726,35 @@ impl KafkaSinkCluster {
...
+
+                match shotover_node {
+                    Some(shotover_node) => {
+                        find_coordinator.host = shotover_node.address.host.clone();
+                        find_coordinator.port = shotover_node.address.port;
+                        find_coordinator.node_id = shotover_node.broker_id;
+                    }
+                    None => {
+                        // Return the COORDINATOR_NOT_AVAILABLE error if all shotover nodes are down
+                        find_coordinator.error_code = 15;
```
rukai, line 2755: Its impossible for every shotover node to be down since the current shotover node must be up, otherwise this code wouldnt be running

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -2738,180 +2764,229 @@ impl KafkaSinkCluster {
...
+
+                    let shotover_node: Option<&&ShotoverNode> = shotover_nodes_in_rack
+                        .get(hash % shotover_nodes_in_rack.len())
+                        .or_else(|| {
+                            shotover_nodes_out_of_rack.get(hash % shotover_nodes_out_of_rack.len())
+                        });
+
+                    match shotover_node {
+                        Some(shotover_node) => {
+                            find_coordinator.host = shotover_node.address.host.clone();
```
rukai, line 2799: here is the encoding problem, we are setting find_coordinator.host/port/node_id instead of coordinator.host/port/node_id

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -2716,19 +2726,35 @@ impl KafkaSinkCluster {
...
-                find_coordinator.node_id = shotover_node.broker_id;
+                let hash = hash_str_bytes(request.key.clone());
+
+                let (shotover_nodes_in_rack, shotover_nodes_out_of_rack): (
+                    Vec<&ShotoverNode>,
+                    Vec<&ShotoverNode>,
+                ) = up_shotover_nodes.iter().partition(|shotover_node| {
+                    coordinator_rack
+                        .map(|rack| rack == &shotover_node.rack)
+                        .unwrap_or(true)
```
justinweng-instaclustr, line 2738: Hey @rukai should it still be `unwrap_or(true)` here? i.e., if rack is none, should we consider shotover nodes in or out of the "rack"?

rukai: yeah, `unwrap_or(true)` still makes sense.
If `coordinator_rack` is None then the kafka broker is configured without a rack.
So the in or out of rack distinction is meaningless and we can just pick any shotover broker to represent it.

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -2374,169 +2376,191 @@ impl KafkaSinkCluster {
...
-                    coordinator.host = shotover_node.address.host.clone();
-                    coordinator.port = shotover_node.address.port;
-                    coordinator.node_id = shotover_node.broker_id;
+                    }) {
+                        coordinator.host = shotover_node.address.host.clone();
+                        coordinator.port = shotover_node.address.port;
+                        coordinator.node_id = shotover_node.broker_id;
+                    } else {
+                        // Return the COORDINATOR_NOT_AVAILABLE error
+                        coordinator.error_code = 15;
```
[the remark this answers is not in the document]
justinweng-instaclustr: added the logic to fallback to out of rack nodes

shotover/src/transforms/kafka/sink_cluster/mod.rs
justinweng-instaclustr: yeah removed this COORDINATOR_NOT_AVAILABLE and updated the comment

shotover/src/transforms/kafka/sink_cluster/mod.rs
justinweng-instaclustr: fixed

shotover-proxy/tests/test-configs/kafka/cluster-3-racks/docker-compose.yaml
```diff
@@ -0,0 +1,122 @@
...
+    networks:
+      cluster_subnet:
+        ipv4_address: 172.16.1.9
+    environment:
+      <<: *environment
+      KAFKA_CFG_ADVERTISED_LISTENERS: "BROKER://172.16.1.9:9092"
+      KAFKA_CFG_NODE_ID: 7
+      KAFKA_CFG_BROKER_RACK: "rack3"
+    volumes: *volumes
+  kafka8:
```
rukai, line 112: CI cant support this many brokers and its causing failures like https://github.com/shotover/shotover-proxy/actions/runs/11340425035/job/31582708235?pr=1770
4 brokers is the highest I've gotten it to go while remaining stable.

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -2594,169 +2611,211 @@ impl KafkaSinkCluster {
...
-                                .map(|rack| rack == &shotover_node.rack)
-                                .unwrap_or(true)
-                        })
-                        .unwrap();
-                    coordinator.host = shotover_node.address.host.clone();
-                    coordinator.port = shotover_node.address.port;
-                    coordinator.node_id = shotover_node.broker_id;
+
+                    let hash = hash_str_bytes(coordinator.key.clone());
+                    let (shotover_nodes_in_rack, shotover_nodes_out_of_rack) =
```
rukai, line 2616: lets move this out of the loop to avoid recreating these vecs every loop.

rukai: Actually... its very rare to actually handle multiple coordinators in a single message.
Maybe we are better off combining partition_shotover_nodes_by_rack and select_shotover_node_by_hash into as single call.

rukai: Oh oops, I missed that partition_shotover_nodes_by_rack and select_shotover_node_by_hash are used outside of this function. Never mind then.

shotover-proxy/tests/test-configs/kafka/cluster-3-racks/docker-compose.yaml
justinweng-instaclustr: ok, sounds like we can only support 2 racks instead of 3 racks? If that's the case, I'll test "Send some requests, kill shotover nodes 2/3 and then send more requests" for 3 shotover nodes in the same rack

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -2860,3 +2925,50 @@ impl<'a> std::fmt::Display for FormatTopicName<'a> {
...
+        None
+    }
+}
+
+/// Returns a set of unique broker IDs from available shotover nodes.
+/// Prioritizes nodes in the rack if available, falls back to nodes out of the rack.
+fn collect_unique_broker_ids(
+    shotover_nodes_in_rack: Vec<&ShotoverNode>,
+    shotover_nodes_out_of_rack: Vec<&ShotoverNode>,
+) -> HashSet<BrokerId> {
```
rukai, line 2966: Instead of converting to a hashset which requires extra allocation and hashing (expensive) we should have an assertion on startup that each ShotoverNode has a unique BrokerId

justinweng-instaclustr: makes sense. will do

shotover-proxy/tests/test-configs/kafka/cluster-3-racks/docker-compose.yaml
justinweng-instaclustr: actually I'll probably keep the 3-rack test cases and only assign 1 kafka broker to each rack to reduce the number of brokers, as 3 rack is a more common deployment configuration than 2 rack and worth testing.

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -2860,3 +2925,50 @@ impl<'a> std::fmt::Display for FormatTopicName<'a> {
...
+    rack: &Option<&StrBytes>,
+) -> (Vec<&'a ShotoverNode>, Vec<&'a ShotoverNode>) {
+    shotover_nodes
+        .iter()
+        .partition(|shotover_node| rack.map(|rack| rack == &shotover_node.rack).unwrap_or(true))
+}
+
+/// Tries to deterministically select a shotover node based on a hash value.
+/// Prioritizes nodes in the rack if available, falls back to nodes out of the rack.
+fn select_shotover_node_by_hash<'a>(
```
rukai, line 2943: Two suggestions here:
1.  lets treat having at least one shotover node as an invariant and panic if that isnt true
2. we can use `.or_else` to avoid the is_empty checks.
```
fn select_shotover_node_by_hash<'a>(
    shotover_nodes_in_rack: Vec<&'a ShotoverNode>,
    shotover_nodes_out_of_rack: Vec<&'a ShotoverNode>,
    hash: usize,
) -> &'a ShotoverNode {
    shotover_nodes_in_rack
        .get(hash % shotover_nodes_in_rack.len())
        .cloned()
        .or_else(|| {
            shotover_nodes_out_of_rack
                .get(hash % shotover_nodes_out_of_rack.len())
                .cloned()
        })
        .expect("There must be at least one shotover node")
}
```

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -2860,3 +2925,50 @@ impl<'a> std::fmt::Display for FormatTopicName<'a> {
...
     }
 }
+
+/// Partitions the given shotover nodes based on their rack placement.
+/// Returns a tuple of two vectors: (shotover_nodes_in_rack, shotover_nodes_out_of_rack).
+/// If rack is None, all shotover nodes are considered to be in the rack.
+fn partition_shotover_nodes_by_rack<'a>(
+    shotover_nodes: &[&'a ShotoverNode],
+    rack: &Option<&StrBytes>,
+) -> (Vec<&'a ShotoverNode>, Vec<&'a ShotoverNode>) {
```
rukai, line 2935: I am very cautious about returning tuples containing values of the same type, since its easy to get it mixed up which value is which.

Can we define a:
```rust
struct ShotoverNodes<'a> {
  in_rack: Vec<&'a ShotoverNode>,
  out_of_rack: Vec<&'a ShotoverNode>
}
```
and return that instead?
It will also mean we can take `ShotoverNodes` as an argument in functions like `select_shotover_node_by_hash`.

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -2594,169 +2611,211 @@ impl KafkaSinkCluster {
...
-                            {
-                                shotover_isr_nodes.push(shotover_node.broker_id);
-                            }
+
+                    // Every replica node has its entire corresponding shotover rack included.
+                    // Since we can set as many replica nodes as we like, we take this all out approach.
+                    // This ensures that:
+                    // * metadata is deterministic and therefore the same on all UP shotover nodes (note that metadata can be different on DOWN shotover nodes)
+                    // * clients evenly distribute their queries across UP shotover nodes
+                    let mut shotover_replica_nodes = vec![];
```
justinweng-instaclustr, line 2712: change this to hashset to avoid duplicate

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -2594,169 +2611,211 @@ impl KafkaSinkCluster {
...
+                .iter()
+                .find(|node| node.broker_id == metadata.controller_id)
+                .map(|x| x.rack.clone())
+            {
+                // If broker has no rack - use the first UP shotover node.
+                // If broker has rack - use the first UP shotover node with the same rack if available,
+                // and fall back to use the first UP shotover node out of the rack otherwise.
+                // This is deterministic because the list of UP shotover nodes is sorted.
+                let (shotover_nodes_in_rack, shotover_nodes_out_of_rack) =
+                    partition_shotover_nodes_by_rack(
```
justinweng-instaclustr, line 2788: confirm if partition ensures the original order

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -2594,169 +2611,211 @@ impl KafkaSinkCluster {
...
+                for partition in &mut topic.partitions {
+                    // Try to deterministically choose a single UP shotover node in the rack as leader based on topic + partition id
+                    if let Some(leader_rack) = self
                         .nodes
                         .iter()
-                        .find(|x| x.broker_id == *replica_node)
+                        .find(|x| x.broker_id == *partition.leader_id)
                         .map(|x| x.rack.clone())
                     {
-                        for shotover_node in &self.shotover_nodes {
```
rukai, line 2675: Looking closer at the original implementation, cant we just keep this and replace self.shotover_nodes with up_shotover_nodes?

It avoids the allocations from partition_shotover_nodes_by_rack and avoids an intermediate HashSet.

justinweng-instaclustr: But in this case, there might be an edge case where `shotover_replica_nodes` is empty because it doesn't fall back to out of rack shotover nodes. Is it possible that brokers all have rack and there's no UP shotover node in rack of each broker? If that's impossible, then we can use this approach and `shotover_replica_nodes` will never be empty

rukai: ah I see what you mean! fair enough.

shotover/src/transforms/kafka/sink_cluster/mod.rs
justinweng-instaclustr: added the check

shotover/src/transforms/kafka/sink_cluster/mod.rs
justinweng-instaclustr: I've changed it to a different version as `shotover_nodes_in_rack.get(hash % shotover_nodes_in_rack.len())` will panic if shotover_nodes_in_rack.len()=0

shotover/src/transforms/kafka/sink_cluster/mod.rs
justinweng-instaclustr: added this struct

shotover/src/transforms/kafka/sink_cluster/mod.rs
justinweng-instaclustr: If my understanding is right, partition should retain the original order as it adds to left and right by `extend_one()` by looking at the [source code](https://doc.rust-lang.org/1.81.0/src/core/iter/traits/iterator.rs.html#2186-2190).

ronycsdu (approved):

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -2613,150 +2628,166 @@ impl KafkaSinkCluster {
...
-
-                // Every replica node has its entire corresponding shotover rack included.
-                // Since we can set as many replica nodes as we like, we take this all out approach.
-                // This ensures that:
-                // * metadata is deterministic and therefore the same on all shotover nodes
-                // * clients evenly distribute their queries across shotover nodes
-                let mut shotover_replica_nodes = vec![];
-                for replica_node in &partition.replica_nodes {
-                    if let Some(rack) = self
+        if !up_shotover_nodes.is_empty() {
```
rukai, line 2644: since the local shotover node will never be down this will always be true.

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -2613,150 +2628,166 @@ impl KafkaSinkCluster {
...
+                // and fall back to use the first UP shotover node out of the rack otherwise.
+                // This is deterministic because the list of UP shotover nodes is sorted and partitioning does not change the order.
+                let shotover_nodes_by_rack = partition_shotover_nodes_by_rack(
+                    &up_shotover_nodes,
+                    &controller_node_rack.as_ref(),
                 );
+                let shotover_node = shotover_nodes_by_rack
+                    .nodes_in_rack
+                    .first()
+                    .or_else(|| shotover_nodes_by_rack.nodes_out_of_rack.first());
```
rukai, line 2766: lets use an `.expect("There will always be at least one up node")` to avoid needing to handle an error that wont happen.

shotover-proxy/tests/kafka_int_tests/mod.rs
```diff
@@ -388,32 +387,127 @@ async fn cluster_2_racks_multi_shotover(#[case] driver: KafkaDriver) {
...
+        );
+    }
 
-    // One shotover instance per rack
+    // Wait for check_shotover_peers to start
+    tokio::time::sleep(Duration::from_secs(15)).await;
+
+    // Send some produce and consume requests
+    let connection_builder = KafkaConnectionBuilder::new(driver, "localhost:9193");
+    produce_consume_partitions3(&connection_builder, "partitions3_rf3", 1, 500).await;
```
rukai, line 458: Its worth noting that I dont think this is actually running against a topic with rf3 and 3 partitions.
Since `admin_setup` hasnt been called, its relying on kafka's auto topic creation which I believe creates a topic with RF1 and 1 partition.

But more importantly I think we need to change these tests to hold a connection open while the shotover node is killed.
See for example `produce_consume_partitions1_kafka_node_goes_down` which:
1. creates a producer and consumer
2. kills a kafka node
3. continues to use the producer and consumer.

Judging from produce_consume_partitions1_kafka_node_goes_down it looks like we'll have to skip this test on the cpp driver since it doesnt handle this case well.

It might be worth also testing creating new connections after the shotover node is killed, as you are doing currently, maybe we could run the standard test suite instead, that way we properly run admin_setup.

shotover-proxy/tests/kafka_int_tests/mod.rs
```diff
@@ -427,33 +521,65 @@ async fn cluster_2_racks_multi_shotover_with_one_shotover_down(#[case] _driver:
...
-    // Kill one shotover node
-    tokio::time::timeout(
-        Duration::from_secs(10),
-        shotovers.remove(0).shutdown_and_then_consume_events(&[]),
-    )
-    .await
-    .expect("Shotover did not shutdown within 10s");
+    // Send some produce and consume requests
+    let connection_builder = KafkaConnectionBuilder::new(driver, "localhost:9192");
+    produce_consume_partitions3(&connection_builder, "partitions3_rf3", 1, 500).await;
```
rukai, line 526: lets do the full cluster_test_suite here.

shotover-proxy/tests/kafka_int_tests/mod.rs
rukai: I'm thinking the best way to implement it would be to define a test_case function similar to produce_consume_partitions1_kafka_node_goes_down which takes a `Vec<BinProcess>` of shotover nodes to kill.

shotover/src/transforms/kafka/sink_cluster/mod.rs
justinweng-instaclustr: removed this if condition and added a comment about it

shotover/src/transforms/kafka/sink_cluster/mod.rs
justinweng-instaclustr: added

shotover-proxy/tests/kafka_int_tests/mod.rs
justinweng-instaclustr: updated

shotover-proxy/tests/kafka_int_tests/test_cases.rs
```diff
@@ -549,6 +551,159 @@ pub async fn produce_consume_partitions1_kafka_node_goes_down(
...
+                message: "initial".to_owned(),
+                key: Some("Key".to_owned()),
+                topic_name: topic_name.to_owned(),
+                offset: Some(0),
+            })
+            .await;
+
+        let num_nodes = shotover_nodes_to_kill.len();
+        // kill shotover node(s)
+        for _ in 0..num_nodes {
```
rukai, line 613: can we just use a regular vec and do `for shotover_node in shotover_nodes_to_kill` ?

shotover-proxy/tests/kafka_int_tests/test_cases.rs
```diff
@@ -549,6 +551,159 @@ pub async fn produce_consume_partitions1_kafka_node_goes_down(
     }
 }
 
+pub async fn produce_consume_partitions1_shotover_nodes_go_down(
+    driver: KafkaDriver,
+    shotover_nodes_to_kill: &mut VecDeque<BinProcess>,
+    connection_builder: &KafkaConnectionBuilder,
+    topic_name: &str,
+) {
+    if driver.is_cpp() {
```
rukai, line 560: Lets remove this since we've disabled the cpp driver from running this at the test level?

rukai (approved):

rukai:
This is great, good work on closing out this feature.

PR #1771, opened 2024-10-11
refactor splitting

rukai:
Attempt at DRY'ing up the splitting logic, seeing if its worth it or not.

fetch splitting deviates too far from the logic of the rest of the splitting methods, so I've not attempted to implement it with the new RequestSplitAndRouter trait.

Since this only saves 40 lines, normally I would say its not worth the complexity of traits here.
However there are two reasons I think we should go ahead with this:
1. The logic that is DRY'd is actually pretty tricky to get right, hence why it has so many comments explaining whats going on. So having that implemented only once is quite valuable, on the other hand while the `RequestSplitAndRouter` trait implementations take ~25 lines each to implement, they are very difficult to get wrong since its mostly just specifying types so the type checker will catch anything wrong here.
2. we are going to need to implement message splitting for at least 2 more message types and likely more. So this will save more lines in the future.

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/split.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/split.rs
@@ -1,0 +1,113 @@
+use super::KafkaSinkCluster;
+use crate::{
+    frame::{
+        kafka::{KafkaFrame, RequestBody},
+        Frame,
+    },
+    message::Message,
+};
+use kafka_protocol::{
+    indexmap::IndexMap,
+    messages::{
+        add_partitions_to_txn_request::AddPartitionsToTxnTransaction,
+        list_offsets_request::ListOffsetsTopic, produce_request::TopicProduceData,
+        AddPartitionsToTxnRequest, BrokerId, ListOffsetsRequest, ProduceRequest, TopicName,
+        TransactionalId,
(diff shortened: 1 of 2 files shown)


shotover/src/transforms/kafka/sink_cluster/split.rs
```diff
@@ -0,0 +1,113 @@
...
+        add_partitions_to_txn_request::AddPartitionsToTxnTransaction,
+        list_offsets_request::ListOffsetsTopic, produce_request::TopicProduceData,
+        AddPartitionsToTxnRequest, BrokerId, ListOffsetsRequest, ProduceRequest, TopicName,
+        TransactionalId,
+    },
+};
+use std::collections::HashMap;
+
+pub trait RequestSplitAndRouter {
+    type SubRequest;
```
justinweng-instaclustr, line 21: Should `SubRequest` be plural instead as it's IndexMap/Vec?

rukai: yeah, that makes sense to me, I've changed it.

justinweng-instaclustr (approved):

ronycsdu (approved):

PR #1773, opened 2024-10-15
Allow disabling shotover peers check

rukai:
This PR makes the check_shotover_peers_delay_ms field optional.
When it is not provided the shotover peers check task is not started meaning that shotover nodes are always considered up and no TCP connections are created against shotover peers.

I think this field should be mandatory, we should force users to configure shotover to route around down shotover nodes, however, for now, lets make it optional to make it an easier transition.


justinweng-instaclustr:
Should `transforms.md` be updated as well to mention `check_shotover_peers_delay_ms` is optional?

rukai:
yeah thats a good idea, done.

justinweng-instaclustr (approved):

ronycsdu (approved):

PR #1780, opened 2024-10-20
Split ShotoverNodeConfig::address into address_for_peers and address_for_client

rukai:
This PR splits `ShotoverNodeConfig::address` into `address_for_peers` and `address_for_clients`.
This allows shotover to support previously impossible networking setups where the shotover cluster communicates between itself using different addresses than is used by the clients to connect to the cluster.

The implementation is very straightforward: just replace the existing field with 2 new fields and use those fields in the appropriate places.

### DIFF
--- a/shotover-proxy/tests/test-configs/kafka/cluster-3-racks/topology-rack1.yaml
+++ b/shotover-proxy/tests/test-configs/kafka/cluster-3-racks/topology-rack1.yaml
@@ -6,13 +6,16 @@ sources:
       chain:
         - KafkaSinkCluster:
             shotover_nodes:
-              - address: "localhost:9191"
+              - address_for_peers: "localhost:9191"
+                address_for_clients: "localhost:9191"
                 rack: "rack1"
                 broker_id: 0
-              - address: "localhost:9192"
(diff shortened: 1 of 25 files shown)


ronycsdu (approved):

shotover-proxy/tests/test-configs/kafka/cluster-3-racks/topology-rack1.yaml
```diff
@@ -6,13 +6,16 @@ sources:
       chain:
         - KafkaSinkCluster:
             shotover_nodes:
-              - address: "localhost:9191"
+              - address_for_peers: "localhost:9191"
```
justinweng-instaclustr, line 9: should we specifically use different addresses for `address_for_peers` in this test case to ensure `check_peers` task can use different addressed?

rukai: good idea, I did it for cluster-2-racks/topology-rack*.yaml

justinweng-instaclustr (approved):

PR #1781, opened 2024-10-21
Consumer latency fix extension

rukai:
This PR demonstrates that the extended consumer latency fix can work, as demonstrated by the `produce_consume_partitions3(connection_builder, "partitions3_case4", 1_000_000, 100)` test case passing.

However I dont think its worth going ahead with at this time.
The extended solution has no performance impact when `fetch.min.bytes = 1` which is the default value.
While at the same time bringing a lot of extra complexity into the implementation which carries some risk.

In particular I would like to see that:
* shotover runs well in production with `fetch.min.bytes = 1`, in regards to both performance and correctness.
* we experience performance issues with `fetch.min.bytes > 1`

Before considering landing this implementation.

For now, the recommendation is to just not configure clients with `fetch.min.bytes > 1` when connecting to shotover.


rukai:
Closing as we are not pursuing this at the moment.
Can be reopened if we find a need for it.

Issue #1764, opened 2024-10-03
KafkaSinkCluster - Split OffsetFetch requests


rukai:
refer to comment:
https://github.com/shotover/shotover-proxy/blob/985668f371b7a5c7353c111ca5fe354514cf375f/shotover/src/transforms/kafka/sink_cluster/mod.rs#L733-L739

similar to https://github.com/shotover/shotover-proxy/issues/1666

PR #1788, opened 2024-10-28
KafkaSinkCluster: Split OffsetFetch

rukai:
closes https://github.com/shotover/shotover-proxy/issues/1764

I was not able to find a way to reproduce this, I suspect that no drivers actually make use of this functionality.
Regardless, I think we should just handle this correctly since its pretty straightforward to implement.
I'm fairly confident it is correct due to its similarity to other split messages like DeleteGroups that also route by group id.

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/mod.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/mod.rs
@@ -19,6 +19,7 @@ use kafka_protocol::messages::fetch_response::LeaderIdAndEpoch as FetchResponseL
 use kafka_protocol::messages::list_offsets_request::ListOffsetsTopic;
 use kafka_protocol::messages::metadata_request::MetadataRequestTopic;
 use kafka_protocol::messages::metadata_response::MetadataResponseBroker;
+use kafka_protocol::messages::offset_fetch_request::OffsetFetchRequestGroup;
 use kafka_protocol::messages::offset_for_leader_epoch_request::OffsetForLeaderTopic;
 use kafka_protocol::messages::produce_request::TopicProduceData;
 use kafka_protocol::messages::produce_response::{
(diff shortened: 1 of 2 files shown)


shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -899,18 +900,18 @@ impl KafkaSinkCluster {
...
+                    if header.request_api_version <= 7 {
+                        let group_id = offset_fetch.group_id.clone();
+                        self.route_to_group_coordinator(message, group_id);
                     } else {
                         // This is possibly dangerous.
                         // The client could construct a message which is valid for a specific shotover node, but not for any single kafka broker.
                         // We may need to add some logic to split the request into multiple messages going to different destinations,
                         // and then reconstruct the response back into a single response
                         //
                         // For now just pick the first group as that is sufficient for the simple cases.
```
justinweng-instaclustr, line 912: Should this line of comment be removed?

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1405,6 +1406,32 @@ impl KafkaSinkCluster {
         result
     }
 
+    /// This method removes all group ids from the DeleteGroups request and returns them split up by their destination.
```
justinweng-instaclustr, line 1409: should change DeleteGroups to OffsetFetch?

shotover/src/transforms/kafka/sink_cluster/mod.rs
rukai: good catch, fixed

shotover/src/transforms/kafka/sink_cluster/mod.rs
rukai: good catch, fixed

justinweng-instaclustr (approved):

ronycsdu (approved):

PR #1790, opened 2024-10-29
KafkaSinkCluster: route ListGroups request

rukai:
This turned out to be more complicated than I initially thought.

When `AdminClient.listConsumerGroups` is called, the java driver sends a `ListGroups` request to all brokers and then combines the responses, removing any duplicate entries.

To ensure that all groups are included shotover should also send requests to all brokers and combine them.
However we should avoid sending requests outside of shotovers rack to reduce the load of the operation.
This should still include all the results as the client will query all shotover nodes, and each shotover node will query a portion of the kafka brokers, ensuring all kafka nodes are queried.

A second commit was added to avoid running the new test case in the tests with a down shotover node, since that resulted in test failures as the down shotover node can not report its racks groups.
It might be easier to review that commit separately from the main commit.

### DIFF
--- a/shotover-proxy/tests/kafka_int_tests/mod.rs
+++ b/shotover-proxy/tests/kafka_int_tests/mod.rs
@@ -426,7 +426,7 @@ async fn cluster_1_rack_multi_shotover_with_1_shotover_down(#[case] driver: Kafk
 
     // create a new connection and produce and consume messages
     let new_connection_builder = KafkaConnectionBuilder::new(driver, "localhost:9193");
-    test_cases::cluster_test_suite(&new_connection_builder).await;
+    test_cases::cluster_test_suite_with_lost_shotover_node(&new_connection_builder).await;
 
     let mut expected_events = multi_shotover_events();
     // Other shotover nodes should detect the killed node at least once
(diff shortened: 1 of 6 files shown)


test-helpers/src/connection/kafka/mod.rs
```diff
@@ -453,6 +453,14 @@ impl KafkaAdmin {
         }
     }
 
+    pub async fn list_groups(&self) -> Vec<String> {
+        match self {
+            #[cfg(feature = "kafka-cpp-driver-tests")]
+            Self::Cpp(_) => panic!("rdkafka-rs driver does not support list_offsets"),
```
justinweng-instaclustr, line 459: list_group?

shotover-proxy/tests/kafka_int_tests/mod.rs
```diff
@@ -485,7 +485,7 @@ async fn cluster_3_racks_multi_shotover_with_2_shotover_down(#[case] driver: Kaf
 
     // create a new connection and produce and consume messages
     let new_connection_builder = KafkaConnectionBuilder::new(driver, "localhost:9193");
-    test_cases::cluster_test_suite(&new_connection_builder).await;
+    test_cases::cluster_test_suite_with_lost_shotover_node(&new_connection_builder).await;
```
justinweng-instaclustr, line 489: Why do `cluster_3_racks_multi_shotover_with_2_shotover_down` and `cluster_3_racks_multi_shotover_with_1_shotover_missing` fail with the list groups test while `cluster_1_rack_multi_shotover_with_1_shotover_down` succeeds?

rukai: `cluster_1_rack_multi_shotover_with_1_shotover_down` should also fail good catch, I think its just intermittent and managed to slip through CI.

test-helpers/src/connection/kafka/mod.rs
rukai: thanks

justinweng-instaclustr (approved):

ronycsdu (approved):

PR #1791, opened 2024-10-30
KafkaSinkCluster return error on ControlledShutdown request

rukai:
Refer to the reasoning in the comment


shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -979,6 +979,22 @@ impl KafkaSinkCluster {
...
+                    // Since the broker ID's known to the client represent shotover nodes,
+                    // the only reasonable interpretation of this request would be for the shotover with that broker id to shutdown.
+                    // But its not appropriate for shotover to shutdown:
+                    // * It cant hand off topics to other nodes like kafka does as part of its controlled shutdown process.
+                    // * It doesnt have any ACL's, any user could send this request regardless of authorization.
+                    //
+                    // So we abort the connection and log an error to signal to the client that the request failed.
+                    // It might be better to instead construct a response containing an error code,
+                    // but its not worth the complexity for such a rare request type.
+                    return Err(anyhow!("Client sent ControlledShutdown request. Shotover cannot handle this request as it is not appropriate for shotover to shutdown. The connection to the client has been closed."));
```
ronycsdu, line 996: Why is this an error for Shotover? The Shotover is unknown to the client and this ControlledShutdown request is for kafka which should be actioned. I don't understand the logic for the error message ....... "Shotover cannot handle this request as it is not appropriate for shotover to shutdown" -> I wouldn't imagine any client sending shutdown request to Shotover

rukai: The key part is:

> Since the broker ID's known to the client represent shotover nodes,

Shotover rewrites the BrokerId in all responses to be the BrokerId's assumed by the different shotover nodes.
So the client has no idea what the broker id is of the actual kafka brokers, it is entirely hidden from them.
If the client did send a ControlledShutdown request we would have to assume the BrokerId they specified was referring to a shotover node.

rukai: > The Shotover is unknown to the client.

While the client is unaware that there is a shotover proxy in place, the client also believes that the shotover nodes ARE the kafka brokers.

justinweng-instaclustr (approved):

shotover/src/transforms/kafka/sink_cluster/mod.rs
ronycsdu: Got it. Thanks.

ronycsdu (approved):

Issue #1793, opened 2024-10-30
gocql client hangs when connecting  to Cassandra via Shotover. The client hangs if a keyspace is specified


jarodabeysinghe28:
## Describe the bug
gocql client hangs when connecting  to Cassandra via Shotover. The client hangs if a keyspace is specified

## To Reproduce
Use gocql client and specify a keyspace using

cluster.Keyspace = "keyspace1"

If you use comment out cluster.Keyspace. The client is able to connect successfully.

## Configuration
```yaml
---
main_log_level: "info,shotover_proxy=info"
observability_interface: "0.0.0.0:9001"
```

### Topology
```yaml
sources:
- Cassandra:
    name: privatelink_chain
    listen_addr: 0.0.0.0:6001
    connection_limit: 3000
    tls: {certificate_path: <cert>, private_key_path: <privatekey path>}
    chain:
    - DebugPrinter
    - CassandraSinkCluster:
        first_contact_points: ['10.1.50.50:9042']
        shotover_nodes:
        - {address: '10.0.103.7:6002', data_center: DATACENTRE_1, rack: us-east-1b,
          host_id: <id 1>}
        - {address: '10.0.147.128:6003', data_center: DATACENTRE_1, rack: us-east-1c,
          host_id: <id 2>}
        - {address: '10.0.43.236:6001', data_center: DATACENTRE_1, rack: us-east-1a,
          host_id: <id 3>}
        connect_timeout_ms: 3000
        local_shotover_host_id: <shotover host id>
        tls: {certificate_authority_path: <cert path>,
          verify_hostname: false}

```

## Expected behaviour
The gocql client should connect to Cassandra without hanging

## Systems and Version:
 - OS - Debian 11.10
 - Shotover Version - 0.4.1
 - Cassandra 4.0.13, 5.0.2

## Additional context
Tested on 4.0.13 and 5.0.2 versions of Cassandra

PR #1795, opened 2024-10-31
Fix hang in gocql init

rukai:
closes https://github.com/shotover/shotover-proxy/issues/1793

This PR reproduces the issue in an integration test and then fixes the issue with a code change.

## The problem
Shotover works well with clients that perform token aware routing of prepared execute requests.
However, if the client is not performing token aware routing, shotover may not be able to route the execute request correctly.
And while the gocql driver is initializing it sends execute requests through its control connection (not routed by token) therefore hitting this issue.
As a result the gocql driver ends up in a loop as it keeps retrying but never succeeding to prepared and execute the query.

The exact issue follows this flow:

1.   client sends request to create prepared query over control connection
2.   shotover routes request to all nodes in its rack
3.   all cassandra nodes in the rack respond with success
4.   shotover combines these responses into a single success response to send back to client
6.   client attempts to execute query over control connection
7.   shotover (correctly) routes the request to its true destination which is outside of shotovers rack.
8.   cassandra responds with error because there is no such prepared query on this cassandra node since it is outside of shotovers rack and only the nodes within shotovers rack have the query prepared.
9.   shotover returns error response to client.
10. client receives error and retries the whole process, returning to step 1

## The original fix

The absolute simplest fix is to have shotover send the prepare request to all cassandra nodes in the cluster, not just the cassandra nodes in shotover's rack.

However, this has some performance issues as we now need to send a message to all nodes in the cluster, likely forcing shotover to open new connections to them, and requiring out of rack communication.
As a result I considered alternative fixes.

## ~~The final fix~~

~~The final fix used in this PR is to route the request to a random node within the rack and have that node route to the correct node.
This:~~
~~* simplifies our routing logic.~~
~~* avoids the performance concerns of routing prepare requests to all nodes.~~
~~* Has the downside that we remove the fallback to out of node requests.~~

~~The only case where this fallback would be used is if all cassandra nodes in a shotover instance's rack are down.
This is very rare, but possible.
However if we have lost an entire rack of cassandra nodes then we have lost a large chunk of the cassandra cluster, so if we start sending out of rack requests we might start to overwhelm any still functioning racks.
Better for shotover to just handle this case as an error.~~

~~The final fix is implemented as a 2nd commit ontop of the original fix which is the 1st commit.~~
Final fix is reverted, it is common to run with a rack of 1 cassandra node, so this is actually quite dangerous.

We will explore any possible performance fixes once we have clusters running with shotover with high cassandra node count.

### DIFF
--- a/shotover/src/transforms/cassandra/sink_cluster/rewrite.rs
+++ b/shotover/src/transforms/cassandra/sink_cluster/rewrite.rs
@@ -119,7 +119,7 @@ impl MessageRewriter {
                 RewriteTableTy::Prepare { clone_index } => {
                     let mut first = true;
                     for node in pool.nodes().iter() {
-                        if node.is_up && node.rack == self.local_shotover_node.rack {
+                        if node.is_up {
                             if first {
                                 let message_id = messages[*clone_index].id();
                                 self.prepare_requests_to_destination_nodes
--- a/shotover-proxy/tests/cassandra_int_tests/mod.rs
+++ b/shotover-proxy/tests/cassandra_int_tests/mod.rs
@@ -467,6 +467,51 @@ async fn cluster_multi_rack_2_per_rack(#[case] driver: CassandraDriver) {
     cluster::multi_rack::test_topology_task(None, expected_nodes, 16).await;
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn cluster_multi_rack_2_per_rack_go_smoke_test() {
+    let _compose = docker_compose(
+        "tests/test-configs/cassandra/cluster-multi-rack-2-per-rack/docker-compose.yaml",
+    );
+
+    let shotover_rack1 = shotover_process(
+        "tests/test-configs/cassandra/cluster-multi-rack-2-per-rack/topology_rack1.yaml",
+    )
+    .with_config("tests/test-configs/shotover-config/config1.yaml")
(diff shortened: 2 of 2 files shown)


ronycsdu (approved):

justinweng-instaclustr (approved):

shotover/src/transforms/cassandra/sink_cluster/rewrite.rs
```diff
@@ -119,7 +119,7 @@ impl MessageRewriter {
                 RewriteTableTy::Prepare { clone_index } => {
                     let mut first = true;
                     for node in pool.nodes().iter() {
-                        if node.is_up && node.rack == self.local_shotover_node.rack {
```
ronycsdu, line 122: Is this going to have a performance impact on non-prepared queries as well? Meaning existing token-aware routing?

rukai: 1. This codepath is only for the routing of prepare requests, it wont affect other message types, and it does not perform token aware routing.
2. Only prepared query execute messages can perform token aware routing.

ronycsdu: Ok. Thanks.

ronycsdu (approved):

PR #1797, opened 2024-11-03
Route ListTransactions request

rukai:
This PR implements routing for ListTransactions requests by applying the same routing logic to ListTransactions as we did for ListGroups in https://github.com/shotover/shotover-proxy/pull/1790

However, I found that while the java driver deduplicates entries in ListGroups responses it does not do the same for ListTransactions responses.

So I had to alter the logic in `split_request_by_routing_to_all_brokers` to avoid duplicates.
As a result ListGroups is actually more efficient as we better distribute the load across shotover instances by avoiding querying the same node twice.

The comment on `split_request_by_routing_to_all_brokers` explains the new logic pretty well:
```rust
/// Route broadcast requests to all brokers split across all shotover nodes.
/// That is, each shotover node in a rack will deterministically be assigned a portion of the rack to route the request to.
/// If a shotover node is the only node in its rack it will route to all kafka brokers in the rack.
/// When combined with a client that is routing this request to all shotover nodes, the request will reach each kafka broker exactly once.
```

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/mod.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/mod.rs
@@ -30,10 +30,11 @@ use kafka_protocol::messages::{
     BrokerId, DeleteGroupsRequest, DeleteGroupsResponse, EndTxnRequest, FetchRequest,
     FetchResponse, FindCoordinatorRequest, FindCoordinatorResponse, GroupId, HeartbeatRequest,
     InitProducerIdRequest, JoinGroupRequest, LeaveGroupRequest, ListGroupsResponse,
-    ListOffsetsRequest, ListOffsetsResponse, MetadataRequest, MetadataResponse, OffsetFetchRequest,
-    OffsetFetchResponse, OffsetForLeaderEpochRequest, OffsetForLeaderEpochResponse, ProduceRequest,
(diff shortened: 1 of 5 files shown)


shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -1421,24 +1431,49 @@ impl KafkaSinkCluster {
...
+                        .rack
+                        .as_ref()
+                        .map(|rack| rack == &self.rack)
+                        // If the cluster is not using racks, include all brokers in the list.
+                        .unwrap_or(true)
+            })
+            // start enumerating AFTER filtering down to brokers in the rack
+            .enumerate()
+            // assign each shotover node in the rack a subset of the brokers in the rack
+            .filter(|(i, _)| i % shotovers_in_rack == local_shotover_index_in_rack)
```
justinweng-instaclustr, line 1474: Would this panic when `shotovers_in_rack=0` (i.e., all shotover nodes in the rack are down)? This scenario could be common during OS upgrades where nodes are rolling restarted.

rukai: shotovers_in_rack will always be at least 1, since the shotover this code is running in will always be included.
I added a comment to make this clear.

justinweng-instaclustr (approved):

ronycsdu (approved):

PR #1809, opened 2024-11-13
Ins 33065 update docker compose for valkey

ronycsdu:
Updated the docker_compose files in integration tests to use Valkey instead of redis.

**Note:** For backward compatibility testing, `shotover-proxy/tests/test-configs/redis/passthrough/docker-compose.yaml` has not been changed to keep the test that uses `library/redis:5.0.9 `.

### DIFF
--- a/shotover-proxy/tests/test-configs/redis/cluster-auth/docker-compose.yaml
+++ b/shotover-proxy/tests/test-configs/redis/cluster-auth/docker-compose.yaml
@@ -1,54 +1,54 @@
 services:
-  redis-node-0:
-    image: &image bitnami/redis-cluster:6.2.12-debian-11-r26
+  valkey-node-0:
+    image: &image bitnami/valkey-cluster:7.2.5
     ports:
       - "2230:6379"
     environment: &environment
-      - 'REDIS_PASSWORD=shotover'
-      - 'REDIS_NODES=redis-node-0 redis-node-1 redis-node-2 redis-node-3 redis-node-4 redis-node-5'
+      - 'VALKEY_PASSWORD=shotover'
+      - 'VALKEY_NODES=valkey-node-0 valkey-node-1 valkey-node-2 valkey-node-3 valkey-node-4 valkey-node-5'
 
-  redis-node-1:
+  valkey-node-1:
     image: *image
     ports:
       - "2231:6379"
     environment: *environment
 
-  redis-node-2:
+  valkey-node-2:
     image: *image
     ports:
       - "2232:6379"
     environment: *environment
 
-  redis-node-3:
+  valkey-node-3:
     image: *image
     ports:
       - "2233:6379"
     environment: *environment
 
-  redis-node-4:
+  valkey-node-4:
     image: *image
     ports:
       - "2234:6379"
     environment: *environment
 
-  redis-node-5:
+  valkey-node-5:
     image: *image
     ports:
       - "2235:6379"
     environment: *environment
 
-  redis-cluster-init:
+  valkey-cluster-init:
     image: *image
     depends_on:
-      - redis-node-0
-      - redis-node-1
-      - redis-node-2
-      - redis-node-3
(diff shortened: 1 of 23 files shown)


test-helpers/src/docker_compose.rs
```diff
@@ -48,6 +48,18 @@ pub static IMAGE_WAITERS: [Image; 11] = [
         log_regex_to_wait_for: r"Cluster state changed|Cluster correctly created",
```
rukai, line 48: we should be able to delete `library/redis:6.2.5`, `bitnami/redis:6.2.13-debian-11-r73` and `bitnami/redis-cluster:6.2.12-debian-11-r26` now

shotover-proxy/tests/test-configs/redis/cluster-auth/docker-compose.yaml
```diff
@@ -1,54 +1,54 @@
 services:
-  redis-node-0:
-    image: &image bitnami/redis-cluster:6.2.12-debian-11-r26
+  valkey-node-0:
+    image: &image bitnami/valkey-cluster:7.2.5
```
rukai, line 3: can we specify the whole image name, e.g. `bitnami/valkey-cluster:7.2.5-debian-12-r3`
It helps prevent CI from breaking in case an issue or incompatibility occurs in a later release of the image.

rukai (commented):
looking good, just two comments and then we should be good to merge.

test-helpers/src/docker_compose.rs
ronycsdu: I can see `redis-cluster:6.2.12-debian-11-r26` referenced in `shotover-proxy/benches/windsock/redis/bench.rs`. Should I still delete?

ronycsdu: I think the other 2 can be deleted as they are not referenced elsewhere.

rukai: good point!
We'll need to keep `redis-cluster:6.2.12-debian-11-r26` in IMAGE_WAITERS until we change windsock benches.

shotover-proxy/tests/test-configs/redis/cluster-auth/docker-compose.yaml
ronycsdu: I believe this is a specific image and versioned pinned. I would think this will alway pull 7.2.5 and not 7.2.5-debian-12-r4 or any other version. Am I understanding this wrong?

ronycsdu: I wanted to use the exact version as used in our managed platform...

ronycsdu: Screenshot 2024-11-13 at 1 13 57 PM

rukai: image
Seeing as the hash of 7.2.5 matches the hash of 7.2.5-debian-12-r4 but not 7.2.5-debian-12-r3, I would have to assume that 7.2.5 tag is changed every time a new `-rN` release is made.

rukai: > `I wanted to use the exact version as used in our managed platform`
Using any 7.2.5 release is going to be equivalent as to whats used in the managed platform.
The `-rN` at the end of the version is used for any changes to the bitnami image, say a more recent version of ubuntu packages or altering how the env vars used for config are processed. It does not indicate any change to the version of valkey that is being packaged.

ronycsdu: Ah. Got it. Will update.

rukai (approved):
Looks good!

conorbros (approved):

PR #1819, opened 2024-11-17
Support kafka java driver 3.8

rukai:
In order to reproduce the issue, this PR bumps the java driver used in integration tests from 3.7 to 3.8.
j4rs does not appear to support multiple versions of dependencies, so I think the best way to get coverage across different protocol versions is to use the java driver (well maintained and kept up to date with the latest kafka protocol) as the upper bound and the cpp driver as the lower bound (basic maintenance and uses far older versions of the kafka protocol)

The cause of the failures on 3.8 is that the driver is now sending out DescribeCluster requests which shotover does not yet support.
So this PR implements support for the DescribeCluster message type.
The implementation is very similar to the Metadata request. In fact the functionality provided by DescribeCluster is just a subset of that in Metadata. (The remaining functionality was split off into DescribeTopicPartitions)
Since they are so similar I was able to share some of the logic between both DescribeCluster and Metadata by adding a `rewrite_controller_id` method to handle the controller_id field rewriting.

### DIFF
--- a/test-helpers/src/connection/kafka/java.rs
+++ b/test-helpers/src/connection/kafka/java.rs
@@ -35,7 +35,7 @@ impl KafkaConnectionBuilderJava {
         // The list of dependencies can be found here: https://repo1.maven.org/maven2/org/apache/kafka/kafka-clients/3.7.0/kafka-clients-3.7.0.pom
         // These are deployed to and loaded from a path like target/debug/jassets
         let jvm = Jvm::new(&[
-            "org.apache.kafka:kafka-clients:3.7.0",
+            "org.apache.kafka:kafka-clients:3.8.1",
             "org.slf4j:slf4j-api:1.7.36",
             "org.slf4j:slf4j-simple:1.7.36",
         ]);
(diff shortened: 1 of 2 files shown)


test-helpers/src/connection/kafka/java.rs
```diff
@@ -35,7 +35,7 @@ impl KafkaConnectionBuilderJava {
         // The list of dependencies can be found here: https://repo1.maven.org/maven2/org/apache/kafka/kafka-clients/3.7.0/kafka-clients-3.7.0.pom
         // These are deployed to and loaded from a path like target/debug/jassets
         let jvm = Jvm::new(&[
-            "org.apache.kafka:kafka-clients:3.7.0",
+            "org.apache.kafka:kafka-clients:3.8.1",
```
ronycsdu, line 38: Will the above change be backward compatible with 3.7.0? Or are you saying that 3.7.0 will not use `RequestBody::DescribeCluster(_) at all?`

rukai: The shotover changes in this PR are compatible with both kafka client 3.7 and 3.8
DescribeCluster is not sent in 3.7, it is only sent in 3.8 onwards.

ronycsdu (approved):

justinweng-instaclustr (approved):

PR #1822, opened 2024-11-19
Updated RedisFrame to ValkeyFrame and all relevant files

ronycsdu:


rukai (approved):
This looks good to me. The remaining renames will be done in a follow up PR.

However, before we merge this, lets merge https://github.com/shotover/shotover-proxy/pull/1824 so that we can do a final 0.5.x release before putting breaking changes on main branch, making the next release 0.6.0

custom-transforms-example/config/topology.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 sources:
-  - Redis:
+  - Valkey:
       name: "redis"
```
justinweng-instaclustr, line 4: Do we want to rename this opportunistically as well? Or it should be left to the following PRs?

ronycsdu: I will do them in follow up PRs

justinweng-instaclustr (approved):

Issue #1763, opened 2024-10-03
KafkaSinkCluster - move destination field from PendingRequestState into PendingRequest


rukai:
Every variant of the PendingRequestState enum contains a `destination` field of the same `Destination` type.
This is redundant and we would be better off just storing this value once in `PendingRequest`.

Doing so should simplify things by avoiding having to pass `destination` between each variant as we move between states.

rukai:
Closed by https://github.com/shotover/shotover-proxy/pull/1828

PR #1828, opened 2024-11-21
KafkaSinkCluster - move destination field from PendingRequestState into PendingRequest

alleyshairu:
This PR address #1763 

It removes the `destination` associated field from all of the variants of the `PendingRequestState` enum, and moves this state to the parent  type `PendingRequest`.

This eliminates the need to pass the `destination` when transitioning between states.


rukai:
Thanks, this is great!
Everything passed CI, I'll hit "Update branch" and if everything still passes we'll approve the PR.
Theres been a lot of activity in this file lately so unfortunately its not unlikely that something will break.

rukai (approved):

ronycsdu (approved):

Issue #1829, opened 2024-11-21
Improve error message when running test `passthrough_nodejs_and_python` and node or python is not installed.


rukai:
When running `cargo nextest run kafka_int_tests::passthrough_nodejs_and_python`, I get the following panic:
```
thread 'kafka_int_tests::passthrough_nodejs_and_python' panicked at /home/rukai/Projects/Crates/shotover/shotover-proxy/test-helpers/src/lib.rs:48:10:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
stack backtrace:
   0: rust_begin_unwind
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/std/src/panicking.rs:662:5
   1: core::panicking::panic_fmt
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:74:14
   2: core::result::unwrap_failed
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/result.rs:1677:5
   3: unwrap<std::process::ExitStatus, std::io::error::Error>
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/result.rs:1102:23
   4: {async_fn#0}
             at /home/rukai/Projects/Crates/shotover/shotover-proxy/test-helpers/src/lib.rs:42:18
   5: {async_fn#0}
             at /home/rukai/Projects/Crates/shotover/shotover-proxy/test-helpers/src/connection/kafka/node.rs:13:50
   6: {async_block#0}
             at ./tests/kafka_int_tests/mod.rs:49:82
   7: poll<&mut dyn core::future::future::Future<Output=()>>
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/future/future.rs:123:9
   8: poll<&mut core::pin::Pin<&mut dyn core::future::future::Future<Output=()>>>
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/future/future.rs:123:9
   9: {closure#0}<core::pin::Pin<&mut core::pin::Pin<&mut dyn core::future::future::Future<Output=()>>>>
             at /home/rukai/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.41.1/src/runtime/scheduler/current_thread/mod.rs:729:57
[...]

PR #1831, opened 2024-11-21
Improve error message when `run_command_async` fails to find a command

alleyshairu:
Improves error output as defined in the #1829.

Here's a contrived sample output.
output

I would like you hear your thoughts on the testcase that I have added. The test is currently asserting on a string value instead of some custom error type. I generally do not like matching against string in my tests, but in this instance we are panicking anyways, so it is fine? or we are better off with a custom error type such as `RunCommandError`


rukai:
I agree that for errors exposed by a library a properly defined error enum is very important.
However, for this kind of adhoc application level error, I find that we are better off with the flexibility of [anyhow](https://docs.rs/anyhow/latest/anyhow).

I think that this error would actually benefit from using anyhow.
Currently we get the error:
```
called `Result::unwrap()` on an `Err` value: Custom { kind: NotFound, error: "Attempted to run the command `npm install` but npm does not exist. Have you installed nodejs?" }
```

However by swapping to anyhow we can make this a little cleaner.
```
called `Result::unwrap()` on an `Err` value: Attempted to run the command `npm install` but npm does not exist. Have you installed nodejs?
```


Personally I wouldnt have written a test here, since the functionality is improving developer experience of the test suite, not the application itself.
Regardless, I'm quite happy to accept one, it will still bring some value.

I see no issue with comparing strings in the test here, since what we want to test is that the error message is written in a specific way.

In fact the one issue with the test is that we arent properly comparing the strings.
Calling the logic under test twice and comparing the results isnt really testing anything at all.
Instead of calling `command_not_found_error_message` in the test, we should provide the specific string `For errors exposed by a library a properly defined error enum is very important` and compare against that.

test-helpers/src/lib.rs
```diff
@@ -38,16 +38,73 @@ pub fn run_command(command: &str, args: &[&str]) -> Result<String> {
...
 pub async fn run_command_async(current_dir: &Path, command: &str, args: &[&str]) {
     let output = tokio::process::Command::new(command)
         .args(args)
         .current_dir(current_dir)
         .kill_on_drop(true)
         .status()
         .await
+        .map_err(|e| {
+            if e.kind() == std::io::ErrorKind::NotFound {
+                return Error::msg(command_not_found_error_message(command, args));
```
rukai, line 70: https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.msg as per the docs on Error::msg, the `anyhow!` macro is generally preferred as it covers both the case of an error string and a wrapped error correctly.

It looks like you've manually used the function calls in the right place though, so I'm not going to block on this.

test-helpers/src/lib.rs
```diff
@@ -38,16 +38,73 @@ pub fn run_command(command: &str, args: &[&str]) -> Result<String> {
     }
 }
 
+fn command_not_found_error_message(command: &str, args: &[&str]) -> String {
```
rukai, line 41: Another nit:
Would probably make sense to return an anyhow::Error here instead of a string.
But I wont block on this either

rukai (approved):
Thanks!

test-helpers/src/lib.rs
alleyshairu: Thank you for your all feedback! Appreciate it.

rukai (approved):
Reapproving after changes.

ronycsdu (approved):

PR #1871, opened 2025-02-26
Change rustfmt edition to 2024

tan3-netapp:


rukai:
Seems like:
* You haven't run `cargo fmt` after changing the field to 2024
* C
* The CI check for formatting has broken somehow because it shouldn't be passing.

I can double check what's going on tomorrow

rukai:
ah, I see what happened, there are actually two edition fields:
1. [edition](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#edition) - this is used to set the rust edition used for parsing the rust code. This is the field we are currently using, but we should probably just remove it, since as long as we call rustfmt via `cargo fmt` it will auto detect the correct edition to use.
2. [style_edition](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#style_edition) - this is used to set the rustfmt edition for how to style the rust code. This is the field we actually want to change to 2024. Changing this will result in the changes to formatting that we expect.

rustfmt.toml
```diff
@@ -1,2 +1,3 @@
 use_field_init_shorthand = true
-edition = "2021"
+edition = "2024"
```
rukai, line 2: I think this could cause issues because the project code itself is not actually edition 2024 yet.
I think its best to remove this line (which will result in `edition = "2021"` being automatically selected)
And then keeping the below line.

rukai (approved):
Looks good!
This is the kind of fix I'm very happy to see in edition 2024:
image

ronycsdu (approved):

PR #1873, opened 2025-03-02
Integration test with syslog-ng

tan3-netapp:


shotover-proxy/tests/test-configs/valkey/syslog-ng/json/log/syslog
```diff
@@ -0,0 +1 @@
+Mar  4 15:46:30 An emergency message (Something went wrong; type syslog_1) has show up.
```
rukai, line 1: Can we add these files to the `.gitignore` instead of checking them in?

rukai:
I recommend configuring your IDE to automatically run rustfmt on save.
This makes it very difficult to push code that fails the rustfmt check.

In VS code this can be done by checking this tickbox:
image

rukai:
This is really useful for helping me understand how syslog-ng works, thanks!
Seems like we'll be able to just log json over the standard syslog protocol using the syslog crate and then syslog-ng can be configured to access the json fields and format + route using those fields.

PR #1877, opened 2025-03-05
Replace the usage of ubuntu 20.04

tan3-netapp:
Ubuntu 20.04 Actions runner image has been deprecated since Feb 2025 and is fully unsupported by April 2025. We will move away from ubuntu 20.04 to either 22.04 and 24.04.
https://github.com/actions/runner-images/issues/11101

### DIFF
--- a/windsock-cloud-docker/src/container.rs
+++ b/windsock-cloud-docker/src/container.rs
@@ -21,7 +21,7 @@ impl Container {
                 "-d",
                 "--name",
                 "windsock-cloud",
-                "ubuntu:24.04",
+                "ubuntu:22.04",
                 "sleep",
                 "infinity",
             ])
--- a/windsock-cloud-docker/src/container.rs
+++ b/windsock-cloud-docker/src/container.rs
@@ -21,7 +21,7 @@ impl Container {
                 "-d",
                 "--name",
                 "windsock-cloud",
-                "ubuntu:20.04",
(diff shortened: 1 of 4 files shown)


windsock-cloud-docker/src/container.rs
```diff
@@ -21,7 +21,7 @@ impl Container {
                 "-d",
                 "--name",
                 "windsock-cloud",
-                "ubuntu:20.04",
```
rukai, line 24: This is not part of CI but a local tool for running benchmarks on AWS.
The version here specifies which version of ubuntu to use in a local docker image for building the benchmarks.

Since its used to build the benchmark+shotover executable locally, the version needs to be below the version of ubuntu used in AWS for the test.
The version of ubuntu used in AWS is defined here: https://github.com/shotover/aws-throwaway/blob/7ea222d48b9cece8307805c69f3d74ae345cd04b/aws-throwaway/src/ec2_instance_definition.rs#L19
As you can see it still uses ubuntu 22.04.
So we should only upgrade this to 22.04 not 24.04.

tan3-netapp: Yeah, gonna update it.

rukai (approved):
Looks good, I'm not seeing anymore results for 20.04 in the repo:
image

ronycsdu (approved):

PR #1883, opened 2025-03-24
Replace Redis async connections with sync ones for pubsub testcases

tan3-netapp:
`redis::aio::Connection` has been deprecated, we need to replaced the testcases that use this connection to new type `redis::aio::MultiplexedConnection` and turn the pubsub testcases to use the sync connection (`redis::Connection`)


shotover-proxy/tests/valkey_int_tests/mod.rs
```diff
@@ -305,8 +389,15 @@ async fn cluster_dr() {
         .start()
         .await;
 
-    async fn new_connection() -> Connection {
-        let mut connection = valkey_connection::new_async("127.0.0.1", 6379).await;
+    async fn new_connection() -> MultiplexedConnection {
+        // let mut connection = valkey_connection::new_async("127.0.0.1", 6379).await;
```
rukai, line 393: commented code

shotover-proxy/tests/valkey_int_tests/mod.rs
```diff
@@ -168,7 +203,14 @@ async fn tls_source_and_tls_single_sink() {
                 .start()
                 .await;
 
-        let mut connection = valkey_connection::new_async_tls("127.0.0.1", 6379).await;
+        // let mut connection = valkey_connection::new_async_tls("127.0.0.1", 6379).await;
```
rukai, line 206: commented code

test-helpers/src/connection/valkey_connection.rs
```diff
@@ -21,55 +17,77 @@ pub fn new(port: u16) -> redis::Connection {
...
-pub async fn new_async_tls(address: &str, port: u16) -> redis::aio::Connection {
-    let certificate_authority_path = "tests/test-configs/valkey/tls/certs/localhost_CA.crt";
-    let certificate_path = "tests/test-configs/valkey/tls/certs/localhost.crt";
-    let private_key_path = "tests/test-configs/valkey/tls/certs/localhost.key";
-
-    let mut builder = SslConnector::builder(SslMethod::tls()).unwrap();
-    builder.set_ca_file(certificate_authority_path).unwrap();
-    builder
-        .set_private_key_file(private_key_path, SslFiletype::PEM)
+fn create_tls_client(address: &str, port: u16) -> Client {
```
rukai, line 28: This function can be simplified to:
```rust
fn create_tls_client(address: &str, port: u16) -> Client {
    let root_cert = std::fs::read("tests/test-configs/valkey/tls/certs/localhost_CA.crt").unwrap();
    let client_cert = std::fs::read("tests/test-configs/valkey/tls/certs/localhost.crt").unwrap();
    let client_key = std::fs::read("tests/test-configs/valkey/tls/certs/localhost.key").unwrap();

    Client::build_with_tls(
        format!("rediss://{address}:{port}/#insecure"),
        TlsCertificates {
            client_tls: Some(ClientTlsConfig {
                client_cert,
                client_key,
            }),
            root_cert: Some(root_cert),
        },
    )
    .unwrap()
}
```

test-helpers/src/connection/valkey_connection.rs
```diff
@@ -1,15 +1,11 @@
...
-use tokio::net::TcpStream;
-use tokio_io_timeout::TimeoutStream;
-use tokio_openssl::SslStream;
 
-pub fn new(port: u16) -> redis::Connection {
-    let address = "127.0.0.1";
+fn new_sync_connection(address: &str, port: u16) -> redis::Connection {
     let connection = Client::open((address, port))
         .unwrap()
         .get_connection()
```
rukai, line 11: Lets call `        .get_connection_with_timeout(Duration::from_secs(5))` instead.
This way if shotover gets stuck we get a test failure instead of a test hang.

tan3-netapp: Just one minor question from me, do the `set_read_timeout` function below do the same thing?

rukai: get_connection_with_timeout sets the timeout for opening the TCP connection.
set_read_timeout sets the timeout for performing a read on an already open TCP connection.

shotover-proxy/tests/valkey_int_tests/mod.rs
```diff
@@ -36,9 +36,21 @@ async fn passthrough_standard() {
...
-    let connection = || valkey_connection::new_async("127.0.0.1", 6379);
-    let mut flusher =
-        Flusher::new_single_connection(valkey_connection::new_async("127.0.0.1", 6379).await).await;
+    let connection = ValkeyConnectionCreator {
+        address: "127.0.0.1".into(),
+        port: 6379,
+        tls: false,
+    };
+    let mut flusher = Flusher::new_single_connection(
+        ValkeyConnectionCreator {
```
rukai, line 45: shall we just call `connection.new_async()` here instead of recreating the same ValkeyConnectionCreator?

tan3-netapp: Ah right, will change them all.

rukai (approved):
LGTM!
In the future we could combine the `address` and `port` into a single `String` field since `Client::open` and `Client::build_with_tls` both support taking a single string.
But for now lets leave it as is to avoid further complicating this PR.

ronycsdu (approved):

PR #1886, opened 2025-03-26
Upgrade dependencies that do not need code changes

tan3-netapp:
Upgrade the following dependencies, which do not need Rust code changes:
- cached
- governor
- rstest
- j4rs
- hex-literal

We use the latest `codspeed` that has the `--jobs` option but it's not tagged as release yet, so we build from the commit as specified in `bench_run.yaml` .  We will double-back to it later after we have the new version.


rukai (approved):
LGTM!

ronycsdu (approved):

PR #1891, opened 2025-04-02
Install cassandra-cpp driver via precompiled package instead of building from source

rukai:
This driver is an optional dependency used by some integration tests.

Currently we build it from source, since the maintainers did not provide .deb packages for recent ubuntu versions.
But since they now provide packages for ubuntu 22.04, this PR changes CI to download the precompiled package instead of building from scratch.

This PR reduces the following CI time:
* ~5s on cached builds (because we skip cmake install, which we could technically do with the current approach)
* ~2m on uncached builds. (because we skip compiling the cassandra-cpp driver.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -29,11 +29,6 @@ jobs:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.runner }}
     steps:
-      - name: cache custom ubuntu packages
-        uses: actions/cache@v4
-        with:
-          path: shotover-proxy/build/packages
-          key: ubuntu-22.04-packages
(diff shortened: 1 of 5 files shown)


rukai:
Because of the way github actions cache works here, it basically never clears, and we only really pay for the full cost of compilation when upgrading to a new version of the cadence driver.
And building from scratch is actually more flexible since we can better support any OS and we can build from latest master without it needing to reach a release.

Also, the cache hit case improvement could be done by just making the cmake install only run when the package doesnt exist in the cache.

So I'm going to close this PR, if we ever do find a reason that we need precompiled packages we can revisit this.

PR #1892, opened 2025-04-02
Upgrade redis to version 0.29.2

tan3-netapp:
Upgrade the `redis` crate to version `0.29.2`.

The `rustls` crate of redis forces the use of `ring` crypto backend while we have moved over to use `aws-lc-rs` for shotover since https://github.com/shotover/shotover-proxy/pull/1885. This change will bring the explicit calls to specific crypto backends (`ring` or `aws-lc-rs`) until the `redis` crate fixes this issue.


rukai:
Adding `default_provider().install_default().unwrap();` has solved the issue that shotover itself is missing a default provider.

However there is a separate issue that the integration tests and benchmarks do not have a default provider set.
Since shotover is run in a separate process by executing the shotover executable, the provider is set only in shotover not the tests/benches.

I think the best way to address this for all tests is to call `default_provider().install_default().ok();` in the `valkey_connection::create_tls_client` function.
Note that we need to use `ok()` instead of `unwrap()` here, since create_tls_client can be called multiple times in a test.

rukai (approved):
LGTM.

The `default_provider().install_default().ok();` situation is a bit unfortunate, its a bit brittle and its possible we might hit the missing provider errors again in the future, if we write a test case that doesnt hit any of the `install_default` calls. But I think that sprinkling around the `install_default` call in commonly called test helpers is the best solution for now.

In the future I expect our dependencies to remove their ring dependency, meaning that we wont have to worry about this kind of thing anymore.
For example: https://github.com/redis-rs/redis-rs/pull/1559

ronycsdu (approved):

PR #1893, opened 2025-04-02
Skip cmake install on package cache hit

rukai:
This saves a few seconds on CI run time.


rukai:
ugh.. this would uninstall cmake on user systems, which would be pretty annoying

Issue #1903, opened 2025-06-10
KafkaSinkCluster: Fix connect error handling when we dont have cluster metadata yet.


rukai:
To reproduce run `kafka_int_tests::cluster_1_rack_single_shotover::case_1_java`, with the docker_compose init commented out.

What happens is the first connection shotover opens against the cluster fails to connect, but our error handling for this assumes that we have node metadata.
When our list of nodes are empty we panic.
Instead we need to gracefully handle this condition in our error handling without panicking.
The line this happens is here: https://github.com/shotover/shotover-proxy/blob/e6587b7bffc190ba1e034727a73bc674bf2e9ca0/shotover/src/transforms/kafka/sink_cluster/mod.rs#L2206
So that unwrap needs to be replaced with an `if let` so that we do nothing when its `None`

## Motivation
This issue obfuscated the real cause while trying to resolve a local setup issue.
I imagine it could also lead to confusion in possible production misconfigurations as well.

PR #1905, opened 2025-06-15
KafkaSinkCluster: Fix connect error handling when we dont have cluster metadata yet.

tan3-netapp:
closes https://github.com/shotover/shotover-proxy/issues/1903

This issue obfuscated the real cause while trying to resolve a local setup issue.
I imagine it could also lead to confusion in possible production misconfigurations as well.
Solution: that unwrap needs to be replaced with an if let so that we do nothing when its None

### DIFF
--- a/shotover/src/transforms/kafka/sink_cluster/mod.rs
+++ b/shotover/src/transforms/kafka/sink_cluster/mod.rs
@@ -2190,21 +2190,23 @@ The connection to the client has been closed."
                     // Instead just directly set the node as down and return the error
 
                     // set node as down
-                    self.nodes
-                        .iter()
-                        .find(|x| match destination {
-                            Destination::Id(id) => x.broker_id == id,
-                            Destination::ControlConnection => {
-                                &x.kafka_address
-                                    == self
-                                        .connections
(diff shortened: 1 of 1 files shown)


rukai (approved):

shotover/src/transforms/kafka/sink_cluster/mod.rs
```diff
@@ -2190,21 +2190,23 @@ The connection to the client has been closed."
...
+                            &x.kafka_address
+                                == self
+                                    .connections
+                                    .control_connection_address
+                                    .as_ref()
+                                    .unwrap()
+                        }
+                    });
+
+                    // Leave the Kafka node as it as in case it's None. It's then bubbled up
```
rukai, line 2205: minor:
The comment should probably mention *why* the node might be None.

Maybe like this:
```rust
// kafka_node will be None when the cluster metadata hasnt been populated yet.
// This is guaranteed to happen when the cluster is down and shotover has never connected to the cluster.
// In that case we dont set anything and allow regular error handling to continue.
```

ronycsdu (approved):

PR #1906, opened 2025-06-17
rust 1.87 CI fix

ric-pro:
Cleared all warnings after the update 1.87.

### DIFF
--- a/shotover/src/message/mod.rs
+++ b/shotover/src/message/mod.rs
@@ -314,7 +314,7 @@ impl Message {
             MessageInner::Modified {
                 frame: Frame::Dummy,
             } => Encodable::Bytes(Bytes::new()),
-            MessageInner::Modified { frame } => Encodable::Frame(frame),
+            MessageInner::Modified { frame } => Encodable::Frame(Box::new(frame)),
         }
     }
 
@@ -616,7 +616,7 @@ pub enum Encodable {
     /// The raw bytes the protocol should send
     Bytes(Bytes),
     /// The Frame that should be processed into bytes and then sent
-    Frame(Frame),
+    Frame(Box<Frame>),
 }
 
 #[derive(Serialize, Deserialize, PartialEq, Debug, Clone)]
--- a/shotover/src/message/mod.rs
+++ b/shotover/src/message/mod.rs
@@ -320,7 +320,7 @@ impl Message {
             MessageInner::Parsed { bytes, .. } => Encodable::Bytes(bytes),
             MessageInner::Modified { frame } => match frame.as_ref() {
                 Frame::Dummy => Encodable::Bytes(Bytes::new()),
-                 _ => Encodable::Frame(frame),
+                _ => Encodable::Frame(frame),
             },
         }
     }
--- a/shotover/src/message/mod.rs
+++ b/shotover/src/message/mod.rs
@@ -320,7 +320,7 @@ impl Message {
             MessageInner::Parsed { bytes, .. } => Encodable::Bytes(bytes),
             MessageInner::Modified { frame } => match *frame {
                 Frame::Dummy => Encodable::Bytes(Bytes::new()),
-                frame => Encodable::Frame(frame),
+                frame => Encodable::Frame(Box::new(frame)),
             },
         }
     }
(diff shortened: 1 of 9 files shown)


rukai (changes requested):
For the lint on the `ParsingResponse` type, I'm not really fussed whether we ignore the lint or fix it, our opensearch support is only a basic prototype so not worth benchmarking to find which approach is actually faster.
Lets keep it as is.

For the lint on the `Encodable` type, your approach makes sense to me and will fit in really well with my recommendation in the next paragraph. Lets keep it as is.

For the lint on the `MessageOrId`, `Response` and `PendingRequestState` types, instead of wrapping `Message` in a Box in those types, I think we should instead reduce the stack size of the Message type itself.
The [Message](https://docs.rs/shotover/latest/shotover/message/struct.Message.html) type is a fundamental part of shotover's design and `Message`s are often moved around, so moving most of Message onto the heap could be a huge win for shotover, even outside of these 3 cases that were specifically picked up by clippy.
Additionally the current approach would lead to moving Messages into and off of the heap, which would likely be a pessimization.

There are a few ways to reduce the stack size of Message.
We could move all of the fields onto a sub-type and then box that.
Or we could take the largest field and then box that.

I've demonstrated a way of reducing the size of Messages via the draft PR: https://github.com/shotover/shotover-proxy/pull/1907
What I did was modify the `MessageInner` stored in the `Message::inner` field, to box its `Frame` which is easily the biggest part of the Message.
As you can see in the codspeed results in the PR, this resulted in a nice win across our benchmark suite.
Important to note that this PR didnt bump the rust compiler version which makes it easier to compare performance of the change, since compiler changes can also affect performance in various ways.

Normally I would seek to land a change like that in its own PR, but I'm not actively contributing to shotover at the moment, while you are, so instead I suggest you pull the changes from https://github.com/shotover/shotover-proxy/pull/1907 into this PR to resolve the lints on `MessageOrId`, `Response` and `PendingRequestState`.

shotover/src/message/mod.rs
```diff
@@ -311,10 +318,10 @@ impl Message {
         match self.inner.unwrap() {
             MessageInner::RawBytes { bytes, .. } => Encodable::Bytes(bytes),
             MessageInner::Parsed { bytes, .. } => Encodable::Bytes(bytes),
-            MessageInner::Modified {
-                frame: Frame::Dummy,
-            } => Encodable::Bytes(Bytes::new()),
-            MessageInner::Modified { frame } => Encodable::Frame(frame),
+            MessageInner::Modified { frame } => match *frame {
+                Frame::Dummy => Encodable::Bytes(Bytes::new()),
+                frame => Encodable::Frame(Box::new(frame)),
```
rukai, line 323: Looking good! one last thing to fix.

we unbox and rebox the frame here which is expensive.
Instead we can access the frame as a reference to check if its a dummy, and then move the original boxed frame into `Encodeable::Frame`.

Like this:
image

rukai (approved):
This LGTM, good work on your first PR!

## Some notes for the next reviewer:

Rust 1.87 started triggering the [large_enum_variant](https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant) lint in new locations in our codebase.
Since this is a performance recommendation its not always going to be reasonable to blindly apply this lint.

In rust boxing a type means to use the [Box](https://doc.rust-lang.org/std/boxed/index.html) type to store a value on the heap instead of the stack.

There were 5 distinct instances of this lint in our codebase, and as a result we introduced 3 new locations where we box types:
1. `State` - a type from our WIP/exploratory opensearch implementation, we decided its not worth trying to figure out what the best approach here, because this implementation is still WIP. So we just marked as `#[ignore(clippy::large_enum_variant)]`
2. `MessageOrId`, `Response`, `PendingRequestState` after looking into these types it was found that they are all hitting this lint due to the large size of the `Message` type. So to resolve all of these and future issues with using the `Message` type, we sought to make the `Message` type smaller by `Box`ing the `Frame` type stored in the Message.
3. `Encodable` - Since we now box the `Frame` stored in the `Message`, it is now cheaper to keep `Frame` boxed when we take it out of the `Message` and store it in `Encodable`. This also resolves the lint which is great!

The rest of the changes in this PR are the changes required to make the existing code work with these types now boxed.

Fixing these lints has actually improved performance of the project.
[Across the board](https://codspeed.io/shotover/shotover-proxy/branches/ric-pro%3Aupdate_rust_1.87_fix) there are many benches with 10-5% improvements.
Ignore the `decode_system.local_result_v4_no_compression`, `decode_system.local_result_v5_no_compression` benches which have regressed due to a compiler bug which will be fixed in the next rust release.

PR #1908, opened 2025-06-23
1.87 performance improvement

ric-pro:


rukai:
This was for testing https://github.com/shotover/shotover-proxy/pull/1906, closing

PR #1912, opened 2025-07-07
Hotreload interface

ric-pro:
# Summary
Implements the initial CLI interface for hot reloading as the first step towards full hot reload implementation. This implementation adds a '--hotreload' flag that enables hot reload mode and establishes the foundation for future hot reload functionality.

## Design Decisions
- Only adds a CLI interface, there is no actual hot reloading logic implemented yet.
- Integration tests verified that the integrations works

## Changes Made
- Added `--hotreload` CLI flag to enable hot reloading mode
- Integrated flag via 'hot reload_enabled' field in 'Shotover' struct
- Updated CLI documentation

## Testing
- Created Integration test to verify CLI flag acceptance.
- 'test_hotreload_cli_flag_accepted' - verifies that shotover accepts '--hotreload' and starts successfully.

## Test Coverage
- CLI flag parsing and acceptance
- Application startup with hot reload enabled

## Future works
- Unix socket communication between shotover instances.
- Socket Handoff implementation for connected transfer.


shotover-proxy/tests/test-configs/shotover-config/config_metrics_disabled.yaml
```diff
@@ -1 +1,3 @@
 main_log_level: "info, shotover::connection_span=debug"
+
```
rukai, line 2: remove these extra lines

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -0,0 +1,25 @@
...
+        .with_config("tests/test-configs/shotover-config/config_metrics_disabled.yaml")
+        .start()
+        .await;
+    // log messages
+    let events = shotover_process
+        .shutdown_and_then_consume_events(&[EventMatcher::new()
+            .with_level(Level::Info)
+            .with_target("shotover::runner")
+            .with_message(
+                "Hot reloading is ENABLED - shotover will support hot reload operations",
```
rukai, line 16: lets just remove the log assertions for now, since we cant assert on a log that occurs before shotover is fully started up.
We might reintroduce them later on when we are working on the implementation if we add a log after startup.

shotover/src/runner.rs
```diff
@@ -45,6 +45,10 @@ struct ConfigOpts {
 
     #[arg(long, value_enum, default_value = "human")]
     pub log_format: LogFormat,
+
+    //hot reloading functionality
+    #[clap(long)]
+    pub hotreload: bool, //added a new boolean flag which can be triggered using '--hotreload'
```
rukai, line 51: `//added a new boolean flag which can be triggered using '--hotreload'` comments should be about what the code does, not how it was changed. Details about how the code is changed can be included in commit messages if desired.

shotover/src/runner.rs
```diff
@@ -45,6 +45,10 @@ struct ConfigOpts {
 
     #[arg(long, value_enum, default_value = "human")]
     pub log_format: LogFormat,
+
+    //hot reloading functionality
```
rukai, line 49: If we use three `/`s the comment will become part of the documentation generated when running `--help`.
lets make use of that to document the flag.
e.g.
image
results in:
image



The existing flags dont have any documentation, those flags are very old code that doesnt get much attention, one day we should add documentation for them but documenting them is out of scope of this PR.

shotover/src/runner.rs
```diff
@@ -148,6 +163,13 @@ impl Shotover {
     pub fn run_block(self) -> ! {
         let (trigger_shutdown_tx, trigger_shutdown_rx) = watch::channel(false);
 
+        if self.hotreload_enabled {
+            info!("Starting Shotover with hotreload enabled");
+            //hot reload
```
rukai, line 168: standard format for these kinds of comments is:
`// TODO: hot reload`

test-helpers/src/shotover_process.rs
```diff
@@ -67,6 +69,12 @@ impl ShotoverProcessBuilder {
         self
     }
 
+    pub fn with_args(mut self, args: &[&str]) -> Self {
```
rukai, line 72: Rather than have each test define the args they provide, `ShotoverProcessBuilder` provides methods for each argument that we want to configure in tests.
So in this case, lets add a method `fn with_hotreload(enabled: bool)` instead of allowing tests to set arbitrary args.

shotover/src/runner.rs
rukai: I'm not sure if this is the correct place for hot reloading logic to go, but for now its as good a place as any, we can move this during implementation if we need to.

shotover/src/runner.rs
```diff
@@ -112,13 +118,22 @@ impl Shotover {
         let tracing = TracingState::new(config.main_log_level.as_str(), params.log_format)?;
         let runtime = Shotover::create_runtime(params.stack_size, params.core_threads);
 
+        if params.hotreload {
```
rukai, line 121: Lets remove these logs, we already log that hot reload is enabled in `run_block` which is enough.

shotover-proxy/tests/test-configs/shotover-config/config_metrics_disabled.yaml
```diff
@@ -1 +1 @@
-main_log_level: "info, shotover::connection_span=debug"
+main_log_level: "info, shotover::connection_span=debug"
```
rukai, line 1: I think you've deleted one too many lines here 😅 

Each line should end with exactly one newline, and this single line file has no newlines at all.

Github shows that the newline was removed via the little icon:
image

I'm not usually fussed about this kind of formatting for new files, but as a reviewer I dont want existing files that have no changes showing up in the diff.

If your confused, you should be able to restore the original file by running: `git checkout main shotover-proxy/tests/test-configs/shotover-config/config_metrics_disabled.yaml` and then just commit that.

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -0,0 +1,10 @@
+use crate::shotover_process;
+#[tokio::test]
+async fn test_hotreload_cli_flag_accepted() {
+    let shotover_process = shotover_process("tests/test-configs/null-valkey/topology.yaml")
+        .with_hotreload(true) // Use the new method we'll create
```
rukai, line 5: `// Use the new method we'll create` this comment only makes sense in the context of when its initially written or reviewed. If I read this code in 5 months, the comment wont make any sense, so lets remove it.

PR #1913, opened 2025-07-11
Hot Reload Interface testing with Valkey Database

ric-pro:
### Overview
Extends the hot reload CLI interface with integration testing using Valkey databases and Docker containers.

### Tests Added
1. 'test_hotreload_basic_redis_connection' - This test validates the '--hotreload' flag works with Valkey database
2. 'test_dual_shotover_instances_with_redis' - Tests multiple shotover instances running in parallel with hot reload enabled.

### Docker Integration
- Made use of 'docker_compose.rs' for reliable container management.

### Test Coverage
- Tested End-to-End hot reload functionality with real database
- Tested multi-instance capability which will be beneficial for future tasks like socket handoff.
- Tested that instances can communicate in between

### DIFF
--- a/shotover-proxy/tests/test-configs/hotreload/docker-compose.yaml
+++ b/shotover-proxy/tests/test-configs/hotreload/docker-compose.yaml
@@ -1,5 +1,7 @@
   services:
-    redis:
-      image: library/redis:5.0.9
+    valkey-one:
+      image: bitnami/valkey:7.2.5-debian-12-r9
       ports:
         - "6379:6379"
+      environment:
+        ALLOW_EMPTY_PASSWORD: "yes"
--- a/shotover-proxy/tests/test-configs/hotreload/topology-alt.yaml
+++ b/shotover-proxy/tests/test-configs/hotreload/topology-alt.yaml
@@ -1,11 +1,10 @@
 sources:
 - Valkey:
-    name: "Valkey-source-alt"
+    name: "valkey-source-alt"
     listen_addr: "127.0.0.1:6381" 
     chain:
       - QueryCounter:
-          name: "Valkey-query-counter-alt"
+          name: "valkey-query-counter-alt"
       - ValkeySinkSingle:
           remote_address: "127.0.0.1:6379" 
           connect_timeout_ms: 5000
-
--- a/shotover-proxy/tests/test-configs/hotreload/topology.yaml
+++ b/shotover-proxy/tests/test-configs/hotreload/topology.yaml
@@ -1,11 +1,10 @@
 sources:
 - Valkey:
-    name: "Valkey-source"
+    name: "valkey-source"
     listen_addr: "127.0.0.1:6380" 
     chain:
       - QueryCounter:
-          name: "Valkey-query-counter"
+          name: "valkey-query-counter"
       - ValkeySinkSingle:
           remote_address: "127.0.0.1:6379" 
           connect_timeout_ms: 5000
-


shotover-proxy/tests/runner/hotreload_redis_tests.rs
```diff
@@ -0,0 +1,100 @@
...
+use std::time::Duration;
+use test_helpers::docker_compose::docker_compose;
+use tokio::time::sleep;
+
+#[tokio::test]
+async fn test_hotreload_basic_redis_connection() {
+    println!(" Test 1: Basic Redis connection with hotreload");
+    let _compose = docker_compose("tests/transforms/docker-compose-redis.yaml");
+    sleep(Duration::from_millis(2000)).await;
+    println!(" Starting shotover with --hotreload...");
```
rukai, line 12: Please remove all the `println`s.
If the test has failed, then we are able to tell how far we got through the test by looking at the stacktrace of the panic.

shotover-proxy/tests/runner/hotreload_redis_tests.rs
```diff
@@ -0,0 +1,100 @@
+use crate::shotover_process;
+use redis::{Client, Commands};
+use std::time::Duration;
+use test_helpers::docker_compose::docker_compose;
+use tokio::time::sleep;
+
```
rukai, line 6: These two tests should be moved into `hotreload_int_tests` and the existing test `test_hotreload_cli_flag_accepted` should be removed.
Since the new tests cover the same functionality and more, there is no point in maintaining it as a seperate test case.

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -1,10 +1,81 @@
...
+use std::time::Duration;
+use test_helpers::docker_compose::docker_compose;
+use tokio::time::sleep;
+
 #[tokio::test]
-async fn test_hotreload_cli_flag_accepted() {
-    let shotover_process = shotover_process("tests/test-configs/null-valkey/topology.yaml")
+async fn test_hotreload_basic_redis_connection() {
+    let _compose = docker_compose("tests/transforms/docker-compose-moto.yaml");
+    sleep(Duration::from_millis(2000)).await;
```
rukai, line 10: docker_compose and shotover_process contain their own logic for waiting until shotover or docker have finished starting up. Can you try removing these sleeps and seeing if it still works fine?

shotover-proxy/tests/transforms/docker-compose-moto.yaml
```diff
@@ -3,3 +3,12 @@ services:
     image: "motoserver/moto"
     ports:
       - "5000:5000"
+  redis:
+    image: library/redis:5.0.9
```
rukai, line 7: we should use valkey for new tests instead of redis.

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -1,10 +1,81 @@
...
+use redis::{Client, Commands};
+use std::time::Duration;
+use test_helpers::docker_compose::docker_compose;
+use tokio::time::sleep;
+
 #[tokio::test]
-async fn test_hotreload_cli_flag_accepted() {
-    let shotover_process = shotover_process("tests/test-configs/null-valkey/topology.yaml")
+async fn test_hotreload_basic_redis_connection() {
+    let _compose = docker_compose("tests/transforms/docker-compose-moto.yaml");
```
rukai, line 9: this is the wrong docker compose file, its for a specific test that needs the "moto" aws mock service, which is unrelated to this test.
Usually we create a new `docker-compose.yaml` file for each test (although sometimes we share them)
In this case we should create a new `docker-compose.yaml` and put it in the same folder as our `topology.yaml` files.

On that note, looking at our `topology.yaml` files we should put them in the same folder, probably create a directory `test-configs/hotreload`.

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -1,10 +1,81 @@
...
+        .expect("Failed to connect to Redis through shotover");
+
+    let _: () = con
+        .set("test_key", "test_value")
+        .expect("Failed to SET key");
+    let result: String = con.get("test_key").expect("Failed to GET key");
+    assert_eq!(result, "test_value");
+    let pong: String = redis::cmd("PING").query(&mut con).expect("Failed to PING");
+
+    assert_eq!(pong, "PONG");
```
rukai, line 29: nit: this will be easier to read like:

```
    let pong: String = redis::cmd("PING").query(&mut con).expect("Failed to PING");
    assert_eq!(pong, "PONG");

    shotover_process.shutdown_and_then_consume_events(&[]).await;
```

Makes it clear that line 29 is related to line 27, not line 30.

shotover-proxy/tests/test-configs/hotreload/docker-compose.yaml
```diff
@@ -0,0 +1,10 @@
+  services:
+    redis:
+      image: library/redis:5.0.9
+      ports:
+        - "6379:6379"
+      healthcheck:
```
rukai, line 6: None of our other `docker-compose.yaml` files use a healthcheck, I dont think its really needed in an integration test scenario, can we remove it?

shotover-proxy/tests/test-configs/hotreload/topology-alt.yaml
```diff
@@ -0,0 +1,11 @@
+sources:
+- Valkey:
+    name: "redis-source-alt"
```
ronycsdu, line 3: are we still calling everything **redis-** instead of **valkey-** ?

shotover-proxy/tests/test-configs/hotreload/docker-compose.yaml
```diff
@@ -0,0 +1,5 @@
+  services:
+    redis:
+      image: library/redis:5.0.9
```
ronycsdu, line 3: what's the reason for using redis image instead of valkey?

rukai: We still use redis in some places to ensure backwards compatibility, so I figured it wasnt an issue to use it here as well.
If you think its important to move new code to valkey as a default, that sounds reasonable to me.

ronycsdu: As long as there are some tests using Valkey as well, I am fine with it.

shotover-proxy/tests/test-configs/hotreload/topology-alt.yaml
```diff
@@ -0,0 +1,11 @@
+sources:
+- Valkey:
+    name: "Valkey-source-alt"
```
ronycsdu, line 3: should be smaller case **v** for the names where we use Valkey?

PR #1914, opened 2025-07-21
Socket Handoff - UNIX SOCKET SERVER

ric-pro:
## UNIX SOCKET SERVER IMPLEMENTATION

Implemented unix socket server for hot reload functionality. 

### What was Implemented ?

1. Added a --hotreload-socket-path for CLI interface enhancement. This path is also configurable where users can specify custom socket paths. This was integrated with the existing CLI system

2. A new `UnixSocketServer` structure was created. It has JSON support from hot reloading messages.

3. `Shotover` struct was modified by adding hot reload fields.

4. Unit testing was done for testing socket creation.

5. Integration test was written for testing request/response protocol, request/response works correctly.


shotover/Cargo.toml
```diff
@@ -33,9 +33,10 @@ kafka = [
     "dep:xxhash-rust",
     "dep:base64",
     "dep:sasl",
+    "dep:serde_json",
 ]
-valkey = ["dep:redis-protocol", "dep:csv", "dep:crc16"]
-opensearch = ["dep:atoi", "dep:http", "dep:httparse"]
+valkey = ["dep:redis-protocol", "dep:csv", "dep:crc16", "dep:serde_json"]
+opensearch = ["dep:atoi", "dep:http", "dep:httparse", "dep:serde_json"]
```
rukai, line 39: This is a good idea, but we should make `serde_json` a non optional dependency instead.
See below `serde_json = { workspace = true, optional = true }`, just remove the `optional` field.

shotover/src/unix_socket_server.rs
```diff
@@ -0,0 +1,128 @@
...
+        }
+        let listener = UnixListener::bind(&self.socket_path).unwrap();
+        self.listener = Some(listener);
+    }
+
+    pub async fn run(&mut self) {
+        let listener = self.listener.as_ref().unwrap();
+        loop {
+            let (stream, _) = listener.accept().await.unwrap();
+            self.handle_connection(stream).await;
```
rukai, line 33: Your current design seems to use a unique connection per request/response.
This is definitely unusual, but could actually be a neat idea!

Usually a client will open a single connection and then the client and server will exchange many request/response pairs over the lifetime of the connection.
This is efficient since we reuse the connection which has a cost to setup.

However, for our use case, efficiency is less important than correctness.
And using one connection per request/response is simpler and therefore easier to get right.

So I'm happy to proceed with your current approach.

shotover/src/unix_socket_server.rs
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+
+    async fn handle_connection(&self, stream: UnixStream) {
```
rukai, line 37: For error handling here, I recommend making this function return an `anyhow::Result`.
Then the function body should return errors using the `?` operator: See: [TRPL](https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html#a-shortcut-for-propagating-errors-the--operator), [Rust reference](https://doc.rust-lang.org/reference/expressions/operator-expr.html#the-question-mark-operator)
And then in `run` add some logic to log the error and continue the loop.

shotover/src/unix_socket_server.rs
```diff
@@ -0,0 +1,128 @@
...
+        let (reader, mut writer) = stream.into_split();
+        let mut reader = BufReader::new(reader);
+        let mut line = String::new();
+
+        reader.read_line(&mut line).await.unwrap();
```
rukai, line 42: Valid json can contain newlines, so splitting by newline isn't compliant with all possible json (although we could possibly make sure we only emit json without newlines)

Since the current implementation only accepts one request per connection, we dont need to split by anything.
However the read_line logic internally will continue reading until a new line occurs, and we do need to continue reading until we have a full JSON message. (if the JSON is large enough, the first packet might not contain the entire JSON message)

So I'm thinking logic like this might be a good idea: (not tested, doesnt compile)

```rust
async fn read_request() -> Request {
  let mut received_bytes = vec!();
  loop {
    reader.read_into(&mut received_bytes).await;
    if let Ok(request) = serde_json::from_slice(&received_bytes)) {
      return request
    }
  }
```

shotover/src/hot_reload.rs
```diff
@@ -0,0 +1,77 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::collections::HashMap;
+
+    #[test]
+
+    fn test_request_serialization() {
```
rukai, line 34: I appreciate that you wrote tests here, but unfortunately what they are testing is "can serde_json round trip some values?", the `serde_json` library already has [many tests](https://github.com/serde-rs/json/tree/master/tests) answering this question, so we dont need to write our own.

I recommend removing these tests as keeping them would be a maintenance burden without any benefit.

shotover/src/runner.rs
```diff
@@ -76,10 +80,11 @@ pub struct Shotover {
     config: Config,
     tracing: TracingState,
     hotreload_enabled: bool,
+    hotreload_socket_path: String,
 }
 
 impl Shotover {
-    #[expect(clippy::new_without_default)]
+    #[allow(clippy::new_without_default)]
```
rukai, line 87: why did you change this?

shotover/src/unix_socket_server.rs
```diff
@@ -0,0 +1,128 @@
...
+impl UnixSocketServer {
+    pub fn new(socket_path: String) -> Self {
+        Self {
+            socket_path,
+            listener: None,
+        }
+    }
+
+    pub async fn start(&mut self) {
+        if Path::new(&self.socket_path).exists() {
```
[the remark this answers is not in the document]
ric-pro: Removed async. Moved all socket setup logic into the `UnixSocketServer::new() `

shotover/src/unix_socket_server.rs
ric-pro: **The initial Logic**
Expected strings to be received in the correct format - but that won't be the case in real world.

**New Logic**
Read bytes from the client to a buffer.
After each reads, we try to parse the buffer.
if parsing fails - message is incomplete - continue reading
if parsing pass - we now have a complete message that is usable
This can be implemented using a helper function which will read and try to parse the bytes until it succeeds.

shotover/src/lib.rs
```diff
@@ -75,6 +76,7 @@ pub mod tcp;
 pub mod tls;
 mod tracing_panic_handler;
 pub mod transforms;
+pub mod unix_socket_server;
```
rukai, line 79: this is specifically for hot reloading, so lets name this something like `hotreload_server`

shotover/src/runner.rs
```diff
@@ -153,21 +168,35 @@ impl Shotover {
...
+            });
         }
 
         // We need to block on this part to ensure that we immediately register these signals.
         // Otherwise if we included signal creation in the below spawned task we would be at the mercy of whenever tokio decides to start running the task.
         let (mut interrupt, mut terminate) = self.runtime.block_on(async {
             (
-                signal(SignalKind::interrupt()).unwrap(),
-                signal(SignalKind::terminate()).unwrap(),
+                tokio::signal::unix::signal(tokio::signal::unix::SignalKind::interrupt()).unwrap(),
```
rukai, line 198: why did you change this?

shotover/src/unix_socket_server.rs
```diff
@@ -0,0 +1,182 @@
...
+        let response = self.process_request(request).await;
+
+        let response_json =
+            serde_json::to_string(&response).context("Failed to serialize response")?;
+        writer
+            .write_all(response_json.as_bytes())
+            .await
+            .context("Failed to write response")?;
+        writer
+            .write_all(b"\n")
```
rukai, line 68: is this needed?

shotover/src/unix_socket_server.rs
```diff
@@ -0,0 +1,182 @@
...
+        let mut server = UnixSocketServer::new(socket_path.to_string()).unwrap();
+        // Start server in background
+        let server_handle = tokio::spawn(async move {
+            tokio::select! {
+                _ = server.run() => {},
+                _ = tokio::time::sleep(tokio::time::Duration::from_secs(2)) => {}
+            }
+        });
+        // Give server time to start
+        tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
```
rukai, line 156: If the system is really slow, the system might not be ready within 100ms. This kind of thing can happen in CI sometimes.

Additionally, sleeping in tests makes our test suite take longer to run, its not a big issue for one test, but it all adds up.

So lets do something like:
```rust
let mut stream = None;
for i in 0..1000 {
   if let Ok(new_stream) = UnixStream::connect(socket_path).await {
      stream = Some(new_stream)
      break;
   }
   tokio::time::sleep(Duration::from_millis(5)).await;
}
let stream = stream.unwrap();
```

It will continue executing as soon as the server has finished starting up.
If the server is completely broken it wait slightly longer than 5 seconds before giving up.

shotover/src/unix_socket_server.rs
```diff
@@ -0,0 +1,182 @@
...
+    async fn test_request_response() {
+        let socket_path = "/tmp/test-shotover-request-response.sock";
+        // Clean up any existing socket
+        let _ = std::fs::remove_file(socket_path);
+        let mut server = UnixSocketServer::new(socket_path.to_string()).unwrap();
+        // Start server in background
+        let server_handle = tokio::spawn(async move {
+            tokio::select! {
+                _ = server.run() => {},
+                _ = tokio::time::sleep(tokio::time::Duration::from_secs(2)) => {}
```
rukai, line 152: What is this for?
We should be able to just run the server directly without the select on the sleep

shotover/src/unix_socket_server.rs
```diff
@@ -0,0 +1,182 @@
...
+        });
+        // Give server time to start
+        tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
+        // Connect as client
+        let mut stream = UnixStream::connect(socket_path).await.unwrap();
+        // Send request
+        let request = Request::SendListeningSockets;
+        let request_json = serde_json::to_string(&request).unwrap();
+        stream.write_all(request_json.as_bytes()).await.unwrap();
+        stream.write_all(b"\n").await.unwrap();
```
rukai, line 163: shouldnt need this \n anymore

shotover/src/hot_reload_server.rs
```diff
@@ -0,0 +1,182 @@
...
+    fn drop(&mut self) {
+        if Path::new(&self.socket_path).exists() {
+            if let Err(e) = std::fs::remove_file(&self.socket_path) {
+                warn!("Failed to remove socket file {}: {:?}", self.socket_path, e);
+            }
+        }
+    }
+}
+
+// helper function for JSON reading
```
rukai, line 93: again here

shotover/src/hot_reload_server.rs
```diff
@@ -0,0 +1,182 @@
...
+
+        drop(server);
+        assert!(!Path::new(socket_path).exists());
+    }
+
+    #[tokio::test]
+    async fn test_request_response() {
+        let socket_path = "/tmp/test-shotover-request-response.sock";
+        let mut server = UnixSocketServer::new(socket_path.to_string()).unwrap();
+        // Start server in background
```
rukai, line 141: Very minor formatting stuff:
You've used small comments to highlight what blocks of code do, which is nice.
But if we better place newlines, it makes these blocks more distinct and therefore easier to read.

I'm thinking like this:
```rust
        // Start server in background
        let server_handle = tokio::spawn(async move {
            server.run().await.ok();
        });

        let mut stream = None;
        for _ in 0..1000 {
            match UnixStream::connect(socket_path).await {
                Ok(s) => {
                    stream = Some(s);
                    break;
                }
                Err(_) => {
                    use std::time::Duration;
                    tokio::time::sleep(Duration::from_millis(5)).await;
                }
            }
        }
        let mut stream =
            stream.expect("Failed to Connect to Hot Reload Unix Socket Server after waiting");
            
        // Send request
        let request = Request::SendListeningSockets;
        let request_json = serde_json::to_string(&request).unwrap();
        stream.write_all(request_json.as_bytes()).await.unwrap();
        
        // Read response
        let mut response_data = Vec::new();
        stream.read_to_end(&mut response_data).await.unwrap();
        let response: Response = serde_json::from_slice(&response_data).unwrap();
        
        // Verify response
        match response {
            Response::SendListeningSockets { port_to_fd } => {
                assert_eq!(port_to_fd.len(), 1);
                assert!(port_to_fd.contains_key(&6380));
            }
            _ => panic!("Wrong response type"),
        }
        
        // Clean up
        server_handle.abort();
        let _ = std::fs::remove_file(socket_path);
```

shotover/src/hot_reload_server.rs
```diff
@@ -0,0 +1,182 @@
...
+            match self.listener.accept().await {
+                Ok((stream, _)) => {
+                    debug!("New connection received on Unix socket");
+                    if let Err(e) = self.handle_connection(stream).await {
+                        error!("Error handling connection: {:?}", e);
+                    }
+                }
+                Err(e) => {
+                    error!("Error accepting connection: {:?}", e);
+                    continue;
```
rukai, line 44: we dont need this `continue`, the loop will continue anyway.

rukai (approved):

tan3-netapp (approved):

shotover/src/hot_reload.rs
```diff
@@ -0,0 +1,25 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct FileDescriptor(pub RawFd);
+
+/// Requests that can be sent from replacement shotover to original shotover
+#[derive(Debug, Serialize, Deserialize)]
+pub enum Request {
+    SendListeningSockets,
+    //BeginDrainingConnections,
+    //ShutdownOriginalNode,
```
tan3-netapp, line 13: Why do we have uncommented code here?

ric-pro: Those are functions which we haven't implemented yet. As part of implementation of the unix socket server client, we only implemented the SendListeningSockets function. Rest of the functions will be implemented in the next stage of development.

PR #1916, opened 2025-07-31
UNIX SOCKET HANDOFF - CLIENT SIDE IMPLEMENTATION

ric-pro:
## Task 3: Unix Socket Client for Communication

This PR implements a task for the socket handoff feature for hot reload. A Unix Socket Client that will enable replacement shotover instance to communicate with original instance during hot reload operations.

### Core Functions

- Unix Socket Client - Replacement shotover connects to Original Shotover via Unix Domain Sockets
- CLI Integration - A new cli arguement '--hotreload-from-socket' implemeted for client mode
- Request/Response Model - Client sends 'SendListeningSockets' request and receives file descriptors
- Startup Integration: Client automatically connects during shotover initialisation.

### Request flow from Client to Server

- Client connects to Unix socket at specified path
- Sends a JSON request for transferring listening sockets
- Receives a JSON response
- Logging of communication that happend

### What's Next
The communication between the client and the server was sucessful. It was tested by trying and transferring a hardcoded fd between the client and the server rather than using real FD.
Next task is to exact acutal file descriptors from the running shotover instance and then implement unix socket file description passing.

### DIFF
--- a/shotover-proxy/tests/runner/hotreload_int_tests.rs
+++ b/shotover-proxy/tests/runner/hotreload_int_tests.rs
@@ -1,6 +1,8 @@
 use crate::shotover_process;
 use redis::{Client, Commands};
+use std::time::Duration;
 use test_helpers::docker_compose::docker_compose;
+use tokio::time::sleep;
 
 #[tokio::test]
 async fn test_hotreload_basic_valkey_connection() {
@@ -24,17 +26,22 @@ async fn test_hotreload_basic_valkey_connection() {
 
 #[tokio::test]
 async fn test_dual_shotover_instances_with_valkey() {
+    let socket_path = "/tmp/test-hotreload.sock";
     let _compose = docker_compose("tests/test-configs/hotreload/docker-compose.yaml");
     let shotover_a = shotover_process("tests/test-configs/hotreload/topology.yaml")
         .with_hotreload(true)
+        .with_hotreload_socket_path(socket_path)
         .with_config("tests/test-configs/shotover-config/config_metrics_disabled.yaml")
         .start()
         .await;
     let client_a = Client::open("valkey://127.0.0.1:6380").unwrap();
     let mut con_a = client_a.get_connection().unwrap();
     let _: () = con_a.set("key_from_a", "value_from_a").unwrap();
+    sleep(Duration::from_secs(5)).await;
+
+    //Second shotover
     let shotover_b = shotover_process("tests/test-configs/hotreload/topology-alt.yaml")
-        .with_hotreload(true)
+        .with_hotreload_from_socket(socket_path)
         .with_config("tests/test-configs/shotover-config/config_metrics_disabled.yaml")
         .start()
         .await;
--- a/shotover-proxy/tests/runner/hotreload_int_tests.rs
+++ b/shotover-proxy/tests/runner/hotreload_int_tests.rs
@@ -1,8 +1,6 @@
 use crate::shotover_process;
 use redis::{Client, Commands};
-use std::time::Duration;
 use test_helpers::docker_compose::docker_compose;
-use tokio::time::sleep;
 
 #[tokio::test]
 async fn test_hotreload_basic_valkey_connection() {
@@ -37,7 +35,6 @@ async fn test_dual_shotover_instances_with_valkey() {
     let client_a = Client::open("valkey://127.0.0.1:6380").unwrap();
     let mut con_a = client_a.get_connection().unwrap();
     let _: () = con_a.set("key_from_a", "value_from_a").unwrap();
-    sleep(Duration::from_secs(5)).await;
 
     //Second shotover
     let shotover_b = shotover_process("tests/test-configs/hotreload/topology-alt.yaml")
--- a/shotover/src/hot_reload_client.rs
+++ b/shotover/src/hot_reload_client.rs
@@ -73,6 +73,36 @@ impl UnixSocketClient {
     }
 }
 
+/// Request listening sockets from an existing Shotover instance during hot reload
+pub async fn request_listening_sockets(socket_path: String) -> Result<()> {
+    info!(
+        "Hot reload CLIENT will request sockets from existing shotover at: {}",
+        socket_path
+    );
+
+    let client = UnixSocketClient::new(socket_path.clone());
+
+    match client
+        .send_request(crate::hot_reload::Request::SendListeningSockets)
+        .await
+    {
+        Ok(crate::hot_reload::Response::SendListeningSockets { port_to_fd }) => {
+            info!(
+                "Successfully received {} file descriptors from hot reload server",
+                port_to_fd.len()
+            );
+            for (port, fd) in &port_to_fd {
+                info!("Received file descriptor {} for port {}", fd.0, port);
+            }
+            Ok(())
+        }
+        Ok(crate::hot_reload::Response::Error(msg)) => {
+            Err(anyhow::anyhow!("Hot reload request failed: {}", msg))
+        }
+        Err(e) => Err(e).context("Failed to communicate with hot reload server"),
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/shotover/src/hot_reload_client.rs
+++ b/shotover/src/hot_reload_client.rs
@@ -75,7 +75,7 @@ impl UnixSocketClient {
 }
 
 /// Request listening sockets from an existing Shotover instance during hot reload
-pub async fn request_listening_sockets(socket_path: String) -> Result<()> {
+pub async fn perform_hot_reloading(socket_path: String) -> Result<()> {
     info!(
         "Hot reload CLIENT will request sockets from existing shotover at: {}",
         socket_path
@@ -143,7 +143,7 @@ mod tests {
         let mut server =
             crate::hot_reload_server::UnixSocketServer::new(socket_path.to_string()).unwrap();
         let server_handle = tokio::spawn(async move {
-            server.run().await.ok();
+            server.run().await.unwrap();
         });
 
         // Wait for server to start
@@ -173,14 +173,11 @@ mod tests {
     async fn test_multiple_client_requests() {
         let socket_path = "/tmp/test-multiple-clients.sock";
 
-        // Clean up any existing socket
-        let _ = std::fs::remove_file(socket_path);
-
         // Start server
         let mut server =
             crate::hot_reload_server::UnixSocketServer::new(socket_path.to_string()).unwrap();
         let server_handle = tokio::spawn(async move {
-            server.run().await.ok();
+            server.run().await.unwrap();
         });
 
         wait_for_unix_socket_connection(socket_path, 2000).await;
--- a/shotover/src/hot_reload_client.rs
+++ b/shotover/src/hot_reload_client.rs
@@ -107,23 +107,19 @@ pub async fn request_listening_sockets(socket_path: String) -> Result<()> {
 #[cfg(test)]
 mod tests {
     use super::*;
-
-    #[tokio::test]
-    async fn test_client_timeout() {
-        // Use a path that will cause the connection to hang, not fail immediately
-        // This simulates a server that accepts connections but doesn't respond
-        let client =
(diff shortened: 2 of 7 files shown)


shotover/src/runner.rs
```diff
@@ -171,6 +186,44 @@ impl Shotover {
...
+                        info!("Received file descriptor {} for port {}", fd.0, port);
+                    }
+                    Some(port_to_fd)
+                }
+                Ok(crate::hot_reload::Response::Error(msg)) => {
+                    error!("Hot reload request failed: {}", msg);
+                    std::process::exit(1);
+                }
+                Err(e) => {
+                    error!(
```
rukai, line 213: error logs are a bit different to println, this code would result in 3 distinct logs about 1 specific error, which is confusing.

Please make these 3 calls into a single `error!()` call, you can use `\n` or multiline strings if needed.
e.g.
```rust
error!(r#"line1
line2
line3"#);
```

shotover/src/runner.rs
```diff
@@ -171,6 +186,44 @@ impl Shotover {
...
+                        port_to_fd.len()
+                    );
+                    for (port, fd) in &port_to_fd {
+                        info!("Received file descriptor {} for port {}", fd.0, port);
+                    }
+                    Some(port_to_fd)
+                }
+                Ok(crate::hot_reload::Response::Error(msg)) => {
+                    error!("Hot reload request failed: {}", msg);
+                    std::process::exit(1);
```
rukai, line 210: `std::process::exit(1);` is a very strong hammer!
It will terminate the process immediately without letting any further error handling or cleanup from occurring.

We do use it in very specific locations in `runner.rs` but those locations are considered the final point of our error handling / cleanup. We should not be introducing more usages.

In this particular case, I agree that an error response to SendListeningSockets is not recoverable for the client and it should give up and shutdown.
However, we should move all of the hot reload logic (the server as well) into `Server::run` so that we can return an error from there without needing to call `std::process::exist`.

shotover/src/runner.rs
```diff
@@ -129,6 +138,11 @@ impl Shotover {
             tracing::info!(
                 "Hot reloading is ENABLED - shotover will support hot reload operations"
             );
+        } else if params.hotreload_from_socket.is_some() {
+            tracing::info!(
+                " Hot reload CLIENT will request sokcets from existing shotover at: {}",
```
rukai, line 143: Stray space before ` Hot`

shotover/src/runner.rs
```diff
@@ -53,6 +53,10 @@ struct ConfigOpts {
     /// Path for Unix socket used in hot reload communication
     #[clap(long, default_value = "/tmp/shotover-hotreload.sock")]
     pub hotreload_socket_path: String,
+
+    ///Connect to existing shotover for hot reloading
+    #[clap(long)]
+    pub hotreload_from_socket: Option<String>,
```
rukai, line 59: The integration test you wrote in our initial PRs should be altered to have the 2nd shotover instance set this flag, so that we can integration test the client and server running together.

shotover/src/hot_reload_client.rs
```diff
@@ -0,0 +1,218 @@
...
+        assert!(
+            result
+                .unwrap_err()
+                .to_string()
+                .contains("Failed to connect")
+        );
+    }
+
+    #[tokio::test]
+    async fn test_client_with_timeout_builder() {
```
rukai, line 137: This is testing implementation details, `timeout_duration` is a private field, so any code outside of this module cannot access it.

So I'd prefer to remove this test.

shotover/src/hot_reload_client.rs
```diff
@@ -0,0 +1,218 @@
...
+
+impl UnixSocketClient {
+    pub fn new(socket_path: String) -> Self {
+        Self {
+            socket_path,
+            timeout_duration: Duration::from_secs(30),
+        }
+    }
+
+    pub fn with_timeout(mut self, timeout: Duration) -> Self {
```
rukai, line 25: Looks like we dont actually call this in shotover, its really used for tests to reduce the timeout time.
Can we mark this as `#[cfg(test)]` to make it obvious that its only used in tests?

shotover/src/runner.rs
```diff
@@ -129,6 +138,11 @@ impl Shotover {
             tracing::info!(
                 "Hot reloading is ENABLED - shotover will support hot reload operations"
             );
+        } else if params.hotreload_from_socket.is_some() {
```
rukai, line 141: This should be written as `if let Some(socket_path) = params.hotreload_from_socket {`, then we can use `socket_path` in the info log below and avoid needing to call `.unwrap()`.

It is always preferable to use `if let` instead of `is_some()` + `unwrap()` since its easier to tell its correct at a glance.

shotover/src/hot_reload_client.rs
```diff
@@ -0,0 +1,218 @@
...
+
+        assert_eq!(client.timeout_duration, Duration::from_secs(5));
+    }
+
+    #[tokio::test]
+    async fn test_client_server_integration() {
+        let socket_path = "/tmp/test-client-server-integration.sock";
+
+        // Clean up any existing socket
+        let _ = std::fs::remove_file(socket_path);
```
rukai, line 149: `UnixSocketServer` already performs this cleanup, no need to do it here.

shotover/src/hot_reload_client.rs
```diff
@@ -0,0 +1,218 @@
...
+
+        // Start server
+        let mut server =
+            crate::hot_reload_server::UnixSocketServer::new(socket_path.to_string()).unwrap();
+        let server_handle = tokio::spawn(async move {
+            server.run().await.ok();
+        });
+
+        // Wait for server to start
+        tokio::time::sleep(Duration::from_millis(100)).await;
```
rukai, line 159: This should do the same wait logic as `hot_reload_server::test_request_response`, can we pull connection creation logic from `test_request_response` into a common function and call it here?

You can add `#[cfg(test)]` to it, to make sure that its used only by tests.

shotover/src/hot_reload_client.rs
```diff
@@ -0,0 +1,218 @@
...
+        let client =
+            UnixSocketClient::new("/dev/null".to_string()).with_timeout(Duration::from_millis(100));
+
+        let result = client.send_request(Request::SendListeningSockets).await;
+        assert!(result.is_err());
+        let error_msg = result.unwrap_err().to_string();
+
+        // The error could be either a timeout or a connection error, both are valid
+        assert!(
+            error_msg.contains("timed out") || error_msg.contains("Failed to connect"),
```
rukai, line 116: Does this ever actually return `timed out`? From my testing it doesnt. If not, I dont think theres any point in this test.

In general, it is very rarely correct to write a test that can accept multiple outcomes.
The main exception I can think of is adding logic that waits for something to become true, but even there you still want to have a single final result.

shotover/src/hot_reload_client.rs
```diff
@@ -0,0 +1,218 @@
...
+        // Read response
+        let response: Response = read_json(&mut reader).await?;
+        debug!("Received response: {:?}", response);
+
+        Ok(response)
+    }
+}
+
+// Reuse the same JSON parsing logic as server
+async fn read_json<T: DeserializeOwned, R: AsyncReadExt + Unpin>(reader: &mut R) -> Result<T> {
```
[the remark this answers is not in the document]
ric-pro: Done. Created a 'json_parsing' module. Imported the 'read_json' logic to both client and server.

shotover/src/runner.rs
ric-pro: Updated the hot reload integration test to use consistent socket paths and added builder methods for both '--hotreload-socket-path' and '--hotreload-from-socket flags'. The test now successfully validates the hot reload client/server interaction as required.

shotover/src/hot_reload_client.rs
ric-pro: Refactored the server wait/connection logic into a shared wait_for_unix_socket_connection helper, and updated both client and server test modules to use it. This removes duplication and ensures robust, consistent test setup.

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -24,17 +26,22 @@ async fn test_hotreload_basic_valkey_connection() {
...
     let shotover_a = shotover_process("tests/test-configs/hotreload/topology.yaml")
         .with_hotreload(true)
+        .with_hotreload_socket_path(socket_path)
         .with_config("tests/test-configs/shotover-config/config_metrics_disabled.yaml")
         .start()
         .await;
     let client_a = Client::open("valkey://127.0.0.1:6380").unwrap();
     let mut con_a = client_a.get_connection().unwrap();
     let _: () = con_a.set("key_from_a", "value_from_a").unwrap();
+    sleep(Duration::from_secs(5)).await;
```
rukai, line 40: This should not be needed.
If the test has a chance of failing we should alter the implementation to ensure that shotover waits for the hot reload init to complete before shotover declares itself to be ready.

shotover/src/hot_reload_client.rs
```diff
@@ -0,0 +1,207 @@
...
+        // Cleanup
+        server_handle.abort();
+    }
+
+    #[tokio::test]
+    async fn test_multiple_client_requests() {
+        let socket_path = "/tmp/test-multiple-clients.sock";
+
+        // Clean up any existing socket
+        let _ = std::fs::remove_file(socket_path);
```
rukai, line 177: again, this shouldn't be needed.

shotover/src/runner.rs
```diff
@@ -245,6 +241,73 @@ impl Shotover {
     }
 }
 
+pub struct Server {
```
rukai, line 244: What is the purpose of this Server struct?
If the purpose is to reduce the number of arguments to `run`, can we just make `run` a member of `Shotover` instead of introducing a new type that we only use once?

shotover/src/hot_reload_client.rs
```diff
@@ -0,0 +1,207 @@
...
+        // Read response
+        let response: Response = read_json(&mut reader).await?;
+        debug!("Received response: {:?}", response);
+
+        Ok(response)
+    }
+}
+
+/// Request listening sockets from an existing Shotover instance during hot reload
+pub async fn request_listening_sockets(socket_path: String) -> Result<()> {
```
rukai, line 78: I'm thinking lets give this a higher level name like `perform_hot_reloading` that way the runner.rs logic is less concerned with what goes on during the hot reload process.

shotover-proxy/tests/runner/hotreload_int_tests.rs
rukai: After looking at the code, to fix this issue you'll probably to setup a [tokio oneshot](https://docs.rs/tokio/latest/tokio/sync/oneshot/index.html). Pass the sender to the server task, and hold the receiver in main task, then you can signal that the server is setup in the server task and wait for a message to come through in the main task.

shotover/src/hot_reload_client.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    #[tokio::test]
+    async fn test_client_server_integration() {
+        let socket_path = "/tmp/test-client-server-integration.sock";
+
+        // Start server
+        let mut server =
+            crate::hot_reload_server::UnixSocketServer::new(socket_path.to_string()).unwrap();
+        let server_handle = tokio::spawn(async move {
+            server.run().await.ok();
```
rukai, line 146: This should be `server.run().await.unwrap();` otherwise failure to startup the hot reload server will be ignored, leading to either:
* a test to pass when it shouldnt
* or a test to fail in a way that is harder to track down than it needs to be.


I think we already made this mistake in a few places in our tests, can you go through and fix them all?

ric-pro: `server.run().await.ok();` was used in the server and the client. Fixed both.

rukai (approved):
LGTM!

I raised https://github.com/shotover/shotover-proxy/issues/1917 for some refactors to do afterwards to avoid making the diff harder to review in this PR.

PR #1921, opened 2025-08-18
Listening Socket Extraction for Hot Reload

ric-pro:
## Recap

The existing hot reload infrastructure was using hardcoded fake file descriptors (FileDescriptor(10)). The hot reload server had no mechanism to access real socket file descriptors from the actual TcpCodecListener instances that own the listening sockets.

### Communication Infrastructure
- Added internal message types for hot reload server to communicate with socket listeners
- Implemented async channel-based request-response pattern
- Created centralised channel management system

### Socket Listener Integration
- Modified TcpCodecListener to accept hot reload channel receivers
- Updated all source types (Cassandra, Kafka, Valkey, OpenSearch) to support hot reload

### Real Socket Extraction
- Implemented tokio::select! logic to handle both normal connections and hot reload requests
- Added actual file descriptor extraction using AsRawFd trait

## What's next?

- New instance needs to use received FDs instead of creating new sockets


shotover/src/sources/valkey.rs
```diff
@@ -60,20 +66,26 @@ impl ValkeySource {
         hard_connection_limit: Option<bool>,
         tls: Option<TlsAcceptorConfig>,
         timeout: Option<u64>,
+        hotreload_enabled: bool,
     ) -> Result<ValkeySource, Vec<String>> {
-        info!("Starting Valkey source on [{}]", listen_addr);
+        let (hot_reload_tx, hot_reload_rx) = tokio::sync::mpsc::unbounded_channel();
+        if !hotreload_enabled {
+            // Leak the sender so the receiver never sees the channel as closed.
+            std::mem::forget(hot_reload_tx.clone());
```
rukai, line 74: This technically works, since it keeps at least one sender instance alive forever.
However, I think we would be better off avoiding the clone, and forgetting the one sender at the point it would be normally dropped.

That point is the `into_join_handle` method.
Since `into_join_handle` takes `self` instead of `&self` we have ownership of self so it is dropped at the end of the method and therefore the contents of self which includes the hot_reload_tx are also dropped.
So lets add the `std::mem::forget` there.

Also, there is no need to pass in the hotreload_enabled bool, since it is always valid to drop the sender.

ric-pro: Refactored to leak hot_reload_tx only in into_join_handle for all sources.

shotover/src/hot_reload/protocol.rs
```diff
@@ -23,3 +23,14 @@ pub enum Response {
...
     Error(String),
 }
+/// Request sent from hot reload server to TcpCodecListener
+pub struct HotReloadListenerRequest {
+    pub return_chan: tokio::sync::oneshot::Sender<HotReloadListenerResponse>,
+}
+
+/// Response sent from TcpCodecListener back to hot reload server
+#[derive(Debug)]
+pub struct HotReloadListenerResponse {
```
rukai, line 33: Currently we set the fd to -1 to indicate an error, but its better to have an explicit error condition rather than magic integer values.
So lets expand this type to support reporting errors.
I am thinking we should turn it into an enum:
```rust
#[derive(Debug)]
pub enum HotReloadListenerResponse {
   HotReloadResponse { port: u16, listener_socket_fd: FileDescriptor },
   NoListenerAvailable,
}
```

shotover/src/server.rs
```diff
@@ -181,61 +190,60 @@ impl<C: CodecBuilder + 'static> TcpCodecListener<C> {
...
+                    // Normal Connection handling
+                    stream_result = Self::accept(&mut self.listener) => {
+                        let stream = stream_result?;
+
+                        debug!("got socket");
+                        self.available_connections_gauge.set(self.limit_connections.available_permits() as f64);
+                        self.connections_opened.increment(1);
+
+                        let client_details = stream.peer_addr()
+                            .map(|addr| addr.to_string())
```
rukai, line 205: Whats the motivation of this change?

shotover/src/server.rs
```diff
@@ -277,6 +285,51 @@ impl<C: CodecBuilder + 'static> TcpCodecListener<C> {
...
+    async fn handle_hot_reload_request(&mut self, request: HotReloadListenerRequest) {
+        let response = if let Some(ref listener) = self.listener {
+            // Extract the file descriptor from the TcpListener
+            let fd = listener.as_raw_fd();
+
+            let port = match listener.local_addr() {
+                Ok(addr) => addr.port(),
+                Err(_) => {
+                    // Fallback: split once from the right to support hostnames and [IPv6]:port formats.
+                    self.listen_addr
```
rukai, line 304: This is a good idea to use this as a fallback.
But we should take it one step further and just always parse the port from the `listen_addr` field. Relying on a rarely used fallback means you might have bugs hidden in the fallback implementation.

Also we should use the parser in the standard library to do this: https://doc.rust-lang.org/std/net/enum.SocketAddr.html

shotover/src/hot_reload/server.rs
```diff
@@ -1,33 +1,38 @@
...
 use std::path::Path;
 use tokio::io::{AsyncWriteExt, BufReader};
 use tokio::net::{UnixListener, UnixStream};
+use tokio::sync::mpsc;
 use tracing::{debug, error, info, warn};
 
 pub struct UnixSocketServer {
     socket_path: String,
     listener: UnixListener,
+    channel_senders: Vec<(String, mpsc::UnboundedSender<HotReloadListenerRequest>)>,
```
rukai, line 16: From just reading this line its hard to tell what is actually stored in this field.
How about we rename the field and pull the tuple out into a separate struct.
```rust
  ...
  sources: Vec<SourceHandle>
}

struct SourceHandle {
  name: String,
  sender: mpsc::UnboundedSender<HotReloadListenerRequest>
}
```

As a rule of thumb, the only time a tuple should be used in a struct/enum field is when the field name describes what is in each element of the tuple. e.g. `xy_coordinates: (i32, i32),`
For variables in a function implementation, this requirement is not needed, since you have the context of creating+using the variable.

ric-pro: updated the code by replacing the tuple with a dedicated SourceHandle struct

shotover/src/hot_reload/server.rs
```diff
@@ -133,7 +195,8 @@ mod tests {
     #[tokio::test]
     async fn test_request_response() {
         let socket_path = "/tmp/test-shotover-request-response.sock";
-        let mut server = UnixSocketServer::new(socket_path.to_string()).unwrap();
+        let source_handles: Vec<SourceHandle> = vec![]; // Empty for test // Empty for test
```
rukai, line 198: Comment written twice

shotover/src/hot_reload/server.rs
```diff
@@ -95,9 +148,17 @@ impl Drop for UnixSocketServer {
         }
     }
 }
-pub fn start_hot_reload_server(socket_path: String) {
+pub fn start_hot_reload_server(socket_path: String, sources: &[Source]) {
+    let source_handles: Vec<SourceHandle> = sources
+        .iter()
+        .map(|x| SourceHandle {
```
rukai, line 154: huh, looking at this, maybe we should combine the SourceHandle and Source types?

Maybe rename `Source` to `SourceHandle` and then we can clone the SourceHandle here instead of constructing a different type?

shotover/src/server.rs
```diff
@@ -277,6 +285,39 @@ impl<C: CodecBuilder + 'static> TcpCodecListener<C> {
...
         }
     }
+
+    /// Handle hot reload request by extracting file descriptor and responding
+    async fn handle_hot_reload_request(&mut self, request: HotReloadListenerRequest) {
+        let response = if let Some(ref listener) = self.listener {
+            // Extract the file descriptor from the TcpListener
+            let fd = listener.as_raw_fd();
+
+            // Always parse port from listen_addr to support hostnames and [IPv6]:port formats.
```
rukai, line 295: The word `Always` only makes sense knowing the context that the previous implementation had a fallback.

The previous comment was something like: `split once from the right to support hostnames and [IPv6]:port formats.`
Mentioning that splitting once from the right lets us support both of these cases is valuable IMO, so lets keep that original comment.

shotover/src/server.rs
```diff
@@ -277,6 +285,39 @@ impl<C: CodecBuilder + 'static> TcpCodecListener<C> {
...
+            let port = self
+                .listen_addr
+                .rsplit_once(':')
+                .and_then(|(_, p)| p.parse::<u16>().ok())
+                .unwrap_or(0);
```
rukai, line 300: `unwrap_or(0)` is concerning.
If its truly impossible to hit this case we should do `.unwrap()`.
If it is possible then we need some kind of error handling rather than returning an invalid port.

Shotover will fail to start if the address does not contain a valid port:
image
So lets just `.unwrap()` here.

carlsandland (approved):

Issue #1924, opened 2025-08-25
Refactor CassandraSource, ValkeySource etc. to use a single Source struct


rukai:
In Shotover a [source](https://shotover.io/docs/latest/sources.html) defines a port to listen for incoming requests.
The incoming requests are then sent through the transform chain, and then finally to the Sink transform which sends the request to the database.

Shotover has unique source types for all the protocols it supports.
This currently looks like:
* a [CassandraConfig](https://github.com/shotover/shotover-proxy/blob/b66e6f0f75fd892b63e2e8a76209ff7999d58197/shotover/src/sources/cassandra.rs#L17), `KafkaConfig` etc. type which we deserialize the [source config](https://github.com/shotover/shotover-examples/blob/038bc4171fdc57a0fb424def728c5978b9bca664/cassandra-1-1/topology.yaml#L4-L6) into from the topology.yaml, which is how users configure shotover.
* a [CassandraSource](https://github.com/shotover/shotover-proxy/blob/b66e6f0f75fd892b63e2e8a76209ff7999d58197/shotover/src/sources/cassandra.rs#L51), `KafkaSource` etc. type which holds a handle to the running source task.

While the `*Config` types need to be different to handle the unique configuration needs of different protocols, the `*Source` types just hold handles to the task that does the real work, so no need to be unique.

So to simplify our implementation and make adding new fields to the `*Source` types easier we should refactor sources to use a single `Source` type to store the task handle.

PR #1925, opened 2025-08-25
Refactor sources to use unified Source struct

ric-pro:
# Summary
- This PR addresses the issue: https://github.com/shotover/shotover-proxy/issues/1924. 
- Consolidated the protocol-specific `*Source` types (CassandraSource, KafkaSource, ValkeySource, OpenSearchSource) into a single unified Source struct, eliminating code duplication while maintaining identical functionality.

### DIFF
--- a/shotover/src/sources/cassandra.rs
+++ b/shotover/src/sources/cassandra.rs
@@ -9,7 +9,6 @@ use serde::{Deserialize, Serialize};
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::sync::{Semaphore, watch};
-use tokio::task::JoinHandle;
 use tracing::{error, info};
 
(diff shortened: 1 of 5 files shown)


rukai (approved):
nice!

rukai:
@carlsandland 
>Makes sense, JOOC is there a mechanism to remove the join_handle duplication (call tokio::spawn) between sources

yep, thats a good point. I think I would achieve that by moving the spawn logic into the `TcpCodecListener::new` call.
This PR unblocks richu's work, so we'll proceed as is, but I'll raise another ticket to move that spawn logic, which can be done later on.

> Was there a need originally to be aware of the type of Config that originated the Source later on?

And yeah, I suspect there was originally a need for protocol specific logic in this outer source layer, but in the current state of things there is no such need.

Issue #1923, opened 2025-08-21
CassandraSinkCluster does not look for peers table when peers_v2 is not available


bassamchahine:
## To Reproduce
1. Configure shotover to sink to a non-apache cassandra cluster
2. Start shotover
...

## Configuration
```
# configure the first `info` to set the log level for dependencies
# configure `shotover=info` to set the log level for shotover itself
# set `shotover::connection_span=info` to `shotover::connection_span=debug` to attach connection info to most log events, this is disabled by default due to a minor performance hit.
main_log_level: "info, shotover=info, shotover::connection_span=info"
observability_interface: "0.0.0.0:9001"
```

### Topology
```
---
sources:
  - Cassandra:
      name: cassandra
      listen_addr: "0.0.0.0:9043"
      chain:
        - Tee:
            behavior: Ignore
            buffer_size: 10000
            chain:
            - CassandraSinkCluster:
                first_contact_points: ["34.220.46.14:9042"]
                local_shotover_host_id: "2dd022d6-2937-4754-89d6-02d2933a8f7a"
                shotover_nodes:
                  - address: "172.31.21.175:9043"
                    data_center: "datacenter1"
                    rack: "rack1"
                    host_id: "2dd022d6-2937-4754-89d6-02d2933a8f7a"
                connect_timeout_ms: 3000

        - CassandraSinkCluster:
            first_contact_points: ["172.17.0.2:9042"]
            local_shotover_host_id: "2dd022d6-2937-4754-89d6-02d2933a8f7a"
            shotover_nodes:
              - address: "172.31.21.175:9043"
                data_center: "dc1"
                rack: "rack1"
                host_id: "2dd022d6-2937-4754-89d6-02d2933a8f7a"
            connect_timeout_ms: 3000
```

## Expected behaviour
Shotover failes to start because it is unable to locate system.peers_v2 table

## Systems and Version:
 - OS = Amazon Linux 22023
 - Shotover Version = 0.6.0
 - Cassandra=6.8.5

PR #1934, opened 2025-09-02
CassandraSinkCluster - fix system.peers fallback

rukai:
closes https://github.com/shotover/shotover-proxy/issues/1923

In shotover's [CassandraSinkCluster](https://shotover.io/docs/latest/transforms.html#cassandrasinkcluster) transform we query `system.peers_v2` to get the full list of cassandra nodes in the cluster.
We specifically prefer `system.peers_v2` over `system.peers` since it includes port information, without which we need to assume the default port of 9042.

This PR fixes the CassandraSinkCluster transform to be compatible with the [DSE](https://www.datastax.com/resources/datasheet/datastax-enterprise) database which does not have a `system.peers_v2` table and only has a `system.peers` table. Consider:

| DB | `system.peers` | `system.peers_v2` | officially support by shotover |
|----|--------|--------|--|
| DSE | yes | no | no |
| Cassandra 3 | yes |no | yes |
| Cassandra 4 | yes | yes | yes |
| Cassandra 5 | yes | yes | yes |

While we already have logic to support Cassandra 3 which falls back to `system.peers` it relied on the exact error message output by Cassandra 3 which was `unconfigured table peers_v2`.
When shotover hits a missing `peers_v2` table in DSE, DSE instead returns the error `table system.peers_v2 does not exist`.
This meant our fallback logic wasnt working for DSE.
To fix this we instead use the error type instead of error message to detect when to fallback, as this supports both DSE and Cassandra 3.
We dont want to have any error result in fallback, as an intermittent failure could result in Shotover running in a degraded state without the extra port information we need from `system.peers_v2`.

This has been manually tested to verify it resolves the issue.
I do not think any additional tests need to be written.
We do not officially support DSE, since running integration tests for it is impractical.
And we have integration tests for all the databases we do support (Cassandra 3, 4 and 5) which demonstrate that system.peers is used for 4 and 5 and that we successfully fallback for 3.

### DIFF
--- a/shotover/src/transforms/cassandra/sink_cluster/topology.rs
+++ b/shotover/src/transforms/cassandra/sink_cluster/topology.rs
@@ -481,6 +481,8 @@ mod system_local {
 }
 
 mod system_peers {
+    use cassandra_protocol::frame::message_error::ErrorType;
+
     use super::*;
 
     pub async fn query(
@@ -532,7 +534,7 @@ mod system_peers {
             ..
         })) = message.frame()
         {
-            return error.message == "unconfigured table peers_v2";
+            return error.ty == ErrorType::Invalid;
         }
 
         false


shotover/src/transforms/cassandra/sink_cluster/topology.rs
```diff
@@ -532,7 +534,7 @@ mod system_peers {
             ..
         })) = message.frame()
         {
-            return error.message == "unconfigured table peers_v2";
+            return error.ty == ErrorType::Invalid;
```
tharinduamila-insta, line 537: Is this type guaranteed to be for this specific situation?
Or are we fine to try when this is detected regardless of the original cause?

tharinduamila-insta (approved):

PR #1935, opened 2025-09-04
Socket Recreation

ric-pro:


shotover/src/hot_reload/fd_utils.rs
```diff
@@ -0,0 +1,249 @@
...
+///
+/// # Arguments
+///
+/// * `source_pid` - Process ID of the source process containing the FD
+/// * `source_fd` - File descriptor number in the source process
+/// * `expected_addr` - The expected address for validation
+///
+/// # Returns
+/// Returns a Result containing the recreated TcpListener or an error if the operation fails.
+pub fn recreate_tcp_listener_from_pidfd(
```
rukai, line 99: Where is this function called?

shotover/Cargo.toml
```diff
@@ -130,6 +130,8 @@ xxhash-rust = { version = "0.8.6", features = ["xxh3"], optional = true }
 dashmap = { version = "6.0.0", optional = true }
 atoi = { version = "2.0.0", optional = true }
 fnv = "1.0.7"
+libc = "0.2"
```
rukai, line 133: We didnt end up using libc, so it can be removed

run-test-in-docker.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
```
rukai, line 1: This script was not intended to be checked in.

It might make sense to check one in in the future, but should be done in a separate PR and should copy in local files instead of cloning from github.

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -23,36 +23,60 @@ async fn test_hotreload_basic_valkey_connection() {
...
+    // Both connections should be able to see all data since they hit the same backend
+    let new_value_from_old_conn: String = con_a.get("new_key").unwrap();
+    assert_eq!(new_value_from_old_conn, "new_value");
+
+    let existing_value_from_new_conn: String = con_new.get("existing_key").unwrap();
+    assert_eq!(existing_value_from_new_conn, "existing_value");
 
-    let final_check: String = con_a.get("key_from_a").unwrap();
-    assert_eq!(final_check, "value_from_a");
+    println!("Socket handoff successful - both old and new connections working");
```
rukai, line 79: Either remove the prints or use tracing::info instead

PR #1940, opened 2025-09-18
Refactor for hot reload to use UDS Crate

ric-pro:
## Summary
- Replaces Tokio Unix socket implementation with UDS crate to prepare for file descriptor passing without sudo requirements. This is the first step toward eliminating the need for pidfd_getfd and sudo access.

## Why this Refactor?
- Unix sockets provide an out of band mechanism for sending file descriptors between processes. (called [ancillary data](https://doc.rust-lang.org/std/os/unix/net/struct.SocketAncillary.html#method.add_fds)) 
- This sends not just the value of the FD but also tells the OS that the destination process now has permission to access the FD. The rust standard library unix sockets do not expose this functionality.
- We implemented sending of FD's on top of the[ rust standard library unix sockets](https://doc.rust-lang.org/std/os/unix/net/struct.UnixStream.html) because we didnt know about ancillary data, however the lack of permissions to access the FD in the 2nd process is causing us problems, so we are rewriting our implementation to use the uds crate which does expose the ancillary data functionality.

## Linux Requirements Impact

This refactor introduces stricter Linux requirements due to:
- [SOCK_SEQPACKET](https://man7.org/linux/man-pages/man7/socket.7.html) dependency: Requires Linux kernel support for sequenced packet sockets
- Ancillary data functionality: Unix domain socket FD passing is Linux-specific
- UDS crate limitations: Only supports Unix-like systems

As a result, existing hot reload tests are now marked as Linux-only to reflect these platform constraints.

This PR does not use the ancillary data yet. We are just moving our implementation to use the [UDS crate](https://docs.rs/uds/latest/uds/tokio/struct.UnixSeqpacketConn.html#method.send_fds).

## Changes
- Replace socket types: tokio::net::UnixListener/UnixStream → uds::tokio::UnixSeqpacketListener/UnixSeqpacketConn
- Switch protocols: SOCK_STREAM → SOCK_SEQPACKET for packet boundaries
- Simplify JSON parsing: Remove streaming parser, use packet-based approach

### DIFF
--- a/shotover-proxy/tests/runner/hotreload_int_tests.rs
+++ b/shotover-proxy/tests/runner/hotreload_int_tests.rs
@@ -1,63 +1,59 @@
-//! Hot reload integration tests - Linux only
-#[cfg(target_os = "linux")]
-mod tests {
-    use crate::shotover_process;
-    use redis::{Client, Commands};
-    use test_helpers::docker_compose::docker_compose;
+use crate::shotover_process;
+use redis::{Client, Commands};
+use test_helpers::docker_compose::docker_compose;
 
-    #[tokio::test]
-    async fn test_hotreload_basic_valkey_connection() {
-        let _compose = docker_compose("tests/test-configs/hotreload/docker-compose.yaml");
-        let shotover_process = shotover_process("tests/test-configs/hotreload/topology.yaml")
-            .with_hotreload(true)
-            .with_config("tests/test-configs/shotover-config/config_metrics_disabled.yaml")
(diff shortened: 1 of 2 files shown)


rukai (approved):
looks great!

The linux requirements will make testing on our macbooks a little trickier, but I cant see a good way to avoid them.
On the brightside, the use of sock_seqpacket simplifies our implementation a little bit.

anamarim-instaclustr (approved):

PR #1941, opened 2025-09-18
Swap all bitnami images over to legacy equivalent

rukai:
Bitnami is deleting ALL of their existing docker image tags and putting them behind a paywall.
We use bitnami images extensively in our integration tests to provide real valkey, cassandra, kafka etc. instances to test shotover with.
bitnami have started performing brownouts of their images and this has impacted @ric-pro ability to work yesterday.
The images will be removed permanently, one week from now.

Bitnami have recreated all previously existing tags under the `bitnamilegacy` account, but will not be creating any new releases under that account.
Since we only use these images for local/CI testing purposes, and have very specific versions tagged anyway, I dont think there is any issue with just swapping to those `bitnamilegacy` images indefinitely.
If we need to start testing a new release of a DB, we can look to migrate off bitnami for that DB at that time.

This PR updates:
* All our `docker-compose.yaml` files we use to spin up DB clusters.
* some benchmark rust files, which manually create a cluster through docker commands instead of using docker-compose
* `test-helpers/src/docker_compose.rs` defines some metadata we use to determine if a docker container created in a test has finished starting up, since the image names are used here too, we also have to update this file.

All integration tests are run in CI, and since CI passed we know this change is working.

More details on the bitnami change: https://github.com/bitnami/containers/issues/83267

### DIFF
--- a/test-helpers/src/docker_compose.rs
+++ b/test-helpers/src/docker_compose.rs
@@ -15,7 +15,7 @@ pub fn new_moto() -> DockerCompose {
     docker_compose("tests/transforms/docker-compose-moto.yaml")
 }
 
-pub static IMAGE_WAITERS: [Image; 10] = [
+pub static IMAGE_WAITERS: [Image; 9] = [
     Image {
         name: "motoserver/moto",
         log_regex_to_wait_for: r"Press CTRL\+C to quit",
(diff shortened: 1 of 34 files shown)


tharinduamila-insta (approved):

test-helpers/src/docker_compose.rs
```diff
@@ -27,22 +27,17 @@ pub static IMAGE_WAITERS: [Image; 10] = [
...
     Image {
-        name: "bitnami/valkey-cluster:7.2.5-debian-12-r4",
+        name: "bitnamilegacy/valkey-cluster:7.2.5-debian-12-r4",
         //`Cluster state changed` is created by the node services
         //`Cluster correctly created` is created by the init service
         log_regex_to_wait_for: r"Cluster state changed|Cluster correctly created",
         timeout: Duration::from_secs(120),
     },
-    Image {
-        name: "bitnami/cassandra:4.0.6",
```
rukai, line 42: This image is no longer used, so I just deleted this definition instead.

anamarim-instaclustr (approved):

Issue #1915, opened 2025-07-25
hot reload TODO list


rukai:
Since the shotover hot reload project is in a fast moving prototype stage, I'm going to keep a list of items here that I discover while reviewing PRs that should be addressed eventually but are not worth worrying about until we have a better idea of the bigger picture.

- [ ] Failure to startup the hot reload server, should be considered a critical error and prevent shotover from starting up. Otherwise a production shotover could be running but in a state where hot reload is impossible.
- [ ] Final cleanup of CLI interface

PR #1942, opened 2025-09-22
Implement FD transmission via Unix Socket Ancillary Data 

ric-pro:
## Summary

Implements actual file descriptor passing using [Unix socket ancillary data](https://doc.rust-lang.org/std/os/unix/net/struct.SocketAncillary.html#method.add_fds), building on the PR -https://github.com/shotover/shotover-proxy/pull/1940's [UDS crate](https://docs.rs/uds/latest/uds/tokio/struct.UnixSeqpacketConn.html#method.send_fds) foundation. This eliminates the need for sudo access by leveraging the kernel's built-in file descriptor transfer mechanism.

## Why This Change?

Previously, our hot reload used pidfd_getfd which required sudo privileges. Unix sockets provide an out-of-band mechanism for sending file descriptors between processes using ancillary data. This tells the OS that the destination process now has permission to access the FD, eliminating the need for privileged access.

## How It Works

-  If FDs are present, uses write_json_with_fds() to send both JSON and CDs
- Client receives both using read_json_with_fds()
- Client updates response with received FDs using replace_fds_with_received()
-  Kernel assigns new FD numbers in receiving process (proving actual transfer)

## Protocol Simplification
Simplified the hot reload protocol by removing the port_to_fd HashMap from the response structure. 

### Protocol Changes:
- Simplified `Response::SendListeningSockets` to have no fields (previously contained `port_to_fd: HashMap<u32, FileDescriptor>`)
- File descriptors are now transmitted separately via Unix domain socket ancillary data
- Port information is extracted directly from file descriptors using `TcpListener::local_addr().port()`

## Safety Improvements
- Implements sound file descriptor handling following Rust's safety principles
- Unsafe operations are properly isolated in dedicated functions
- Uses OwnedFd instead of RawFd in public APIs to ensure memory safety

## Benefits

- Eliminates sudo requirement for hot reload operations
- Kernel manages FD permissions automatically
- Type-safe file descriptor handling
- Simplified protocol reduces complexity

### DIFF
--- a/shotover/src/hot_reload/json_parsing.rs
+++ b/shotover/src/hot_reload/json_parsing.rs
@@ -1,6 +1,3 @@
-// Allow unsafe code in this module for file descriptor handling
-#![allow(unsafe_code)]
-
 use anyhow::{Context, Result};
 use serde::Serialize;
 use serde::de::DeserializeOwned;
@@ -45,7 +42,8 @@ pub async fn read_json_with_fds<T: DeserializeOwned>(
         .context("Failed to read packet with FDs from socket")?;
 
     // Convert RawFds to OwnedFds immediately after receiving from recv_fds
-    // This is safe because we just received these FDs from the OS via recv_fds
+    // SAFETY: This is safe because we just received these FDs from the OS via recv_fds
+    #[allow(unsafe_code)]
     let owned_fds: Vec<OwnedFd> = fd_buf[..fds_read]
         .iter()
         .map(|&raw_fd| unsafe { OwnedFd::from_raw_fd(raw_fd) })
--- a/shotover/src/hot_reload/json_parsing.rs
+++ b/shotover/src/hot_reload/json_parsing.rs
@@ -1,6 +1,7 @@
 use anyhow::{Context, Result};
 use serde::Serialize;
 use serde::de::DeserializeOwned;
+use std::os::unix::io::RawFd;
 use uds::tokio::UnixSeqpacketConn;
 
 /// Read a complete JSON message from a SEQPACKET socket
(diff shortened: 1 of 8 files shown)


shotover/src/hot_reload/protocol.rs
```diff
@@ -23,6 +23,36 @@ pub enum Response {
...
+        match self {
+            Response::SendListeningSockets { port_to_fd } => {
+                port_to_fd.values().map(|fd| fd.0).collect()
+            }
+            Response::Error(_) => Vec::new(),
+        }
+    }
+
+    /// Replace file descriptors in the response with received FDs from ancillary data
+    pub fn replace_fds_with_received(&mut self, received_fds: Vec<RawFd>) {
```
rukai, line 39: I'm surprised we need this, I assumed that the FD's we receive in the JSON and the FD's we receive in the ancillary data would be the same.
Did you confirm they differed by logging them or something?

ric-pro: The FD numbers are different. When the kernel transfers FD 9 via ancillary data, seems like it assigns a new number (FD 10) when receiving. The JSON still contains the original FD 9, so we need to update it with the actual usable FD numbers.

shotover/src/hot_reload/json_parsing.rs
```diff
@@ -28,3 +29,46 @@ pub async fn write_json<T: Serialize>(conn: &mut UnixSeqpacketConn, data: &T) ->
 
     Ok(())
 }
+
+/// Read a complete JSON message from a SEQPACKET socket with file descriptors
+pub async fn read_json_with_fds<T: DeserializeOwned>(
+    conn: &mut UnixSeqpacketConn,
+) -> Result<(T, Vec<OwnedFd>)> {
+    let mut buf = vec![0u8; 65536]; // Large buffer for JSON messages
```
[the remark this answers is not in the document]
ric-pro: We limited FD buffer to 250 because there is a linux OS limitation of 253. We have the OS limitation with the message buffer size as well.  We had already identified this issue and have added it to #1915(list of things that needs to be addressed later). We are thinking of adjusting the protocol to send multiple SendListeningSocket responses as a solution to this issue. Currently the only variable length part of our protocol is `Response::Error` which we could just truncate the error message. But we might add more variable length parts to our protocol where that is not suitable.

tharinduamila-insta (approved):

PR #1944, opened 2025-09-27
Listening Socket Recreation

ric-pro:


shotover/src/sources/opensearch.rs
```diff
@@ -21,14 +23,38 @@ pub struct OpenSearchConfig {
 
 impl OpenSearchConfig {
     pub async fn get_source(
```
rukai, line 25: This function is unused, why is it here?
Also check for other similar cases in the PR.

rukai: Rust will give a warning if it detects unused functions, but this only works for private functions, declaring a function as `pub` disables this warning.

shotover/src/config/topology.rs
```diff
@@ -34,6 +36,15 @@ impl Topology {
     pub async fn run_chains(
```
rukai, line 36: This function is only called in tests, which is potentially dangerous, we should always be testing the functions that we are actually using in the code.

So please have the tests call `run_chains_with_listeners` instead and remove this function.

ric-pro: Done.

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -24,38 +26,74 @@ async fn test_hotreload_basic_valkey_connection() {
...
+    );
+
+    // Verify ping works
+    let pong: String = redis::cmd("PING").query(&mut con_new).unwrap();
+    assert_eq!(pong, "PONG");
 
-    let final_check: String = con_a.get("key_from_a").unwrap();
-    assert_eq!(final_check, "value_from_a");
+    // Cleanup
+    shotover_old.shutdown_and_then_consume_events(&[]).await;
```
rukai, line 93: We should open a new connection after shutting down old shotover but before shutting down new shotover.
This will make sure that the hot reload actually occurred.

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -13,49 +51,127 @@ async fn test_hotreload_basic_valkey_connection() {
...
+    // Test setting and getting a new key through old connection
+    let _: () = con_old
+        .set("old_post_reload", "value_set_after_reload")
+        .unwrap();
+
+    // Verify new instance can handle new operations
+    let _: () = con_new.set("new_key", "data_from_new_instance").unwrap();
+
+    // Test that we can increment the counter
+    let counter_value: i32 = con_new.incr("counter", 1).unwrap();
```
rukai, line 140: can this go in `assert_valkey_connection_works` as well? Maybe replace the `connection.get("counter")` logic.

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -2,6 +2,44 @@ use crate::shotover_process;
...
+    connection: &mut redis::Connection,
+    expected_counter: Option<i32>,
+    expected_data: &[(&str, &str)],
+) -> Result<(), Box<dyn std::error::Error>> {
+    // Test basic PING
+    let pong: String = redis::cmd("PING").query(connection)?;
+    assert_eq!(pong, "PONG");
+
+    // Test basic SET/GET operation
+    let test_key = format!("test_key_{}", std::process::id());
```
rukai, line 16: Why do we use the process ID as part of the key name?

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -13,49 +51,120 @@ async fn test_hotreload_basic_valkey_connection() {
...
+    )
+    .unwrap();
+
+    // Test that old connection can still perform operations
+    let _: () = con_old
+        .set("old_connection_key", "old_still_works")
+        .unwrap();
+
+    // Test increment operation on old connection
+    let old_incr: i32 = con_old.incr("counter", 10).unwrap();
```
rukai, line 122: this is covered by `assert_valkey_connection_works` now, please remove

ric-pro: Done.

rukai (approved):
Nice work!
This is a really nice milestone.

shotover/src/hot_reload/client.rs
```diff
@@ -43,7 +45,7 @@ impl UnixSocketClient {
         }
     }
 
-    async fn send_request_inner(&self, request: Request) -> Result<Response> {
+    async fn send_request_inner(&self, request: Request) -> Result<HashMap<u16, TcpListener>> {
```
rukai, line 48: Its worth explicitly calling out that having this function return a specific response instead of a generic `Response` enum is not ideal. It kind of confuses the "request in, response out" abstraction used here.
We will want to have a new Response type that can contain any kind of response, including this HashMap response.

That said, we also only have a single kind of response so far, so hard to predict what our needs will be for a response type.
So I think its best to just wait till we implement a new request/response type in the next chunk of work.

No changes needed.
You can just resolve this comment when its time to merge.

tharinduamila-insta (approved):

PR #1949, opened 2025-10-17
Using OwnedFD instead of RawFD

ric-pro:
## Replace RawFd with OwnedFd for hot reload socket transfers
This PR refactors the hot reload mechanism to use OwnedFd instead of RawFd for transferring listening socket file descriptors between shotover instances.

### Key changes:
- Proper ownership semantics: Replaced FileDescriptor([RawFd](https://doc.rust-lang.org/std/os/fd/type.RawFd.html)) wrapper with direct [OwnedFd](https://doc.rust-lang.org/beta/std/os/fd/struct.OwnedFd.html) usage, ensuring proper lifetime management of file descriptors
- Prevents resource leaks: The old instance now properly closes and unbinds listening sockets after transfer by taking ownership via `listener.take()` and converting to OwnedFd
- Fixed shutdown bug: Added `futures::pending().await `after hot reload to prevent the listener from attempting to recreate the socket after it's been transferred
- Updated tests: Modified test mocks to create valid OwnedFd instances using Unix socket pairs instead of dummy raw file descriptor integers
- Added explanatory comment: Documented the non-obvious behaviour where `return Ok(())` inside `tokio::select!` only returns to the macro level, not from the enclosing function, explaining why we use `pending().await` instead

### Technical details:

- OwnedFd is converted to RawFd only during the actual ancillary data transfer
- File descriptors are explicitly dropped after transfer to immediately close the originals (kernel already duplicated them during transfer)
- The comment clarifies that once the listener FD is transferred, the TcpCodecListener has no more work to do and should wait indefinitely rather than continue its loop

### DIFF
--- a/shotover/src/server.rs
+++ b/shotover/src/server.rs
@@ -262,9 +262,8 @@ impl<C: CodecBuilder + 'static> TcpCodecListener<C> {
                             self.handle_hot_reload_request(request).await;
                             // Wait forever once the FD has been sent. This prevents the loop from continuing
                             // and attempting to recreate the listener.
-                            // The TcpCodecListener has no more work to do once it has handed off its listener.
-                            // Note: The Ok(()) below only returns to the tokio::select! macro level, not from
-                            // the run function, attempting to return from run would shut down shotover anyway.
+                            // This is fine, since the TcpCodecListener has no more work to do once it has handed off its listener.
+                            // Unfortunately, simply returning from `run` would not work as that would cause shotover to shutdown since there are no more sources running.
                             futures::future::pending().await
                         }
                         Ok::<(), anyhow::Error>(())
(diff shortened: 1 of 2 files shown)


shotover/src/server.rs
```diff
@@ -263,7 +260,12 @@ impl<C: CodecBuilder + 'static> TcpCodecListener<C> {
                     hot_reload_request = self.hot_reload_rx.recv() => {
                         if let Some(request) = hot_reload_request{
                             self.handle_hot_reload_request(request).await;
-                            return Ok(());
+                            // Wait forever once the FD has been sent. This prevents the loop from continuing
```
rukai: Some tweaks:
```
// Wait forever once the FD has been sent. This prevents the loop from continuing
// and attempting to recreate the listener.
// This is fine, since the TcpCodecListener has no more work to do once it has handed off its listener.
// Unfortunately, simply returning from `run` would not work as that would cause shotover to shutdown since there are no more sources running.
```

Better to focus on "returning from `run` not being an option", instead of the `Ok(())` bit which would only be relevant if returning were an option.

Issue #1945, opened 2025-10-01
HotReload gradual shutdown


rukai:
Shotovers hot reload functionality needs to be able to automate the graceful shutdown of existing connections on the old shotover instance before finally shutting down the old shotover instance.

To initiate this process a new request/response needs to be added to the hot reload unix socket protocol.
Once initiated the old shotover instance should start gracefully shutting down chunks of connections until there are none left.
For now, lets implement this as shutting down 10% of connections every 10s, so it will take ~100s to completely shutdown the old shotover.
In the future we may want to make this configurable but lets start with a simple hardcoded solution.

An alternative solution could have the new shotover instance drive the shutdown process, sending individual requests to shutdown individual chunks of connections.
This has the advantage that if something goes wrong with the new instance the old instance would still be serving some of the connections.
But I also think that the extra complexity here is not really worth it, seems more likely to introduce bugs into the process than to help with robustness.
It might be worth further exploring this in the future though.

To gracefully shutdown a process means TODO...

It would be reasonable to ungracefully shutdown the TCP sockets if that made an initial PR simpler.
Then follow up with a 2nd PR to gracefully shutdown the connections.

ric-pro:
# Plan for PR1: Immediate Shutdown of Old Shotover Instance

This PR will focus on enabling the new shotover instance to trigger a shutdown of the old instance during hot reload. This phase will not implement a graceful shutdown, it will be an immediate shutdown. The graceful shutdown will be addressed in the next PR.

## Shutdown Protocol

The existing hot reload communication protocol needs to be modified to include a new request that will trigger the shutdown of the old shotover instance. This can be achieved by:

- Extending the `Request` Enum to add a new variant which will be used by the new shotover instance to pass the shutdown request to the old one.
- Extending the `Response` Enum to add a new variant which will be used by the old shotover instance to confirm that the shutdown request have been received.

## Client Side Shutdown Trigger Logic

The client shotover instance will send the shutdown request once it starts listening to the new connections through the recreated listening sockets.
- A new function to be created to send the shutdown request
- This new function will be triggered once the shotover starts listening to new connections

By this, the client shotover will be able to communicate with the server shotover to trigger the shutdown

## Server Side Shutdown Logic

The server shotover will handle the shutdown request received from the client shotover by triggering the existing shutdown mechanism.

- Modify the call to `start_hot_reload_server` within `run_inner` to pass a clone of the `trigger_shutdown_tx` sender.
[...]

PR #1952, opened 2025-10-27
Gradual Connection Draining

ric-pro:
closes https://github.com/shotover/shotover-proxy/issues/1945

### Summary
This PR implements gradual connection draining for the old Shotover instance after a hot reload handoff, replacing the previous immediate shutdown behavior. Connections are now drained in a controlled manner at 10% every 10 seconds, ensuring smoother transitions and reducing connection disruptions during hot reloads.

### Core Implementation

1. Hot Reload Protocol Changes (protocol.rs)
- Renamed ShutdownOriginalNode request/response to GradualShutdown to better reflect the new behavior
- Added GradualShutdownRequest struct for communication between the hot reload server and sources

2. Server Connection Tracking (server.rs)

- Introduced TrackedConnection struct to wrap connection handles with abort handles for controlled termination
- Each connection now has its own watch::Sender<bool> for controlled termination
- Connections can be checked for completion and shut down individually

#### Added connection draining logic that:
- Waits 10 seconds between drain cycles
- Drains 10% of active connections per cycle (minimum 1 connection)
- Aborts connections using task abort handles
- Continues until all connections are drained
- Added socket_handed_off flag to prevent recreating listeners after socket handoff
- Added draining_complete_rx channel to signal when draining is complete
- Modified run() loop to handle three concurrent operations:
- Accepting new connections (until socket handoff)
- Processing hot reload requests
- Processing gradual shutdown requests

- Removed global trigger_shutdown_rx from TcpCodecListener
- Each connection now gets a unique watch::Receiver<bool> for shutdown signalling
- Connections are no longer globally terminated but individually managed

3. Hot Reload Server (server.rs)

- Updated UnixSocketServer to track gradual shutdown state
- Removed dependency on shutdown signal watch channel
- Implemented gradual shutdown request handling that:
- Sends shutdown requests to all sources
- Waits for acknowledgments with 5-second timeout
- Sets gradual_shutdown_initiated flag to stop accepting new requests
- Server exits gracefully after initiating gradual shutdown

4. Hot Reload Client (client.rs)

- Updated request_shutdown_old_instance() to send GradualShutdown request instead of immediate shutdown
- Improved logging to indicate gradual shutdown process

#### Updated all source types (Cassandra, Kafka, OpenSearch, Valkey) to:
- Create and pass gradual_shutdown_tx channel to listeners
- Provide access to gradual shutdown channel via get_gradual_shutdown_tx()

Runner (runner.rs)
- Removed trigger_shutdown_tx parameter from hot reload server initialization
- Hot reload server no longer controls global shutdown signal

### Enhanced test to verify gradual draining behavior:
- Creates 13 pre-handoff connections instead of a single connection
- Tests each connection over 7 rounds (5 seconds each, 35 seconds total)
- Verifies that connections are gradually terminated (not all at once)
- Confirms new instance continues working throughout draining
- Waits for old instance to complete draining with 100-second timeout
- Added comprehensive logging to track draining progress

### New test_hot_reload_with_zero_connections:
- Tests hot reload when old instance has zero active connections
- Verifies old instance shuts down quickly (within 5 seconds) without wasting time on drain cycles
- Confirms new instance continues working correctly after old instance shutdown

#### Draining Schedule:

- Every 10 seconds: Drain 10% of remaining connections (minimum 1)

#### During Hot Reload:
- New Shotover instance starts and requests socket handoff
- Old instance hands off listening socket and stops accepting new connections
- New instance begins accepting new connections
- New instance requests gradual shutdown of old instance
- Old instance begins draining connections at 10% per 10 seconds
- Once all connections are drained, old instance shuts down completely

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -144,4 +144,3 @@ jobs:
             git status
             exit 1
           fi
-          


shotover/src/hot_reload/server.rs
```diff
@@ -39,12 +38,18 @@ impl UnixSocketServer {
...
             sources,
-            shutdown_tx,
+            gradual_shutdown_initiated: false,
         })
     }
 
     pub async fn run(&mut self) -> Result<()> {
         loop {
+            // Exit if gradual shutdown has been initiated
+            if self.gradual_shutdown_initiated {
```
rukai: Do we need to manually shutdown the hot reload server task?
I see no harm in just leaving it running until shotover shutsdown.

rukai (changes requested):
This is a good start, I've got some (mostly) high level feedback.
Will take a deeper look once those are resolved.

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -95,21 +117,75 @@ async fn test_hot_reload_with_old_instance_shutdown() {
...
+    );
+
+    // Verify new connections still work
+    info!("Verifying new shotover instance still works");
+    assert_valkey_connection_works(&mut con_new, Some(1), &[("key_0", "value_0")]).unwrap();
+    info!("New shotover instance verified successfully");
+
+    // Wait for the old shotover to complete draining
+    info!("Waiting for old shotover to complete draining and consume remaining events");
+    shotover_old.consume_remaining_events(&[]).await;
```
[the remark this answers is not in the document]
ric-pro: Done

shotover/src/hot_reload/server.rs
ric-pro: Done. Removed manual shutdown. hot reload server running til shotover shutdown.

shotover/src/server.rs
```diff
@@ -233,50 +345,134 @@ impl<C: CodecBuilder + 'static> TcpCodecListener<C> {
...
+                            info!("[{}] Gradual shutdown initiated", self.source_name);
+                        } else {
+                            // Channel was closed, but this is expected during normal shutdown
+                            // Just continue the loop - shotover shutdown will be triggered via the shutdown signal
+                            debug!("[{}] Gradual shutdown channel closed", self.source_name);
+                        }
+                        Ok::<bool, anyhow::Error>(false)
+                    },
+                    // Gradual shutdown draining - runs every 10 seconds when in shutdown mode
+                    _ = self.drain_interval.tick(), if self.gradual_shutdown_in_progress => {
```
rukai: Do we need this `_ = self.drain_interval.tick(), if self.gradual_shutdown_in_progress => {` state?
Cant we just run the shutdown logic immediately in `gradual_shutdown_request = self.gradual_shutdown_rx.recv() => { .. }`?

shotover/src/server.rs
```diff
@@ -212,11 +291,44 @@ impl<C: CodecBuilder + 'static> TcpCodecListener<C> {
...
+                                }
+                            }
+                        } else {
+                            self.limit_connections.clone().acquire_owned().await?
+                        };
+
+                        // Only accept new connections if socket hasn't been handed off
+                        if self.socket_handed_off {
+                            // Socket handed off, do not accept new connections
+                            futures::future::pending::<Result<(TcpStream, OwnedSemaphorePermit)>>().await
```
rukai: This will prevent gradual shutdown from occuring if 1 second has passed since the socket was handed off.

shotover/src/server.rs
```diff
@@ -212,11 +291,44 @@ impl<C: CodecBuilder + 'static> TcpCodecListener<C> {
...
-            async {
+            let should_return = async {
                 tokio::select! {
-                    // Normal Connection handling
-                    stream_result = Self::accept(&mut self.listener) => {
-                        let stream = stream_result?;
+                    // Wait for a permit to become available and accept new connection
+                    stream_result = async {
+                        // Wait for a permit to become available
+                        let permit = if self.hard_connection_limit {
```
rukai: why is this logic in the select now?
Can it be moved back to where it was before?

shotover/src/server.rs
```diff
@@ -233,50 +272,120 @@ impl<C: CodecBuilder + 'static> TcpCodecListener<C> {
...
+                            self.socket_handed_off = true;
+                            self.listener = None;
                         }
-                        Ok::<(), anyhow::Error>(())
-                    }
-                }
+                        Ok::<bool, anyhow::Error>(false)
+                    },
+                    // Gradual shutdown request handling
+                    gradual_shutdown_request = self.gradual_shutdown_rx.recv() => {
```
rukai: Can we put everything inside the `{ .. }`, into a function?

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -95,21 +118,81 @@ async fn test_hot_reload_with_old_instance_shutdown() {
...
+    )
+    .await
+    .unwrap();
+
+    info!("Old shotover draining completed");
 
     // Cleanup
+    info!("Shutting down new shotover instance");
     shotover_new.shutdown_and_then_consume_events(&[]).await;
+    info!("Test completed successfully");
```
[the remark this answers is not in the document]
ric-pro: Logs were added to debug the issues we encountered during the early stages of the PR. Even though we have resolved that issue, since we are still working on this PR and lot of changes are being made, I thought of keeping it. I was planning to remove it just before passing the PR for a second review.

shotover/src/hot_reload/server.rs
```diff
@@ -145,20 +142,43 @@ impl UnixSocketServer {
...
+            Request::GradualShutdown => {
+                info!(
+                    "Processing GradualShutdown request - initiating gradual connection draining"
+                );
+
+                // Send gradual shutdown requests to all sources
+                for source in &self.sources {
+                    let (response_tx, response_rx) = tokio::sync::oneshot::channel();
+                    let shutdown_request = GradualShutdownRequest {
+                        return_chan: response_tx,
```
rukai: Should we remove this return_chan?
I dont think it provides any value, since any source that has received the request will immediately respond, and we can already tell which sources wont receive it by checking for the error returned by `source.gradual_shutdown_tx.send(shutdown_request)` which we already check for.

rukai: We could change the return_chan logic to instead indicate when the source has completed shutdown (instead of just started shutting down). But I dont think there's any need for this logic to block on the sources shutting down.

shotover/src/server.rs
```diff
@@ -233,50 +265,110 @@ impl<C: CodecBuilder + 'static> TcpCodecListener<C> {
...
+
+    pub async fn gradual_shutdown(&mut self) {
+        info!(
+            "[{}] Gradual shutdown initiated - draining connections",
+            self.source_name
+        );
+
+        loop {
+            // Remove finished connections
+            self.connection_handles.retain(|tc| !tc.is_finished());
```
rukai: Is this needed?
We already drain the connections below?

shotover/src/server.rs
```diff
@@ -233,50 +265,110 @@ impl<C: CodecBuilder + 'static> TcpCodecListener<C> {
...
+                info!(
+                    "[{}] All connections drained, shutting down listener",
+                    self.source_name
+                );
+                return;
+            }
+
+            // Calculate 10% of connections to drain (at least 1 if there are any connections)
+            let connections_to_drain =
+                std::cmp::max(1, (total_connections as f64 * 0.1).ceil() as usize);
```
rukai: this 10% will shrink every iteration.
Say we have 100 connections.
The first iteration will close 10 connections.
The next iteration will close 9 connections.
And third will close 8 connections. (probably, the math gets more complex at this point)
And so on.

The most predictable behavior would be for each iteration to remove the same amount.
Lets
* move `let total_connections = self.connection_handles.len();` out of the loop, to fix this.
* And then change the `loop { ..; if total_connections == 0 {` into a `while !self.connection_handles.empty()` style loop.

shotover/src/server.rs
```diff
@@ -276,7 +294,64 @@ impl<C: CodecBuilder + 'static> TcpCodecListener<C> {
...
+            join_all(connections_to_close.iter_mut().map(|tc| &mut tc.handle)).await;
+
+            info!(
+                "[{}] {} connections remaining after drain",
+                self.source_name,
+                self.connection_handles.len()
+            );
+
+            // Don't sleep after draining the last batch
+            if !self.connection_handles.is_empty() {
```
rukai: good thinking!

I also appreciate fixing the `broadcast::Receiver` ->  `watch::Receiver` below.

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -96,19 +107,99 @@ async fn test_hot_reload_with_old_instance_shutdown() {
...
+        for (i, con) in connections.iter_mut().enumerate() {
+            if con.get::<_, String>(format!("key_{}", i)).is_err() {
+                connections_failed += 1;
+            }
+        }
+    }
+
+    // After sufficient time, most or all old connections should have been drained
+    assert!(
+        connections_failed > 0,
```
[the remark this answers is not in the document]
ric-pro: I have updated the test to verify 10% drain rate. The new approach is that there is a 5s Sleep to position in the middle of the drain cycle to avoid race conditions. Then verifies the drain cycles, checking after each 10s cycle. Then an assertion that cycle N should have drained atleast 2N connections. 

With 13 connections, rounding makes it ~7 cycles to fully drain. At realistic counts (100+), it's consistently ~10 cycles for 10% drain rate.

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -96,19 +107,115 @@ async fn test_hot_reload_with_old_instance_shutdown() {
...
+        let mut connections_failed = 0;
+        for (i, con) in connections.iter_mut().enumerate() {
+            if con.get::<_, String>(format!("key_{}", i)).is_err() {
+                connections_failed += 1;
+            }
+        }
+
+        let expected_drained = expected_drain_per_cycle * cycle;
+        assert!(
+            connections_failed >= expected_drained,
```
rukai: My thinking is that since we have such long cycle times (10s), we should be able to assert on an exact number?
Can you give this a go and report back what happens?

shotover/src/hot_reload/server.rs
```diff
@@ -12,21 +14,17 @@ use uds::tokio::{UnixSeqpacketConn, UnixSeqpacketListener};
 pub struct SourceHandle {
     pub name: String,
     pub sender: mpsc::UnboundedSender<HotReloadListenerRequest>,
+    pub gradual_shutdown_tx: mpsc::UnboundedSender<GradualShutdownRequest>,
```
tharinduamila-insta: should these channels be oneshot?

shotover/src/sources/opensearch.rs
```diff
@@ -30,6 +31,8 @@ impl OpenSearchConfig {
         info!("Starting OpenSearch source on [{}]", self.listen_addr);
 
         let (hot_reload_tx, hot_reload_rx) = tokio::sync::mpsc::unbounded_channel();
+        let (gradual_shutdown_tx, mut gradual_shutdown_rx) =
+            tokio::sync::mpsc::unbounded_channel::<GradualShutdownRequest>();
```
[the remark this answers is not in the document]
ric-pro: Only the first shutdown request is processed. Any subsequent request sits in the channel buffer forever and ignored. The `tokio::select!` is not in a loop. It is a one shot handler.

shotover/src/hot_reload/server.rs
rukai: Theres also some existing channels that could be oneshots as well, I've added an item to https://github.com/shotover/shotover-proxy/issues/1915

Since you've given approval, I'll merge the PR as is, and that item can be addressed in the future.

Issue #1953, opened 2025-11-06
Hotreload - improve CLI interface


rukai:
## Current design
The current CLI interface for hot reload has three CLI flags:
* `--hotreload` enable hot reload functionality
* `--hotreload_socket_path` Path to create a unix socket to listen for new shotover instances that want to hot reload from this one
* `--hotreload_from_socket` Path to connect to a unix socket of a shotover instance that this instance will hot reload from

## Issues with current design

1. `--hotreload` is redundant, we should be able to infer this one from the state of the other two.
2. Chaining hot reloads together is complex:
```
shotover-proxy --hotreload  --hotreload_socket_path /tmp/socket1
shotover-proxy --hotreload  --hotreload_socket_path /tmp/socket2 --hotreload_from_socket /tmp/socket1
shotover-proxy --hotreload  --hotreload_socket_path /tmp/socket3 --hotreload_from_socket /tmp/socket2
...
```

## improved design

Hot reload uses a single flag: `--hotreload_socket /tmp/socket`

By default hot reload is disabled, by providing this flag hot reload is enabled at the provided socket.

This new design will allow hot reloads to be chained like:
```
shotover-proxy --hotreload_socket /tmp/socket
shotover-proxy --hotreload_socket /tmp/socket
shotover-proxy --hotreload_socket /tmp/socket
...
```

## Handling initial instance

To allow the initial shotover to be started with hot reload enabled, if the specified socket does not exist, hot reload will not be performed, but the shotover instance will create the socket so that it can be hot reloaded from.

In a scenario like:
```
shotover-proxy --hotreload_socket /tmp/socket
shotover-proxy --hotreload_socket /tmp/socket1
```
The second shotover instance will not be able to create listener sockets since its ports are already taken by the first shotover, when this occurs the 2nd shotover instance will fail to startup.
In this way shotover will fail with a somewhat meaningful error, that the ports are already in use.

## handling handover of hot reload unix socket

The naive approach would result in the following:
```
shotover-proxy --hotreload_socket /tmp/socket # create unix socket at /tmp/socket
shotover-proxy --hotreload_socket /tmp/socket # attempt to create unix socket at /tmp/socket but fail because it already exists.
```
To avoid this issue I think the best solution would be to add a retry to our hot reload server task loop.
Once the original shotover instance has shutdown the new shotover instances hot reload server loop will finally successfully create the socket after many retries.

[...]

PR #1954, opened 2025-11-13
CLI CleanUp

ric-pro:
closes: https://github.com/shotover/shotover-proxy/issues/1953

## Simplified Hot Reload CLI

This update streamlines how hot reload is configured and managed.

### What Changed:

Instead of having multiple confusing command-line flags (--hotreload, --hotreload-socket-path, and --hotreload-from-socket), there's now just one simple option: --hotreload-socket.

### How It Works Now:

When you provide a socket path:
- If the socket already exists, the system automatically knows to connect to the running instance for a smooth handoff
- If the socket doesn't exist, the system creates it and sets up hot reload capability
- If the socket is temporarily unavailable, the system now waits and retries instead of failing immediately

### Why This Matters:

- Simpler to use: One flag instead of three means less confusion and fewer mistakes
- Smarter behavior: The system figures out what to do automatically based on whether the socket exists
- More reliable: Better retry logic means hot reloads are less likely to fail due to timing issues

This makes zero-downtime deployments easier to set up and more reliable.

### DIFF
--- a/shotover-proxy/tests/kafka_int_tests/mod.rs
+++ b/shotover-proxy/tests/kafka_int_tests/mod.rs
@@ -13,6 +13,7 @@ use test_helpers::connection::kafka::python::run_python_bad_auth_sasl_scram;
 use test_helpers::connection::kafka::python::run_python_smoke_test_sasl_scram;
 use test_helpers::connection::kafka::{KafkaConnectionBuilder, KafkaDriver};
 use test_helpers::docker_compose::docker_compose;
+use test_helpers::metrics::assert_metrics_contains_keys;
 use test_helpers::shotover_process::{Count, EventMatcher};
 use tokio_bin_process::event::Level;
 
@@ -822,6 +823,9 @@ async fn cluster_sasl_scram_over_mtls_nodejs_and_python() {
         run_node_smoke_test_scram("127.0.0.1:9192", "super_user", "super_password").await;
         run_python_smoke_test_sasl_scram("127.0.0.1:9192", "super_user", "super_password").await;
 
+        // verify metrics are being recorded
+        assert_delegation_token_creation_seconds_metric_emitted().await;
+
         tokio::time::timeout(
             Duration::from_secs(10),
             shotover.shutdown_and_then_consume_events(&[]),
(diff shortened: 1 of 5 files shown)


shotover/src/hot_reload/server.rs
```diff
@@ -184,15 +185,57 @@ pub fn start_hot_reload_server(socket_path: String, sources: &[Source]) {
         .collect();
 
     tokio::spawn(async move {
-        match UnixSocketServer::new(socket_path, source_handles) {
-            Ok(mut server) => {
-                info!("Unix socket server started for hot reload communication");
-                if let Err(e) = server.run().await {
-                    error!("Unix socket server error: {:?}", e);
+        const MAX_RETRIES: u32 = 100;
```
rukai: There should be no maximum retries, shotover should continue retrying forever until this succeeds.
Usually a max retries is a good idea, but in this case it is critical that this succeeds eventually and there is only one instance of the loop so it wont impact performance.

shotover/src/hot_reload/server.rs
```diff
@@ -184,15 +185,47 @@ pub fn start_hot_reload_server(socket_path: String, sources: &[Source]) {
...
+
+                    info!("Hot reload server at {} has stopped", socket_path);
+                    break;
+                }
+                Err(e) => {
+                    retry_count += 1;
+
+                    // Adaptive logging
+                    if retry_count <= 3 || retry_count % 10 == 0 {
+                        debug!(
```
rukai: I think this is important enough that it should be `info!` level, so that it actually shows up in standard configurations of shotover.

Everything below `info!` (`debug` and `trace`) are not emitted unless shotover is specifically configured to do so.

shotover/src/hot_reload/server.rs
```diff
@@ -184,15 +185,47 @@ pub fn start_hot_reload_server(socket_path: String, sources: &[Source]) {
         .collect();
 
     tokio::spawn(async move {
-        match UnixSocketServer::new(socket_path, source_handles) {
-            Ok(mut server) => {
-                info!("Unix socket server started for hot reload communication");
-                if let Err(e) = server.run().await {
-                    error!("Unix socket server error: {:?}", e);
+        const RETRY_DELAY: std::time::Duration = std::time::Duration::from_secs(3);
+        let mut retry_count = 0;
```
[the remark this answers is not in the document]
rukai: If you want to reduce the amount of times that the `Failed to bind socket at ...` log is emitted, you could increase the `RETRY_DELAY` a bit more, but I'm personally not fussed as long as its >= 3s

shotover/src/runner.rs
```diff
@@ -47,17 +47,11 @@ struct ConfigOpts {
     #[arg(long, value_enum, default_value = "human")]
     pub log_format: LogFormat,
 
-    /// Enable hot reloading functionality
+    /// Path for Unix socket used in hot reload communication.
+    /// If the socket exists, shotover will connect to it for hot reload.
+    /// If the socket doesn't exist, shotover will create it and enable hot reload server.
```
rukai: The documentation should also include that the hot reload server will also be started (with retry on socket availability) when the the socket doesnt exist at first.

shotover/src/runner.rs
```diff
@@ -131,31 +122,42 @@ impl Shotover {
...
-        } else if let Some(socket_path) = &params.hotreload_from_socket {
-            tracing::info!(
-                "Hot reload CLIENT will request sockets from existing shotover at: {}",
-                socket_path
-            );
-        } else {
-            tracing::debug!("Hot reloading is disabled");
-        }
+        // Determine hot reload mode based on socket existence
+        let (hotreload_socket, hotreload_client) = match params.hotreload_socket {
```
rukai: I suggest this change to simplify this logic:

Image

The changes include:
* remove `hotreload_client` field, instead derive it from `hotreload_socket` once we need the client.
* Move the `path::exists` check into the client constructor, this keeps the client logic in the client module and fits in well with the logic in run_inner.
* one downside of this approach is we lose the explicit `REPLACEMENT` mode vs `INITIAL mode` logs, but we already have plenty of logs walking through the hot reload process and I think its worth the simplification.

rukai (changes requested):
oops, noticed another issue

shotover/src/hot_reload/client.rs
```diff
@@ -1,5 +1,5 @@
-//This module gives client-side implementation for socket handoff as part of hot reloading
-//Client will connect to existing shotovers and requests for FDs
+//! This module gives client-side implementation for socket handoff as part of hot reloading.
```
rukai: good catch

rukai (approved):
reapproved

shotover/src/hot_reload/server.rs
```diff
@@ -184,15 +187,31 @@ pub fn start_hot_reload_server(socket_path: String, sources: &[Source]) {
         .collect();
 
     tokio::spawn(async move {
-        match UnixSocketServer::new(socket_path, source_handles) {
-            Ok(mut server) => {
-                info!("Unix socket server started for hot reload communication");
-                if let Err(e) = server.run().await {
-                    error!("Unix socket server error: {:?}", e);
+        const RETRY_DELAY: std::time::Duration = std::time::Duration::from_secs(3);
```
carlsandland: why 3? why not 10 or 5? should this be another cli arg?
would we ever need to backoff?
should there be a bail counter? (ie if it fails 100 times give up and panic).

carlsandland (approved):

PR #1960, opened 2025-11-25
User Configurable Gradual Shutdown

### LINKED ISSUE #1958: hot reload configuration design
As an extension of the hot reload functionality recently introduced, we want to give the user control over various timers and counters used in the gradual shutdown process.

The replacement shotover will take a flag `--hot-reload-gradual-shutdown-seconds N` which will configure the number of seconds that the shotover instance it is hot reloading from (not itself) should spend in the gradual shutdown state.
To achieve this we will need to pass the duration over to the original instance in the gradual shutdown request.
The motivation for passing it to the original shotover is because the ideal value might change in between the first shotover being started and the hot reload being performed. e.g. months have passed and shotover is now under a much higher load.

We will then hardcode the time spend on each chunk (currently 10s) to a much smaller value, say 200ms, then we can derive a number of chunks via `hot-reload-gradual-shutdown-seconds / 200ms`, e.g. `--hot-reload-shutdown-seconds` 10, would result in 50 chunks.
I think that a smaller number for time per chunk is always better (connection shutdowns are better distributed across the shutdown) so its better to not expose this to the user where they could accidentally pick a number that is too high.

TODO: I will think if there any other flags we need.

ric-pro:
closes #1958 
## Summary
Adds a --hot-reload-gradual-shutdown-seconds flag to control how long the original shotover instance spends draining connections during hot reload. Default: 60 seconds.

## Motivation
The ideal shutdown duration depends on current system load and operational requirements, which may change between the original deployment and hot reload. This flag allows the replacement instance to configure the original instance's shutdown duration, adapting to conditions at reload time.

## Changes
- New CLI flag: --hot-reload-gradual-shutdown-seconds N (default: 10)
- Improved shutdown algorithm: Changed from percentage-based (10% in every 10s) to chunk-based distribution (200ms) for more predictable and evenly distributed connection draining
- Protocol update: Duration is passed from replacement to original instance via hot reload protocol
- Test updates: Enhanced tests to validate chunk-based draining behavior

## Behavior
The shutdown now divides the total duration into 200ms chunks and distributes connection shutdowns evenly across them. For example, with 10 seconds and 50 active connections: 50 chunks, ~1 connection per chunk.

## Testing
- Updated test_hot_reload_with_old_instance_shutdown to verify chunk-based draining

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -51,14 +51,93 @@ jobs:
           # downsides:
           # * PRs that update rust version or changes deps will be slow to iterate on due to changes not being cached.
           save-if: ${{ github.ref == 'refs/heads/main' }}
+      - name: cache custom ubuntu packages
+        uses: actions/cache@v4
+        with:
+          path: shotover-proxy/build/packages
+          key: ubuntu-24.04-packages
+      - name: Install ubuntu packages
+        run: shotover-proxy/build/install_ubuntu_packages.sh
       - name: Install nextest
         uses: taiki-e/install-action@v2
         with:
           tool: nextest@0.9.70
 
+        # It is currently impossible to combine j4rs, rust-cache and nextest:
+        # * j4rs needs to store jars somewhere, by default this is in target/debug/jassets, we do have the option to move this somewhere outside of target.
+        # * rust-cache will delete all files in target/debug other than `build`, `deps` and `.fingerprint`
+        # * nextest will only archive files within the target directory
+        # There is no way of combining all of these requirements.
+        # We will need to find one of these projects that is suitable to have its requirements loosened.
+        # I suspect that rust-cache is the project that needs to change, maybe add a config field to include extra paths in the cache, similar to nextest's archive.include.
+        # This is going to be tricky and require discussion with the various upstream projects, and will take a while to land anything.
+        #
+        # So for now we need a quick workaround.
+        # This workaround is to force j4rs to be rebuilt from scratch via cargo clean.
+        # This has a cost on CI runtime and in the future we should find another solution as discussed above.
+      - name: Workaround j4rs cache issue
+        run: cargo clean -p j4rs ${{ matrix.cargo_flags }}
+
       - name: Build tests
         run: |
-          cargo nextest run ${{ matrix.cargo_flags }} --features alpha-transforms --all-targets hotreload_int_tests
+          cargo test --doc ${{ matrix.cargo_flags }} --all-features -- --show-output --nocapture
+          cargo nextest archive --archive-file nextest-${{ matrix.profile }}.tar.zst ${{ matrix.cargo_flags }} --all-features --all-targets
+      - name: Upload built tests to workflow
+        uses: actions/upload-artifact@v4
+        with:
+          name: nextest-${{ matrix.profile }}
+          path: nextest-${{ matrix.profile }}.tar.zst
+      - name: Report disk usage
+        run: |
+          df -h
+          
+          echo -e "\ntarget dir usage:"
+          du -h $PWD/target
+          
+          echo -e "\n.cargo dir usage:"
+          du -h ~/.cargo
+      - name: Cleanup archive
+        run: rm nextest-${{ matrix.profile }}.tar.zst
+      - name: Ensure that tests did not create or modify any files that arent .gitignore'd
+        run: |
+          if [ -n "$(git status --porcelain)" ]; then
+            git status
+            exit 1
+          fi
+
+  run_tests_partitioned:
+    strategy:
+      fail-fast: false
+      matrix:
+        partition: [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15 ]
+        profile: [ "release", "debug" ]
+    name: Test ${{ matrix.profile}} ${{ matrix.partition }}/15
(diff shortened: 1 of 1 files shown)


shotover/src/hot_reload/protocol.rs
```diff
@@ -2,12 +2,17 @@
...
 
 /// Requests that can be sent from replacement shotover to original shotover
 #[derive(Debug, Serialize, Deserialize)]
 pub enum Request {
     SendListeningSockets,
-    GradualShutdown,
+    /// GradualShutdown request with the total duration for the shutdown process.
+    /// The duration specifies how long the original instance should spend draining connections.
+    GradualShutdown {
+        duration_secs: u64,
```
rukai: does this work as a `Duration` instead of u64? It might not work because Serde might not support serializing/deserializing it, but worth checking.

shotover/src/hot_reload/server.rs
```diff
@@ -145,14 +145,18 @@ impl UnixSocketServer {
                 );
                 (Response::SendListeningSockets, collected_fds)
             }
-            Request::GradualShutdown => {
+            Request::GradualShutdown { duration } => {
                 info!(
-                    "Processing GradualShutdown request - initiating gradual connection draining"
+                    "Processing GradualShutdown request - initiating gradual connection draining over {}s",
+                    duration.as_secs()
```
rukai: Can we pass in the duration directly without first converting to seconds?

shotover/src/server.rs
```diff
@@ -303,20 +303,29 @@ impl<C: CodecBuilder + 'static> TcpCodecListener<C> {
...
-            (self.connection_handles.len() as f64 * 0.1).ceil() as usize,
+        // Chunk duration is fixed at 200ms for better distribution of connection shutdowns
+        const CHUNK_DURATION: Duration = Duration::from_millis(200);
+
+        // Calculate number of chunks based on total duration
+        let num_chunks = (shutdown_duration.div_duration_f64(CHUNK_DURATION) as u32).max(1);
+
+        // Calculate connections to drain per chunk (at least 1 if there are any connections)
+        let connections_to_drain =
+            ((self.connection_handles.len() as f64 / num_chunks as f64).ceil() as usize).max(1);
```
rukai: we can avoid the back and forth conversions with div_ceil:
```rust
        let connections_to_drain = self.connection_handles.len().div_ceil(num_chunks).max(1);
```

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -113,66 +113,76 @@ async fn test_hot_reload_with_old_instance_shutdown() {
...
-    );
-
-    // Calculate how many drain cycles we need to verify
-    // We'll verify enough cycles to drain at least 50% of connections
-    let num_cycles_to_verify = std::cmp::min(
-        5,
-        (total_connections as f64 / expected_drain_per_cycle as f64 / 2.0).ceil() as usize,
+    let shutdown_duration_secs = 10;
+    let chunk_duration_ms = 200;
+    let num_chunks = (shutdown_duration_secs * 1000 / chunk_duration_ms) as usize;
```
rukai: I think this math would be easier if we used `Duration`s

shotover/src/runner.rs
```diff
@@ -72,6 +80,7 @@ impl Default for ConfigOpts {
             stack_size: 2097152,
             log_format: LogFormat::Human,
             hotreload_socket: None,
+            hot_reload_gradual_shutdown_seconds: 10,
```
rukai: this should match the default above

... actually, I'm not sure why we even have this default implementation, it seems unused.
Can you remove the whole `impl Default for ConfigOpts { .. }` block.

shotover-proxy/tests/runner/hotreload_int_tests.rs
```diff
@@ -113,66 +113,73 @@ async fn test_hot_reload_with_old_instance_shutdown() {
...
-        num_cycles_to_verify
-    );
-
-    // Verify each drain cycle
-    for cycle in 1..=num_cycles_to_verify {
+    // We'll check the state after a few drain cycles
+    // Each cycle consists of multiple chunks, so we need to wait appropriately
+    // Let's verify after approximately 20%, 40%, 60%, and 80% of the shutdown period
+    let check_interval = chunk_duration * 10; // 10 chunks = ~2 seconds per check
+    let check_intervals = [
```
rukai: do we need this array if all the intervals are the same?
Maybe just do `for i in 0..4 {` for the loop below?

shotover/src/server.rs
```diff
@@ -303,20 +303,32 @@ impl<C: CodecBuilder + 'static> TcpCodecListener<C> {
...
 
-        // Calculate 10% of connections to drain (at least 1 if there are any connections)
-        let connections_to_drain = std::cmp::max(
-            1,
-            (self.connection_handles.len() as f64 * 0.1).ceil() as usize,
+        // Chunk duration is fixed at 200ms for better distribution of connection shutdowns
+        const CHUNK_DURATION: Duration = Duration::from_millis(200);
+
+        // Calculate number of chunks based on total duration
+        let num_chunks = (shutdown_duration.div_duration_f64(CHUNK_DURATION) as u32).max(1);
```
rukai: you can do `as u32` -> ` as usize` to avoid the usize conversion later on.

rukai (approved):
nice!

PR #1968, opened 2026-01-07
Fix tls configuration mismatch connection close behaviour

tharinduamila-insta:


shotover-proxy/tests/cassandra_int_tests/mod.rs
```diff
@@ -573,6 +574,29 @@ async fn source_tls_and_cluster_tls(#[case] driver: CassandraDriver) {
...
+async fn server_tls_no_client_tls() {
+    test_helpers::cert::generate_cassandra_test_certs();
+    let _compose = docker_compose("tests/test-configs/cassandra/tls/docker-compose.yaml");
+
+    let shotover = shotover_process("tests/test-configs/cassandra/tls/topology.yaml")
+        .start()
+        .await;
+
+    let result = AssertUnwindSafe(CassandraConnectionBuilder::new("127.0.0.1", 9042, Java).build())
+        .catch_unwind()
```
rukai: The cases where its reasonable to catch an unwind are exceptionally rare.
Instead we should find what is triggering the panic (probably an unwrap()) and then change it to bubble up a `Result::Err` instead.

tharinduamila-insta:
Not required to do any changes here

PR #1971, opened 2026-02-02
Ins 38005 kafka close waits

tharinduamila-insta:
The Problem
When a Kafka broker closed a connection and Shotover wasn't actively polling that connection (no pending requests), the writer_task would remain blocked on out_rx.recv().await indefinitely, keeping the TCP write half open. This caused CLOSE-WAIT state connections to accumulate.

Additionally it was also identified by @rukai that we were not sending a notification to the transform chain when reader errors are met. 
We have added the notification behaviour to awaken the transform chain in shotover::connection::spawn_read_write_tasks and also updated the definition of shotover::transforms::kafka::sink_cluster::KafkaSinkCluster::recv_responses to cleanup the broken connections regardless pending messages state

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,8 @@
 /target
 .idea/
 *.gz
-/.vscode
+/.vscode/*
+!/.vscode/extensions.json
 /shotover-proxy/tests/test-configs/valkey/tls/certs
 /shotover-proxy/tests/test-configs/valkey/tls2/certs
 /shotover-proxy/tests/test-configs/cassandra/tls/certs
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,5 @@
+{
+    "recommendations": [
+        "rust-lang.rust-analyzer",
+    ],
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -926,9 +926,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.10.1"
+version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d71b6127be86fdcfddb610f7182ac57211d4b18a3e9c82eb2d17662f2227ad6a"
+checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
 dependencies = [
  "serde",
 ]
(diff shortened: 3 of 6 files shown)


shotover/src/connection.rs
```diff
@@ -424,29 +442,62 @@ async fn writer_task<C: CodecBuilder + 'static, W: AsyncWrite + Unpin + Send + '
...
     loop {
-        if let Some(messages) = out_rx.recv().await {
-            request_pending.add(messages.len() as u64);
-            match writer.send(messages).await {
-                Err(CodecWriteError::Encoder(err)) => {
-                    return Err(ConnectionError::MessageEncode(Arc::new(err)));
+        // When reader has shutdown, only process already-buffered messages (non-blocking).
+        // This prevents CLOSE-WAIT leaks while still allowing pending sends to complete.
+        if reader_has_shutdown {
+            match out_rx.try_recv() {
```
rukai: If the read side has shutdown, I dont think theres any point in trying to send pending requests.
We need a response from the server to actually confirm that the operation occurred. So if theres no way to read the response, we may as well just give up immediately.

shotover/src/connection.rs
```diff
@@ -281,28 +281,31 @@ fn spawn_read_write_tasks<
...
+    // Destination closes connection (proactive cleanup via force_run_chain):
+    // 1.   The reader task detects that the destination has closed the connection via reader returning None
     // 1.1. connection_closed_tx is sent `ConnectionError::OtherSideClosed`
-    // 1.2. in_tx and the first out_tx are dropped
-    // 2.   The `Connection::recv/recv_try` detects that in_tx is dropped by in_rx returning None and returns the ConnectionError::OtherSideClosed received from connection_closed_rx.
-    // 2.1. `Connection::recv/recv_try` prevents any future sends or receives by storing the ConnectionError
-    // 3.   Once the user handles the error by dropping the Connection out_tx is dropped, the writer task detects this by out_rx returning None causing the task to terminate.
-    // 3.1. The writer task could also close early by detecting that the client has closed the connection via writer returning BrokenPipe
+    // 1.2. in_tx is dropped, closing the channel
+    // 1.3. force_run_chain.notify_one() is called to wake up the transform chain
```
rukai: I think we should include the termination of the read task here.

rukai (approved):
This PR is great, really good investigation work on this one!

I ran the new test locally with:
* the `connection.rs` fix reverted
* the `sink_cluster/mod.rs` fix reverted
And both cases the test failed, indicating that the test is doing its job and that both fixes are required.

I appreciate the updated internal documentation.

